PR #390, opened 2014-12-11
GEOMESA-446 Wire up TDI via WPS

kevinwheeler:
The only issue I know of is that the TDI process must be ran with xml as the output type currently. json does not work. Currently, we embed the json timeseries inside of the xml output (unless there is a query hint specifying to return an encoded time series object rather than the timeseries in json form). I'm looking into fixing it.


geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/TemporalDensityIterator.scala
```diff
@@ -156,8 +164,8 @@ object TemporalDensityIterator extends Logging {
 
   val INTERVAL_KEY = "geomesa.temporal.density.bounds"
   val BUCKETS_KEY = "geomesa.temporal.density.buckets"
-  val ENCODED_TIME_SERIES: String = "timeseries"
-  val TEMPORAL_DENSITY_FEATURE_STRING = s"$ENCODED_TIME_SERIES:String,geom:Geometry"
+  val TIME_SERIES: String = "timeseries"
```
kevinwheeler: I followed the convention found in the DensityIterator object. I have found that other objects in the iterators package use lower camel case for vals, for example in the TimestampSetIterator, TServerClassLoader, and AggregatingKeyIterator objects. The DensityIterator object is a bit different in that it uses lower camel case and also capslock with underscores for vals (like I did her). Perhaps we should just change everything in this package to use capslock with underscores.

geomesa-core/src/test/scala/org/locationtech/geomesa/core/process/temporaldensity/TemporalDensityProcessTest.scala
```diff
@@ -0,0 +1,339 @@
...
+
+  import org.locationtech.geomesa.utils.geotools.Conversions._
+
+  private val tableName = "tableTestTDIP"
+  private val sftName = "sftTestTDIP"
+
+  def createDataStore(sft: SimpleFeatureType, i: Int = 0): DataStore = {
+    val mockInstance = new MockInstance("dummy" + i)
+    val c = mockInstance.getConnector("user", new PasswordToken("pass".getBytes))
+    c.tableOperations.create(tableName)
```
kevinwheeler: I copied this code from DensityIteratorTest.scala (and also TemporalDensityIteratorTest.scala) and am honestly not sure exactly what it is doing. Should I remove it from both of these locations as well as here?

geomesa-core/src/test/resources/process/temporaldensity/temporaldensity.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?><wps:Execute version="1.0.0" service="WPS" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.opengis.net/wps/1.0.0" xmlns:wfs="http://www.opengis.net/wfs" xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" xmlns:wcs="http://www.opengis.net/wcs/1.1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 http://schemas.opengis.net/wps/1.0.0/wpsAll.xsd">
```
kevinwheeler: I tried copying the license comment found in geomesa-unique.xml / wps-unique.xml and kept getting [this error message.] when trying to use this xml file to make a wps request to geoserver. (https://gist.github.com/kevinwheeler/950d781e225f95dd8bf4) Do you want me to add it anyways? Or I could delete it from both of the *-unique xml files.

kevinwheeler: I tried copying the license comment found in geomesa-unique.xml / wps-unique.xml and kept getting [this error message.](https://gist.github.com/kevinwheeler/950d781e225f95dd8bf4) when trying to use this xml file to make a wps request to geoserver. Do you want me to add it anyways? Or I could delete it from both of the *-unique xml files.

kevinwheeler: Or as a third option, we could add it below the first line of the xml file, like where we place the geoserver curl commands -- these: ""

geomesa-core/src/main/scala/org/locationtech/geomesa/core/process/temporalDensity/TemporalDensityProcess.scala
```diff
@@ -0,0 +1,106 @@
...
+               buckets: Int
+
+               ): SimpleFeatureCollection = {
+    logger.info("Attempting Geomesa temporal density on type " + features.getClass.getName)
+
+    if(features.isInstanceOf[ReTypingFeatureCollection]){
+      logger.warn("WARNING: layer name in geoserver must match feature type name in geomesa")
+    }
+
+   val interval = new Interval(startDate.getTime, endDate.getTime)
```
elahrvivaz: [picky] formatting is a bit off here

geomesa-core/src/main/scala/org/locationtech/geomesa/core/process/temporalDensity/TemporalDensityProcess.scala
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  var resultCalc: TDResult = new TDResult(manualVisitResults)
+
+  override def getResult: CalcResult = resultCalc
+
+  def setValue(r: SimpleFeatureCollection) = resultCalc = TDResult(r)
+
+  def query(source: SimpleFeatureSource, query: Query) = {
+    logger.info("Running Geomesa temporal density process on source type " + source.getClass.getName)
```
elahrvivaz: this is probably more a debug level message

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -350,9 +351,6 @@ object SimpleFeatureTypes {
     val cardinality = Cardinality.UNKNOWN
 
     override def toAttribute: AttributeDescriptor = {
-      if (!(srid == 4326 || srid == -1)) {
-        throw new IllegalArgumentException(s"Invalid SRID '$srid'. Only 4326 is supported.")
-      }
```
elahrvivaz: What is the reason for removing this check?

elahrvivaz:
Looks good, i had a few minor comments. @kevinwheeler, could you merge in the latest accumulo1.5/1.x code to resolve the merge conflicts? Thanks.

geomesa-core/src/test/resources/process/temporaldensity/temporaldensity.xml
jahhulbert-ccri: yeah it needs to be below the XML...technically the first characters must be <?xml I think

geomesa-core/src/main/scala/org/locationtech/geomesa/core/process/temporalDensity/TemporalDensityProcess.scala
kevinwheeler: If so, we should change it in the rest of the *Process.scala files as well.

PR #407, opened 2014-12-18
GEOMESA-583 Adding customized encoding/decoding for index entries

elahrvivaz:
* Replacing IndexEntry encode/decode with IndexValueEncoder
* Adding ability to specify attributes to encode in st index value to SimpleFeatureTypes, SftBuilder
* Selecting index iterator in the case of a transform covered by the custom index value
* Improving method profiling


ronq:
Looks good, with good comments in new code added. The new files still have the license header dated from 2014 --- which I feel is appropriate as they were written in 2014.

ronq:
Merged as commit 0dc16c2bc1d84ce142592c691d167034e1f0e541

PR #414, opened 2014-12-29
GEOMESA-607 enabling caching by default

elahrvivaz:
* Also updating deprecated synchronization in metadata

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
@@ -121,7 +121,10 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
         .map(FeatureEncoding.withName)
         .getOrElse(FeatureEncoding.AVRO)
 
-    val collectStats = !useMock && Try(statsParam.lookUp(params).asInstanceOf[java.lang.Boolean] == true).getOrElse(true)
+    val collectStats = !useMock &&
+        Option(statsParam.lookUp(params).asInstanceOf[java.lang.Boolean]).forall(_ == true)
+    val caching =
+      Option(cachingParam.lookUp(params).asInstanceOf[java.lang.Boolean]).getOrElse(!useMock) == true
 
     if (collectStats) {
       new AccumuloDataStore(connector,
@@ -132,7 +135,7 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
         queryThreadsParam.lookupOpt(params),
         recordThreadsParam.lookupOpt(params),
         writeThreadsParam.lookupOpt(params),
-        cachingParam.lookUp(params).asInstanceOf[Boolean],
+        caching,
         featureEncoding) with StatWriter
     } else {
       new AccumuloDataStore(connector,
@@ -143,7 +146,7 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
         queryThreadsParam.lookupOpt(params),
         recordThreadsParam.lookupOpt(params),
         writeThreadsParam.lookupOpt(params),
-        cachingParam.lookUp(params).asInstanceOf[Boolean],
+        caching,
         featureEncoding)
     }
   }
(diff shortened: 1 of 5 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
```diff
@@ -121,7 +121,10 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
         .map(FeatureEncoding.withName)
         .getOrElse(FeatureEncoding.AVRO)
 
-    val collectStats = !useMock && Try(statsParam.lookUp(params).asInstanceOf[java.lang.Boolean] == true).getOrElse(true)
+    val collectStats = !useMock &&
+        Option(statsParam.lookUp(params).asInstanceOf[java.lang.Boolean]).forall(_ == true)
+    val caching =
+      Option(cachingParam.lookUp(params).asInstanceOf[java.lang.Boolean]).getOrElse(!useMock) == true
```
jnh5y: Do you need the == true?  Is it a Java/Scala boolean issue?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureSource.scala
```diff
@@ -67,10 +67,11 @@ trait AccumuloAbstractFeatureSource extends AbstractFeatureSource with Logging w
     new AccumuloFeatureCollection(self, query)
   }
 
-  override def getFeatures(query: Query): SimpleFeatureCollection = tryLoggingFailures(getFeaturesNoCache(query))
+  override def getFeatures(query: Query): SimpleFeatureCollection =
+    tryLoggingFailures(getFeaturesNoCache(query))
 
   override def getFeatures(filter: Filter): SimpleFeatureCollection =
-    tryLoggingFailures(getFeatures(new Query(getSchema().getTypeName, filter)))
+    getFeatures(new Query(getSchema().getTypeName, filter))
```
jnh5y: Didn't like the tryLoggingFailures?  

We added this to try and log underlying errors (which GT otherwise swallows).

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreTest.scala
```diff
@@ -1091,6 +1094,54 @@ class AccumuloDataStoreTest extends Specification {
...
+        "instanceId"        -> "mycloud",
+        "zookeepers"        -> "zoo1:2181,zoo2:2181,zoo3:2181",
+        "user"              -> "myuser",
+        "password"          -> "mypassword",
+        "auths"             -> "A,B,C",
+        "tableName"         -> sftName,
+        "useMock"           -> "true",
+        "caching"           -> false,
+        "featureEncoding"   -> "avro")).asInstanceOf[AccumuloDataStore].cachingConfig must beFalse
+      DataStoreFinder.getDataStore(Map(
```
jnh5y: I'd add a space before this line to help with readability.

jnh5y:
I've read through and this looks reasonable. (Only picky comments.)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
elahrvivaz: ah, good point. don't think so, it was just left over from the old code

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureSource.scala
elahrvivaz: there are 2 try logging failures - since it's calling a delegate method. we don't need 2 i assume...

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreTest.scala
elahrvivaz: done

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
elahrvivaz: hmm, actually it does seem to be a java/scala issue... i don't see another option that is clearer/cleaner than this.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureSource.scala
jnh5y: Fair enough....  But what if the first tryLoggingFailures throws an exception;)

elahrvivaz:
I'm going to close this - caching can cause OOM errors with large layers, so it probably shouldn't be enabled by default.

PR #415, opened 2014-12-29
GEOMESA-548 Projection Check in SimpleRasterIngest

aannex-ccri:
* Added Check in SimpleRasterIngest to ensure the data is EPSG:WGS 84


ronq:
Merged into f_wcs_3 as commit c38dc223e5b6a32143c644457e5c3d1993db30b9

PR #422, opened 2015-01-07
GEOMESA-617 WCS mosaiced image's imageType is hardcoded

jdk2pq:
Refactored mosaic code in the way a new, empty image is retrieved. Instead of harcoding the imageType, it depends on the colorModel and sampleModel from
the first raster chunk in the raster iterator and then continues mosaicing as it had before. Tested on GeoServer to return the exact image ingested.


ronq:
Merged into f_wcs_3 as commit 1156b275b21f2ff5f4ce583e6696ba3c25e18fca

PR #423, opened 2015-01-08
GEOMESA-522 - Either depend on xml-apis 1.3.04 or exclude sun files from...

mm64z:
... 1.0.b2

Added a dependency on xml-apis 1.3.04 to override transitive
depencies on xml-apis 1.0.b2.
This conforms with a requirement to remove licensed Sun files.


jnh5y:
Committed as 66634cc7e546a1acae673f1d3d5a5d1d173c16dd on 1.5.

PR #425, opened 2015-01-09
GEOMESA-625 Remove org.jaitools jt-zonalstats and jt-utils from jars

mm64z:
jt-zonalstats and jt-utils are excluded from gt-coverage.


jnh5y:
Committed as 1c794c930d56d459ef4417f381e1acfacf67e0a7 on 1.5.

PR #426, opened 2015-01-12
GEOMESA-628 GeohashUtils.getUniqueGeohashSubstringsInPolygon dies when b...

cne1x:
...its > 5

This fix represents work performed by John Armstrong during a joint
debugging session on the afternoon of Friday, January 9 2015.  There
is also a new unit test that failed on the old code, but passes on
the new fix.


elahrvivaz:
5fe2581f8669e959cfb078e788f52babd5ee230a

PR #439, opened 2015-01-23
GEOMESA-651 Adding kryo serialization for simple features

elahrvivaz:
* New KryoSimpleFeature class
* Default encoding is now kryo
* Still works with old avro-encoded schemas
* Compared to avro, rough estimates:
** 34% faster encoding, 24% faster decoding
** 17% smaller size on disk for list/map attributes

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
@@ -117,10 +117,9 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
     val modifiedParams = params ++ Map(authsParam.key -> auths.mkString(","))
     authorizationsProvider.configure(modifiedParams)
 
-    val featureEncoding =
-      featureEncParam.lookupOpt[String](params)
-        .map(FeatureEncoding.withName)
-        .getOrElse(FeatureEncoding.AVRO)
+    val featureEncoding = featureEncParam.lookupOpt[String](params)
+      .map(FeatureEncoding.withName)
+      .getOrElse(FeatureEncoding.KRYO)
 
     val collectStats = !useMock && Try(statsParam.lookUp(params).asInstanceOf[java.lang.Boolean] == true).getOrElse(true)
 
@@ -207,7 +206,7 @@ object AccumuloDataStoreFactory {
     val statsParam          = new Param("collectStats", classOf[java.lang.Boolean], "Toggle collection of statistics", false, java.lang.Boolean.TRUE)
     val cachingParam        = new Param("caching", classOf[java.lang.Boolean], "Toggle caching of results", false, java.lang.Boolean.TRUE)
     val mockParam           = new Param("useMock", classOf[String], "Use a mock connection (for testing)", false)
-    val featureEncParam     = new Param("featureEncoding", classOf[String], "The feature encoding format (text or avro). Default is Avro", false, "avro")
+    val featureEncParam     = new Param("featureEncoding", classOf[String], "The feature encoding format (kryo, avro or text). Default is Kryo", false, "kryo")
   }
 
   import org.locationtech.geomesa.core.data.AccumuloDataStoreFactory.params._
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
@@ -119,7 +119,7 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
 
     val featureEncoding = featureEncParam.lookupOpt[String](params)
       .map(FeatureEncoding.withName)
-      .getOrElse(FeatureEncoding.KRYO)
+      .getOrElse(DEFAULT_ENCODING)
 
     val collectStats = !useMock && Try(statsParam.lookUp(params).asInstanceOf[java.lang.Boolean] == true).getOrElse(true)
 
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
@@ -1,9 +1,8 @@
 package org.locationtech.geomesa.core.transform
 
 import org.geotools.process.vector.TransformProcess
+import org.locationtech.geomesa.feature.FeatureEncoding.FeatureEncoding
 import org.locationtech.geomesa.feature._
-import FeatureEncoding.FeatureEncoding
-import org.locationtech.geomesa.feature.kryo.{KryoSimpleFeature, KryoSimpleFeatureFactory}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.JavaConversions._
@@ -26,13 +25,13 @@ object TransformCreator {
     featureEncoding match {
       case FeatureEncoding.KRYO | FeatureEncoding.AVRO =>
         (feature: SimpleFeature) => {
-          val newSf = new KryoSimpleFeature(feature.getIdentifier.getID, targetFeatureType)
+          val newSf = new ScalaSimpleFeature(feature.getIdentifier.getID, targetFeatureType)
           defs.foreach { t => newSf.setAttribute(t.name, t.expression.evaluate(feature)) }
           encoder.encode(newSf)
         }
 
       case FeatureEncoding.TEXT =>
-        val builder = KryoSimpleFeatureFactory.featureBuilder(targetFeatureType)
+        val builder = ScalaSimpleFeatureFactory.featureBuilder(targetFeatureType)
         (feature: SimpleFeature) => {
           defs.foreach { t => builder.set(t.name, t.expression.evaluate(feature)) }
           val newFeature = builder.buildFeature(feature.getID)
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
@@ -3,6 +3,7 @@ package org.locationtech.geomesa.core.transform
 import org.geotools.process.vector.TransformProcess
 import org.locationtech.geomesa.feature._
 import FeatureEncoding.FeatureEncoding
+import org.locationtech.geomesa.feature.kryo.{KryoSimpleFeature, KryoSimpleFeatureFactory}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.JavaConversions._
@@ -18,27 +19,25 @@ object TransformCreator {
    */
   def createTransform(targetFeatureType: SimpleFeatureType,
                       featureEncoding: FeatureEncoding,
-                      transformString: String): (SimpleFeature => Array[Byte]) =
+                      transformString: String): (SimpleFeature => Array[Byte]) = {
+
+    val encoder = SimpleFeatureEncoder(targetFeatureType, featureEncoding)
+    val defs = TransformProcess.toDefinition(transformString)
     featureEncoding match {
-      case FeatureEncoding.AVRO =>
-        val encoder = SimpleFeatureEncoder(targetFeatureType, featureEncoding)
-        val defs = TransformProcess.toDefinition(transformString)
+      case FeatureEncoding.KRYO | FeatureEncoding.AVRO =>
         (feature: SimpleFeature) => {
-          val newSf = new AvroSimpleFeature(feature.getIdentifier, targetFeatureType)
+          val newSf = new KryoSimpleFeature(feature.getIdentifier.getID, targetFeatureType)
           defs.foreach { t => newSf.setAttribute(t.name, t.expression.evaluate(feature)) }
           encoder.encode(newSf)
         }
 
       case FeatureEncoding.TEXT =>
-        val defs = TransformProcess.toDefinition(transformString)
-        val encoder = SimpleFeatureEncoder(targetFeatureType, featureEncoding)
-        val builder = AvroSimpleFeatureFactory.featureBuilder(targetFeatureType)
+        val builder = KryoSimpleFeatureFactory.featureBuilder(targetFeatureType)
         (feature: SimpleFeature) => {
-          builder.reset()
           defs.foreach { t => builder.set(t.name, t.expression.evaluate(feature)) }
           val newFeature = builder.buildFeature(feature.getID)
           encoder.encode(newFeature)
         }
     }
-
+  }
 }
--- a/geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/ComplexFeatureTest.scala
+++ b/geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/ComplexFeatureTest.scala
@@ -35,6 +35,8 @@ import scala.collection.JavaConverters._
 @RunWith(classOf[JUnitRunner])
 class ComplexFeatureTest extends Specification with TestWithDataStore {
 
+  sequential
+
   override def spec =
     """
       |names:List[String],
--- a/geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/ComplexFeatureTest.scala
+++ b/geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/ComplexFeatureTest.scala
@@ -35,6 +35,11 @@ import scala.collection.JavaConverters._
 @RunWith(classOf[JUnitRunner])
 class ComplexFeatureTest extends Specification with TestWithDataStore {
 
+  // Due to the implementation of mock connectors, this test needs to be sequential.
+  // Our kryo decoding of the accumulo key/value pairs actually mutates the byte array as it
+  // decodes. The array is returned to it's original state after decoding - however, the mock
+  // connector returns the same byte array object every time for a given accumulo value. This causes
+  // serialization errors if we run the test non-sequentially.
   sequential
 
   override def spec =
--- a/geomesa-feature/pom.xml
+++ b/geomesa-feature/pom.xml
@@ -31,13 +31,13 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
-
-        <!-- Test dependencies -->
         <dependency>
             <groupId>com.esotericsoftware</groupId>
             <artifactId>kryo</artifactId>
             <version>3.0.0</version>
         </dependency>
+
+        <!-- Test dependencies -->
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
--- a/geomesa-feature/pom.xml
+++ b/geomesa-feature/pom.xml
@@ -34,6 +34,11 @@
 
         <!-- Test dependencies -->
         <dependency>
+            <groupId>com.esotericsoftware</groupId>
+            <artifactId>kryo</artifactId>
+            <version>3.0.0</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
--- a/geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
+++ b/geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
@@ -21,6 +21,7 @@ import java.io.{ByteArrayInputStream, ByteArrayOutputStream, InputStream}
 import org.apache.avro.io._
 import org.geotools.data.DataUtilities
 import org.locationtech.geomesa.feature.FeatureEncoding.FeatureEncoding
+import org.locationtech.geomesa.feature.kryo.KryoFeatureSerializer
 import org.locationtech.geomesa.utils.text.ObjectPoolFactory
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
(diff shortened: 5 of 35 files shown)


geomesa-feature/src/test/scala/org/locationtech/geomesa/feature/SerializationTester.scala
```diff
@@ -57,35 +58,63 @@ object SerializationTester {
...
+//    decode kryo: 7.625 24% faster!
+
+//    encode/decode avro: 16.02
+//    encode/decode kryo: 12.53 28% faster!
+
+//    complex attributes size avro: 1301239
+//    complex attributes size kryo: 1111239 17% smaller
+
+//    complex attributes encode/decode avro: 30.980
+//    complex attributes encode/decode kryo: 17.326 79% faster!
```
jahhulbert-ccri: nice

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
```diff
@@ -117,10 +117,9 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
     val modifiedParams = params ++ Map(authsParam.key -> auths.mkString(","))
     authorizationsProvider.configure(modifiedParams)
 
-    val featureEncoding =
-      featureEncParam.lookupOpt[String](params)
-        .map(FeatureEncoding.withName)
-        .getOrElse(FeatureEncoding.AVRO)
+    val featureEncoding = featureEncParam.lookupOpt[String](params)
+      .map(FeatureEncoding.withName)
+      .getOrElse(FeatureEncoding.KRYO)
```
jnh5y: (idly) I wonder if we could factor out a 'default encoding' reference.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
```diff
@@ -18,27 +19,25 @@ object TransformCreator {
...
           encoder.encode(newSf)
         }
 
       case FeatureEncoding.TEXT =>
-        val defs = TransformProcess.toDefinition(transformString)
-        val encoder = SimpleFeatureEncoder(targetFeatureType, featureEncoding)
-        val builder = AvroSimpleFeatureFactory.featureBuilder(targetFeatureType)
+        val builder = KryoSimpleFeatureFactory.featureBuilder(targetFeatureType)
         (feature: SimpleFeature) => {
-          builder.reset()
```
jnh5y: Did you absorb 'reset' into 'encode'?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/ComplexFeatureTest.scala
```diff
@@ -35,6 +35,8 @@ import scala.collection.JavaConverters._
 @RunWith(classOf[JUnitRunner])
 class ComplexFeatureTest extends Specification with TestWithDataStore {
 
+  sequential
```
jnh5y: Did you find this necessary? 

If so, why?  Parallel tests are faster:)

geomesa-feature/pom.xml
```diff
@@ -34,6 +34,11 @@
 
         <!-- Test dependencies -->
         <dependency>
+            <groupId>com.esotericsoftware</groupId>
+            <artifactId>kryo</artifactId>
+            <version>3.0.0</version>
```
jnh5y: Two questions:
1.  Does anything else in sight use Kryo?  Specifically, what about our listed Spark version?  

2.  Do we need a new CQ for this dependency?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
```diff
@@ -57,16 +58,17 @@ trait SimpleFeatureDecoder extends HasEncoding {
 }
 
 object SimpleFeatureDecoder {
+
   def apply(sft: SimpleFeatureType, encoding: FeatureEncoding) =
     encoding match {
+      case FeatureEncoding.KRYO => new KryoFeatureEncoder(sft, sft)
       case FeatureEncoding.AVRO => new AvroFeatureDecoder(sft)
       case FeatureEncoding.TEXT => new TextFeatureDecoder(sft)
```
jnh5y: (idly) I wonder if there's a Scala compiler check to make sure that all Values in an Enumeration/Value class/whatnot are handled in a match statement.

I'm thinking along the lines of a sealed trait...

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoFeatureSerializer.scala
```diff
@@ -0,0 +1,120 @@
...
+  private val serializer = if (sft.eq(decodeAs)) {
+    new SimpleFeatureSerializer(sft)
+  } else {
+    new TransformingSimpleFeatureSerializer(sft, decodeAs)
+  }
+
+  private val kryo = new Kryo()
+
+  kryo.setReferences(false)
+  kryo.register(classOf[KryoSimpleFeature], serializer,  kryo.getNextRegistrationId)
```
jnh5y: @elahrvivaz Have you used GeoMesa Spark?  We should test briefly with that; I'm wondering how registering these classes will interact with the current Kryo registration which happens to run a Spark job/shell.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoFeatureSerializer.scala
```diff
@@ -0,0 +1,120 @@
...
+ * @param sft
+ * @param decodeAs
+ */
+case class KryoFeatureSerializer(sft: SimpleFeatureType, decodeAs: SimpleFeatureType) {
+
+  private val serializer = if (sft.eq(decodeAs)) {
+    new SimpleFeatureSerializer(sft)
+  } else {
+    new TransformingSimpleFeatureSerializer(sft, decodeAs)
+  }
```
jnh5y: (picky) Alternatively, you could add multiple constructors which would sort out the serializer.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoFeatureSerializer.scala
```diff
@@ -0,0 +1,120 @@
...
+   * @return
+   */
+  def readId(value: Array[Byte]): String = {
+    input.setBuffer(value)
+    kryo.readObject(input, classOf[KryoFeatureId]).id
+  }
+}
+
+object KryoFeatureSerializer {
+  def apply(sft: SimpleFeatureType): KryoFeatureSerializer = apply(sft, sft)
```
jnh5y: Well, here's the second constructor.

Potentially refactoring the serializer construction to here might separate concerns better(?).

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeature.scala
```diff
@@ -0,0 +1,132 @@
...
+    setAttribute(index, value)
+  }
+  override def setAttribute(index: Int, value: Object) = {
+    val binding = sft.getDescriptor(index).getType.getBinding
+    values(index) = Converters.convert(value, binding).asInstanceOf[AnyRef]
+  }
+
+  // following methods delegate to setAttribute to get type conversion
+  override def setAttributes(vals: JList[Object]) =
+    vals.zipWithIndex.foreach { case (v, i) => setAttribute(i, v) }
```
jnh5y: From personal experience, I'm against zipWithIndex.  It instantiates additional objects for this control structure which slows things down. 

Since we are pushing for speed, I'd suggest trying a more 'Java' approach to the iteration.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeature.scala
```diff
@@ -0,0 +1,132 @@
...
+    case _ => new ReferencedEnvelope(sft.getCoordinateReferenceSystem)
+  }
+
+  override def getDefaultGeometryProperty =
+    if (geomDesc == null) null else new GeometryAttributeImpl(getDefaultGeometry, geomDesc, null)
+  override def setDefaultGeometryProperty(geoAttr: GeometryAttribute) =
+    if (geoAttr == null) setDefaultGeometry(null) else setDefaultGeometry(geoAttr.getValue)
+
+  override def getProperties: JCollection[Property] =
+    getAttributes.zip(sft.getAttributeDescriptors).map {
```
jnh5y: Same idea as above.  I'd suggest trying out the mutable/Java approach.  I'd be interested to see if the benchmarks...

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeatureFactory.scala
```diff
@@ -0,0 +1,74 @@
...
+}
+
+object KryoSimpleFeatureFactory {
+
+  import scala.collection.mutable.Map
+
+  private val hints = new Hints(Hints.FEATURE_FACTORY, classOf[KryoSimpleFeatureFactory])
+  private val featureFactory = CommonFactoryFinder.getFeatureFactory(hints)
+
+  private val cache = new ThreadLocal[Map[SimpleFeatureType, SoftReference[SimpleFeatureBuilder]]] {
```
jnh5y: Nice!  We're getting fancy! 

Lemme see if I have this right...

SimpleFeatureBuilders are not thread safe, hence the ThreadLocal.  We wish to cache them by SFT, so that's the Map.  The SoftReference allows GC to reclaim the SFB, right?  

If that happens, are we reasonably sure that the Map with work appropriately?  (I ask since I haven't used this combination of devices/tricks.)

If all this works as expected, is it (or something close enough) wrapped up into a library?  If this is profitable, part of me wants to see something like
```
class SoftThreadLocalCache[A, B](builder: A => B) { ... }
```

I wonder if one could just ThreadLocal a regular Guava cache?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -0,0 +1,346 @@
...
+import scala.collection.JavaConversions._
+import scala.ref.SoftReference
+
+/**
+ * Kryo serialization implementation for simple features. This class shouldn't be used directly -
+ * see @KryoFeatureSerializer
+ *
+ * @param sft
+ */
+class SimpleFeatureSerializer(sft: SimpleFeatureType) extends Serializer[SimpleFeature] {
```
jnh5y: I wonder if this class need the word 'Kryo' in its name?  I only hesitate since extending com.esotericsoftware.kryo.Serializer is likely clear.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -0,0 +1,346 @@
...
+
+  type Encoding = (Output, SimpleFeature) => Unit
+  type Decoding = ((Input) => AnyRef, Int)
+
+  // encodings are cached per-thread to avoid synchronization issues
+  // we use soft references to allow garbage collection as needed
+  private val encodingsCache = new ThreadLocal[Map[String, SoftReference[Seq[Encoding]]]] {
+    override def initialValue = Map.empty[String, SoftReference[Seq[Encoding]]]
+  }
+  private val decodingsCache = new ThreadLocal[Map[String, SoftReference[Seq[Decoding]]]] {
```
jnh5y: Ok, ok, if we are doing it *three or more places*....  I think we should extract the class I suggested...

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -0,0 +1,346 @@
...
+   * @param clas
+   * @param metadata
+   * @return
+   */
+  def matchEncode(clas: Class[_], metadata: jMap[AnyRef, AnyRef]): (Output, AnyRef) => Unit = clas match {
+
+    case c if classOf[String].isAssignableFrom(c) =>
+      (out: Output, value: AnyRef) => out.writeString(value.asInstanceOf[String])
+
+    case c if classOf[java.lang.Integer].isAssignableFrom(c) =>
```
jnh5y: I wonder if there is a way to cut down on the repetition and/or to share this more broadly.  

If Kryo is better than Avro, it might be good to discuss serialization options with the raster team.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -0,0 +1,346 @@
...
+        out.writeInt(0, true): Unit
+      } else {
+        val bytes = WKBUtils.write(value.asInstanceOf[Geometry])
+        out.writeInt(bytes.length, true)
+        out.write(bytes)
+      }
+
+    case c if classOf[jList[_]].isAssignableFrom(c) =>
+      val subtype = metadata.get("subtype").asInstanceOf[Class[_]]
+      val subEncoding = matchEncode(subtype, null)
```
jnh5y: Out of curiosity, do we support List[List[...]] or Map[List[...], ...] or other complex combinations?  Either way, it might be good to document that.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -0,0 +1,346 @@
...
+        } else {
+          val list = new java.util.ArrayList[Object](length)
+          (0 until length).foreach(_ => list.add(subDecoding(in)))
+          list
+        }
+      }
+
+    case c if classOf[jMap[_, _]].isAssignableFrom(c) =>
+      val keyClass      = metadata.get("keyclass").asInstanceOf[Class[_]]
+      val valueClass    = metadata.get("valueclass").asInstanceOf[Class[_]]
```
jnh5y: (picky) Factoring out the magic strings?  If you want to...  See also, AvroSimpleFeatureUtils and SimpleFeatureTypes.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -0,0 +1,346 @@
...
+      val keyDecoding   = matchDecode(keyClass, null)
+      val valueDecoding = matchDecode(valueClass, null)
+
+      (in: Input) => {
+        val length = in.readInt()
+        if (length < 0) {
+          null
+        } else {
+          val map = new java.util.HashMap[Object, Object](length)
+          (0 until length).foreach(_ => map.put(keyDecoding(in), valueDecoding(in)))
```
jnh5y: From a cursory look on the interwebs, foreach in Scala is usually much slower than the similar while loop.  

Since we are in a race for speed, I'd consider ditching 'inner' foreachs...

pom.xml
```diff
@@ -436,6 +436,11 @@
                         <groupId>thirdparty</groupId>
                         <artifactId>jgrapht-jdk1.6</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <!-- replaced by minlog 1.3 under com.esotericsoftware -->
+                        <groupId>com.esotericsoftware.minlog</groupId>
+                        <artifactId>minlog</artifactId>
+                    </exclusion>
```
jnh5y: Does that mean that the new minlog with be a transitive dependency to register a CQ for?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
elahrvivaz: yeah, i was thinking that but wasn't sure how often it was going to change... I'll go ahead and add one.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
elahrvivaz: SimpleFeatureBuilder.buildFeature() calls reset internally (line 356), so the call is redundant...

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/ComplexFeatureTest.scala
elahrvivaz: I'll add a comment. This was due to that issue I mentioned to you about how a mock connector returns the exact same byte[] for multiple scans, which causes kryo to occasionally get thread conflicts which causes OOM errors when it tries to read the length of an array to allocate memory and it ends up getting some huge number... I'm pretty sure this is only a test problem though...

geomesa-feature/pom.xml
elahrvivaz: Good point, I need to double check this... it is used already by scalding I know.

elahrvivaz: ah, i also shouldn't have put this under the 'test' section...

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoFeatureSerializer.scala
elahrvivaz: I haven't tried spark... I will see if I can test it out. I believe that since this is a separate kryo instance it won't interfere with anything else.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoFeatureSerializer.scala
elahrvivaz: Will do.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeature.scala
elahrvivaz: I blame @jahhulbert-ccri, whose avro feature i blatantly ripped off :)
I'll update it.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeature.scala
elahrvivaz: will do.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
jnh5y: Yeah, we definitely don't have to change it.  On the plus side, further code can use the global default reference.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeatureFactory.scala
elahrvivaz: Yeah, guava caches would work in a thread local. However, I assume there is some overhead since the caches are meant to be shared between threads. This implementation is probably overkill... I did it this way based off of some kryo sample code here: https://github.com/EsotericSoftware/kryo#pooling-kryo-instances, was just trying to make it at fast as possible. I can extract it out into a class, though.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
elahrvivaz: I was trying to avoid ambiguity between this and the KryoFeatureSerializer... Also, this can serialize any simple feature, not just a kryo one (although it deserializes to a kryo impl). Since it's not used directly anywhere except in the KryoFeatureSerializer, I would prefer to keep it as is.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoSimpleFeature.scala
elahrvivaz: I think I might rename this class to just 'ScalaSimpleFeature', since it's not really kryo-specific at all, except for the fact that it's efficient to instantiate it from kryo.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
elahrvivaz: Will do.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
elahrvivaz: I see this pattern of class matching->functions a lot, but I'm not sure there is a better/faster way to do it... if you think of something, I'd be glad to change it.
I think in general kryo tends to outperform avro, but specific cases should be tested out before switching.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
elahrvivaz: We don't support nested lists or maps. I think this is documented in SimpleFeatureTypes (and createType will fail if you try to do a nested List)

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
elahrvivaz: Will do.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
elahrvivaz: will do.

geomesa-feature/src/test/scala/org/locationtech/geomesa/feature/SerializationTester.scala
elahrvivaz: of course, this is probably due to my implementing it poorly in avro to begin with...

pom.xml
elahrvivaz: I'll check

elahrvivaz:
We'll need new CQs for kryo 3.0.0 (should already have one for 2.2.1, which we are still using), and for minlog 1.3.0 (should already have one for 1.2, which we are replacing)

jnh5y:
@elahrvivaz can you put in the CQs?  If not, lemme know and I'll try and sort them.

elahrvivaz:
@jnh5y I've been out for a bit, I assume you haven't updated CQs?
If there are no other issues, I think it's good to merge.

geomesa-feature/src/test/scala/org/locationtech/geomesa/feature/SerializationTester.scala
jahhulbert-ccri: i think one day i'll do the whole compiled avro vs thrift vs kryo thing too. see what pops out.

jahhulbert-ccri:
looks good. using it for my project in prod now seems to have gone OK so far for a few days.

jahhulbert-ccri:
i think it would be nice to throw a command line arg to select feature encoding eventually to make testing/ingest more fun.

elahrvivaz:
@jnh5y created the CQs and I attached source code.

cne1x:
Merged as affcd057c6681958abae6dd3071aafa78dbe6365

PR #440, opened 2015-01-23
GEOMESA-652 Enable per feature visibilities

anthonyccri:
Users can now set visibilities on a per feature level
by setting a property in the feature user data.  There
is a utility class, SecurityUtils, that exposes the key
and convenience methods for setting visibilities.


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureWriter.scala
```diff
@@ -173,29 +179,30 @@ class ModifyAccumuloFeatureWriter(featureType: SimpleFeatureType,
...
         List.empty
       }
     stWriter ::: attrWriters
   }
 
+  import scala.collection.JavaConversions._
   override def remove() =
     if (original != null) {
-      removers.foreach { r => r(original) }
+      removers.foreach { r => r(original, original.getUserData.getOrElse(security.SecurityUtils.FEATURE_VISIBILITY, visibility).asInstanceOf[String]) }
```
elahrvivaz: i don't think that we currently encode user data with the feature - i suspect that updating a feature will result in the old row not getting deleted because of that... Could you add a unit test to verify one way or the other?

elahrvivaz:
Looks good. I had one comment about updates/deletes.
It might be user-friendly if we allowed something in the SFT that specified a particular attribute to be a classification/capco marking. Then we could run something (in a pluggable fashion) to turn that label into visibilities for them.
I imagine we might need something like that for the command line tools ingest as well, where we'll need to set the user data based on a column in the file.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
```diff
@@ -39,8 +39,9 @@ import scala.collection.JavaConversions._
 import scala.io.Source
 import scala.util.Success
 
+case class TypeSchema(name: String, schema: String)
 package object csv extends Logging {
-  case class TypeSchema(name: String, schema: String)
```
jnh5y: Thanks; I think some of us had some IntelliJ trouble due to this.

jnh5y:
I've read through, and it looks good.  @elahrvivaz's comment brings up an excellent point.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureWriter.scala
anthonyccri: Hmm, good point.  Checking now.

elahrvivaz:
Updates for deletes look good.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -154,28 +156,54 @@ case class QueryPlanner(schema: String,
...
+
+
+  def adaptIteratorForStandardQuery(accumuloIterator: CloseableIterator[Entry[Key, Value]],
+                                    query: Query,
+                                    decoder: SimpleFeatureDecoder): CloseableIterator[SimpleFeature] = {
+    val features = accumuloIterator.map { kv =>
+      val ret = decoder.decode(kv.getValue.get)
+      val visibility = kv.getKey.getColumnVisibility
+      if(visibility != null && !EMPTY_VIZ.equals(visibility)) {
+        ret.getUserData.put(SecurityUtils.FEATURE_VISIBILITY, visibility.toString)
```
jnh5y: Nice; I like the sequence of 'adapt' functions.  And that the vis is coming back to the user.

jnh5y:
Committed as b182f6bc867f239c135f4db0e55b88ac2958ac2d on 1.5.

PR #442, opened 2015-01-26
GEOMESA-656 Add ability to rewrite Query filter in CNF and AndSplittingFilter

jdk2pq:
Separated rewriteFilter function into rewriteFilterInCNF and rewriteFilterInDNF and added AndSplittingFilter for when filters are rewritten into CNF

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/filter/package.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/filter/package.scala
@@ -25,8 +25,8 @@ package object filter {
    * @param filter An arbitrary filter.
    * @return       A filter in DNF (described above).
    */
-  def rewriteFilter(filter: Filter)(implicit ff: FilterFactory): Filter = {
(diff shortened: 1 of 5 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/filter/package.scala
```diff
@@ -48,20 +48,75 @@ package object filter {
...
+        case f: Filter => List(List(not))
+      }
+
+    case f: Filter => List(List(f))
+  }
+
+  /**
+   * This function rewrites a org.opengis.filter.Filter in terms of a top-level AND with children filters which
+   * 1) do not contain further ANDs, (i.e., ANDs bubble up)
+   * 2) only contain at most one AND which is at the top of their 'tree'
```
cne1x: Doesn't this mean, "only contain at most one OR which is at the top of their 'tree'"?

jnh5y:
Committed as d72a09f604e3ceaa7332221954ce0080eca3f7b3 on 1.5.

PR #443, opened 2015-01-28
GEOMESA-658 Infinite loop in GeoMesaCoverageReader

mm64z:
ERROR: function getOriginalEnvelope called itself
CHANGE: now calls getBounds
GOTO ERROR

### DIFF
--- a/geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/wcs/GeoMesaCoverageReader.scala
+++ b/geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/wcs/GeoMesaCoverageReader.scala
@@ -63,7 +63,7 @@ class GeoMesaCoverageReader(val url: String, hints: Hints) extends AbstractGridC
     true
   }
 
-  override def getOriginalEnvelope = getBounds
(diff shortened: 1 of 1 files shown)


geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/wcs/GeoMesaCoverageReader.scala
```diff
@@ -63,7 +63,7 @@ class GeoMesaCoverageReader(val url: String, hints: Hints) extends AbstractGridC
     true
   }
 
-  override def getOriginalEnvelope = this.getOriginalEnvelope
+  override def getOriginalEnvelope = getBounds
```
aannex-ccri: actually, it would probably be better to just to set it to "this.originalEnvelope", as the call to getBounds is occurring at init anyways.

mm64z:
I see the initialization now.  Pushed an update, I believe this reflects the original intention.

jnh5y:
Committed on f_wcs_4 as 319116d1ba143acd1999e9a4c639a1745778614f.

PR #449, opened 2015-02-02
GEOMESA-670 Timing raster ingest adding up

mm64z:
This fix makes a new timer for each raster ingest as opposed to one long-lived one.
Times no longer add up, and reported numbers correspond to individual ingests


jnh5y:
Committed as c117bec6a49d75090d961be2c072656e411a409d on f_wcs_4.

PR #453, opened 2015-02-04
GEOMESA-676 Remove exclude epsg-hsql jar from geomesa-raster and related modules

ronq:
This is needed to prevent issues with CRS lookups in the runtime jar. The blanket exclusion may cause trouble with raster ingest so needs to be tested first.


ronq:
@aannex-ccri found issues on raster ingest with the blanket exclusion of epsg-hsql. I've instead just excluded it from the runtime jar.

jnh5y:
Committed as 221e4f88d49efb6b5016f36abbe18a74ae223fe6 on f_wcs_4.

PR #454, opened 2015-02-05
GEOMESA-677 csv to gml

elahrvivaz:
* Adding assemble jar for geomesa-csv that can be dropped into geoserver
* Adding csv-gml endpoint to geomesa-csv
* Adding parsing of different geometry types to geomesa-csv

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
@@ -18,18 +18,18 @@ package org.locationtech.geomesa.core
 
 import java.io._
 import java.lang.{Double => jDouble, Integer => jInt}
-import java.util.{Date, Iterator => jIterator}
 import java.util.zip.{ZipEntry, ZipOutputStream}
+import java.util.{Date, Iterator => jIterator}
 
 import com.typesafe.scalalogging.slf4j.Logging
-import com.vividsolutions.jts.geom.{Coordinate, GeometryFactory, Point}
-import org.apache.commons.csv.{CSVRecord, CSVFormat}
+import com.vividsolutions.jts.geom.{Coordinate, Geometry, GeometryFactory, Point}
+import org.apache.commons.csv.{CSVFormat, CSVRecord}
 import org.apache.commons.io.FilenameUtils
 import org.geotools.data.DefaultTransaction
 import org.geotools.data.shapefile.{ShapefileDataStore, ShapefileDataStoreFactory}
-import org.geotools.data.simple.{SimpleFeatureStore, SimpleFeatureCollection}
-import org.geotools.feature.simple.SimpleFeatureBuilder
+import org.geotools.data.simple.{SimpleFeatureCollection, SimpleFeatureStore}
 import org.geotools.feature.DefaultFeatureCollection
+import org.geotools.feature.simple.SimpleFeatureBuilder
 import org.locationtech.geomesa.core.csv.CSVParser._
 import org.locationtech.geomesa.core.util.SftBuilder
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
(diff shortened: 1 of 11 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
```diff
@@ -39,84 +39,75 @@ import scala.collection.JavaConversions._
...
-    val typename = FilenameUtils.getBaseName(csvFile.getName)
-    val reader = Source.fromFile(csvFile).bufferedReader()
-    val guess = guessTypes(typename, reader, hasHeader)
-    reader.close()
-    guess
+  protected[csv] def getParser[A](clas: Class[A]) = clas match {
+    case c if c.isAssignableFrom(classOf[jInt])     => IntParser
+    case c if c.isAssignableFrom(classOf[jDouble])  => DoubleParser
+    case c if c.isAssignableFrom(classOf[Date])     => TimeParser
+    case c if c.isAssignableFrom(classOf[Point])    => PointParser
```
jnh5y: Do LineString, Polygon, etc. parsers need to be listed here?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
```diff
@@ -167,14 +156,14 @@ package object csv extends Logging {
...
     val records = CSVFormat.DEFAULT.parse(reader).iterator()
     if (hasHeader) { records.next } // burn off header rather than try (and fail) to parse it.
-    for {
-      record <- records
-      f      <- buildFeature(record, fb, parsers, latlonIdx) // logs and discards lines that fail to parse but keeps processing
-    } fc.add(f)
+    records.foreach { record =>
+      // logs and discards lines that fail to parse but keeps processing
+      buildFeature(record, fb, parsers, latlonIdx).foreach(fc.add)
+    }
```
jnh5y: More idiomatic might be 
```
for { 
  record <- records
  feature <- buildFeature(record, fb, parsers, latlonIdx).foreach(fc.add) 
} {
  fc.add(feature)
}
```
Of course, at this point, whenever I see foreach/zip, etc, I start to wonder if a while loop would be faster...

geomesa-core/src/test/scala/org/locationtech/geomesa/core/csv/CSVParserTest.scala
```diff
@@ -43,30 +43,61 @@ class CSVParserTest extends Specification {
...
+    }
+    "parse Geometries"  in {
+      val wkt = "POINT(1 1)"
+      val geom = WKTUtils.read(wkt)
+      GeometryParser.parse(wkt) must beASuccessfulTry(geom)
+    }
+    "parse strings" in {
+      StringParser.parse(s) must beASuccessfulTry(s)
     }
+  }
```
jnh5y: Is it worth testing failure?  I can see it either way.

geomesa-web/geomesa-web-csv/pom.xml
```diff
@@ -15,29 +15,93 @@
         <dependency>
             <groupId>org.locationtech.geomesa</groupId>
             <artifactId>geomesa-web-core-accumulo1.5</artifactId>
-            <version>1.0.0-rc.4-SNAPSHOT</version>
+            <version>${project.parent.version}</version>
```
jnh5y: Should we use this syntax for all the sub-module poms?  I imagine the maven release plugin will handle this either way.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
elahrvivaz: yes, good catch.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
elahrvivaz: yeah, I don't really like for comprehensions for some reason... I'll change it back.

geomesa-web/geomesa-web-csv/pom.xml
elahrvivaz: I think it's better to use the variable, that way they can't get out of sync (this happened in the schema changes branch).

geomesa-core/src/test/scala/org/locationtech/geomesa/core/csv/CSVParserTest.scala
elahrvivaz: Ah, yeah, probably a good idea.

geomesa-web/geomesa-web-csv/pom.xml
jnh5y: If that's the case, should we do it for all the other sub-projects?

elahrvivaz: I think it is like that in most places already... it probably isn't worth changing wholesale, we can address it on a case-by-case basis.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
```diff
@@ -39,84 +39,80 @@ import scala.collection.JavaConversions._
...
-      case 'p' =>
-        if (defaultGeomSet) sftb.geometry(field)
-        else {
-          sftb.point(field, default = true)
-          defaultGeomSet = true
-        }
-      case 's' =>
-        sftb.stringType(field)
-    }}
+    headers.zip(parsers).foreach { case (field, parser) =>
```
cne1x: I suppose this works, because each of independent parser objects is declared inside the ```CSVParser``` object that gives them all access to the same shared builder, right?  This strikes me as somewhat less intuitive than the old approach in which the ```SftBuilder``` was declared locally.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
```diff
@@ -39,84 +39,80 @@ import scala.collection.JavaConversions._
...
-    val (header, record) = sampleRecords(records, hasHeader)
-    val typeChars = typeData(record.iterator)
+                 format: CSVFormat = CSVFormat.DEFAULT,
+                 numSamples: Int = 5): TypeSchema = {
+    assert(numSamples > 0)
+    val records = format.parse(csvReader).iterator.take(numSamples + 1).toSeq
+    assert(records.size > 1 || (!hasHeader && records.size > 0))
+
+    val headers = guessHeaders(records(0), hasHeader)
+    val sample = records.drop(1)
```
cne1x: [minor]  Do we always want to drop the first record, or only when the first record is known to be a header?  (This is way over thinking things, I suspect.)  This also affects line 76's ```+ 1```, I suspect.

elahrvivaz: This is just for guessing the types, so it shouldn't matter that we drop the header - we just want a few rows we can look at. It would make sense to use the 'hasHeader' flag, but if somehow there was a header when we thought there wasn't, the type inference wouldn't work. So, I can see it both ways...
What do you think?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
elahrvivaz: The goal was to try to not duplicate/split logic about types. Since each parser already knows it's type, it knows how to define it with the SftBuilder. It made more sense to me to keep all the type logic in the parsers, vs splitting some of it out to here based on the parser 'flag'. Also, once I added all the other geometry types, the flags became unwieldy.
The SftBuilder is still declared locally, it gets passed in to the parser to call the correct method.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/csv/package.scala
elahrvivaz: although the assert right above that doesn't match the current behavior, as it expects the header to only be dropped if hasHeader.

cne1x:
Merged as 615e8ec9dfb69e26d36cf13485b5cdd4100e46c7

PR #459, opened 2015-02-10
GEOMESA-568 Configure Parameters in Rasterstore

aannex-ccri:
* Exposed Query, Shard, and Write parameters in RasterStore


jnh5y:
Committed as 7af2150c09397edeec45d34dbbd2c45427537759 on f_wcs_4.

PR #463, opened 2015-02-11
GEOMESA-685 Versioning Iterator Settings Change

aannex-ccri:
* Versioning Iterator now set to only store one version of a chunk.


jnh5y:
Committed as f1678b448c5001e0f2414d17f34a6bf29704a0ba on f_wcs_4.

PR #464, opened 2015-02-11
GEOMESA-686 DelimitedTextConverter fails to process records bigger than buffer size properly

anthonyccri:
Separate PipedReader and PipedWriter into multiple threads to avoid
blocking when writing records larger than PipedReader's buffer size.


geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
```diff
@@ -79,6 +100,7 @@ class DelimitedTextConverter(format: CSVFormat,
   }
 
   override def close(): Unit = {
+    es.shutdown()
```
elahrvivaz: I think shutdown won't terminate the running thread, so it will just keep looping/polling. You should use shutdownNow() or add some termination logic to the writing thread.

elahrvivaz:
6fa3e458eed56f062c804ea3d111786fd8a210df

PR #465, opened 2015-02-11
GEOMESA-687 adding cardinality hints to attributes for query planning

elahrvivaz:
* Query planning uses a cardinality hint when determining attribute queries
* Valid values are 'high', 'low' or 'unknown'
* Current impl uses user data for cardinality hints
** In SimpleFeatureTypes, you can specify 'attr1:index=true:cardinality=high'
** In SftBuilder, you can use: stringType("attr1", Opts(index = true, cardinality = Cardinality.HIGH))
** Manually, you can use:'attributeDescriptor.getUserData.put("cardinality", "high")

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
@@ -659,7 +659,7 @@ class AccumuloDataStore(val connector: Connector,
    */
   def getFeatureEncoding(sft: SimpleFeatureType): FeatureEncoding = {
     metadata.read(sft.getTypeName, FEATURE_ENCODING_KEY)
-      .map(FeatureEncoding.withName(_)).getOrElse(FeatureEncoding.TEXT)
+      .map(FeatureEncoding.withName).getOrElse(FeatureEncoding.TEXT)
   }
 
   // We assume that they want the bounds for everything.
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
@@ -78,7 +78,7 @@ class AccumuloDataStore(val connector: Connector,
                         val writeThreadsConfig: Option[Int] = None,
                         val cachingConfig: Boolean = false,
                         val featureEncoding: FeatureEncoding = FeatureEncoding.AVRO)
-    extends AbstractDataStore(true) with AccumuloConnectorCreator with Logging {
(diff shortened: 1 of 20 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
```diff
@@ -652,16 +652,14 @@ class AccumuloDataStore(val connector: Connector,
...
 
   /**
    * Reads the feature encoding from the metadata. Defaults to TEXT if there is no metadata.
    */
   def getFeatureEncoding(sft: SimpleFeatureType): FeatureEncoding = {
-    val encodingString = metadata.read(sft.getTypeName, FEATURE_ENCODING_KEY)
-                         .getOrElse(FeatureEncoding.TEXT.toString)
-    FeatureEncoding.withName(encodingString)
+    metadata.read(sft.getTypeName, FEATURE_ENCODING_KEY)
+      .map(FeatureEncoding.withName(_)).getOrElse(FeatureEncoding.TEXT)
```
cne1x: [style] In this sort of situation, you can omit the explicit argument, and use the shorthand, ```.map(FeatureEncoding.withName)```.

elahrvivaz: Good point. I usually do that, but for some reason I feel like it didn't compile or something... but may have just been intellij's flakiness with enums. I'll try it again and see.

elahrvivaz: fixed.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
```diff
@@ -591,4 +461,214 @@ object AttributeIndexStrategy {
...
+      case f: Filter =>
+        (Some(f), Seq())
+    }
+
+    val nonIndexFilters = filterListAsAnd(cqlFilter).getOrElse(Filter.INCLUDE)
+
+    val newQuery = new Query(query)
+    newQuery.setFilter(nonIndexFilters)
+
+    if (indexFilter.isEmpty) {
```
ddseapy: i'd probably prefer do a .getOrElse(throw new Exception...) and return that, than do a .get below.

elahrvivaz: fixed.

jnh5y:
Merged as beebc0581c3fba27644ac4faf0047b75e81a05fd on 1.5.

PR #466, opened 2015-02-12
GEOMESA-688 adding implicit sft wrapper generation

elahrvivaz:


geomesa-utils/README.md
```diff
@@ -0,0 +1,36 @@
+# GeoMesa Utils
+
+## Simple Feature Wrapper Generation
+Tools can generate wrapper classes for simple feature types defined in TypeSafe Config files. Your
+config files should be under src/main/resources. Add the following snipped to your pom, specifying
```
BrittEvans: snipped -> snippet

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/GenerateFeatureWrappers.scala
```diff
@@ -0,0 +1,176 @@
...
+    import scala.collection.JavaConversions._
+
+    val attrs = sft.getAttributeDescriptors.map(AttributeDetails(_, sft))
+
+    val sb = new StringBuilder()
+    sb.append(s"${tab}implicit class ${sft.getTypeName}")
+    sb.append("(val sf: org.opengis.feature.simple.SimpleFeature) extends AnyVal {\n")
+    attrs.foreach { a =>
+      sb.append("\n")
+      sb.append(s"$tab  ${a.getter}\n")
```
BrittEvans: Do we want to mix the getters/setters instead of having them grouped together?

elahrvivaz: I changed it so we wouldn't have to loop 3 times. If it bothers you though, I can change back.

BrittEvans: It's fine, not a big deal.

geomesa-utils/README.md
elahrvivaz: good catch. fixed.

BrittEvans:
Maybe we should include a nice toString() method in these classes that print the whole feature pipe-delimited?  I do that often for debugging.

elahrvivaz:
Do you mean in the generated classes?

BrittEvans:
yeah

elahrvivaz:
Yeah, I can add that. So you want something like:
"SimpleFeature <name>|ID|attr1=value|etc"
?

BrittEvans:
yep

elahrvivaz:
@BrittEvans added to 'debug' method (toString can't be easily overridden in an implicit) - let me know if it works for you. Output is like:
test1:fid-1|attr=test|dtg=Fri Feb 13 11:20:39 EST 2015|geom=POINT (10 10)

jnh5y:
This looks like an interesting tool.  Just to be clear, it is meant to be run manually and the output would be added to a source file?  I suppose I'm curious about the intended use case.

elahrvivaz:
If you follow the example in the readme, it gets added to the pom and then runs automatically during the 'generate-sources' phase. The output can either be checked in or just generated each build.
It only works with SFTs that are declared as typesafe config files (which is the thresher use case).
Of course, you could run it however you wanted...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/GenerateFeatureWrappers.scala
```diff
@@ -0,0 +1,185 @@
...
+      }
+    } else {
+      file.listFiles().flatMap(findFormatFiles)
+    }
+  }
+
+  /**
+   * Creates implicit wrappers for any typesafe config format files found under src/main/resources
+   *
+   * @param args (0) - base directory for the maven project
```
anthonyccri: Not for this commit, but at some point we ought to make this a full blown maven plugin.  We can keep the main method around for one-off use cases but it can just wrap the maven plugin.

jnh5y:
Committed as e17db121c2e76f0f2f4f83a9318a5aaedabf4a7c on 1.5.

PR #468, opened 2015-02-13
GEOMESA-690 Converters should only parse the input fields necessary for populating the SFT

anthonyccri:
Changed ToSimpleFeatureConverter so that it determines the input fields
and the necessary dependencies required to populate the target SFT.


elahrvivaz:
Looks good to me.

jnh5y:
Committed as f6e5bae41abc0f9f6062e50ba8c9fa5ce1fbe9cd on 1.5.

PR #470, opened 2015-02-18
GEOMESA-680 AccumuloFeatureStore fails writing features to Accumulo tabl...

wjp22903:
...e

*Fix the bug of using interator of features more than one time
*Unify shape file ingestion to only through general shape file ingestion implementation


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
```diff
@@ -41,8 +41,10 @@ class AccumuloFeatureStore(val dataStore: AccumuloDataStore, val featureName: Na
     val fids = Lists.newArrayList[FeatureId]()
     if (collection.size > 0) {
       writeBounds(collection.getBounds)
-      writeTimeBounds(collection)
 
+      val dateField = org.locationtech.geomesa.core.index.getDtgFieldName(collection.getSchema)
+      val minMaxVisitorO: Option[MinMaxTimeVisitor] =
+        if (dateField.isDefined) Some(new MinMaxTimeVisitor(dateField.get)) else None
```
anthonyccri: A more idiomatic use of options is as follows:

```scala
val minMaxVisitorO = index.getDtgFieldName.map { dtgField => new MinMaxTimeVisitor(dtgField) }
```

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
```diff
@@ -50,6 +52,8 @@ class AccumuloFeatureStore(val dataStore: AccumuloDataStore, val featureName: Na
         val feature = iter.next()
         val newFeature = fw.next()
 
+        minMaxVisitorO.foreach ( _.visit(feature) )
```
anthonyccri: Rather than calling this option which may be ```None``` for each iteration of the loop, I'd suggest traversing the collection a second time...

```scala
val iter = collection.features()
while(iter.hasNext) {
   val feature = iter.next()
   val newFeature = iter.next()
   ...
   // write feature
}

minMaxVisitorO.foreach { visitor = > collection.features().foreach { f => visitor.visit(f) } }
```

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
```diff
@@ -41,8 +41,10 @@ class AccumuloFeatureStore(val dataStore: AccumuloDataStore, val featureName: Na
     val fids = Lists.newArrayList[FeatureId]()
```
anthonyccri: Did you add to the test cases for this bug that you found?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
wjp22903: Good catch. Will do.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
wjp22903: Will do.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
wjp22903: Need to think a bit. I discussed it with other guys. The problem here is collection.features is an iterator and is empty when traversing for the second time.

wjp22903: Hmm, the codes here are wrong (different from my local branch) for some reason. It should be:
minMaxVisitorO.foreach ( _.visit(feature) )
I'going to change.

wjp22903: NVM. The codes are fine.

jnh5y: A feature collection can have an underlying iterator which is consumed by the first traversal.

If we want to save on the 'option check', let's build up a little lambdas for the two cases and then call it in the inner loop.
```
val write: SimpleFeature => Unit = if(dtg) { feature => updateBounds(feature); write(feature) } else write(feature)
while ( iter.hasNext) { 
  write(iter.next) 
 }
```

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
```diff
@@ -41,36 +42,50 @@ class AccumuloFeatureStore(val dataStore: AccumuloDataStore, val featureName: Na
...
     if (collection.size > 0) {
       writeBounds(collection.getBounds)
-      writeTimeBounds(collection)
 
+      val minMaxVisitorO = index.getDtgFieldName(collection.getSchema).map { dateField => new MinMaxTimeVisitor(dateField) }
```
jnh5y: (picky) Try out
```
index.getDtgFieldName(collection.getSchema).map { new MinMaxTimeVisitor(_) }
```
(I'm mainly mentioning this for line length.)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureStore.scala
```diff
@@ -41,36 +42,50 @@ class AccumuloFeatureStore(val dataStore: AccumuloDataStore, val featureName: Na
...
 
-        val useExisting = java.lang.Boolean.TRUE.equals(feature.getUserData.get(Hints.USE_PROVIDED_FID).asInstanceOf[java.lang.Boolean])
-        if (getQueryCapabilities().isUseProvidedFIDSupported && useExisting) {
-          newFeature.getIdentifier.asInstanceOf[FeatureIdImpl].setID(feature.getID)
+      val write: SimpleFeature => FeatureId =
+        if (minMaxVisitorO.isDefined) { feature =>
+          updateTimeBounds(feature)
+          writeFeature(fw, feature)
         }
+        else { feature => writeFeature(fw, feature) }
```
jnh5y: (Picky) Our not-documented-at-all style convention is that functions/work go on newlines.  For example,
```
else { feature => 
  writeFeature(fw, feature) 
}
```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/GeneralShapefileIngest.scala
```diff
@@ -67,6 +67,9 @@ object GeneralShapefileIngest {
         features.getSchema
 
     val featureTypeName = featureType.getName.getLocalPart
+    if (ds.getSchema(featureTypeName) != null) {
+      throw new Exception(s"Type name $featureTypeName already exists in data store...shape file ingest cannot continue")
```
jnh5y: (Picky) shapefile is usually one word.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
```diff
@@ -33,7 +35,9 @@ class IngestCommand(parent: JCommander) extends Command(parent) with Logging {
     val fmt = Option(params.format).getOrElse(getFileExtension(params.files(0)))
     fmt match {
       case CSV | TSV => new DelimitedIngest(params).run()
-      case SHP       => new ShpIngest(params).run()
```
jnh5y: Ah, nice!  Thanks for reducing duplicate code.

jnh5y:
Committed as 37249b067617cdbc0bf4ba3fcf9176241ea5757e on 1.5.

PR #472, opened 2015-02-19
GEOMESA-560 Improved Raster Query Ranges

AndrewAnnex:
* Added unit tests.
* Removed offsets in query params.
* Added touching geohashes to raster query planner.
* Made hash enumeration more selective.
* Geometry test in RFI is updated to be more constraining.
* Lengths of geohashes are stored in the CF of Bounds table as lexiencoded ints.
* Bounds information used in planning is now cached.
* RasterFilteringIterator update to be more selective.

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
@@ -72,6 +72,13 @@ case class BoundingBox(ll: Point, ur: Point) {
    */
   def contains(otherGeom: Geometry): Boolean = geom.contains(otherGeom)
 
+  /**
+   * This bounding box equals bbox if the lower left and upper right points are the same.
+   * @param bbox
+   * @return
+   */
+  def equals(bbox: BoundingBox): Boolean = (ll == bbox.ll) && (ur == bbox.ur)
+
   lazy val longitudeSize = ur.getX - ll.getX
   lazy val latitudeSize = ur.getY - ll.getY
 
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
@@ -77,7 +77,8 @@ case class BoundingBox(ll: Point, ur: Point) {
    * @param bbox
    * @return
    */
-  def equals(bbox: BoundingBox): Boolean = (ll == bbox.ll) && (ur == bbox.ur)
+  def equals(bbox: BoundingBox): Boolean = (ll.getX == bbox.ll.getX) && (ll.getY == bbox.ll.getY) &&
+                                            (ur.getX == bbox.ur.getX) && (ur.getY == bbox.ur.getY)
 
   lazy val longitudeSize = ur.getX - ll.getX
   lazy val latitudeSize = ur.getY - ll.getY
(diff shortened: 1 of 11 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
```diff
@@ -72,6 +72,13 @@ case class BoundingBox(ll: Point, ur: Point) {
    */
   def contains(otherGeom: Geometry): Boolean = geom.contains(otherGeom)
 
+  /**
+   * This bounding box equals bbox if the lower left and upper right points are the same.
+   * @param bbox
+   * @return
+   */
+  def equals(bbox: BoundingBox): Boolean = (ll == bbox.ll) && (ur == bbox.ur)
```
jnh5y: My spidey sense is going off.  Which notion of equality is '==' for Points?  And does that make sense for what we are doing?

jnh5y:
The PR looks good.  I want to hear a quick defense of the overriding/implementing equals.  (Since that can be a tricky thing: http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode%28%29.)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
AndrewAnnex: I think it is correct to compare the lower left and upper right points, although I could be doing the incorrect thing here and it may be better to directly compare the x and y values for both points. Otherwise it is simple enough to remove this implementation/overriding.

jnh5y: The comparison of the points is logical and correct.  I just confused since there are a few notions of equality on the JVM.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/BoundingBox.scala
```diff
@@ -72,6 +72,14 @@ case class BoundingBox(ll: Point, ur: Point) {
    */
   def contains(otherGeom: Geometry): Boolean = geom.contains(otherGeom)
 
+  /**
+   * This bounding box equals bbox if the lower left and upper right points are the same.
+   * @param bbox
+   * @return
+   */
+  def equals(bbox: BoundingBox): Boolean = (ll.getX == bbox.ll.getX) && (ll.getY == bbox.ll.getY) &&
```
cne1x: [picky]  Does this invite the possibility of rounding errors?  For example ```(0.1).toFloat = 0.1```, but ```(0.1).toFloat.toDouble = 0.10000000149011612```, which seems to suggest that initializing from a Float (that gets upcast to a Double) might not be the same as initializing from (what was meant to be) the same Double literal.  It should be possible to pick a reasonable tolerance for this comparison, yes?

AndrewAnnex: I am not sure how to address this yet... The BoundingBox uses two Point objects to denote lower left and upper right coordinates. Those Points in turn store their information in a Coordinate object, which uses doubles. For the purposes of equals as used here, I think that *strictness* is required. Anything else would be an "approximately equals" or an "equals within tolerance". I think that this may justify a new ticket.

jnh5y:
Committed as e857b082c8345004e9ad4c44575a40870c4d2edb on f_wcs_4.

PR #473, opened 2015-02-20
GEOMESA-592,GEOMESA-696,GEOMESA-697

dmcwhorter:
* Adding support for delimited ingest of Map attributes
* Adding the MapAggregatingIterator to support aggregating Map attributes for maps of type AnyRef -> Int
* Refactoring the DensityIterator, TemporalDensityIterator, and MapAggregatingIterator to share common code in the FeatureAggregatingIterator class
* Fixing the encodings and decodings cache key in SimpleFeatureSerializer to return the correct encodings for different SFT's with the same field names


geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -38,7 +38,9 @@ import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import org.opengis.filter.sort.{SortBy, SortOrder}
 
+import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
+import scalaz.Scalaz._
```
jnh5y: What is Scalaz giving you?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
```diff
@@ -163,6 +160,22 @@ trait Strategy {
       configureFeatureType(cfg, featureType)
 
       Some(cfg)
-    } else None
+    case mapAggregationQuery if mapAggregationQuery.getHints.containsKey(MAP_AGGREGATION_KEY) =>
```
jnh5y: I feel like this needs to be refactored.  Doesn't have to be you.  

If you don't make this better, can you add a comment referencing this ticket: https://geomesa.atlassian.net/browse/GEOMESA-698 ?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/FeatureAggregatingIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    projectedSFT = SimpleFeatureTypes.createType(simpleFeatureType.getTypeName, projectedSFTDef)
+
+    // Use density SFT for the encoder since we are transforming the feature into
+    // a sparse matrix as the result type of this iterator
+    featureEncoder = SimpleFeatureEncoder(projectedSFT, encodingOpt)
+    featureBuilder = AvroSimpleFeatureFactory.featureBuilder(projectedSFT)
+  }
+
+  def initProjectedSFTDefClassSpecificVariables(source: SortedKeyValueIterator[Key, Value],
```
jnh5y: Since this a function to be overridden, I'd love to see a comment explaining what's going on.

jnh5y:
@dmcwhorter I am excited for this, thanks.  Would it be possible to submit the work for GM-592 and 697 separately first?  

I don't think there's anything in there to change.  For the MapAggregatingIterator, I'd like to get input for others since it may evolve into an integration point.

geomesa-core/pom.xml
```diff
@@ -103,6 +103,10 @@
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.scalaz</groupId>
+            <artifactId>scalaz-core_2.10</artifactId>
+        </dependency>
```
elahrvivaz: We'd need to open a CQ with eclipse for this.

jnh5y: Maybe.  I talked to David when he was working on the original work, and I tried to figure out if he really needed scalaz.  I don't think it is strictly required.  (See my other comments.)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
dmcwhorter: |+| operator on maps

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
dmcwhorter: can you say more about how you'd like this refactored?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/FeatureAggregatingIterator.scala
dmcwhorter: I gave it that long name mostly so that it'd be obvious what it was there for, but I can put in a comment explaining as well.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
jnh5y: I'm thinking that we may be served by a general getQueryHints which returns an object to case match on.  I suppose I'm trying to hide the 'if's a little bit.

dmcwhorter:
will reopen as 2 p/r's

dmcwhorter:
reopened as https://github.com/locationtech/geomesa/pull/476

PR #476, opened 2015-02-23
GEOMESA-696 Map Aggregating Iterator

dmcwhorter:
* Adding the MapAggregatingIterator to support aggregating Map attributes for maps of type AnyRef -> Int
* Refactoring the DensityIterator, TemporalDensityIterator, and MapAggregatingIterator to share common code in the FeatureAggregatingIterator class


geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
```diff
@@ -163,6 +160,22 @@ trait Strategy {
...
+
+      val cfg = new IteratorSetting(iteratorPriority_AnalysisIterator,
+        "topfilter-" + randomPrintableString(5),
+        clazz)
+
+      val mapAttribute = query.getHints.get(MAP_AGGREGATION_KEY).asInstanceOf[String]
+
+      MapAggregatingIterator.configure(cfg, mapAttribute)
+
+
```
jnh5y: (picky) Extra newlines.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -210,6 +213,34 @@ case class QueryPlanner(schema: String,
...
+                                          decoder: SimpleFeatureDecoder): CloseableIterator[SimpleFeature] = {
+    val aggregateKeyName = query.getHints.get(MAP_AGGREGATION_KEY).asInstanceOf[String]
+
+    val maps = accumuloIterator.map { kv =>
+      decoder
+        .decode(kv.getValue.get)
+        .getAttribute(aggregateKeyName)
+        .asInstanceOf[JMap[AnyRef, Int]]
+        .asScala
+        .toMap // TODO: Slow?
```
elahrvivaz: Is the point of the toMap just to make it immutable?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
```diff
@@ -118,17 +117,16 @@ trait Strategy {
                      schema: String,
                      featureEncoding: FeatureEncoding,
                      query: Query): Option[IteratorSetting] = {
-    if (iteratorConfig.useSFFI) {
-      Some(configureSimpleFeatureFilteringIterator(featureType, ecql, schema, featureEncoding, query))
-    } else None
+    if (iteratorConfig.useSFFI) Some(configureSimpleFeatureFilteringIterator(featureType, ecql, schema, featureEncoding, query))
```
elahrvivaz: Our style is to use brackets in places like this.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
cne1x: Just to be sure, this line is only necessary to convert from a ```collection.mutable.Map``` to ```collection.immutable.Map```?

dmcwhorter: yes, thats right

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
dmcwhorter: ok, just so i know for future reference, are you referring to a style guide or is it just a convention in practice?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/FeatureAggregatingIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+  def init(source: SortedKeyValueIterator[Key, Value],
+           options: JMap[String, String],
+           env: IteratorEnvironment): Unit = {
+    this.source = source
+
+    val simpleFeatureTypeSpec = options.get(GEOMESA_ITERATORS_SIMPLE_FEATURE_TYPE)
+    simpleFeatureType = SimpleFeatureTypes.createType(this.getClass.getCanonicalName, simpleFeatureTypeSpec)
+    simpleFeatureType.decodeUserData(options, GEOMESA_ITERATORS_SIMPLE_FEATURE_TYPE)
+
+    // default to text if not found for backwards compatibility
```
cne1x: [picky]  This is an issue with the original (pasted) fragment, but I do not believe that text is the default encoding any longer, is it?

elahrvivaz: good point... I think that the idea is that if it's not text then the option will be Some - but that's not reflected in the code.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
elahrvivaz: just convention... we're not super picky about it, but since it looks like the actual code didn't change I would roll it back.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
dmcwhorter: since, i've gotten rid of scalaz, i think i can get rid of these .toMap calls also...

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/FeatureAggregatingIterator.scala
dmcwhorter: yeah default is just DEFAULT_ENCODING, so i'll take the comment out

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/TemporalDensityIterator.scala
```diff
@@ -30,125 +28,62 @@ import org.apache.commons.codec.binary.Base64
...
-import org.locationtech.geomesa.feature._
-import org.locationtech.geomesa.utils.geotools.{SimpleFeatureTypes, TimeSnap}
-import org.opengis.feature.simple.SimpleFeatureType
+import org.locationtech.geomesa.core.index.getDtgFieldName
+import org.locationtech.geomesa.core.iterators.FeatureAggregatingIterator.Result
+import org.locationtech.geomesa.core.iterators.TemporalDensityIterator.TimeSeries
+import org.locationtech.geomesa.utils.geotools.TimeSnap
 
-import scala.util.Random
+import scala.collection.mutable
```
cne1x: Could this import have unintended side-effects?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIteratorTest.scala
```diff
@@ -0,0 +1,344 @@
...
+      instanceIdParam.key -> "dummy",
+      userParam.key       -> "user",
+      passwordParam.key   -> "pass",
+      tableNameParam.key  -> tableName,
+      mockParam.key       -> "true").asJava)
+    ds
+  }
+
+  def loadFeatures(ds: DataStore, sft: SimpleFeatureType, encodedFeatures: Array[_<:Array[_]]): SimpleFeatureStore = {
+    val builder = AvroSimpleFeatureFactory.featureBuilder(sft)
```
cne1x: [picky]  It doesn't really matter, but isn't our default encoding Kryo?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIteratorTest.scala
```diff
@@ -0,0 +1,344 @@
...
+    q.getHints.put(QueryHints.MAP_AGGREGATION_KEY, "map")
+    q
+  }
+
+  "MapAggregatingIterator" should {
+    val spec = "id:Integer,map:Map[String,Integer],dtg:Date,geom:Geometry:srid=4326"
+    val typeName = "test"
+    val sft = SimpleFeatureTypes.createType(typeName, spec)
+    sft.getUserData.put(Constants.SF_PROPERTY_START_TIME, "dtg")
+    val random = new Random()
```
cne1x: Do we want to specify a seed so that this test is deterministic?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIteratorTest.scala
elahrvivaz: Just a warning, the kryo encoding doesn't play nice with non-sequential unit tests, as the mock connection will return the exact same byte array object for a given key/value pair, which messes up kryo as it tries to manipulate the byte array.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/TemporalDensityIterator.scala
dmcwhorter: i don't think so, this is just to make it so that you can refer to mutable.HashMap instead of collection.mutable.HashMap

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIteratorTest.scala
dmcwhorter: ill try to go to kryo, i think using isolated instead of sequential deals with that problem

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIteratorTest.scala
dmcwhorter: sure

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIterator.scala
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2014 Commonwealth Computer Research, Inc.
```
jahhulbert-ccri: change to 2015...and elsewhere.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -210,6 +213,33 @@ case class QueryPlanner(schema: String,
...
+                                          query: Query,
+                                          returnSFT: SimpleFeatureType,
+                                          decoder: SimpleFeatureDecoder): CloseableIterator[SimpleFeature] = {
+    val aggregateKeyName = query.getHints.get(MAP_AGGREGATION_KEY).asInstanceOf[String]
+
+    val maps = accumuloIterator.map { kv =>
+      decoder
+        .decode(kv.getValue.get)
+        .getAttribute(aggregateKeyName)
+        .asInstanceOf[JMap[AnyRef, Int]]
```
jahhulbert-ccri: question: could we support the other numeric types instead of just int? I'm assuming that Int is the thing for this use case but if it were some other measurement stored in a Map it may be a double or something.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/maps/package.scala
```diff
@@ -0,0 +1,12 @@
+package org.locationtech.geomesa.utils
```
jahhulbert-ccri: license header needed

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/Strategy.scala
```diff
@@ -127,8 +126,8 @@ trait Strategy {
                     geometryToCover: Geometry,
                     schema: String,
                     featureEncoding: FeatureEncoding,
-                    featureType: SimpleFeatureType) = {
-    if (query.getHints.containsKey(DENSITY_KEY)) {
+                    featureType: SimpleFeatureType) = query match {
+    case densityQuery if densityQuery.getHints.containsKey(DENSITY_KEY) =>
```
jahhulbert-ccri: perhaps use ```case _``` similar to syntax here for consistency on the unused name ```densityQuery```:
https://github.com/locationtech/geomesa/pull/476/files#diff-82e60bfcc5552adf547b65dcc6bea863L241
``` case _: Query if query.getHints.containsKey(DENSITY_KEY) =>```

File: geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/DensityIterator.scala [hunk 12]
```diff
@@ -18,173 +18,119 @@
...
   /** for a given polygon, take the centroid of each polygon from the BBOX coverage grid
     * if the given polygon contains the centroid then it is passed on to addResultPoint */
-  def handlePolygon(inPolygon: Polygon) = {
+  def handlePolygon(result: SparseMatrix, inPolygon: Polygon) = {
     val grid = snap.generateCoverageGrid
     val featureIterator = grid.getFeatures.features
     featureIterator
       .filter{ f => inPolygon.intersects(f.polygon) }
-      .foreach{ f=> addResultPoint(f.polygon.getCentroid) }
+      .foreach{ f=> addResultPoint(result, f.polygon.getCentroid) }
```
jahhulbert-ccri: style space after variable: ```f =>```

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/DensityIterator.scala
```diff
@@ -192,28 +138,7 @@ class DensityIterator(other: DensityIterator, env: IteratorEnvironment) extends
...
-
-  def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] = new DensityIterator(this, env)
-
-  def next(): Unit = if(!source.hasTop) {
-    topDensityKey = None
-    topDensityValue = None
-  } else {
-    findTop()
-  }
+  override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] = new DensityIterator(this, env)
```
elahrvivaz: deepCopy should call init on the iterator it's returning. Can't quite tell if that is happening somewhere or not...

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/DensityIterator.scala
```diff
@@ -294,5 +219,12 @@ object DensityIterator extends Logging {
     }
     table
   }
+}
 
+case class DensityIteratorResult(geometry: Geometry,
+                                 densityGrid: SparseMatrix = HashBasedTable.create[Double, Double, Long]()) extends Result {
+  override def addToFeature(featureBuilder: SimpleFeatureBuilder): Unit = {
+    featureBuilder.add(DensityIterator.encodeSparseMatrix(densityGrid))
+    featureBuilder.add(geometry)
```
elahrvivaz: do we need to keep the geometry around here? maybe we could just use a static geom? that way we wouldn't have to keep copying the result up in handleKeyValue:107

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  override def handleKeyValue(resultO: Option[MapAggregatingIteratorResult],
+                              topSourceKey: Key,
+                              topSourceValue: Value): MapAggregatingIteratorResult = {
+
+    val feature = originalDecoder.decode(topSourceValue.get)
+    val currCounts = feature.getAttribute(mapAttribute).asInstanceOf[JMap[AnyRef, Int]].asScala
+
+    val result = resultO.getOrElse(MapAggregatingIteratorResult(mapAttribute))
+    result.copy(countMap = sumIntValueMaps(Seq(result.countMap, currCounts)))
```
elahrvivaz: I would use a mutable map here and not copy it each time.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] = new MapAggregatingIterator(this, env)
```
elahrvivaz: as above, deepCopy should call init on the the iter being returned.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/TemporalDensityIterator.scala
```diff
@@ -30,125 +28,62 @@ import org.apache.commons.codec.binary.Base64
...
   /** take a given Coordinate and add 1 to the result time that it corresponds to via the snap time */
-  def addResultDate(date: DateTime) = {
+  def addResultDate(date: DateTime, result: TimeSeries): Unit = {
     val t: DateTime = snap.t(snap.i(date))
-    val cur: Long = result.get(t).getOrElse(0L)
+    val cur: Long = result.getOrElse(t, 0L)
     result.put(t, cur + 1L)
   }
 
   override def seek(range: ARange,
```
elahrvivaz: shouldn't need this method anymore as it's implemented in the super class, right?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/TemporalDensityIterator.scala
```diff
@@ -30,125 +28,62 @@ import org.apache.commons.codec.binary.Base64
...
     findTop()
   }
 
-  def hasTop: Boolean = topTemporalDensityKey.nonEmpty
-
-  def getTopKey: Key = topTemporalDensityKey.orNull
-
-  def getTopValue = topTemporalDensityValue.orNull
-
   def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] = new TemporalDensityIterator(this, env)
```
elahrvivaz: ditto init in deepCopy

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/MapAggregatingIteratorTest.scala
```diff
@@ -0,0 +1,346 @@
...
... 38.0478368848681,-78.4699356469289 38.0472075403136,-78.4698289155783 38.0470086499745,-78.4695339345729 38.0468371439102,-78.4693053388149 38.0469393179596,-78.468407168336 38.0466034706965,-78.4676917660222 38.0462473734836,-78.4661192632977 38.045714987292,-78.4653292114835 38.047182785812,-78.4651074150484 38.048250382979,-78.4647453057632 38.0482502765916,-78.4642891032747 38.048663190151,-78.4639468173163 38.0492436680877,-78.4638729309242 38.0495125312664,-78.4573545185689 38.0497324748685,-78.4570999398137 38.0492609003817,-78.4571048708452 38.0499860033047,-78.4572469354638 38.0504060779264,-78.458898424969 38.0510299931243,-78.4567841773194 38.0556302795318,-78.4561273995112 38.0553332011668,-78.4557521877327 38.0549991258425,-78.4555445617933 38.0547458706718,-78.4552967746776 38.0545217622445,-78.4550136860866 38.0541576231591,-78.4547304813727 38.053989517875,-78.4542703578379 38.053569291964,-78.454093163388 38.053765231087,-78.4543764171386 38.0538493662526,-78.4544118048045 38.0539053700334,-78.4546595354771 38.0541574891887,-78.4541635943456 38.0544373334645,-78.4548714741119 38.0550816694308,-78.4492025035279 38.0589437591411,-78.4480347940116 38.0574871002506))"
```
elahrvivaz: I forget, is it ok for us to reference stuff like cville?

jnh5y: I think so.  My push-back would be to factor this out and only have it one place in the test code base.  

*shrugs*  No big deal; either way is fine.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/maps/package.scala
```diff
@@ -0,0 +1,12 @@
...
+
+package object maps {
+
+  def aggregateMaps[K,V](maps: TraversableOnce[collection.Map[K,V]], operation: (V,V) => V, emptyVal: V): Map[K,V] =
+    maps.flatMap(_.toSeq).foldLeft(Map[K,V]()){ case (mapSoFar, (k, v)) =>
+      mapSoFar.updated(k, operation(mapSoFar.getOrElse(k, emptyVal), v))
+    }
```
elahrvivaz: I think it would be more efficient to use a mutable map and add everything to it. Seems like this could create a lot of intermediate maps.

elahrvivaz:
I've looked through and it looks solid. I had some performance suggestions to use mutable state instead of copying immutable state.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/maps/package.scala
dmcwhorter: My general impression/understanding of immutable data structures in scala is that when you create a "new" immutable map, etc, by copying an old one, the new one internally reuses most of the data structures from the old one, only changing the parts of the map where data is different.  That being said, I don't have a good feel for how to tell when it is actually more efficient to use a mutable map, so I'll go ahead and change it here and above.

elahrvivaz: ah, maybe others can chime in...

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/DensityIterator.scala
dmcwhorter: so, this is the way the DensityIterator was working before -- returning the geometry from the last feature, I am not sure if this is an essential part of its behavior or not.  Can anyone who knows weigh in? (@jnh5y?)

elahrvivaz: Looking at org.locationtech.geomesa.core.index.QueryPlanner#adaptIteratorForDensityQuery and org.locationtech.geomesa.core.iterators.DensityIterator#expandFeature, it seems to be discarded except in a failure case where it is the backup geometry returned.

jnh5y: The one hinky thing about this procedure is that we have to return *some* geometry.  Which one is more or less irrelevant.

dmcwhorter:
should be good to go

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -210,6 +213,33 @@ case class QueryPlanner(schema: String,
...
+        .asScala
+    }
+
+    if(maps.nonEmpty) {
+      val reducedMap = sumNumericValueMutableMaps(maps.toIterable).toMap // to immutable map
```
jnh5y: I'm trying to figure out the .toIterable and the implicit conversion happen with the calls to Iterator below.  

I think there may be a little bit of clean-up we can do here.

dmcwhorter: ok, i messed with it for a while and came up with this, but if you've got something better feel free

dmcwhorter:
this p/r should be good to go, if you need me to do something to set the signed-off bit on the commit let me know, i am not sure if i should rewrite the history given the comments and follow-on commits on the p/r

elahrvivaz:
d8f228d754814591ca4bc0c2bb1210bbf8e9d2c2

PR #481, opened 2015-02-24
GEOMESA-640 Adding support for LineStrings in bin output

elahrvivaz:
* When exporting a LineString, the 'dtg' field that is specified in the parameters should point to a List[Date] field, where each point in the LineString corresponds to a value in the list.

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
@@ -1,0 +1,229 @@
+/*
+ * Copyright 2015 Commonwealth Computer Research, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.locationtech.geomesa.filter.function
+
+import java.io.OutputStream
+import java.util.Date
+
+import com.typesafe.scalalogging.slf4j.Logging
+import com.vividsolutions.jts.geom.{Geometry, LineString}
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.locationtech.geomesa.utils.geotools.Conversions.RichSimpleFeatureIterator
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.ListAttributeSpec
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+
+import scala.collection.JavaConversions._
+import scala.collection.immutable.TreeSet
+
+
+object BinaryOutputEncoder extends Logging {
+
+  import org.locationtech.geomesa.filter.function.AxisOrder._
+
+  case class ValuesToEncode(lat: Float, lon: Float, dtg: Long, track: Option[String], label: Option[String])
+
+  implicit val ordering = new Ordering[ValuesToEncode]() {
+    override def compare(x: ValuesToEncode, y: ValuesToEncode) = x.dtg.compareTo(y.dtg)
+  }
+
+  /**
+   * Encodes a feature collection to bin format
+   *
+   * @param fc
+   * @param output
+   * @param dtgField
+   * @param trackIdField
+   * @param labelField
+   * @param latLon
+   * @param axisOrder
+   * @param sort
+   */
+  def encodeFeatureCollection(
+      fc: SimpleFeatureCollection,
+      output: OutputStream,
+      dtgField: String,
+      trackIdField: Option[String],
+      labelField: Option[String] = None,
+      latLon: Option[(String, String)] = None,
+      axisOrder: AxisOrder = LatLon,
+      sort: Boolean = false) = {
+
+    val sft = fc.getSchema
+    val isLineString = sft.getGeometryDescriptor.getType.getBinding == classOf[LineString]
+
+    // validate our input data
+    validateDateAttribute(dtgField, sft, isLineString)
+    validateLabels(trackIdField, labelField, sft)
+    validateLatLong(latLon, sft)
+
+    val sysTime = System.currentTimeMillis
+    val dtgIndex = sft.indexOf(dtgField)
+
+    // get date function based
+    val getDtg: (SimpleFeature) => Long = (f) => {
+      val date = f.getAttribute(dtgIndex).asInstanceOf[Date]
+      if (date == null) sysTime else date.getTime
+    }
+    // for line strings, we need an array of dates corresponding to the points in the line
+    val getLineDtg: (SimpleFeature) => Array[Long] = (f) => {
+      val dates = f.getAttribute(dtgIndex).asInstanceOf[java.util.List[Date]]
+      if (dates == null) Array.empty else dates.map(_.getTime).toArray
+    }
+
+    // get lat/lon as floats
+    val getLatLon: (SimpleFeature) => (Float, Float) =
+      latLon.map { case (lat, lon) =>
+        // if requested, export arbitrary fields as lat/lon
+        val latIndex = sft.indexOf(lat)
+        val lonIndex = sft.indexOf(lon)
+        val binding = sft.getType(latIndex).getBinding
+        if (binding == classOf[java.lang.Float]) {
+          (f: SimpleFeature) =>
+            (f.getAttribute(latIndex).asInstanceOf[Float], f.getAttribute(lonIndex).asInstanceOf[Float])
+        } else if (binding == classOf[java.lang.Double]) {
+          (f: SimpleFeature) =>
+            (f.getAttribute(latIndex).asInstanceOf[Double].toFloat, f.getAttribute(lonIndex).asInstanceOf[Double].toFloat)
+        } else {
+          (f: SimpleFeature) =>
+            (f.getAttribute(latIndex).toString.toFloat, f.getAttribute(lonIndex).toString.toFloat)
+        }
+      }.getOrElse {
+        // default is to use the geometry
+        // depending on srs requested and wfs versions, axis order can be flipped
+        axisOrder match {
+          case LatLon => (f) => {
+            val geom = f.getDefaultGeometry.asInstanceOf[Geometry].getInteriorPoint
+            (geom.getX.toFloat, geom.getY.toFloat)
+          }
+          case LonLat => (f) => {
+            val geom = f.getDefaultGeometry.asInstanceOf[Geometry].getInteriorPoint
+            (geom.getY.toFloat, geom.getX.toFloat)
+          }
+        }
+      }
+    // for linestrings, we return each point - use an array so we get constant-time lookup
+    // depending on srs requested and wfs versions, axis order can be flipped
+    val getLineLatLon: (SimpleFeature) => Array[(Float, Float)] = axisOrder match {
+      case LatLon => (f) => {
+        val geom = f.getDefaultGeometry.asInstanceOf[LineString]
+        val points = (0 until geom.getNumPoints).map(geom.getPointN(_))
+        points.map(p => (p.getX.toFloat, p.getY.toFloat)).toArray
+      }
+      case LonLat => (f) => {
+        val geom = f.getDefaultGeometry.asInstanceOf[LineString]
+        val points = (0 until geom.getNumPoints).map(geom.getPointN(_))
+        points.map(p => (p.getY.toFloat, p.getX.toFloat)).toArray
+      }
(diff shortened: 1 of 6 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,229 @@
...
+      }
+      case None => (v) => {
+        val toEncode = BasicValues(v.lat, v.lon, v.dtg, v.track)
+        Convert2ViewerFunction.encode(toEncode, output)
+      }
+    }
+
+    val iter = if (isLineString) {
+      // expand the line string into individual points to encode
+      new RichSimpleFeatureIterator(fc.features()).flatMap { sf =>
```
jnh5y: Any reason to prefer calling the wrapper directly?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,229 @@
...
+        val dtg = getDtg(sf)
+        val trackId = getTrackId(sf)
+        val label = getLabel(sf)
+        ValuesToEncode(lat, lon, dtg, trackId, label)
+      }
+    }
+    if (sort) {
+      val set = new scala.collection.mutable.TreeSet[ValuesToEncode]()
+      iter.foreach(set.add)
+      set.foreach(encode)
```
jnh5y: (idly) I wonder if this is the best way to sort.

geomesa-plugin/src/test/scala/org/locationtech/geomesa/plugin/wfs/output/BinaryViewerOutputFormatTest.scala
```diff
@@ -17,7 +17,15 @@
...
 
+import java.io.ByteArrayOutputStream
+import java.text.SimpleDateFormat
+
+import org.geotools.data.collection.ListFeatureCollection
+import org.geotools.feature.simple.SimpleFeatureBuilder
 import org.junit.runner.RunWith
+import org.locationtech.geomesa.filter.function.{BasicValues, Convert2ViewerFunction, ExtendedValues}
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.locationtech.geomesa.utils.text.WKTUtils
```
jnh5y: Did the import really change for this file?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/ExportCommand.scala
```diff
@@ -56,6 +56,9 @@ class ExportCommand(parent: JCommander) extends CommandWithCatalog(parent) with
       exporter.flush()
       exporter.close()
     }
+
+    //necessary to avoid StatsWriter exception when exporting
+    Thread.sleep(1000)
```
jnh5y: Which exception does this avoid?  Maybe something with interrupting Zookeeper?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,229 @@
...
+
+import scala.collection.JavaConversions._
+import scala.collection.immutable.TreeSet
+
+
+object BinaryOutputEncoder extends Logging {
+
+  import org.locationtech.geomesa.filter.function.AxisOrder._
+
+  case class ValuesToEncode(lat: Float, lon: Float, dtg: Long, track: Option[String], label: Option[String])
```
jnh5y: Floats, not doubles?  Is there a preference to write fewer bytes (since the precision is likely unnecessary)?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,229 @@
...
+   */
+  def encodeFeatureCollection(
+      fc: SimpleFeatureCollection,
+      output: OutputStream,
+      dtgField: String,
+      trackIdField: Option[String],
+      labelField: Option[String] = None,
+      latLon: Option[(String, String)] = None,
+      axisOrder: AxisOrder = LatLon,
+      sort: Boolean = false) = {
```
jnh5y: It took me a few minutes to figure out the type...  Might be worth marking as unit since the function is so long.  Also, it may make sense to add a line to the comment block indicating that it writes directly to the output stream.

jnh5y:
Looks good.  Most comments are minor.

Generally, does this PR interact with the Kafka DataStore?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: This is a legacy format that we already have viewers for - I believe you are correct about the preference for fewer bytes over precision.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: will do.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: nope... will fix.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: Probably not... http://stackoverflow.com/questions/3607593/is-it-faster-to-add-to-a-collection-then-sort-it-or-add-to-a-sorted-collection
I'll change.

geomesa-plugin/src/test/scala/org/locationtech/geomesa/plugin/wfs/output/BinaryViewerOutputFormatTest.scala
elahrvivaz: Fixed.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/ExportCommand.scala
elahrvivaz: Yeah, if we don't sleep, zookeeper will log an interrupted error that we can't catch. I'll update the comment.

elahrvivaz:
It's an output format, so you should be able to use it with any data store. But specifically it doesn't have any ties to the kafka ds.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,226 @@
...
+    val sysTime = System.currentTimeMillis
+    val dtgIndex = sft.indexOf(dtgField)
+
+    // get date function based
+    val getDtg: (SimpleFeature) => Long = (f) => {
+      val date = f.getAttribute(dtgIndex).asInstanceOf[Date]
+      if (date == null) sysTime else date.getTime
+    }
+    // for line strings, we need an array of dates corresponding to the points in the line
+    val getLineDtg: (SimpleFeature) => Array[Long] = (f) => {
```
apc3n: shouldn't the type signature reflect that the SimpleFeature should be a line feature then? Not too familiar with the API

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,226 @@
...
+      val dates = f.getAttribute(dtgIndex).asInstanceOf[java.util.List[Date]]
+      if (dates == null) Array.empty else dates.map(_.getTime).toArray
```
apc3n: why not an Array of sysTime like the other function?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    // get lat/lon as floats
+    val getLatLon: (SimpleFeature) => (Float, Float) =
+      latLon.map { case (lat, lon) =>
+        // if requested, export arbitrary fields as lat/lon
+        val latIndex = sft.indexOf(lat)
+        val lonIndex = sft.indexOf(lon)
+        val binding = sft.getType(latIndex).getBinding
+        if (binding == classOf[java.lang.Float]) {
```
apc3n: does isInstanceOf not work here?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: There's no way to have a type signature for SimpleFeature with a LineString. SimpleFeature is pretty much just a map of String->Any.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: binding is a Class[T], we'd have to use a classTag or something to check the sub-type. We could check each simple feature as it comes through, but in this case we can do 1 check up front instead of 1 per feature.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,226 @@
...
+      }
+      case LonLat => (f) => {
+        val geom = f.getDefaultGeometry.asInstanceOf[LineString]
+        val points = (0 until geom.getNumPoints).map(geom.getPointN(_))
+        points.map(p => (p.getY.toFloat, p.getX.toFloat)).toArray
+      }
+    }
+
+    // gets the track id from a feature
+    val getTrackId: (SimpleFeature) => Option[String] = trackIdField match {
```
apc3n: not sure I have a huge opinion on it... but you can convert all these functions that use a match in them to PartialFunctions and get rid of the match. Your choice

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: This output format is designed to visualized points by time, so putting all points at the same time wouldn't be useful. Having no date(s) and using systime is a failure condition. They will get skipped them down on line 170.
The other function above uses systime just as an easy error handling condition - it's not consistent between the linestrings and points, as you pointed out, but Longs can't be null so we need some kind of default value.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
elahrvivaz: hmm, I'm still not that familiar with partial functions... this is actually a match that results in a function... does that still apply?

apc3n: ah nvm these are matches that generate functions. You're good. I think a match statement actually creates an anonymous partial function just to let you know. So if you do i match { case i: Int i + 1} what you are actually doing is val anonoymousFunction = PartialFunction[Int, Int] { case i: Int => i + }; anonoymousFunction(i)}

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -0,0 +1,226 @@
...
+          (f: SimpleFeature) =>
+            (f.getAttribute(latIndex).toString.toFloat, f.getAttribute(lonIndex).toString.toFloat)
+        }
+      }.getOrElse {
+        // default is to use the geometry
+        // depending on srs requested and wfs versions, axis order can be flipped
+        axisOrder match {
+          case LatLon => (f) => {
+            val geom = f.getDefaultGeometry.asInstanceOf[Geometry].getInteriorPoint
+            (geom.getX.toFloat, geom.getY.toFloat)
```
apc3n: maybe extract function: 

  val pointToXY: Point => (Float, Float) = 
      p => (p.getX.toFloat, p.getY.toFloat)

and use thoroughout. For the YX case you can use the same function and then call swap on the tuple

elahrvivaz: fixed.

jnh5y:
Committed as f12e539b1a416c592154900f77a28ec7ada9f0a2 on 1.5.

PR #483, opened 2015-02-25
GEOMESA-671 Add check for Single Raster Iterator in Mosaic Function

AndrewAnnex:
* Mosaic code now checks for (and handles) the case when only one raster is in the iterator.


jnh5y:
Committed as f109dd1c75bd7157a3d84e02d767c5a6e48ed95a on f_wcs_4.

PR #487, opened 2015-03-02
GEOMESA-704 Get data into cards via getFeatureInfo request.

kevinwheeler:


geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -1,39 +1,43 @@
...
-                    zoom: 8,
-                    maxZoom: 18,
-                    minZoom: 3,
-                    attributionControl: false,
-                    layers: [baseLayer, wmsLayer]
-                });
+            link: function (scope, element, attrs) {
+                scope.cards = [
+                    {
+                        Start: ' Click a data point to view it\'s attributes.'
```
jnh5y: (Picky) "it's" means "it is".  You want "its" which possessive.

geomesa-ui/src/main/ui/Gruntfile.js
```diff
@@ -56,6 +56,14 @@ module.exports = function (grunt) {
                 expand: true
             }]
         },
+        build_vendorcss: {
```
mforkin: We don't want to include vendor CSS files in this way anymore. Instead add the necessary files to the main.styl file

mforkin: this whole task can be taken out

geomesa-ui/src/main/ui/geomesa.config.js
```diff
@@ -45,10 +45,11 @@ module.exports = {
             'vendor/bower/moment-timezone/builds/moment-timezone-with-data-2010-2020.min.js',
             'vendor/bower/leaflet-dist/leaflet.js',
             'vendor/bower/leaflet-providers/leaflet-providers.js',
+            'vendor/bower/openlayers/build/ol.js',
             // TODO - test only
             'vendor/bower/angular-mocks/angular-mocks.js',
         ],
-        css: [],
+        css: ['vendor/bower/openlayers/css/ol.css'],
```
mforkin: remove this whole option so people aren't confused in the future

mforkin: maybe add a comment to point them to main.styl for vender css files

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -1,39 +1,43 @@
...
 
-    .directive('geomesaMap', [function () {
+    .directive('geomesaMap', ['$http', function ($http) {
+
         return {
             restrict: 'E',
             scope: {
                 map: '=?',
                 api: '=',
+                cards: '=',
```
mforkin: I think that instead of calling this cards we should just call it selectedPoints or something like that to keep it general. Most people won't know what a map "card" is.

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -1,39 +1,43 @@
...
+                    }
+                ];
 
-                scope.map.on('click', function (evt) {
-                    scope.$apply(function () {
-                        scope.selectedPoint = {
-                            lat: evt.latlng.lat,
-                            lng: evt.latlng.lng
-                        };
+                var baseLayer = new ol.layer.Tile({
```
mforkin: define all variables on first line

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -42,6 +46,20 @@ angular.module('geomesa.map', [])
                     }
                 };
 
+                scope.map.on('singleclick', function(evt) {
+                    var viewResolution = olView.getResolution();
```
mforkin: combine variables into single line

mforkin: should only have one "var" keywords per scope and it should be on the first line.

geomesa-ui/src/main/ui/src/app/map/map.js
mforkin: What about making these things configurable? Maybe that is a separate PR

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -42,6 +46,20 @@ angular.module('geomesa.map', [])
...
+                    var url = wmsSource.getGetFeatureInfoUrl(
+                        evt.coordinate, viewResolution, 'EPSG:3857',
+                        {'INFO_FORMAT': 'application/json', FEATURE_COUNT: 50}
+                    );
+                    $http.get(url).success(function(data, status, headers, config) {
```
mforkin: Does it make sense to separate out this logic into a service. There is not much going on here currently, but it seems like some of this could be organized separately.

mforkin: Perhaps waiting until the next feature to refactor would make sense, then we could get a better idea of the abstraction necessary.

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -1,39 +1,43 @@
...
+                        wmsLayer = new ol.layer.Tile({
+                        source: wmsSource
+                    }),
+                    olView = new ol.View({
+                        center: ol.proj.transform([37.41, 8.82], 'EPSG:4326', 'EPSG:3857'),
+                        zoom : 4,
+                        maxResolution : 40075016.68557849 / screen.width,
                     });
+
+                scope.map = new ol.Map({
```
mforkin: did we discuss setting the maxExtent to prevent repeating tiles on far zoom out

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -42,6 +46,20 @@ angular.module('geomesa.map', [])
                     }
                 };
 
+                scope.map.on('singleclick', function(evt) {
+                    var viewResolution = olView.getResolution();
+                    var url = wmsSource.getGetFeatureInfoUrl(
+                        evt.coordinate, viewResolution, 'EPSG:3857',
+                        {'INFO_FORMAT': 'application/json', FEATURE_COUNT: 50}
```
jdk2pq: Are these params correct? Should this be configurable?

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -1,39 +1,43 @@
...
+                        url: 'http://geomesa:8080/geoserver/geomesa/wms',
+                        params: {LAYERS: 'QuickStart'}
+                    }),
+                        wmsLayer = new ol.layer.Tile({
+                        source: wmsSource
+                    }),
+                    olView = new ol.View({
+                        center: ol.proj.transform([37.41, 8.82], 'EPSG:4326', 'EPSG:3857'),
+                        zoom : 4,
+                        maxResolution : 40075016.68557849 / screen.width,
```
jdk2pq: Would be nice to have these be configurable.

geomesa-ui/src/main/ui/bower.json
```diff
@@ -8,12 +8,11 @@
     "angular-resource": "~1.2.13",
     "bootstrap": "~3.1.1",
     "d3": "~3.4.2",
-    "jquery": "~2.1.0",
     "font-awesome": "~4.0.3",
+    "jquery": "~2.1.0",
```
jdk2pq: Did moving this dependency down change anything?

geomesa-ui/src/main/ui/geomesa.config.js
mforkin: Also remove the leaflet css from the main.styl file

geomesa-ui/src/main/ui/bower.json
```diff
@@ -8,12 +8,11 @@
...
     "lodash": "~2.4.1",
     "moment": "~2.7.0",
     "moment-timezone": "~0.1.0",
-    "leaflet-dist": "~0.7.2",
-    "leaflet-providers": "~1.0.1"
+    "openlayers" : "https://github.com/openlayers/ol3/releases/download/v3.2.1/v3.2.1.zip"
```
mforkin: this is perfectly fine, but I think you can point to something like "https://github.com/openlayers/ol3#v3.2.1"  or "https://github.com/openlayers/ol3#v3.2.x" (not sure if those work, but something like that should). though I am thinking that we should remove all the "~" from above and just depend on a specific version of everything.

geomesa-ui/src/main/ui/bower.json
kevinwheeler: I figure it should be organized alphabetically.

jdk2pq: Nice. I definitely agree.

geomesa-ui/src/main/ui/bower.json
kevinwheeler: The open layers repo doesn't seem to include the built artifacts, only source code, however, they are included in the zip files that can be found with their releases (IE there are download links to the zip files here https://github.com/openlayers/ol3/releases). Do you know of a way to get at these zip files in the manner you were speaking of?

mforkin: Thought I saw it in there, but I was mistaken. Keeping this line is fine.

geomesa-ui/src/main/ui/src/app/map/map.js
kevinwheeler: I had originally only used 'maxResolution' to get this effect, and just now added 'extent' to the tile layers because it was being buggy when the user would slide the map off the page such that they would get a new copy of the map rotated in from the other side. Just as a small note, we are using open layers 3 and there is no maxExtent in map anymore.

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -1,39 +1,47 @@
...
+                ];
 
-                scope.map.on('click', function (evt) {
-                    scope.$apply(function () {
-                        scope.selectedPoint = {
-                            lat: evt.latlng.lat,
-                            lng: evt.latlng.lng
-                        };
+                var projection = ol.proj.get('EPSG:3857'),
+                    extent = projection.getExtent(),
```
kevinwheeler: I definitely agree that these things should be configurable. I am going to work on the temporal density iterator for a bit and will get back to you on this. Some initial thoughts would be, the only two sources of configuration I can think of would be to either get configuration info passed to us in the scope, or to get it via a service. Services are singletons and this would seem to restrict us to only one map instance. Who knows, maybe some day we will want to have two maps side by side so we can compare the two. So maybe we should get configuration info via the scope. If we get configuration info via the scope, I guess we'll need to decide what goes in the link function vs what we get passed to us via the scope. Should we just get everything from the scope and have an empty link function? I need to take some time to think about all of this, but I thought I'd let you know where my thoughts are starting.

geomesa-ui/src/main/ui/src/app/map/map.js
```diff
@@ -42,6 +50,20 @@ angular.module('geomesa.map', [])
                     }
                 };
 
+                scope.map.on('singleclick', function(evt) {
```
kevinwheeler: test

jdk2pq:
Committed as 2a4a1348810dfad50dd0e6fa0f8c30bbb516115f

PR #488, opened 2015-03-03
GEOMESA-521 Remove objenesis from main pom

jdk2pq:
Excluded objenesis from kryo and removed objenesis dependency from main pom. Now is only included as a transitive dependency from specs2


AndrewAnnex:
:+1:

jnh5y:
Committed as c1cd649a89c58850a9d5e06ef7f8d8e4af254efd on 1.5.

PR #489, opened 2015-03-03
GEOMESA-705 Don't encode null attributes in the index

elahrvivaz:


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/tables/AttributeTable.scala
```diff
@@ -164,20 +167,22 @@ object AttributeTable extends GeoMesaTable with Logging {
...
-  def encode(valueOption: Option[Any], descriptor: AttributeDescriptor): Seq[String] = {
-    val value = valueOption.getOrElse(nullString)
-    if (descriptor.isCollection) {
+  def encode(value: Any, descriptor: AttributeDescriptor): Seq[String] = {
+    if (value == null) {
+      Seq.empty
+    } else if (descriptor.isCollection) {
       // encode each value into a separate row
-      value.asInstanceOf[JCollection[_]].asScala.toSeq.map(Option(_).getOrElse(nullString)).map(typeEncode)
+      value.asInstanceOf[JCollection[_]].toSeq.flatMap(Option(_).map(typeEncode).filterNot(_.isEmpty))
```
apc3n: what about filter(_.nonEmpty)

elahrvivaz: I feel like there is some minor cost involved in the implicit conversion to a GenTraversableOnce with .nonEmpty, but I can't really confirm that... Any thoughts?

apc3n: oh yeah gotcha. I didn't see that it was a JCollection. Yeah avoid the conversion and leave as it

apc3n: nvm again... yeah I see that it is a string.. It's fine the way it is

jdk2pq:
Read through this -- looks good. Built, and all went as expected.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/process/unique/UniqueProcess.scala
```diff
@@ -233,13 +233,11 @@ object AttributeVisitor {
...
   /**
    * Returns a filter that is equivalent to Filter.INCLUDE, but against the attribute index.
-   * Excludes null values, since we don't care about those anyway.
    *
    * @param attribute
    * @return
    */
-  def getIncludeAttributeFilter(attribute: String) =
-    ff.greater(ff.property(attribute), ff.literal(AttributeTable.nullString))
+  def getIncludeAttributeFilter(attribute: String) = ff.greater(ff.property(attribute), ff.literal(""))
```
jnh5y: Quick question: is "" the 'least' literal/string/lexicoded value?

elahrvivaz: yeah, it ends up acting as a 'prefix' function, e.g. if the rows are:
<sft><attr-name><encoded value>
we create
<sft><attr-name>
which will sort before any encoded value (empty strings are handled as prefixes on AttributeTable#202 and then AttributeIdxStrategy#379)

jnh5y:
Merged as 8af364710c120eb3d979a164b41f5f2964c8caf0 on 1.5.

jnh5y:
Actually, d6882306cccde71a4b011f0ba8253a717f478bb3.

PR #493, opened 2015-03-06
GEOMESA-479 - adding covering indices for attributes

elahrvivaz:
A 'covering' index means that we store the entire simple feature in the index, and thus we won't do a join against the record table.
You can now indicate covered indices in the SFT spec by using the notation:
'attr:String:index=full'
options are:
full -> fully covering index
join -> partial index, most queries will require a join on the record table
none -> not attribute indexed
true/false -> legacy fallbacks equivalent to join/none

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/core.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/core.scala
@@ -42,6 +42,7 @@ package object core {
   val GEOMESA_ITERATORS_SFT_NAME            = "geomesa.iterators.sft-name"
   val GEOMESA_ITERATORS_SFT_INDEX_VALUE     = "geomesa.iterators.sft.index-value-schema"
   val GEOMESA_ITERATORS_ATTRIBUTE_NAME      = "geomesa.iterators.attribute.name"
+  val GEOMESA_ITERATORS_ATTRIBUTE_COVERAGE  = "geomesa.iterators.attribute.coverage"
   val GEOMESA_ITERATORS_ECQL_FILTER         = "geomesa.iterators.ecql-filter"
   val GEOMESA_ITERATORS_TRANSFORM           = "geomesa.iterators.transform"
   val GEOMESA_ITERATORS_TRANSFORM_SCHEMA    = "geomesa.iterators.transform.schema"
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/core.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/core.scala
@@ -42,7 +42,7 @@ package object core {
   val GEOMESA_ITERATORS_SFT_NAME            = "geomesa.iterators.sft-name"
   val GEOMESA_ITERATORS_SFT_INDEX_VALUE     = "geomesa.iterators.sft.index-value-schema"
   val GEOMESA_ITERATORS_ATTRIBUTE_NAME      = "geomesa.iterators.attribute.name"
-  val GEOMESA_ITERATORS_ATTRIBUTE_COVERAGE  = "geomesa.iterators.attribute.coverage"
+  val GEOMESA_ITERATORS_ATTRIBUTE_COVERED   = "geomesa.iterators.attribute.covered"
   val GEOMESA_ITERATORS_ECQL_FILTER         = "geomesa.iterators.ecql-filter"
   val GEOMESA_ITERATORS_TRANSFORM           = "geomesa.iterators.transform"
   val GEOMESA_ITERATORS_TRANSFORM_SCHEMA    = "geomesa.iterators.transform.schema"
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIterator.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIterator.scala
@@ -16,11 +16,15 @@
 
 package org.locationtech.geomesa.core.iterators
 
+import java.util.Date
+
 import com.typesafe.scalalogging.slf4j.Logging
+import com.vividsolutions.jts.geom.Geometry
 import org.apache.accumulo.core.data._
 import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
 import org.locationtech.geomesa.core._
 import org.locationtech.geomesa.core.data.tables.AttributeTable._
+import org.locationtech.geomesa.utils.stats.IndexCoverage
 import org.opengis.feature.`type`.AttributeDescriptor
 
 import scala.util.{Failure, Success}
@@ -38,12 +42,16 @@ class AttributeIndexIterator
     with HasIndexValueDecoder
     with HasFeatureDecoder
     with HasSpatioTemporalFilter
+    with HasEcqlFilter
     with HasTransforms
     with Logging {
 
   // the following fields get filled in during init
   var attributeRowPrefix: String = null
   var attributeType: Option[AttributeDescriptor] = null
+  var dtgIndex: Option[Int] = None
+
+  var setTopFunction: () => Unit = null
 
   override def init(source: SortedKeyValueIterator[Key, Value],
                     options: java.util.Map[String, String],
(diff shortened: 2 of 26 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIterator.scala
```diff
@@ -56,16 +64,44 @@ class AttributeIndexIterator
...
-  override def setTopConditionally() {
+  override def setTopConditionally() = setTopFunction()
+
+  /**
+   * Each value is the fully encoded simple feature
+   */
+  def setTopFullCoverage(): Unit = {
+    val dataValue = source.getTopValue
+    val sf = featureDecoder.decode(dataValue.get)
+    val meetsStFilter = stFilter.forall(fn => fn(sf.getDefaultGeometry.asInstanceOf[Geometry],
```
apc3n: could you extract the arguments to the functions and give them a name? It looks like they'll be the same for each function "fn"

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorTrigger.scala
```diff
@@ -94,25 +95,32 @@ object IteratorTrigger extends Logging {
    */
   def useIndexOnlyIterator(ecqlPredicate: Option[Filter],
                            query: Query,
-                           sourceSFT: SimpleFeatureType,
+                           sft: SimpleFeatureType,
                            indexedAttribute: Option[String] = None): Boolean = {
+    if (useDensityIterator(query: Query)) {
```
apc3n: what's up with the argument to useDenesityIterator... I feel like I'm reading this wrong or something

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -94,13 +100,78 @@ object Conversions {
...
+
+    def getMapTypes(): Option[(Class[_], Class[_])] = for {
+      keyClass   <- Option(ad.getUserData.get(USER_DATA_MAP_KEY_TYPE))
+      valueClass <- Option(ad.getUserData.get(USER_DATA_MAP_VALUE_TYPE))
+    } yield {
+      (keyClass.asInstanceOf[Class[_]], valueClass.asInstanceOf[Class[_]])
+    }
+
+    def isIndexed = getIndexCoverage() match {
+      case IndexCoverage.FULL | IndexCoverage.JOIN => true
```
apc3n: does | do OR in case statements?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIterator.scala
elahrvivaz: the reason I didn't was b/c stFilter is an option, so it will only be evaluated once, and I wanted to avoid the overhead of creating the vals outside the option if the option was None. I could make them lazy, but there's still some overhead involved...
These methods have actually been completely replaced in some pending work I haven't put up for PRs yet.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorTrigger.scala
elahrvivaz: Oh, yeah I didn't mean to do that -  I'll fix it. It is technically valid, I guess, but you're right it looks weird.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: yeah, you just can't bind to a variable if you use it:
http://stackoverflow.com/questions/1837754/match-multiple-cases-classes-in-scala

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/ScaldingDelimitedIngestJobTest.scala
```diff
@@ -41,7 +41,7 @@ import scala.util.Try
 
 @RunWith(classOf[JUnitRunner])
 class ScaldingDelimitedIngestJobTest extends Specification{
-  isolated
+  sequential
```
elahrvivaz: this fixes the stack trace spam in the unit tests.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIterator.scala
elahrvivaz: PR that replaces this part is up: https://github.com/locationtech/geomesa/pull/496

geomesa-core/src/main/scala/org/locationtech/geomesa/core/core.scala
```diff
@@ -42,6 +42,7 @@ package object core {
   val GEOMESA_ITERATORS_SFT_NAME            = "geomesa.iterators.sft-name"
   val GEOMESA_ITERATORS_SFT_INDEX_VALUE     = "geomesa.iterators.sft.index-value-schema"
   val GEOMESA_ITERATORS_ATTRIBUTE_NAME      = "geomesa.iterators.attribute.name"
+  val GEOMESA_ITERATORS_ATTRIBUTE_COVERAGE  = "geomesa.iterators.attribute.coverage"
```
anthonyccri: Rename this ```geomesa.iterators.attribute.covered```

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorTrigger.scala
```diff
@@ -94,25 +95,32 @@ object IteratorTrigger extends Logging {
    */
   def useIndexOnlyIterator(ecqlPredicate: Option[Filter],
                            query: Query,
-                           sourceSFT: SimpleFeatureType,
+                           sft: SimpleFeatureType,
                            indexedAttribute: Option[String] = None): Boolean = {
+    if (useDensityIterator(query)) {
+      // the Density Iterator is run in place of the SFFI. If it is requested we keep the SFFI
+      // config in the stack, and do NOT run the IndexIterator.
+      return false
```
anthonyccri: [style nitpick] Don't use ```return``` keyword.

```scala
if(useDensityIteratory(query)) true
else if(indexedAttribute.exists(a => sft.getDescriptor(a).getIndexCoverage() == IndexCoverage.FULL))
   false
else {

}
```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -144,46 +121,49 @@ object SimpleFeatureTypes {
...
-          ad.getUserData.get(USER_DATA_LIST_TYPE).asInstanceOf[Class[_]],
-          ad.getUserData.getOrElse(OPT_INDEX, false).asInstanceOf[Boolean],
-          getCardinality(ad)
+          ad.getCollectionType().get,
+          ad.getIndexCoverage(),
+          ad.getCardinality()
         )
 
       case t if t.getBinding.equals(classOf[java.util.Map[_, _]]) =>
+        val mapTypes = ad.getMapTypes().get
```
anthonyccri: Use decomposition for readability:

```scala
val Some(left, right) = ad.getMapTypes()
```

anthonyccri:
Are we reusing the serialized feature when we write to the covering index versus the spatial index rather than serializing twice?

elahrvivaz:
Currently yes (and again when we write to the record table). I've address this in another PR: https://github.com/locationtech/geomesa/pull/499

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: Ah, nice didn't know you could do that without a match.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/core.scala
elahrvivaz: fixed.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorTrigger.scala
elahrvivaz: fixed.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: fixed.

anthonyccri:
148bf05

PR #495, opened 2015-03-10
GEOMESA-709 Speeding up delimited exports

elahrvivaz:
In the cloud I was testing, delimited export was much slower than bin export. This was due to the map creation, update, etc for each feature. After this ticket, delimited export was just as fast as bin export, which means the client code was no longer the bottleneck in this particular cloud.
idAttribute, latAttribute, lonAttribute, dtgAttribute are all bin export specific, and don't need to be considered here. As a future ticket, we might want to break apart bin export from other exports so that those arguments aren't confusing.

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
@@ -63,7 +63,8 @@ object BinaryOutputEncoder extends Logging {
       sort: Boolean = false): Unit = {
 
     val sft = fc.getSchema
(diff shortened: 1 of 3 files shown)


AndrewAnnex:
this looks good to me :+1: .

elahrvivaz:
Created a PR for breaking out bin command: https://geomesa.atlassian.net/browse/GEOMESA-714

jdk2pq:
Also looks good to me.

jnh5y:
Merged as 8af364710c120eb3d979a164b41f5f2964c8caf0 on 1.5.

jnh5y:
Committed as e6768bef2ee29a3a62f462857f25341bb68d9502 on 1.5.

PR #496, opened 2015-03-11
GEOMESA-710, GEOMESA-408, GEOMESA-200 Optimizing iterator stack filters and transforms

elahrvivaz:
* Pre-computing optimal iterator stack function in iterator init
* Kryo encoding index values
* Refactored query planner and removed IndexSchema as a wrapper for query planning to make codebase easier to grok
* Moved explain query to data store level
* Exposed direct kryo serialization of geometries (vs geometries in simple features)
* Fixed code smell in timing calculation in feature reader - GEOMESA-408
* Spatio-temporal planning only considers filters against the primary date and geom - GEOMESA-200


elahrvivaz:
Rebased on top of latest accumulo1.5/1.x

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureReader.scala
```diff
@@ -20,32 +20,24 @@ import org.geotools.data.{FeatureReader, Query}
...
 import org.locationtech.geomesa.feature.SimpleFeatureEncoder
 import org.locationtech.geomesa.utils.stats.{MethodProfiling, TimingsImpl}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 class AccumuloFeatureReader(dataStore: AccumuloDataStore,
                             query: Query,
-                            indexSchemaFmt: String,
                             sft: SimpleFeatureType,
-                            featureEncoder: SimpleFeatureEncoder)
+                            indexSchemaFmt: String,
```
anthonyccri: Probably for a later sprint . . . we need to pass around materialized ```IndexSchema```s rather than the format string to reduce the number of times we parse this string.

anthonyccri: Could you add a ticket for that issue?

elahrvivaz: Created here: https://geomesa.atlassian.net/browse/GEOMESA-718

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureReader.scala
```diff
@@ -20,32 +20,24 @@ import org.geotools.data.{FeatureReader, Query}
...
-      val cc = new ExplainingConnectorCreator(o)
-      queryPlanner.getIterator(cc, sft, query, strategyHints, o)
-    }, "explain")
-    o(s"Query Planning took ${timings.time("explain")} milliseconds.")
-  }
-
-  private lazy val iter = profile(queryPlanner.query(query, dataStore, strategyHints), "planning")
+  private val hints = dataStore.strategyHints(sft)
+  private val planner = new QueryPlanner(sft, featureEncoding, indexSchemaFmt, dataStore, hints, version)
+  private val iter = profile(planner.query(query), "planning")
```
anthonyccri: Definitely for another sprint . . . can you add a ticket about hooking up dropwizard metrics into this profile stuff here?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureReader.scala
```diff
@@ -64,11 +56,31 @@ class AccumuloFeatureReader(dataStore: AccumuloDataStore,
                     QueryStatTransform.filterToString(query.getFilter),
                     QueryStatTransform.hintsToString(query.getHints),
                     timings.time("planning"),
-                    //TODO: planning time gets added to scan time due to lazy val... Revisit in GEOMESA-408
-                    timings.time("next") + timings.time("hasNext") - timings.time("planning"),
+                    timings.time("next") + timings.time("hasNext"),
```
anthonyccri: Ditto re: dropwizard metrics.

anthonyccri: Maybe we should have an ```AccumuloReporter``` instance of a ```MetricsReporter``` interface?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/package.scala
```diff
@@ -76,7 +76,9 @@ import scala.collection.JavaConversions._
   // 0 == old single table style
   // 1 == multi-table style
   // 2 == sorted keys in the STIDX table
-  val INTERNAL_GEOMESA_VERSION = 2
+  // skipping 3 for integration...
+  // 4 is kryo encoded index values
+  val INTERNAL_GEOMESA_VERSION = 4
```
anthonyccri: Is this ticket backwards compatible and is there a test for that?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/filter/PropertyLiteral.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.locationtech.geomesa.core.filter
+
+import org.opengis.filter.expression.Literal
+
+case class PropertyLiteral(name: String, literal: Literal, secondary: Option[Literal], flipped: Boolean = false)
```
anthonyccri: Maybe throw this in the package object?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
```diff
@@ -161,13 +165,21 @@ trait AttributeIdxStrategy extends Strategy with Logging {
...
-    configureEcqlFilter(cfg, ecqlFilter.map(ECQL.toCQL))
-    configureAttributeName(cfg, attributeName)
-    configureIndexCoverage(cfg, featureType.getDescriptor(attributeName).getIndexCoverage())
     configureIndexValues(cfg, featureType)
+    configureAttributeName(cfg, attributeName)
+    configureIndexCoverage(cfg, coverage)
+    configureVersion(cfg, version)
+    if (coverage == IndexCoverage.FULL) {
+      // combine filters into one check
+      configureEcqlFilter(cfg, filterListAsAnd(Seq(stFilter ++ ecqlFilter).flatten).map(ECQL.toCQL))
```
anthonyccri: How does query planning work when we have a covering index?  Are we still taking into account the fact that the st index is covering and that the cardinalities need to be respected?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureReader.scala
elahrvivaz: Added ticket here: https://geomesa.atlassian.net/browse/GEOMESA-719

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/package.scala
elahrvivaz: It is backwards compatible (as long as the jars match up). I will add a unit test.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/IndexValueEncoder.scala
```diff
@@ -20,20 +20,156 @@ import java.nio.ByteBuffer
...
+
+  def apply(sft: SimpleFeatureType, version: Int): IndexValueEncoder = apply(sft, None, version)
+
+  def apply(sft: SimpleFeatureType, transform: SimpleFeatureType, version: Int): IndexValueEncoder =
+    apply(sft, Some(transform), version)
+
+  def apply(sft: SimpleFeatureType, transform: Option[SimpleFeatureType], version: Int): IndexValueEncoder = {
+    val key = {
+      // cache key should be unique per feature - we use name + attributes + bindings
+      val descriptors = sft.getAttributeDescriptors.map(d => s"${d.getLocalName}:${d.getType.getBinding}}")
```
anthonyccri: Could we just have a cache of ```String -> SimpleFeatureType``` and put the indexed attributes in the user data of the SFT?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/filter/PropertyLiteral.scala
elahrvivaz: Fixed.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryStrategyDecider.scala
```diff
@@ -78,51 +67,35 @@ object QueryStrategyDecider {
...
     // scan the query and identify the type of predicates present
 
     // record strategy takes priority
-    val recordStrategies =
-      filters.toStream.flatMap(f => getRecordIdxStrategy(f, sft).map(StrategyAndFilter(_, f)))
-    if (!recordStrategies.isEmpty) {
-      return recordStrategies(0).strategy
+    val recordStrategy = filters.iterator.flatMap(f => RecordIdxStrategy.getStrategy(f, sft, hints)).headOption
+    if (recordStrategy.isDefined) {
+      return recordStrategy.get.strategy
```
anthonyccri: Don't use ```return``` keyword.  I prefer that a function only return from one place.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
```diff
@@ -70,11 +71,12 @@ class STIdxStrategy extends Strategy with Logging with IndexFilterHelpers {
...
                           featureType: SimpleFeatureType,
+                          version: Int,
                           output: ExplainerOutputType) = {
-    val schema          = iqp.schema
+    val schema          = iqp.stSchema
     val featureEncoding = iqp.featureEncoding
-    val keyPlanner      = IndexSchema.buildKeyPlanner(iqp.schema)
-    val cfPlanner       = IndexSchema.buildColumnFamilyPlanner(iqp.schema)
+    val keyPlanner      = IndexSchema.buildKeyPlanner(schema)
+    val cfPlanner       = IndexSchema.buildColumnFamilyPlanner(schema)
```
anthonyccri: We should be able to pass these around fully materialized right?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IndexIterator.scala
```diff
@@ -32,46 +32,32 @@ import org.locationtech.geomesa.utils.stats.MethodProfiling
...
-      if (meetsIndexFilters) { // we hit a valid geometry, date and id
-        val transformedFeature = encodeIndexValueToSF(decodedValue)
-        // update the key and value
-        // copy the key because reusing it is UNSAFE
-        topKey = Some(indexKey)
-        topValue = transform.map(fn => new Value(fn(transformedFeature)))
-            .orElse(Some(new Value(featureEncoder.encode(transformedFeature))))
-      }
+    setTopOptimized = (stFilter, transform, checkUniqueId) match {
+      case (null, null, null) => setTopIndexInclude
```
anthonyccri: Ditto above . . . comments would be valuable here.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorExtensions.scala
```diff
@@ -61,80 +55,36 @@ trait HasFeatureType {
...
 }
 
-/**
- * Provides an index value decoder
- */
-trait HasIndexValueDecoder extends IteratorExtensions {
+trait HasVersion extends IteratorExtensions {
 
-  var indexEncoder: IndexValueEncoder = null
+  var version: Int = -1
```
anthonyccri: Any reason this isn't a ```val```?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/GeomesaFilteringIterator.scala
```diff
@@ -82,6 +75,12 @@ abstract class GeomesaFilteringIterator
    */
   def setTopConditionally(): Unit
 
-  override def deepCopy(env: IteratorEnvironment) =
+  override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] =
     throw new UnsupportedOperationException("GeoMesa iterators do not support deepCopy")
 }
+
+trait HasSourceIterator {
```
anthonyccri: Ok, I see what's going on here . . . the actual iterator is composed and shares these vars which get set in various places.  I find this a bit confusing - there's a lot of places where a mutation occurs - but off the top of my head I can't think of a better way.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorFunctions.scala
```diff
@@ -0,0 +1,260 @@
+/*
```
anthonyccri: Perhaps add a Readme.md in geomesa-core that describes how the iterator stack is built up?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RowSkippingIterator.scala
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
+import org.apache.hadoop.io.Text
+import org.locationtech.geomesa.core.iterators.RowSkippingIterator._
+
+import scala.collection.SortedSet
+
+/**
+ * Iterator that will skip rows based on a set of suffix strings.
+ */
+class RowSkippingIterator extends GeomesaFilteringIterator {
```
anthonyccri: Nice

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
elahrvivaz: Covering vs non-covering indices isn't included in the query planning at this point. Priorities are:
1. if there are ID filters -> RecordStrategy
2. if there are high-cardinality attribute filters -> attribute strategy
3. if there are spatial filters -> STIDX strategy
4. Otherwise use attribute strategy if possible, else STIDX strategy
Should I add some logic for covering indices?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RowSkippingIterator.scala
anthonyccri: Is there a RowSkippingIterator test?  If not, add a ticket for making one.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryStrategyDecider.scala
elahrvivaz: Eh, I prefer it to deeply nested statements... but I'll update :)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/package.scala
anthonyccri: Thanks

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
anthonyccri: No, just note it in a query planning ticket.  @jnh5y do you have thoughts about where to capture this info?

anthonyccri: Also, for map-reduce jobs . . . we typically want to use covering indexes whether ST or ATTR.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryStrategyDecider.scala
anthonyccri: Maybe split it into more functions then?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
jnh5y: For capture, I'd suggest cut and pasting some of this conversation into a JIRA ticket.  I have been suggesting adding comments in the code which link to the ticket.

If you meant capturing info about the query planner, I suggest creating versioned documentation.  The query explainer should spit out the metadata which could be used, and then move on to explaining the plan.  

In general, I imagine there will be a 'covering index' trait for the ST and ATTR tables which will help organize this bit of the query planning.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
```diff
@@ -285,18 +298,18 @@ trait GeoHashPlanner extends Logging {
...
+      // that fall inside the given polygon; in this case, return the LL, UR
+      // GeoHash endpoints of the entire range (which could encompass many
+      // more GeoHashes than we wish, but can only be better than (or equal
+      // to) a full-table scan)
+      val env = geom.getEnvelopeInternal
+      val ghLL = GeoHash(env.getMinX, env.getMinY)
+      val ghUR = GeoHash(env.getMaxX, env.getMaxY)
+      KeyRange(ghLL.hash, ghUR.hash)
+    } else {
+      KeyList(subHashes)
```
jnh5y: I notice that the removed code above sorted subHashes.  Is that important

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
```diff
@@ -304,11 +317,11 @@ trait GeoHashPlanner extends Logging {
     case SpatialFilter(geom)                => polyToPlan(geom, offset, bits)
     case SpatialDateFilter(geom, _)         => polyToPlan(geom, offset, bits)
     case SpatialDateRangeFilter(geom, _, _) => polyToPlan(geom, offset, bits)
-    case DateRangeFilter(_, _)              => KeyAccept
-    case AcceptEverythingFilter             => KeyAccept
+    case DateRangeFilter(_, _)              => KeyAccept(bits)
+    case AcceptEverythingFilter             => KeyAccept(bits)
```
jnh5y: Does KeyAccept(bits) help with table sharing, etc?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
```diff
@@ -316,16 +329,16 @@ case class GeoHashKeyPlanner(offset: Int, bits: Int) extends KeyPlanner with Geo
...
+        output(s"GeoHashKeyPlanner (${keys.size}): ${keys.take(20).mkString(",")}")
         KeyListTiered(keys)
 
       case KeyRange(keyMin, keyMax) =>
         output(s"GeoHashKeyPlanner: KeyRange $keyMin to $keyMax")
         KeyRangeTiered(keyMin, keyMax)
 
-      case KeyAccept =>
+      case ka: KeyAccept =>
         output("GeoHashKeyPlanner: KeyAccept")
```
jnh5y: Since KeyAccept has info...  we should update this output.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryStrategyDecider.scala
```diff
@@ -78,51 +67,35 @@ object QueryStrategyDecider {
...
-      case (Some(st), Some(attr)) =>
-        if (filters.indexOf(st.filter) < filters.indexOf(attr.filter)) {
-          st.strategy
-        } else {
-          attr.strategy
-        }
-    }
+    // finally, prefer spatial filters if available
+    val stStrategy = filters.iterator.flatMap(f => STIdxStrategy.getStrategy(f, sft, hints)).headOption
+    stStrategy.orElse(attributeStrategies.headOption).map(_.strategy).getOrElse(new STIdxStrategy)
```
jnh5y: I think I follow this.  I wonder if there is a more clear way to express this.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IndexIterator.scala
```diff
@@ -32,46 +32,32 @@ import org.locationtech.geomesa.utils.stats.MethodProfiling
...
+      case (null, null, _)    => setTopIndexUnique
+      case (_, null, null)    => setTopIndexFilter
+      case (_, null, _)       => setTopIndexFilterUnique
+      case (null, _, null)    => setTopIndexTransform
+      case (null, _, _)       => setTopIndexTransformUnique
+      case (_, _, null)       => setTopIndexFilterTransform
+      case (_, _, _)          => setTopIndexFilterTransformUnique
```
jnh5y: All 8 combinations are enumerations here and elsewhere.  It may be worth wrapping this up in a trait.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RecordTableIterator.scala
```diff
@@ -26,36 +26,29 @@ import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIt
...
-      // current entry matches our filter - update the key and value
-      topKey = Some(source.getTopKey)
-      // return the value or transform it as required
-      topValue = transform.map(fn => new Value(fn(feature)))
-          .orElse(Some(sourceValue))
+    setTopOptimized = (filter, transform) match {
+      case (null, null) => setTopInclude
+      case (_, null)    => setTopFilter
+      case (null, _)    => setTopTransform
+      case (_, _)       => setTopFilterTransform
```
jnh5y: Yeah, ok, here's the pattern;)
```
trait SetTopFilterTransformFunc { 
  self: SetTopInclude with SetTopFilter with SetTopTransform with SetTopFilterTransform =>
   def foo(filter, transform) = (filter, transform) match {
    case (null, null) => setTopInclude
    case (_, null) => setTopFilter
    case (null, _) => setTopTransform
    case (_, _) => setTopFilterTransform
  }
}
```

jnh5y: Of course, we could just roll these into a little bigger traits;)  Or have traits which just serve as unions...  Not sure which is cleanest.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/SpatioTemporalIntersectingIterator.scala
```diff
@@ -54,27 +51,18 @@ class SpatioTemporalIntersectingIterator
...
-        }
-      }
+    setTopOptimized = (filter, transform, checkUniqueId) match {
+      case (null, null, null) => setTopInclude
+      case (null, null, _)    => setTopUnique
+      case (_, null, null)    => setTopFilter
+      case (_, null, _)       => setTopFilterUnique
+      case (null, _, null)    => setTopTransform
+      case (null, _, _)       => setTopTransformUnique
+      case (_, _, null)       => setTopFilterTransform
```
jnh5y: This function is not included by the traits extended, right?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/FeatureWritersTest.scala
```diff
@@ -103,7 +103,6 @@ class FeatureWritersTest extends Specification {
 
         /* write the feature to the store */
         fs.addFeatures(featureCollection)
-        fs.flush()
```
jnh5y: I've lost track of when we want to flush.  I suppose this does not matter since it is using MockAccumulo?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/index/IndexValueEncoderTest.scala
```diff
@@ -88,10 +89,10 @@ class IndexValueEncoderTest extends Specification {
 
       // requirements
       decoded must not beNull;
-      decoded.attributes must haveSize(3)
-      decoded.id mustEqual id
-      decoded.geom mustEqual geom
-      decoded.date mustEqual Some(dt)
+      decoded.getAttributeCount mustEqual(2)
```
jnh5y: Did the previous attribute count include the id?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIteratorTest.scala
```diff
@@ -115,12 +119,10 @@ class AttributeIndexIteratorTest extends Specification with TestWithDataStore {
         "name = 'b'",
         "name < 'b'",
         "name > 'b'",
-        "name is NULL",
         "dtg TEQUALS 2014-01-01T12:30:00.000Z",
         "dtg = '2014-01-01T12:30:00.000Z'",
         "dtg BETWEEN '2012-01-01T12:00:00.000Z' AND '2013-01-01T12:00:00.000Z'",
-        "age < 10",
-        "name = 'b' AND BBOX(geom, 30, 30, 50, 50)"
```
jnh5y: How is this filter handled?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorExtensions.scala
elahrvivaz: It has to get set in the init method below.

jnh5y:
I'm read through, and I think most things are small.  @elahrvivaz should have a round of commenting and/or tweaking.

jnh5y:
I'm read through, and I think most things are small.  @elahrvivaz should have a round of commenting and/or tweaking.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryStrategyDecider.scala
elahrvivaz: fixed.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/IndexValueEncoder.scala
elahrvivaz: Originally it was similar to that, but we ran into issues b/c of re-used sft names.
The user data doesn't get persisted right now, so we'd have to add that to SimpleFeatureTypes... I think I prefer keeping it localized to here.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
elahrvivaz: This query strategy decision is pretty well commented in the QueryStrategyDecider. I've expanded on the comments there a bit too.

elahrvivaz: Created ticket here:
https://geomesa.atlassian.net/browse/GEOMESA-720

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RowSkippingIterator.scala
elahrvivaz: Created ticket here: https://geomesa.atlassian.net/browse/GEOMESA-721

This iterator isn't used anywhere, but I figured I would check it in since I had written it.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorFunctions.scala
elahrvivaz: Added.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RowSkippingIterator.scala
jnh5y: Would this be handy for something like 'decimation'?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
elahrvivaz: It was being sorted twice. Note that subHashes is already sorted when it's created on line 299.

jnh5y: Thanks.  I'm glad that's sorted.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
elahrvivaz: It's not really for table sharing - but in order to join keys in tiers, keyAccept needs to know how many chars it contains. I'm not sure it was really affecting us in the default case, but in the time first index work I did I noticed it was broken, so I brought over the fix.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
elahrvivaz: fixed.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryStrategyDecider.scala
elahrvivaz: I refactored this a bit per anthony's comments. Let me know once I push the changes if you think it could still use some clarification.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/SpatioTemporalIntersectingIterator.scala
elahrvivaz: Which function do you mean?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RecordTableIterator.scala
elahrvivaz: I think only one of these functions is actually copied once - the other are all different... think it's still worthwhile to create the traits?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/FeatureWritersTest.scala
elahrvivaz: The flush here was actually some implicit specs2 conversion to ConsoleOutput, which has a flush method...

geomesa-core/src/test/scala/org/locationtech/geomesa/core/index/IndexValueEncoderTest.scala
elahrvivaz: Yeah, previously it was a map of key values (including the ID, as you guessed), which we later turned into a simple feature. Now it decodes directly into a simple feature, which lets us standardize the interface and use kryo.

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/AttributeIndexIteratorTest.scala
elahrvivaz: We prioritize spatial filters now (unless the attribute has been marked high-cardinality). Using the order was problematic through geoserver, since geoserver doesn't respect the order of your cql when passing it along.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RowSkippingIterator.scala
elahrvivaz: The original idea was to use less ranges and then skip over rows we know aren't part of our query area. But in practice it wasn't any faster in my (limited) testing, so we decided to hold off. It probably warrants further testing in geobench or the stuff @jahhulbert-ccri is working on.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanners.scala
elahrvivaz: :)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RecordTableIterator.scala
elahrvivaz: Actually, maybe there's a bit more overlap... let me try the traits

elahrvivaz: Hmm, yeah the only overlap is between AttributeIndexIterator with full coverage and SpatioTemporalIterator. I dunno, I think adding a trait with the function matching might be more confusing... what do you think?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
elahrvivaz: We could, but they only get used here I believe... this can be addressed as part of the ticket I opened for passing around materialized IndexSchema objects. I'd have to look closer to see exactly how often we parse them.

elahrvivaz:
I believe all comments so far have been addressed. Let me know if any of the changes aren't satisfactory.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/RecordTableIterator.scala
jnh5y: Since they are different, I agree with you.  

I just saw eight lines of match statements and assumed something similar was happening...  I don't know if a comment would help or not.

anthonyccri:
I think its good and I'd like to get this one merged sooner rather than later so we have more time to flush out any issues before we cut rc.5.

anthonyccri:
Merged 09d3a2f

PR #497, opened 2015-03-12
GEOMESA-706 Supporting 'is not null' attribute queries

elahrvivaz:


AndrewAnnex:
This all looks good to me.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
```diff
@@ -331,6 +332,18 @@ object AttributeIndexStrategy extends StrategyProvider {
         checkOrder(f.getExpression1, f.getExpression2).filter(indexed)
             .map(p => StrategyDecision(new AttributeIdxRangeStrategy, costp(p)))
 
+      // not check - we only support 'not null' for an indexed attribute
+      case n: Not =>
+        Option(n.getFilter).collect { case f: PropertyIsNull => f }.flatMap { f =>
```
jnh5y: Is the Option just to get us in the right monad?  If not, I'm fine with some defensiveness.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
```diff
@@ -467,41 +482,60 @@ object AttributeIndexStrategy extends StrategyProvider {
...
         throw new RuntimeException(msg)
     }
 
   private def greaterThanRange(sft: SimpleFeatureType, prop: String, lit: AnyRef): AccRange = {
-    val rowIdPrefix = getTableSharingPrefix(sft)
     val start = new Text(getEncodedAttrIdxRow(sft, prop, lit))
-    val endPrefix = AttributeTable.getAttributeIndexRowPrefix(rowIdPrefix, sft.getDescriptor(prop))
-    val end = AccRange.followingPrefix(new Text(endPrefix))
+    val end = upperBound(sft, prop)
     new AccRange(start, false, end, false)
```
jnh5y: Ah!  Nice; since we don't have to deal with nulls, <, > is easier?  Is that what's going on?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
elahrvivaz: Yeah, it's so we can collect on the filter we want - it could have alternatively been written as a if(.isInstanceOf[PropertyIsNull]), but that seemed uglier...

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/AttributeIdxStrategy.scala
elahrvivaz: not really actually... I think the old way nulls would be counted as < everything, so the ranges haven't changed. I just extracted out the common code here and in the big match above into methods.

jnh5y:
Merged as 31ace116d304f00aecd6d7ab12675210186629e3 on 1.5.

PR #500, opened 2015-03-18
GEOMESA-722 Ordering of Point Pairs

mm64z:
Add a Hint to force Longitude to be first,
 as expected by downstream processes.

Signed-off-by: Michael Matthews <mmatthews@ccri.com>

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
@@ -20,6 +20,7 @@ import java.util.{Date, Locale, UUID}
 
 import com.typesafe.config.{Config, ConfigFactory}
 import com.vividsolutions.jts.geom._
+import org.geotools.factory.Hints
 import org.geotools.feature.AttributeTypeBuilder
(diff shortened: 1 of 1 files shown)


elahrvivaz:
Hmm, do we want to force this to be called every time we create a simple feature type? We already call it in AccumuloDataStore, so maybe it wouldn't matter.
@mm64z What downstream process is being affected by this?

mm64z:
The AnalyticWPS project was being affected by this.  I believe several other people found similar issues, I'm not sure of the specifics though.
If there's another commit that should have a similar effect, I can test that out.

elahrvivaz:
I'm hesitant to make a potentially sweeping change like this without pretty thorough testing... Could the code snippet go in the analytics WPS project instead?

mm64z:
From my testing, the snippet could be here or in each test that was failing. There wasn't a location inside the regular WPS process that worked, I believe since we set up variables from GeoMesa in the test before going into the WPS code.

cne1x:
@elahrvivaz @jnh5y @mm64z Can we chat briefly about this today (Monday, 30 March 2015)?  I'd like to go ahead and get this resolved one way or the other so that the AnalyticWPS can start working on RC4.  Thanks!

elahrvivaz:
The underlying problem stems from EPSG encodings - it's fixed when an EPSG jar is available on the classpath. I'm going to close this unmerged.

PR #502, opened 2015-03-19
GEOMESA-724 Map/reduce input and output formats

elahrvivaz:
This stacks on top of https://github.com/locationtech/geomesa/pull/499, please only consider the last commit(s).


elahrvivaz:
Need to fix a TODO...

PR #503, opened 2015-03-20
GEOMESA-276 Adding save method to allow saving SimpleFeature RDDs to Geo...

dmcwhorter:
...Mesa tables, adding simple unit test for GeoMesaSpark reading and writing


elahrvivaz:
Looks reasonable. We should probably test out the stack (iterators, geoserver, etc) with the new guava version.

pom.xml
```diff
@@ -535,7 +535,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>11.0.2</version>
+                <version>14.0.1</version>
```
elahrvivaz: Does this require an IP CQ?

jnh5y: It will require a piggyback CQ.

geomesa-compute/src/test/scala/org/locationtech/geomesa/compute/spark/GeoMesaSparkTest.scala
```diff
@@ -0,0 +1,147 @@
+package org.locationtech.geomesa.compute.spark
```
jahhulbert-ccri: header!

jnh5y:
Merged as efa705f8f5c5dc77592f10edd0096cf05ac58e6f on 1.5.

PR #504, opened 2015-03-20
GEOMESA-725 Add configurable cache and TTL to KafkaDataStore

jdk2pq:
Added expiry and expirationPeriod options to KafkaDataStore params. If expiry is set to true, then
the queue of features uses a Guava Cache with an expiration period set to the expirationPeriod value in milliseconds.

Signed-off-by: Jake Kenneally <jdk2pq@ccri.com>

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
@@ -21,7 +21,7 @@ import java.util
 import java.util.Properties
 import java.util.concurrent.{Executors, TimeUnit}
 
-import com.google.common.cache.{Cache, CacheBuilder}
+import com.google.common.cache.{RemovalNotification, RemovalListener, Cache, CacheBuilder}
 import com.google.common.eventbus.{EventBus, Subscribe}
 import com.vividsolutions.jts.geom.{Envelope, Geometry}
 import com.vividsolutions.jts.index.quadtree.Quadtree
@@ -71,6 +71,13 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
 
   if (expiry) {
     cb.expireAfterWrite(expirationPeriod, TimeUnit.MILLISECONDS)
+      .removalListener(
+        new RemovalListener[String, FeatureHolder] {
+          def onRemoval(removal: RemovalNotification[String, FeatureHolder]) = {
+            removeFeature(new Delete(removal.getKey))
+          }
+        }
+      )
   }
 
   val features: Cache[String, FeatureHolder] = cb.build()
@@ -88,16 +95,15 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
   def processNewFeatures(update: CreateOrUpdate): Unit = {
     val sf = update.f
     val id = update.id
-    Option(features.asMap().get(id)).foreach {  old => qt.remove(old.env, old.sf) }
+    Option(features.getIfPresent(id)).foreach {  old => qt.remove(old.env, old.sf) }
     val env = sf.geometry.getEnvelopeInternal
     qt.insert(env, sf)
     features.put(sf.getID, FeatureHolder(sf, env))
   }
 
   def removeFeature(toDelete: Delete): Unit = {
-    Option(features.asMap().get(toDelete.id)).foreach { holder =>
-      qt.remove(holder.env, holder.sf)
-    }
+    val id = toDelete.id
+    Option(features.getIfPresent(id)).foreach {  old => qt.remove(old.env, old.sf) }
     features.invalidate(toDelete.id)
   }
 
@@ -134,7 +140,7 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
   def include(i: IncludeFilter) = new DFR(schema, new DFI(features.asMap().valuesIterator.map(_.sf)))
 
   def fid(ids: FidFilterImpl): FR = {
-    val iter = ids.getIDs.flatMap(id => Option(features.asMap().get(id.toString)).map(_.sf)).iterator
+    val iter = ids.getIDs.flatMap(id => Option(features.getIfPresent(id.toString)).map(_.sf)).iterator
     new DFR(schema, new DFI(iter))
   }
 
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
@@ -19,8 +19,9 @@ package org.locationtech.geomesa.kafka
 import java.nio.charset.StandardCharsets
 import java.util
 import java.util.Properties
-import java.util.concurrent.Executors
+import java.util.concurrent.{ConcurrentMap, Executors, TimeUnit}
 
+import com.google.common.cache.{Cache, CacheBuilder}
 import com.google.common.collect.Maps
 import com.google.common.eventbus.{EventBus, Subscribe}
 import com.vividsolutions.jts.geom.{Envelope, Geometry}
@@ -50,7 +51,9 @@ import scala.collection.JavaConversions._
 class KafkaConsumerFeatureSource(entry: ContentEntry,
                                  schema: SimpleFeatureType,
                                  eb: EventBus,
-                                 query: Query)
+                                 query: Query, 
+                                 expiry: Boolean,
+                                 expirationPeriod: Long)
   extends ContentFeatureStore(entry, query) {
 
   type FR = FeatureReader[SimpleFeatureType, SimpleFeature]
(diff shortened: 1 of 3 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -65,7 +68,14 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
...
   }
 
-  val features = Maps.newConcurrentMap[String, FeatureHolder]()
+  val features = if (expiry) {
+    CacheBuilder
+      .newBuilder()
+      .expireAfterWrite(expirationPeriod, TimeUnit.MILLISECONDS)
+      .build()
+  } else {
+    Maps.newConcurrentMap[String, FeatureHolder]()
```
elahrvivaz: To simplify, I would just make it always use a cache. If there is no expiry, then don't use the expireAfterWrite option. But then you don't need all the duplicate code below for the 2 paths, plus all the 'asInstanceOf' (which you should usually try to avoid).

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
```diff
@@ -158,10 +162,14 @@ class KafkaDataStoreFactory extends DataStoreFactorySpi {
       if(IS_PRODUCER_PARAM.lookUp(params) == null) java.lang.Boolean.FALSE
       else IS_PRODUCER_PARAM.lookUp(params).asInstanceOf[java.lang.Boolean]
 
-    val partitions  = Option(TOPIC_PARTITIONS.lookUp(params)).map(_.toString.toInt).getOrElse(1)
-    val replication = Option(TOPIC_REPLICATION.lookUp(params)).map(_.toString.toInt).getOrElse(1)
+    val partitions       = Option(TOPIC_PARTITIONS.lookUp(params)).map(_.toString.toInt).getOrElse(1)
+    val replication      = Option(TOPIC_REPLICATION.lookUp(params)).map(_.toString.toInt).getOrElse(1)
+    val expiry =
+      if(EXPIRY.lookUp(params) == null) java.lang.Boolean.FALSE
+      else EXPIRY.lookUp(params).asInstanceOf[java.lang.Boolean]
```
elahrvivaz: Could use an option here for consistency.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/KafkaDataStoreTest.scala
```diff
@@ -93,6 +104,17 @@ class KafkaDataStoreTest extends Specification with Logging {
       fw.write()
       Thread.sleep(2000)
 
+      //Check LoadingCache expiration
+      {
+        import cachedConsumerFS._
+        val cachedFeatures = cachedConsumerFS.features.asInstanceOf[Cache[String, FeatureHolder]]
+        cachedFeatures.cleanUp() //remove old entries...hopefully there aren't any
+        cachedFeatures.size() must be equalTo 1
+        Thread.sleep(1100)
```
elahrvivaz: Might be better to break this out into a separate test method and turn off the expiry for the normal test - otherwise you might get some unexpected expirations if someone tries to update the test.

elahrvivaz:
Looks good.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -80,20 +88,21 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
   def processNewFeatures(update: CreateOrUpdate): Unit = {
     val sf = update.f
     val id = update.id
-    Option(features.get(id)).foreach {  old => qt.remove(old.env, old.sf) }
+    Option(features.asMap().get(id)).foreach {  old => qt.remove(old.env, old.sf) }
```
jnh5y: (Picky) Inconsistent style between this and line 98.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/KafkaDataStoreTest.scala
```diff
@@ -169,6 +175,34 @@ class KafkaDataStoreTest extends Specification with Logging {
...
+      val featureCache = cachedConsumerFS.features
+
+      //Write test feature
+      val fw = producerDS.getFeatureWriter("testExpiration", null, Transaction.AUTO_COMMIT)
+      val sf = fw.next()
+      sf.setAttributes(Array("jones", 30, DateTime.now().toDate).asInstanceOf[Array[AnyRef]])
+      sf.setDefaultGeometry(gf.createPoint(new Coordinate(0.0, 0.0)))
+      fw.write()
+
+//      featureCache.cleanUp() //remove stale entries, BUT there shouldn't be anything to remove
```
jnh5y: (Picky) remove or reformat comment.

jnh5y:
@anthonyccri @elahrvivaz Can one of you confirm that adding expiry in this PR does not impact other use cases for the Kafka datastore?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
elahrvivaz: It would probably be better to use features.getIfPresent() - asMap might have some hidden cost we can avoid. Applies to the other uses of asMap too.

elahrvivaz:
It's a configurable option, so it should be fine.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -65,7 +67,13 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
     }
   }
 
-  val features = Maps.newConcurrentMap[String, FeatureHolder]()
+  val cb = CacheBuilder.newBuilder()
+
+  if (expiry) {
+    cb.expireAfterWrite(expirationPeriod, TimeUnit.MILLISECONDS)
```
elahrvivaz: You should add a RemovalListener here that will remove the feature from the QuadTree when it gets evicted from the cache.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -65,7 +67,20 @@ class KafkaConsumerFeatureSource(entry: ContentEntry,
...
+      .removalListener(
+        new RemovalListener[String, FeatureHolder] {
+          def onRemoval(removal: RemovalNotification[String, FeatureHolder]) = {
+            removeFeature(new Delete(removal.getKey))
```
elahrvivaz: Hmm, I don't think this will work - removeFeature only removes from the QT if the feature is still in the cache, which it might not be at this point... You can just remove it directly from the QT here.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/KafkaDataStoreTest.scala
```diff
@@ -169,6 +175,33 @@ class KafkaDataStoreTest extends Specification with Logging {
...
+
+      //Write test feature
+      val fw = producerDS.getFeatureWriter("testExpiration", null, Transaction.AUTO_COMMIT)
+      val sf = fw.next()
+      sf.setAttributes(Array("jones", 30, DateTime.now().toDate).asInstanceOf[Array[AnyRef]])
+      sf.setDefaultGeometry(gf.createPoint(new Coordinate(0.0, 0.0)))
+      fw.write()
+
+      featureCache.size() must eventually(5, 500.millis)(beEqualTo(1))
+      Thread.sleep(3005) //sleep enough time to reach the expirationPeriod
```
elahrvivaz: You could set this lower (like 1000 maybe?) to speed up the test a bit.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
elahrvivaz: Maybe add a test that doesn't call cleanup() on the cache and see if it works or not.

elahrvivaz:
latest commits look good.

elahrvivaz:
Merged as f3192a8d91fbaccdf53a9b73f04909e9ba0271e8

PR #505, opened 2015-03-23
GEOMESA-724 Map/reduce input and output formats

elahrvivaz:
Stacks on top of https://github.com/locationtech/geomesa/pull/499 - please only consider the last commit(s).

* Split query planning and query execution into two separate steps so that m/r can use query planning
* Split feature writing into creating mutations and applying them so that m/r can use the mutations directly


elahrvivaz:
Closing - will replace with a merged PR that includes scalding and map/reduce

PR #507, opened 2015-03-24
GEOMESA-729 Scalding geomesa source/sink

elahrvivaz:
Stacks on top of https://github.com/locationtech/geomesa/pull/505, please only consider the last commit(s).
* Also split AccumuloSource options into input and output specific flavors


elahrvivaz:
Closing - will replace with a merged PR that includes scalding and map/reduce

PR #511, opened 2015-03-24
GEOMESA-724,GEOMESA-729 Map/reduce and scalding input and output formats

elahrvivaz:
* Split query planning and query execution into two separate steps so that m/r can use query planning
* Split feature writing into creating mutations and applying them so that m/r can use the mutations directly
* Split AccumuloSource options into input and output specific flavors


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
```diff
@@ -215,41 +216,35 @@ class AccumuloDataStore(val connector: Connector,
...
    * Read SpatioTemporal Index table name from store metadata
    */
-  def getSpatioTemporalIdxTableName(featureName: String): String =
-    if (getGeomesaVersion(featureName) > 0) {
-      metadata.readRequired(featureName, ST_IDX_TABLE_KEY)
-    } else {
-      catalogTable
-    }
+  def getSpatioTemporalTable(featureName: String): String =
+    metadata.readRequired(featureName, ST_IDX_TABLE_KEY)
```
jnh5y: Does this break backward compatibility with version 0?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
```diff
@@ -169,9 +169,7 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
       cachingParam
     )
 
-  val pi = getParametersInfo
```
jnh5y: Was pi unused?  Hmm... seems odd.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureWriter.scala
```diff
@@ -142,8 +117,7 @@ abstract class AccumuloFeatureWriter(sft: SimpleFeatureType,
       feature
     }
 
-    val toWrite = new FeatureToWrite(withFid, defaultVisibility, encoder, indexValueEncoder)
-    writers.foreach(write => write(toWrite))
+    writer(new FeatureToWrite(withFid, defaultVisibility, encoder, indexValueEncoder))
```
jnh5y: Does introducing FeatureToWrite objects cost us anything in terms of object instantiation/GC?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/tables/AttributeTable.scala
```diff
@@ -42,37 +43,33 @@ import scala.util.{Failure, Success, Try}
...
+  def attributeWriter(sft: SimpleFeatureType): Option[FeatureToMutations] = {
+
+    val indexedAttributes = SimpleFeatureTypes.getSecondaryIndexedAttributes(sft)
+    if (indexedAttributes.isEmpty) {
+      None
+    } else {
+      val rowIdPrefix = org.locationtech.geomesa.core.index.getTableSharingPrefix(sft)
+
+      val indexesOfIndexedAttributes = indexedAttributes.map { a => sft.indexOf(a.getName) }
+      val attributesToIdx = indexedAttributes.zip(indexesOfIndexedAttributes)
```
jnh5y: I'm tempted to suggest dropping the zip, passing the two Seqs, and rewriting the flatMap to use a while loop...  

That might be a later optimization though...  Not sure how much overhead the new Seq creates and if the flatmap is more expensive than a while loop.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -65,13 +65,12 @@ object GeoMesaSpark {
...
 
-    val planner = new STIdxStrategy
-    val qp = planner.buildSTIdxQueryPlan(query, queryPlanner, sft, version, ExplainPrintln)
+    val qp = new STIdxStrategy().getQueryPlan(query, queryPlanner, ExplainPrintln)
 
     ConfiguratorBase.setConnectorInfo(classOf[AccumuloInputFormat], conf, ds.connector.whoami(), ds.authToken)
     ConfiguratorBase.setZooKeeperInstance(classOf[AccumuloInputFormat], conf, ds.connector.getInstance().getInstanceName, ds.connector.getInstance().getZooKeepers)
 
-    InputConfigurator.setInputTableName(classOf[AccumuloInputFormat], conf, ds.getSpatioTemporalIdxTableName(sft))
+    InputConfigurator.setInputTableName(classOf[AccumuloInputFormat], conf, ds.getSpatioTemporalTable(sft))
```
jnh5y: This might be more appropriate for another PR, but this should change to use the GeoMesaInputFormat, no?

geomesa-jobs/src/main/java/org/locationtech/geomesa/jobs/interop/mapred/FeatureCountJob.java
```diff
@@ -0,0 +1,83 @@
...
+        static enum CountersEnum { FEATURES }
+
+        public void map(Text key,
+                        SimpleFeature value,
+                        OutputCollector<Text, Text> output,
+                        Reporter reporter) throws IOException {
+            Counter counter = reporter.getCounter(CountersEnum.class.getName(),
+                                                  CountersEnum.FEATURES.toString());
+            counter.increment(1);
+            output.collect(new Text(value.getID()), new Text(value.getAttribute("geom").toString()));
```
jnh5y: Rather than getAttribute("geom") which may not exist, prefer getDefaultGeometry?

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/GeoMesaConfigurator.scala
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Common place for setting and getting values out of the mapreduce config
+ */
+object GeoMesaConfigurator {
+
+  private val prefix         = "org.locationtech.geomesa"
+
+  private val dsInParams     = s"$prefix.params.in."
+  private val dsOutParams    = s"$prefix.params.out."
+  private val dsInRegex      = dsInParams.replaceAll("\\.", "\\.") + ".+"
```
jnh5y: Ok, just so I know, what is this doing?  Escaping .'s?

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
```diff
@@ -0,0 +1,281 @@
...
+    val zookeepers = AccumuloDataStoreFactory.params.zookeepersParam.lookUp(dsParams).asInstanceOf[String]
+    InputFormatBase.setZooKeeperInstance(job, instance, zookeepers)
+
+    val auths = Option(AccumuloDataStoreFactory.params.authsParam.lookUp(dsParams).asInstanceOf[String])
+    auths.foreach(a => InputFormatBase.setScanAuthorizations(job, new Authorizations(a.split(","): _*)))
+
+    // run an explain query to set up the iterators, ranges, etc
+    val ecql = filter.map(ECQL.toFilter).getOrElse(Filter.INCLUDE)
+    val query = new Query(featureTypeName, ecql)
+    val queryPlans = ds.explainQuery(featureTypeName, query, ExplainNull)
```
jnh5y: I definitely understand this since I've read along the whole time.  

Getting a queryPlan from a function called 'explainQuery' seems odd though. 

Maybe we could rename explainQuery to getQuery?  A new ds.explainQuery method could call getQuery with a Logging/PrintlnExplainer?  Meh, maybe it isn't worth sorting out.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
```diff
@@ -0,0 +1,281 @@
...
+
+    // see if the plan is something we can execute from a single table
+    val tryPlan = if (queryPlans.length > 1) None else queryPlans.headOption.filter {
```
jnh5y: I wonder if this logic should go to the query planner?  

We could expose a new function or a flag to convey the notion of a MR/Spark compatible query plan?

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
```diff
@@ -0,0 +1,281 @@
...
+      val indexSchema = ds.getIndexSchemaFmt(featureTypeName)
+      val hints = ds.strategyHints(sft)
+      val version = ds.getGeomesaVersion(sft)
+      val queryPlanner = new QueryPlanner(sft, featureEncoding, indexSchema, ds, hints, version)
+      new STIdxStrategy().getQueryPlan(query, queryPlanner, ExplainNull)
+    }
+
+    // use the explain results to set the accumulo input format options
+    InputFormatBase.setInputTableName(job, queryPlan.table)
+    if (!queryPlan.ranges.isEmpty) {
```
jnh5y: If the ranges are empty, does that imply a full-table scan?

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  override def write(key: Text, value: SimpleFeature) = {
+    val sft = value.getFeatureType
+    val sftName = sft.getTypeName
+
+    // ensure that the type has been created if we haven't seen it before
+    if (sftCache.add(sftName)) {
+      // this is a no-op if schema is already created, and should be thread-safe from different mappers
+      ds.createSchema(value.getFeatureType)
+      // short sleep to ensure that feature is fully written if it is happening in some other thread
```
jnh5y: ^feature^feature type

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
```diff
@@ -0,0 +1,148 @@
...
+      val stEncoder = IndexSchema.buildKeyEncoder(sft, ds.getIndexSchemaFmt(sft.getTypeName))
+      val stWriter = SpatioTemporalTable.spatioTemporalWriter(stEncoder)
+      val stTable = new Text(ds.getSpatioTemporalTable(sft))
+      val recWriter = RecordTable.recordWriter(sft)
+      val recTable = new Text(ds.getRecordTable(sft))
+      AttributeTable.attributeWriter(sft) match {
+        case None => Seq((stTable, stWriter), (recTable, recWriter))
+        case Some(attrWriter) =>
+          val attrTable = new Text(ds.getAttributeTable(sft))
+          Seq((stTable, stWriter), (recTable, recWriter), (attrTable, attrWriter))
```
jnh5y: Can we extract this function, shove it in the DataStore or something?

I guess the argument against is that anyone would be able to get access to our writers easily?  

Also, this bypasses any metrics writing.  Hmm....

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -0,0 +1,275 @@
...
+    val auths = Option(AccumuloDataStoreFactory.params.authsParam.lookUp(dsParams).asInstanceOf[String])
+    auths.foreach(a => InputFormatBase.setScanAuthorizations(job, new Authorizations(a.split(","): _*)))
+
+    // run an explain query to set up the iterators, ranges, etc
+    val ecql = filter.map(ECQL.toFilter).getOrElse(Filter.INCLUDE)
+    val query = new Query(featureTypeName, ecql)
+    val queryPlans = ds.explainQuery(featureTypeName, query, ExplainNull)
+
+    // see if the plan is something we can execute from a single table
+    val tryPlan = if (queryPlans.length > 1) None else queryPlans.headOption.filter {
```
jnh5y: Oh!  Wow!  We have to re-write these classes to compile against the different FQCNs?

That's unfortunate.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/scalding/package.scala
```diff
@@ -34,16 +39,84 @@ package object scalding {
...
-    val ACCUMULO_PASSWORD   = "geomesa.accumulo.password"
-    val AUTHORIZATIONS      = "geomesa.accumulo.authorizations"
-    val VISIBILITIES        = "geomesa.accumulo.visibilities"
-    val FEATURE_NAME        = "geomesa.feature.name"
-    val CATALOG_TABLE       = "geomesa.feature.tables.catalog"
-    val SCAN_RANGES         = "geomesa.scan.ranges"
-    val MOCK                = "geomesa.mock"
+
+    // old arg strings - replaced with standardized input/output ones below
+    @Deprecated val ACCUMULO_INSTANCE_OLD = "geomesa.accumulo.instance"
```
jnh5y: I don't think any of our users picked up on our nascent Scalding support.

I'm tempted to say that deprecating this is overkill.

jnh5y:
Looks great.  Most of my comments are small/quick.  I'd like to get another +1.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
elahrvivaz: Yeah, but it's already broken due to schema_changes, with no upgrade path available... Is that something we need to worry about?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreFactory.scala
elahrvivaz: Yeah, it was unused... maybe it was there for some side-effect? But it seems fine without it...

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloFeatureWriter.scala
elahrvivaz: Yeah, it is another object. The alternatives seem bad though - make the writer take the encoded feature values (which causes duplicated code and they all might not be used in every case), or break up the writer interface into specific impls for each table. What do you think?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/tables/AttributeTable.scala
elahrvivaz: You mean the flatMap in getAttributeIndexMutations? That could maybe be optimized, although in general i think map/flatMap is not *that* expensive relative to a while loop.
The snippet of code right here only gets run once per feature writer (not per feature), so optimizing this bit would not be worthwhile.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
elahrvivaz: Yeah, I will open a ticket if you want.

geomesa-jobs/src/main/java/org/locationtech/geomesa/jobs/interop/mapred/FeatureCountJob.java
elahrvivaz: Yeah, I can change that. I'll add a comment on the expected SFT that this job runs on - it's kind of implied but not specified.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/GeoMesaConfigurator.scala
elahrvivaz: Yeah, exactly. Plus it adds a '.+' on the end, so it will match anything starting with 'org.locationtech.geomesa.params.in.' Then it's used in the Configuration.getValByRegex method.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
elahrvivaz: yeah, maybe we could call it getQueryPlan? I basically hijacked the explainQuery method to also return the query it had just explained...
I'll add another method and make explainQuery not return anything.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
elahrvivaz: I don't really like the idea of the query planner having any notion of m/r. But we do have to duplicate some code down below to avoid that. Hmm, let me think about it.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
elahrvivaz: Yeah - I don't think that should ever happen. It would at least have a single range that spans the whole table. This is more a fail safe since I don't think you're supposed to pass empty ranges to InputFormatBase.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
elahrvivaz: Fixed, thanks. This is sort of a hack but I couldn't think of anything better.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
elahrvivaz: Yeah, I will extract this somewhere. What metrics writing do you mean?

elahrvivaz: Trying to extract this got real complicated... I'll open a ticket instead.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
elahrvivaz: Yeah, mapred vs mapreduce... I tried to share code, but in the end it was a lot easier just to duplicate it due to every single interface being different.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/scalding/package.scala
elahrvivaz: Heh, I was trying to stay back compatible :) But i can remove it fo sho

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
elahrvivaz: Created https://geomesa.atlassian.net/browse/GEOMESA-736

jnh5y:
@anthonyccri @elahrvivaz This got merged a-ok, right?  I'm just curious why it is still open.  If no one gets to it first, I can sort it out in the morning.

elahrvivaz:
Yeah, closing now.
7ed08cf602b83a5a440271728462ed95ea615b64

PR #516, opened 2015-03-26
F convert add linenumber and filename

anthonyccri:
For comments


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -284,13 +286,33 @@ class IdFunctionFactory extends TransformerFunctionFactory {
...
 
 }
+
+class FileInfoFunctionFactory extends TransformerFunctionFactory {
+  override def functions = Seq(FileNameFn(), LineNumberFn())
+
+  case class LineNumberFn() extends TransformerFn {
+    override def getInstance: LineNumberFn = LineNumberFn()
+    override def name: String = "lineNo"
+    override def eval(args: Any*)(implicit ec: EvaluationContext): Any = ec.incCount()
```
anthonyccri: This should just access the current line number, not increment it.  Increment the line number in the converter as it processes each line.

PR #518, opened 2015-03-30
GEOMESA-741 for comments on API changes

mmatz-ccri:
proposed api change for objects SimpleFeatureEncoder and SimpleFeatureDecoder

### DIFF
--- a/geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
+++ b/geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
@@ -289,7 +289,7 @@ class KryoFeatureEncoder(sft: SimpleFeatureType, projected: SimpleFeatureType,
     this(sft, sft, options)
   }
 
-  val encoder = KryoFeatureSerializer(sft, projected)
+  val encoder = KryoFeatureSerializer(sft, projected, options)
 
   override val encoding = FeatureEncoding.KRYO
   override def encode(feature: SimpleFeature) = encoder.write(feature)
--- a/geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
+++ b/geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
@@ -30,6 +30,10 @@ trait HasEncoding {
   def encoding: FeatureEncoding
 }
 
+trait HasEncodingOptions {
+  def options: Set[EncodingOption]
+}
+
 /**
  * Interface to encode SimpleFeatures with a configurable serialization format.
  *
@@ -39,7 +43,7 @@ trait HasEncoding {
  * SimpleFeatureEncoder classes may not be thread safe and should generally be used
  * as instance variables for performance reasons.
  */
-trait SimpleFeatureEncoder extends HasEncoding {
+trait SimpleFeatureEncoder extends HasEncoding with HasEncodingOptions {
   def encode(feature: SimpleFeature): Array[Byte]
 }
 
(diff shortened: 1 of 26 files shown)


geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
```diff
@@ -101,6 +148,19 @@ object FeatureEncoding extends Enumeration {
...
+ * Options to be applied when encoding.  The same options must be specified when decoding.
+ */
+object EncodingOption extends Enumeration {
+  type EncodingOption = Value
+
+  /**
+   * If this [[EncodingOption]] is specified then the security marking associated with the sample feature will be
+   * serialized and deserialized.
+   */
+  val SECURE = Value("secure")
```
elahrvivaz: I would make this more explicit - maybe 'encode_visibilities' or something.

elahrvivaz:
seems reasonable.

jahhulbert-ccri:
yeah seems like that would work...as emilio suggested i think ```EncodeVisibilities``` is better than ```Secure```

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
mmatz-ccri: I think that it should not be a verb.  What about WITH_VISIBILITIES?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
```diff
@@ -59,38 +59,85 @@ trait SimpleFeatureDecoder extends HasEncoding {
...
-  def apply(sft: SimpleFeatureType, encoding: FeatureEncoding) =
+  /**
+   * Decode without projecting.
+   *
+   * @param sft the encoded simple feature type to be decode
+   * @param encoding the encoding that was used to encode
+   * @param options any options that were used to encode
+   * @return a new [[SimpleFeatureDecoder]]
+   */
+  def apply(sft: SimpleFeatureType, encoding: FeatureEncoding, options: EncodingOption*) =
```
jnh5y: I haven't seen us use varargs much in Scala...

mmatz-ccri: Should I avoid it?  I way trying to stay flexible and avoid having duplicate sets of methods with and without option(s).

elahrvivaz: A scala-y alternative would be to provide a default value. Something like:
options: Set[EncodingOption] = Set.empty

mmatz-ccri: Sound good.  I'll make the change.

mmatz-ccri: Error:(60, 8) in object SimpleFeatureDecoder, multiple overloaded alternatives of method apply define default arguments.
object SimpleFeatureDecoder {
       ^

elahrvivaz: I think you've essentially created 2 methods with the same signature. If you have two methods: a(s:String) a(s:String,i: Int = 0), it can't disambiguate a call to a("")

mmatz-ccri: The signitures are different:

apply(sft: SimpleFeatureType, encoding: FeatureEncoding, options: Set[EncodingOptions] = Set.empty)

apply(sft: SimpleFeatureType, encoding: String, options: Set[EncodingOptions] = Set.empty)

but the complier doesn't allow multiple overloaded methods with default arguments:

https://groups.google.com/forum/#!msg/scala-user/FyQK3-cqfaY/fXLHr8QsW_0J

elahrvivaz: interesting.

elahrvivaz: maybe ask in the scala chat room, see if anyone has any suggestions.

PR #519, opened 2015-03-30
GEOMESA-740: Improve Configure Behavior re: Env Variables

JakeKurzer:
This small fix causes the configuration script for geo tools
    to correctly prompt the user if the home or libs variables
    are already set. Without this, it's possible to end up
    incorrectly configuring the paths as a result of prior runs.

    This is unlikely to affect users, but could lead to weirdness
    in certain specific development situations.

Signed-off-by: Jake Kurzer <jake.kurzer@ccri.com>


JakeKurzer:
I've rolled a CLA now. What do I need to do to fix the failed check?

geomesa-tools/bin/geomesa
```diff
@@ -36,14 +36,36 @@ if [ -z "$GEOMESA_HOME" ]; then
     export PATH=${GEOMESA_HOME}/bin:$PATH
     echo >&2 "Warning: GEOMESA_HOME is not set, using $GEOMESA_HOME"
 else
-    echo >&2 "Using GEOMESA_HOME $GEOMESA_HOME"
+    echo >&2 "Using GEOMESA_HOME as set: $GEOMESA_HOME"
+    read -p "Is this intentional? Y\n " -n 1 -r
```
elahrvivaz: so this will make the user confirm their geomesa home every time? I don't like this...

geomesa-tools/bin/geomesa
```diff
@@ -36,14 +36,36 @@ if [ -z "$GEOMESA_HOME" ]; then
...
+    fi
+    echo >&2 ""
 fi
 
 # GEOMESA paths, GEOMESA_LIB should live inside GEOMESA_HOME, but can be pointed elsewhere
 if [ -z "$GEOMESA_LIB" ]; then
     GEOMESA_LIB=${GEOMESA_HOME}/lib
+else
+    message="Warning: GEOMESA_LIB already set, probably by a prior configuration."
+    message="${message}\n Current value is ${GEOMESA_LIB}."
```
elahrvivaz: Ditto here. We don't want to make the tools super annoying for normal use case, which would be get installed correctly once and not touch again.

elahrvivaz:
Maybe instead of this we could create a 'troubleshoot' function or something that will walk the user through the appropriate settings.

geomesa-tools/bin/geomesa
JakeKurzer: Just caught that myself, actually. I'll twiddle it around so that the check occurs once. There's probably a need to do a bit more serious rewriting.

JakeKurzer:
@elahrvivaz I agree. This should probably happen for Configure, and can probably be moved there. Likewise, handling the configure command-line arg should probably be the first thing the script does. Sound fair?

elahrvivaz:
@JakeKurzer yeah, sounds good.

JakeKurzer:
@elahrvivaz I've switched it to an elif check against the configure arg, and factored out the behavior for setting GeoHome. Take a look?

geomesa-tools/bin/geomesa
```diff
@@ -35,15 +30,48 @@ if [ -z "$GEOMESA_HOME" ]; then
...
+JAVA_LIBRARY_PATH=""
+
+
+
+# Define GEOMESA_HOME and update the PATH if necessary.
+if [ -z "$GEOMESA_HOME" ]; then
+setGeoHome
 else
-    echo >&2 "Using GEOMESA_HOME $GEOMESA_HOME"
+    echo >&2 "Using GEOMESA_HOME as set: $GEOMESA_HOME"
```
elahrvivaz: [picky] I prefer removing the 'as set:'

geomesa-tools/bin/geomesa
```diff
@@ -35,15 +30,48 @@ if [ -z "$GEOMESA_HOME" ]; then
...
+    if [[ $1 = configure ]]; then
+        read -p "Is this intentional? Y\n " -n 1 -r
+        if [[  $REPLY =~ ^[Nn]$ ]]; then
```
elahrvivaz: [picky] probably safer to check for yes, and assume other things are no. Also, checking for (y|yes) (ignore case) would be nice.

geomesa-tools/bin/geomesa
```diff
@@ -35,15 +30,48 @@ if [ -z "$GEOMESA_HOME" ]; then
...
 if [ -z "$GEOMESA_LIB" ]; then
     GEOMESA_LIB=${GEOMESA_HOME}/lib
+elif [[ $1 = configure ]]; then
+    message="Warning: GEOMESA_LIB already set, probably by a prior configuration."
+    message="${message}\n Current value is ${GEOMESA_LIB}."
+    echo >&2 ""
+    echo -e >&2 "$message"
+    echo >&2 ""
+    read -p "Is this intentional? Y\n " -n 1 -r
+    if [[  $REPLY =~ ^[Nn]$ ]]; then
```
elahrvivaz: ditto yes/no above

elahrvivaz:
yeah, looks good. I had some nitpicky things.

geomesa-tools/bin/geomesa
JakeKurzer: =~ (?i)^(y|yes)$  will work for case insensitive checks, portably, right?

elahrvivaz: eh, i'm not sure... bash regexes are wonky :P
if it works in your terminal, that's good enough for me

JakeKurzer:
Resolved, I think.

geomesa-tools/bin/geomesa
```diff
@@ -35,15 +30,49 @@ if [ -z "$GEOMESA_HOME" ]; then
...
 # GEOMESA paths, GEOMESA_LIB should live inside GEOMESA_HOME, but can be pointed elsewhere
 if [ -z "$GEOMESA_LIB" ]; then
     GEOMESA_LIB=${GEOMESA_HOME}/lib
+elif [[ $1 = configure ]]; then
+    message="Warning: GEOMESA_LIB already set, probably by a prior configuration."
+    message="${message}\n Current value is ${GEOMESA_LIB}."
+    echo >&2 ""
+    echo -e >&2 "$message"
+    echo >&2 ""
+    read -p "Is you want to reset this? Y\n " -n 1 -r
```
AndrewAnnex: what?

JakeKurzer:
@AndrewAnnex Fixed, I am thinks.

jdk2pq:
This looks good to me, and I tested it out as well. You should rebase onto the current accumulo1.5.x/1.x branch and check that things are still working as expected. It worked for me, there was a small conflict that I had to solve, but everything seemed to work as expected.

jdk2pq:
Committed as 261e52a612f882c664a4bab939b304e0dfd514bf

PR #523, opened 2015-04-01
GEOMESA-750 Making default index schema use hour precision

elahrvivaz:
Putting the hour in the default index schema dramatically speeds up queries that span less than a day, and doesn't seem to negatively impact larger queries (which might be assumed, as we will create more rows)


AndrewAnnex:
:+1: looks good

jdk2pq:
Committed as c9f96d0475b130c83ce7a12886725fa218ec51ca

PR #525, opened 2015-04-06
GEOMESA-741 Propagate security markings when serializing SimpleFeatures

mmatz-ccri:
Security markings for SimpleFeatures can now be propagated using kryo, avro, or text encoding by specifying an EncodingOption when creating the FeatureEncoder.  The encoding option must be specified when creating the FeatureDecoder as well.


geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/FeatureSpecificReader.scala
```diff
@@ -25,18 +25,20 @@ import org.apache.avro.io.{BinaryDecoder, DatumReader, Decoder, DecoderFactory}
...
-class FeatureSpecificReader(oldType: SimpleFeatureType, newType: SimpleFeatureType)
+class FeatureSpecificReader(oldType: SimpleFeatureType, newType: SimpleFeatureType, opts: Set[EncodingOption] = Set.empty)
   extends DatumReader[AvroSimpleFeature] {
 
   def this(sft: SimpleFeatureType) = this(sft, sft)
 
-  var oldSchema = generateSchema(oldType)
-  val newSchema = generateSchema(newType)
+  private val includeVis = opts.contains(EncodingOption.WITH_VISIBILITIES)
+  var oldSchema = generateSchema(oldType, includeVis)
```
anthonyccri: This should be a ```val```.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/SimpleFeatureEncoder.scala
```diff
@@ -52,58 +57,114 @@ trait SimpleFeatureEncoder extends HasEncoding {
...
+ * @param sft the simple feature type to decode
+ * @param options the options what were applied when encoding
+ */
+class TextFeatureDecoder(sft: SimpleFeatureType, val options: Set[EncodingOption] = Set.empty)
+  extends SimpleFeatureDecoder {
+
+  private val expectVis = options.contains(EncodingOption.WITH_VISIBILITIES.asInstanceOf[EncodingOption])
+
+  override def decode(bytes: Array[Byte]): SimpleFeature = {
+    var encoded = new String(bytes, StandardCharsets.UTF_8)
```
anthonyccri: Use guava Strings and some scala syntax to clean this up

```scala
val (encodedString, vis) = 
   if(expectVis) {
      val (e, v) = TextEncoding.splitVisibility(new String(bytes, ....))
      (e, Option(Strings.emptyToNull(v)))
   } else {
      (new String(bytes, ...), None)
   }
```

anthonyccri: err, better yet, make this two different methods as vals and pick the right one . . . 

```scala
val defaultDecode: Array[Byte] => SimpleFeature = { bytes =>
   // do default decode
}

val withVizDecode: Array[Byte] => SimpleFeature = { bytes =>
   // do metadata decode
}

override val decode: Array[Byte] => SimpleFeature = if(expectViz) metadataDecode else defaultDecode
```

I think this would be more performant bc it would happen at instantiation time and then the if test would never occur again.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -130,10 +130,63 @@ class TransformingSimpleFeatureSerializer(sft: SimpleFeatureType, transform: Sim
...
+  }
+
+  def readHeader(input: Input): (Int, String, Option[String]) = {
+    val version = input.readInt(true)
+    val id = input.readString()
+
+    val vis = {
+      if (includeVis) {
+        val v = input.readString()
+        if (v.isEmpty) Option.empty else Option(v)
```
anthonyccri: Use guava .... ```Option(Strings.emptyToNull(v))```

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -466,35 +505,32 @@ object SimpleFeatureSerializer {
...
-
-    in.readInt(true) match {
-      case WKBConstants.wkbPoint => factory.createPoint(readCoordinate(in, csFactory))
+  def readGeometry(in: Input, factory: GeometryFactory, csFactory: CoordinateSequenceFactory): Geometry =
+    nullableReader( (in: Input, version: Int) => {
+      in.readInt(true) match {
+        case WKBConstants.wkbPoint => factory.createPoint(readCoordinate(in, csFactory))
 
-      case WKBConstants.wkbLineString => factory.createLineString(readCoordinateSequence(in, csFactory))
+        case WKBConstants.wkbLineString => factory.createLineString(readCoordinateSequence(in, csFactory))
```
anthonyccri: What's going on with these whitespace only changes?  Can we avoid them?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -466,35 +505,32 @@ object SimpleFeatureSerializer {
...
-      case WKBConstants.wkbGeometryCollection =>
-        val geoms = readGeometryCollection[Geometry](in, factory, csFactory)
-        factory.createGeometryCollection(geoms)
-    }
-  }
+        case WKBConstants.wkbGeometryCollection =>
+          val geoms = readGeometryCollection[Geometry](in, factory, csFactory)
+          factory.createGeometryCollection(geoms)
+      }
+  })(in, 1)
```
anthonyccri: Oh I see . . . this application is very confusing.  Can we create some intermediate variables here to make it easier to read?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/security/SimpleFeatureSecurityWrapper.scala
```diff
@@ -0,0 +1,33 @@
+package org.locationtech.geomesa.utils.security
```
anthonyccri: More licensing.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/security/SimpleFeatureSecurityWrapper.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.opengis.feature.simple.SimpleFeature
+
+import scala.language.implicitConversions
+
+/**
+ * Wraps a [[SimpleFeature]] allowing visibility to be accessed and modified.
+ * 
+ * Created by mmatz on 4/3/15.
+ */
+class SimpleFeatureSecurityWrapper(val sf: SimpleFeature) {
```
anthonyccri: You can use an implicit class instead of the Class+Object method:

```scala
implicit class SecurityWrappedSimpleFeature(val sf: SimpleFeature) extends SimpleFeature {
   def getViz = SecurityUtils.getViz(sf)
}```

Then to use, just import it:

```scala
import org.locationtech....SecurityWrappedSimpleFeature
val sf = ....
val viz = sf.getViz
```

elahrvivaz: Should make this class implicit and extend AnyVal - that way, it will not create actual objects, the calls will just be static methods. The class has to be moved inside an object.

mmatz-ccri: Any objections to moving the visibility methods to RichSimpleFeature?

elahrvivaz: nope, that sounds like a good idea.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
mmatz-ccri: This needed some help before I got started.  I tried to clean it up a little (by eliminating the duplicate null handling logic).  Given that its not directly related to my ticket, should I revert it?

PR #537, opened 2015-04-10
GEOMESA-741 Propagate user data when serializing SimpleFeatures

mmatz-ccri:
SimpleFeatures user data can now be propagated using kryo or avro encoding by specifying an EncodingOption when creating the FeatureEncoder.  The encoding option must be specified when creating the FeatureDecoder as well.

Text encoding has been removed.


geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/AccumuloDataStore.scala
```diff
@@ -684,11 +681,13 @@ class AccumuloDataStore(val connector: Connector,
...
   /**
-   * Reads the feature encoding from the metadata. Defaults to TEXT if there is no metadata.
+   * Reads the feature encoding from the metadata.
+   *
+   * @throws NoSuchElementException if the feature encoding is missing
    */
   def getFeatureEncoding(sft: SimpleFeatureType): FeatureEncoding = {
     metadata.read(sft.getTypeName, FEATURE_ENCODING_KEY)
-      .map(FeatureEncoding.withName).getOrElse(FeatureEncoding.TEXT)
+      .map(FeatureEncoding.withName).get
```
elahrvivaz: there's a readRequired method on metadata you can use here instead of the .get

geomesa-core/src/main/scala/org/locationtech/geomesa/core/transform/TransformCreator.scala
```diff
@@ -38,21 +38,13 @@ object TransformCreator {
...
-          defs.foreach { t => builder.set(t.name, t.expression.evaluate(feature)) }
-          val newFeature = builder.buildFeature(feature.getID)
-          encoder.encode(newFeature)
-        }
-    }
+
+    val newSf = new ScalaSimpleFeature("reusable", targetFeatureType)
+
+    (feature: SimpleFeature) => {
+        newSf.getIdentifier.setID(feature.getIdentifier.getID)
```
elahrvivaz: for future reference, you can just call feature.getID which is a shortcut to getIdentifier.getID

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/EncodingOption.scala
```diff
@@ -0,0 +1,57 @@
...
+ * Options to be applied when encoding.  The same options must be specified when decoding.
+ */
+object EncodingOption extends Enumeration {
+  type EncodingOption = Value
+
+  /**
+   * If this [[EncodingOption]] is specified then all user data of the simple feature will be
+   * serialized and deserialized.
+   */
+  val WITH_USER_DATA = Value("withUserData")
```
elahrvivaz: I would leave out the string arg to Value - that way the value will match the enum name, which is more user-friendly. Also, the convention in scala is to use camel case for constants/enums - e.g. WithUserData (not sure if we have followed that everywhere though...)

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/FeatureSpecificReader.scala
```diff
@@ -21,22 +21,22 @@ import java.util.{Date, UUID}
...
+import org.locationtech.geomesa.feature.EncodingOption.EncodingOptions
 import org.locationtech.geomesa.feature.serde.{ASFDeserializer, Version1Deserializer, Version2Deserializer}
+import org.locationtech.geomesa.feature.serialization.avro.AvroSimpleFeatureDecodingsCache
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.JavaConversions._
 
-class FeatureSpecificReader(oldType: SimpleFeatureType, newType: SimpleFeatureType)
-  extends DatumReader[AvroSimpleFeature] {
+class FeatureSpecificReader(oldType: SimpleFeatureType, newType: SimpleFeatureType, opts: EncodingOptions = EncodingOptions.none) {
```
elahrvivaz: [style] long line

elahrvivaz: In general, we use the github line length, which corresponds to about 110 characters - you can set this up in intellij under editor -> code style

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/SimpleFeatureDecodings.scala
```diff
@@ -0,0 +1,74 @@
...
+import collection.JavaConversions._
+import CacheKeyGenerator.cacheKeyForSFT
+
+import scala.ref.SoftReference
+
+
+/** Provides access to the decodings for a [[SimpleFeatureType]].
+  *
+  */
+class SimpleFeatureDecodings[Reader](val datumReaders: AbstractReader[Reader], val sft: SimpleFeatureType) {
```
elahrvivaz: This seems like a good candidate for a case class

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/FeatureSpecificReader.scala
```diff
@@ -96,9 +94,12 @@ class FeatureSpecificReader(oldType: SimpleFeatureType, newType: SimpleFeatureTy
   lazy val v1fieldreaders = buildFieldReaders(Version1Deserializer)
   lazy val v2fieldreaders = buildFieldReaders(Version2Deserializer)
 
-  def read(reuse: AvroSimpleFeature, in: Decoder): AvroSimpleFeature = {
-    // Read the version first and choose the proper deserializer
+  def defaultRead(reuse: AvroSimpleFeature, in: Decoder): AvroSimpleFeature = {
+    // read and store the version
     val serializationVersion = in.readInt()
+    AvroSimpleFeatureDecodingsCache.getAbstractReader.version = serializationVersion
```
elahrvivaz: this seems like a bad idea to update a static variable and then reference it later. Also, I believe technically the soft reference could get invalidated, and the version could get lost. Is there a better way you could expose this in the DatumReader class?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/SimpleFeatureDecodings.scala
elahrvivaz: oh yeah, the other thing you get with case classes is pattern matching unapply, e.g. 
obj match {
case SimpleFeatureDecodings(readers, sft) =>

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/FeatureSpecificReader.scala
```diff
@@ -113,6 +114,23 @@ class FeatureSpecificReader(oldType: SimpleFeatureType, newType: SimpleFeatureTy
     sf
   }
 
+  def readWithUserData(reuse: AvroSimpleFeature, in: Decoder): AvroSimpleFeature = {
+    val sf = defaultRead(reuse, in)
+
+    val ar = AvroSimpleFeatureDecodingsCache.getAbstractReader
+
+    val userData = ar.readGenericMap(in)
+    sf.getUserData.clear()
```
elahrvivaz: shouldn't need to clear the user data here, since it will be a newly-created avroSimpleFeature

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/KryoFeatureSerializer.scala
```diff
@@ -53,10 +51,14 @@ object KryoFeatureSerializer {
...
-  def apply(sft: SimpleFeatureType): KryoFeatureSerializer = apply(new SimpleFeatureSerializer(sft))
+  def apply(sft: SimpleFeatureType, options: EncodingOptions = EncodingOptions.none): KryoFeatureSerializer =
+    apply(new SimpleFeatureSerializer(sft, options))
 
-  def apply(sft: SimpleFeatureType, decodeAs: SimpleFeatureType): KryoFeatureSerializer =
-    if (sft.eq(decodeAs)) apply(sft) else apply(new TransformingSimpleFeatureSerializer(sft, decodeAs))
+  def apply(sft: SimpleFeatureType, decodeAs: SimpleFeatureType, options: EncodingOptions): KryoFeatureSerializer = {
+    if (sft.eq(decodeAs))
+      apply(sft, options)
+    else apply(new TransformingSimpleFeatureSerializer(sft, decodeAs, options))
```
elahrvivaz: good to add {} if you're going to multi-line the if.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -86,14 +80,18 @@ class FeatureIdSerializer extends Serializer[KryoFeatureId] {
  * @param sft
  * @param transform
  */
-class TransformingSimpleFeatureSerializer(sft: SimpleFeatureType, transform: SimpleFeatureType)
-    extends SimpleFeatureSerializer(sft) {
+class TransformingSimpleFeatureSerializer(sft: SimpleFeatureType, transform: SimpleFeatureType, options: EncodingOptions)
```
elahrvivaz: [style] long line

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/kryo/SimpleFeatureSerializer.scala
```diff
@@ -116,431 +112,80 @@ class TransformingSimpleFeatureSerializer(sft: SimpleFeatureType, transform: Sim
...
-    var i = 0
-    while (i < g.getNumGeometries) {
-      writeGeometry(out, g.getGeometryN(i))
-      i += 1
-    }
-  }
+  def defaultRead(kryo: Kryo, input: Input, typ: Class[SimpleFeature]): SimpleFeature = {
+    // read and store version
+    val version = input.readInt(true)
+    KryoSimpleFeatureDecodingsCache.getAbstractReader.version = version
```
elahrvivaz: same comment about finding a different way to pass the version.

elahrvivaz: default read is pretty simple... maybe just duplicate the code in the readWithVis method.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/FeatureSpecificReader.scala
mmatz-ccri: I agree that this isn't great.  However, it is not static - there is one instance per thread.  And, at least for Kryo, there is a hard reference through the "val encodings".

Initially, I was passing the version as a parameter to the DatumReader but when debugging I noticed a lot of boxing and unboxing.  And, the version is rarely used which is why I removed it.

A cleaner solution is to specify the version at the time that the attribute decodings are built and to cache the decodings for multiple versions.  I will make this change.

elahrvivaz: hmm, version is on a per-feature level... don't think we can keep them with the decodings.

mmatz-ccri: Why not?  There would be a complete set of decodings for each version needed.

read version
get decodings(version) // will be lazily created and cached
continue reading using decodings

elahrvivaz: oh, i thought decodings were created once per reader? Still working my way through the pr... If it works, that could be ok. could we get around the un/boxing by passing a java Int or something?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/AbstractReader.scala
```diff
@@ -0,0 +1,120 @@
...
+  def selectReader(clazz: Class[_], metadata: JMap[_ <: AnyRef, _ <: AnyRef] = JCollections.emptyMap(),
+                   isNullable: isNullableFn = notNullable): DatumReader[Reader, AnyRef] = {
+
+    val reader: DatumReader[Reader, AnyRef] = clazz match {
+      case cls if classOf[java.lang.String].isAssignableFrom(cls) => readString
+      case cls if classOf[java.lang.Integer].isAssignableFrom(cls) => readInt.asInstanceOf[DatumReader[Reader, AnyRef]]
+      case cls if classOf[java.lang.Long].isAssignableFrom(cls) => readLong.asInstanceOf[DatumReader[Reader, AnyRef]]
+      case cls if classOf[java.lang.Float].isAssignableFrom(cls) => readFloat.asInstanceOf[DatumReader[Reader, AnyRef]]
+      case cls if classOf[java.lang.Double].isAssignableFrom(cls) => readDouble.asInstanceOf[DatumReader[Reader, AnyRef]]
+      case cls if classOf[java.lang.Boolean].isAssignableFrom(cls) => readBoolean.asInstanceOf[DatumReader[Reader, AnyRef]]
```
elahrvivaz: [style] long lines

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/AbstractReader.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+   * @param clazz the [[Class]] of the object to be read
+   * @return a [[DatumReader]] capable of reading object of the given ``clazz``
+   */
+  def selectReader(clazz: Class[_], metadata: JMap[_ <: AnyRef, _ <: AnyRef] = JCollections.emptyMap(),
+                   isNullable: isNullableFn = notNullable): DatumReader[Reader, AnyRef] = {
+
+    val reader: DatumReader[Reader, AnyRef] = clazz match {
```
elahrvivaz: this might be more readable as an if/else, since you're not unapplying or anything.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/AbstractWriter.scala
```diff
@@ -0,0 +1,151 @@
...
+  def writeGenericMap: DatumWriter[Writer, JMap[AnyRef, AnyRef]] = (writer, map) => {
+
+    // may not be able to write all entries - must pre-filter to know correct count
+    import collection.JavaConverters.mapAsScalaMapConverter
+    val filtered = map.asScala.filter {
+      case (key, value) =>
+        if (canSerialize(key)) {
+          true
+        } else {
+          logger.warn(s"Can't serialize Map entry ($key,$value).  The map entry will be skipped.")
```
elahrvivaz: i would reduce this logging level - aren't some of the likely keys like PROVIDED_FID not going to be serialized?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/AbstractWriter.scala
```diff
@@ -0,0 +1,151 @@
...
+  /*
+    /**
+   * Finds an encoding function based on the input type
+   *
+   * @param clas
+   * @param metadata
+   * @return
+   */
+  def matchEncode(clas: Class[_], metadata: JMap[AnyRef, AnyRef]): AttributeWriter = clas match {
+   */
```
elahrvivaz: is this commented out? can remove it.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/CollectionWriter.scala
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+package org.locationtech.geomesa.feature.serialization
+
+import java.util.{List => JList, Map => JMap}
+
+/** [[DatumReader]]s for reading [[java.util.Map]]s and [[java.util.List]]s.
+ *
+ */
+trait CollectionWriter[Writer] extends PrimitiveWriter[Writer] {
```
elahrvivaz: does it make sense to combine the reader/writer classes, so that the read and write logic is next to each other? or maybe just put them in the same scala file?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/SimpleFeatureEncodings.scala
```diff
@@ -0,0 +1,74 @@
...
+   * @return a seq of functions to encode the attributes of simple feature
+   */
+  lazy val attributeEncodings: Array[AttributeEncoding] =
+    sft.getAttributeDescriptors.zipWithIndex.toArray.map { case (d, i) =>
+
+      // This method is needed to capture the type of the attribute in order to type the writer.
+      @inline def encode[T](attribClass: Class[T]): AttributeEncoding = {
+        val attribWriter: DatumWriter[Writer, T] =
+          datumWriters.selectWriter(attribClass, d.getUserData, datumWriters.standardNullable)
+        (writer: Writer, sf: SimpleFeature) => attribWriter(writer, attribClass.cast(sf.getAttribute(i)))
```
elahrvivaz: could you do sf.getAttribute(i).asInstanceOf[T] instead?

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/SimpleFeatureEncodings.scala
```diff
@@ -0,0 +1,74 @@
...
+  protected def datumWritersFactory: () => AbstractWriter[Writer]
+
+  /** Gets a sequence of functions to encode the attributes of a simple feature. */
+  def get(sft: SimpleFeatureType): SimpleFeatureEncodings[Writer] =
+    encodingsCache.getOrElseUpdate(cacheKeyForSFT(sft), {
+      new SimpleFeatureEncodings[Writer](getAbstractWriter, sft)
+    })
+  
+  def getAbstractWriter: AbstractWriter[Writer] = {
+    Option(writersCache.get()).flatMap(_.get).getOrElse {
```
elahrvivaz: might be useful to pull out a soft reference thread local as a class, similar to the cache class but without a key - i've found myself using that in places.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/avro/AvroSimpleFeatureEncodingsCache.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.locationtech.geomesa.feature.serialization.avro
+
+import org.apache.avro.io.Encoder
+import org.locationtech.geomesa.feature.serialization.{AbstractWriter, SimpleFeatureEncodingsCache}
+
+/** Concrete [[SimpleFeatureEncodingsCache]] for Avro. */
+object AvroSimpleFeatureEncodingsCache extends SimpleFeatureEncodingsCache[Encoder] {
+
+  override protected val datumWritersFactory: () => AbstractWriter[Encoder] = () => new AvroWriter()
```
elahrvivaz: can probably combine some of these classes into 1 file - it's kind of a fine line in scala - some people go overboard and put everything into 1 file.

geomesa-feature/src/main/scala/org/locationtech/geomesa/feature/serialization/AbstractWriter.scala
mmatz-ccri: I made USE_PROVIDE_FID serializable with the ability to add additional hint keys as we find that we need them.

I thought that we would make at least the most common keys serializable.  So, anything that isn't serializable would be rare.  Is this not the case?

elahrvivaz: There's another key named 'PROVIDED_FID' that also gets used - i don't think it's normally part of our workflow, but it's pretty common... not sure if there are others or not.

elahrvivaz: maybe we should just serialize that one too

elahrvivaz:
looking good - did you run any timing tests compared to the old version?

mmatz-ccri:
I have not run any timing tests yet.  Do we have any tools or standard set of data for timing tests?

elahrvivaz:
no, just use that unit test i pointed to. it doesn't have to be anything exhaustive, just make sure it's not obviously slower.

mmatz-ccri:
I ran org.locationtech.geomesa.feature.SerializationTester 3 times with the old code and 3 times with the new code on my local machine.

**Before**

Run | Kryo    | Avro
----- | -------- | --------
    1 | 19.510 | 43.450
    2 | 19.226 | 41.193
    3 | 19.228 | 40.841
Avg | 19.321 | 41.828

**After**

Run | Kryo    | Avro
----- | -------- | --------
    1 | 19.373 | 41.844
    2 | 19.385 | 41.054
    3 | 19.777 | 42.467
Avg | 19.512 | 41.788

elahrvivaz:
Cool, looks within the margin of error.

elahrvivaz:
merged as e18b151bc000e439b627248871b79ba2c9229110

PR #538, opened 2015-04-10
Histogram Job results are written to accumulo in a {featureName}_stats table

jdk2pq:
For comments only. I'm probably doing things wrong.

### DIFF
--- a/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
+++ b/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
@@ -19,7 +19,9 @@ package org.locationtech.geomesa.jobs.analytics
 import com.twitter.algebird.Aggregator
 import com.twitter.scalding._
 import com.twitter.scalding.typed.UnsortedGrouped
+import org.apache.accumulo.core.data.{Mutation, Range => AcRange, Value}
 import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.io.Text
 import org.geotools.data.DataStoreFinder
 import org.locationtech.geomesa.core.data._
 import org.locationtech.geomesa.jobs.GeoMesaBaseJob
@@ -30,7 +32,6 @@ import org.locationtech.geomesa.jobs.scalding._
 import org.opengis.feature.simple.SimpleFeature
 
 import scala.collection.JavaConversions._
-import scala.collection.JavaConverters._
 import scala.math.Ordering
 
 /**
@@ -47,14 +48,18 @@ class HistogramJob(args: Args) extends GeoMesaBaseJob(args) {
   val groupBy    = args.nonStrictList(GROUP_BY)
   val uniqueBy   = args.nonStrictList(UNIQUE_BY)
   val transforms = Option(args.list(TRANSFORM_IN).toArray).filter(_.nonEmpty)
+  val writeToAccumulo  = args.boolean(WRITE_TO_ACCUMULO)
 
   val input = GeoMesaInputOptions(dsParams, feature, filter, transforms)
+  lazy val dsParamsAccOutput = dsParams.updated("tableName", s"${dsParams.get("tableName").get}_stats")
+  lazy val accOutput = AccumuloOutputOptions(dsParamsAccOutput)
+
   val output = args(FILE_OUT)
 
   // TODO we could look up attribute values by index for performance gain
   // verify input params - inside a block so they don't get serialized
   {
-    val ds = DataStoreFinder.getDataStore(dsParams.asJava).asInstanceOf[AccumuloDataStore]
+    val ds = DataStoreFinder.getDataStore(dsParams).asInstanceOf[AccumuloDataStore]
     val sft = ds.getSchema(feature)
     assert(sft != null, s"The feature '$feature' does not exist in the input data store")
     val descriptors = sft.getAttributeDescriptors.map(_.getLocalName)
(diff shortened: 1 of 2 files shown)


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
```diff
@@ -47,14 +48,18 @@ class HistogramJob(args: Args) extends GeoMesaBaseJob(args) {
   val groupBy    = args.nonStrictList(GROUP_BY)
   val uniqueBy   = args.nonStrictList(UNIQUE_BY)
   val transforms = Option(args.list(TRANSFORM_IN).toArray).filter(_.nonEmpty)
+  val writeToAccumulo  = args(WRITE_TO_ACCUMULO).toBoolean
 
   val input = GeoMesaInputOptions(dsParams, feature, filter, transforms)
+  val dsParamsAccOutput = dsParams.updated("tableName", s"${dsParams.get("tableName").get}_stats")
+  val accOutput = AccumuloOutputOptions(dsParamsAccOutput, createTable = true)
```
elahrvivaz: might make these lazy for the non-accumulo output case.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
```diff
@@ -47,14 +48,18 @@ class HistogramJob(args: Args) extends GeoMesaBaseJob(args) {
   val groupBy    = args.nonStrictList(GROUP_BY)
   val uniqueBy   = args.nonStrictList(UNIQUE_BY)
   val transforms = Option(args.list(TRANSFORM_IN).toArray).filter(_.nonEmpty)
+  val writeToAccumulo  = args(WRITE_TO_ACCUMULO).toBoolean
```
elahrvivaz: use args.boolean

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
```diff
@@ -151,11 +166,13 @@ object HistogramJob {
              feature: String,
              attribute: String,
              groupBy: List[String] = List.empty,
-             uniqueBy: List[String] = List.empty) = {
+             uniqueBy: List[String] = List.empty,
+             writeToAccumulo: List[String] = List.empty) = {
```
elahrvivaz: can make this a boolean and just put it into a list for the map below (scalding requires everything to be a list, but this interface doesn't have to reflect that)

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
```diff
@@ -151,11 +166,13 @@ object HistogramJob {
...
     val args = Seq(FEATURE_IN -> List(feature),
-                   ATTRIBUTE  -> List(attribute),
-                   GROUP_BY   -> groupBy,
-                   UNIQUE_BY  -> uniqueBy).toMap ++ toInArgs(dsParams)
+      ATTRIBUTE  -> List(attribute),
+      GROUP_BY   -> groupBy,
+      UNIQUE_BY  -> uniqueBy,
+      WRITE_TO_ACCUMULO -> writeToAccumulo).toMap ++ toInArgs(dsParams)
```
elahrvivaz: [picky] line up ->

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/scalding/AccumuloSourceOptions.scala
```diff
@@ -26,21 +26,13 @@ import org.apache.log4j.Level
...
 sealed trait AccumuloSourceOptions {
-  def instance: String
-  def zooKeepers: String
-  def user: String
-  def password: String
-  def table: String
+  def dsParams: Map[String, String]
   def logLevel: Option[Level]
-  override def toString = s"${getClass.getSimpleName}[$instance,$table]"
+  override def toString = s"${getClass.getSimpleName}[${dsParams.getOrElse("instanceId", "None")},${dsParams.getOrElse("tableName", "None")}]"
```
elahrvivaz: i would roll back these changes and keep things as a simple case class, instead of a wrapper for a map - that way we get easy access to each param, instead of all hte 'getOrElse' with strings. (also, the accumulo formats shouldn't be tied to geomesa at all). To get the same functionality, add an object for each case class with an apply method that takes a Map and pulls out the appropriate values.

elahrvivaz:
Looks good - have you discussed the stat table layout with @jnh5y? Also, I would revert the changes to the AccumuloSourceOptions classes and just add an apply method that handles data store maps.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/scalding/AccumuloSourceOptions.scala
```diff
@@ -51,6 +51,35 @@ case class AccumuloInputOptions(
...
+            localIterators: Option[Boolean]   = None,
+            offlineTableScan: Option[Boolean] = None,
+            scanIsolation: Option[Boolean]    = None,
+            logLevel: Option[Level]           = None) = {
+    new AccumuloInputOptions(
+      dsParams.getOrElse("instanceId", "None"),
+      dsParams.getOrElse("zookeepers", "None"),
+      dsParams.getOrElse("user", "None"),
+      dsParams.getOrElse("password", "None"),
+      dsParams.getOrElse("table", "None"),
```
elahrvivaz: i would change default to null

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
```diff
@@ -96,16 +101,26 @@ class HistogramJob(args: Args) extends GeoMesaBaseJob(args) {
   val aggregates: UnsortedGrouped[Product, Long] = groups.aggregate(Aggregator.size)
 
   // write out the histogram to a tsv file - we create the tabs ourselves since length isn't known
-  aggregates.toTypedPipe.map { case (group, count) => s"${group.productIterator.mkString("\t")}\t$count" }
+  if (writeToAccumulo) {
+    aggregates.toTypedPipe
+      .flatMap { case (group, count) =>
+      val mut = new Mutation(new Text(feature))
+      mut.put(EMPTY_COLF, new Text(s"${group.productIterator.mkString("\t")}"), new Value(count.toString.getBytes))
+      Seq(mut).map((null: Text, _))
```
elahrvivaz: so i think for the stats use case, the group will be just the attribute we are histogramming. so how about:
row: feature~attribute~group.last
cf: group(0 to n-2) as a string (maybe ~ separated to be consistent with our other schemas?) - will usually be empty
cq: count
value: empty
I think we would also want things like:
feature~attribute::cardinality::<cardinality>
feature~attribute::min::<min>
feature~attribute::max::<max>
but those might be for a different PR

jdk2pq:
Closing in favor of https://github.com/locationtech/geomesa/pull/567

PR #539, opened 2015-04-10
GEOMESA-765 The '--force' parameter does not appear to work for the "del...

cne1x:
...etecatalog" command

One-line fix, verified locally.


AndrewAnnex:
Committed as: 2fd9d432233cb3209bdba0d9a78c23bf5f908b4a

PR #541, opened 2015-04-14
GEOMESA-766 Separate WPS processes so that users don't have to deploy th...

anthonyccri:
...e gs-wps plugin when installing the geomesa plugin


geomesa-plugin/pom.xml
```diff
@@ -188,16 +188,6 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.geoserver.extension</groupId>
```
jnh5y: Great!

geomesa-wps/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-core-accumulo1.5</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geoserver.extension</groupId>
+            <artifactId>gs-wps-core</artifactId>
+        </dependency>
```
jnh5y: Thanks for this; it'll make it easier to ship over the necessary components to the GeoServer module.

geomesa-wps/src/main/scala/org/locationtech/geomesa/process/ProcessFactory.scala
```diff
@@ -15,7 +15,7 @@
  */
 
 
-package org.locationtech.geomesa.plugin.process
+package org.locationtech.geomesa.process
```
jnh5y: Great name change.

elahrvivaz: should we try to make the module name and package name consistent? I think all our other modules are like that.

geomesa-wps/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+
```
elahrvivaz: So the wps jar will depend on the plugin jar? since it doesn't make a shaded jar.
We should add a README with install instructions/update the docs/tutorials/whatever.

anthonyccri: In general, I think we should move away from shaded jars for anything other than the distributed runtime.  We should follow the model of geoserver wps extension which is distributed as a zip with all the dependencies packaged in there.  Thoughts?

anthonyccri: But yes, a README for this project seems like a good thing.

geomesa-wps/src/main/scala/org/locationtech/geomesa/process/ProcessFactory.scala
anthonyccri: Good point.  I'll make that change.

geomesa-wps/pom.xml
elahrvivaz: yeah, i think a zip would be good... esp now that we have multiple geoserver jars, there would be a lot of overlapping dependencies.

elahrvivaz: we could even have additional zips for accumulo 1.5.2, hadoop 2.x, etc to make it easier for people to install the required dependencies.

elahrvivaz:
merged as ae1e44d5322a611ebf4d07429b2b680379ff504f

PR #546, opened 2015-04-17
GEOMESA-774,775 Fix CAGH and Return null for invalid Mosaic

AndrewAnnex:
* Made getQueryPlan return a Option[QueryPlan] to deal with invalid queries.
* Fixed closest acceptable geohash to return None when given a bad GeoHash.
* The GeoMesaCoverageReader now returns nulls coverages when mosaic could not be constructed.


AndrewAnnex:
merged in 9ea579372fc4890e20a15a00959cddd94e3fe952

PR #547, opened 2015-04-17
WIP First pass at making stat collection optional, configurable via checkbox...

AndrewAnnex:
... in GeoServer Store page. For comments only.


AndrewAnnex:
So I need to test the latest commit but this is looking much better. The only thing I am thinking of changing now is that maybe the collect stats should be a property added to the table config instead of using the check-box in the geoserver UI. That would mean probably adding a tool that enables/disables collecting stats for a raster table.

AndrewAnnex:
merged in 93984a0f7de8403782815980c673bf4a03d6c931

PR #549, opened 2015-04-20
GEOMESA-742 add a visibility filter as a security extension to GeoServer

mmatz-ccri:
Visibility filtering can be applied using combination of the DataStoreSecurityService by the data store, the geomesa-web-security plugin, and GeoServer configuration.  See README.md for more details.


anthonyccri:
Merged

PR #551, opened 2015-04-21
GEOMESA-781 Fix non-Accumulo general shapefile ingest

jwkessi:
Issue was being caused by datastores throwing exceptions when calling getSchema when the schema did not exist, rather than returning null.


AndrewAnnex:
This PR looks good to me.

AndrewAnnex:
committed as: aa16c4f5829aea0b77ac1f94646fdbf80cea66e6

PR #554, opened 2015-04-22
GEOMESA-782 improve query plan for multiple attribute queries

mmatz-ccri:
AttributeIdxStrategy now selects the best (lowest cost) attribute filter as the scan range.

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
@@ -205,6 +205,70 @@ object FilterHelper {
       }
     }
 
+  /**
+   * Finds the filter with the lowest known cost and returns the rest in the same order they were in.  If
+   * there are multiple filters with the same lowest cost then the first will be selected.  If no filters
+   * have a known cost or if ``s`` is empty then (None, s) will be returned.
+   */
+  def findBest(cost: Filter => Option[Long])(s: Seq[Filter]): CostAnalysis = {
+    if (s.isEmpty) {
+      CostAnalysis.unknown(s)
+    } else {
+      val head = s.head
+      val tail = s.tail
+
+      val headAnalysis = cost(head).map(c => new KnownCost(head, tail, c)).getOrElse(CostAnalysis.unknown(s))
+      val tailAnalysis = findBest(cost)(tail)
+
+      if (headAnalysis <= tailAnalysis) {
+        headAnalysis
+      } else {
+        // tailAnaysis must have a known cost
+        val ta = tailAnalysis.asInstanceOf[KnownCost]
+        new KnownCost(ta.best, head +: ta.otherFilters, ta.cost)
+      }
+    }
+  }
+
+  /**
+    * @param bestFilter the [[Filter]] with the lowest cost
+    * @param otherFilters all other [[Filter]]s
+    */
+  sealed abstract case class CostAnalysis(bestFilter: Option[Filter], otherFilters: Seq[Filter]) {
+
+    /**
+      * @param rhs the [[CostAnalysis]] to compare to
+      * @return ``true`` if ``this`` has a lower or the same cost as ``rhs``
+      */
+    def <=(rhs: CostAnalysis): Boolean
+
+    def extract: (Option[Filter], Seq[Filter]) = (bestFilter, otherFilters)
+  }
+
+  class KnownCost(val best: Filter, others: Seq[Filter], val cost: Long) extends CostAnalysis(Some(best), others) {
+
+    def <=(rhs: CostAnalysis): Boolean = rhs match {
+      case knownRhs: KnownCost =>
+        this.cost <= knownRhs.cost
+      case _ =>
+        // always less than an unknown cost
+        true
+    }
+  }
+
+  class UnknownCost(filters: Seq[Filter]) extends CostAnalysis(None, filters) {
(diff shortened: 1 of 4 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
```diff
@@ -205,6 +205,41 @@ object FilterHelper {
...
+   * Finds the filter with the lowest known cost and returns the rest in the same order they were in.  If
+   * there are multiple filters with the same lowest cost then the first will be selected.  If no filters
+   * have a known cost or if ``s`` is empty then (None, s) will be returned.
+   */
+  def findBest(cost: Filter => Option[Long])(s: Seq[Filter]): (Option[Filter], Seq[Filter]) = {
+    val (best, rest, _) = findBestWithCost(cost, s)
+    (best, rest)
+  }
+
+  private def findBestWithCost(cost: Filter => Option[Long], s: Seq[Filter]): (Option[Filter], Seq[Filter], Option[Long]) = {
```
anthonyccri: Instead of returning tuples, create a new case class.  Maybe ```CostAnalysis```.  It'll be more readable that way.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
```diff
@@ -205,6 +205,39 @@ object FilterHelper {
       }
     }
 
+  /**
+   * Finds the filter with the lowest known cost and returns the rest in the same order they were in.
```
AndrewAnnex: is there a particular reason that *rest* portion is not also sorted best to worse?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
```diff
@@ -205,6 +205,41 @@ object FilterHelper {
...
+  }
+
+  private def findBestWithCost(cost: Filter => Option[Long], s: Seq[Filter]): (Option[Filter], Seq[Filter], Option[Long]) = {
+    if (s.isEmpty) {
+      (None, s, None)
+    } else {
+      val head = s.head
+      val tail = s.tail
+
+      val (tailBest, tailOther, tailCost) = findBestWithCost(cost, tail)
```
anthonyccri: Would a minBy work?

```scala
val (filterWithMinCost, minCost) = s.zipWith(s.map { filter => cost(filter) }).minBy { case (_, cost) => cost }
```

geomesa-core/src/test/scala/org/locationtech/geomesa/core/index/AttributeIndexStrategyTest.scala
```diff
@@ -129,6 +130,29 @@ class AttributeIndexStrategyTest extends Specification {
       success
     }
 
+    "partition a query by selecting the best filter" >> {
```
AndrewAnnex: It may be worth adding a few more unit tests like this that demonstrate this working like the cases you added in FilterHelperTest but this could be considered redundant.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
mmatz-ccri: I was following the pattern of findFirst.  Should I replace findBest with sortBy?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
mmatz-ccri: That could work, but I still need to get the rest.  I could just sort but it would be more expensive.  Is there an advantage to having the rest sorted?

anthonyccri: No, I didn't realize you had to save all the rest.  Method is fine as is.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
mmatz-ccri: OK.  I'll change it.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
```diff
@@ -205,6 +205,70 @@ object FilterHelper {
...
+    /**
+      * @param rhs the [[CostAnalysis]] to compare to
+      * @return ``true`` if ``this`` has a lower or the same cost as ``rhs``
+      */
+    def <=(rhs: CostAnalysis): Boolean
+
+    def extract: (Option[Filter], Seq[Filter]) = (bestFilter, otherFilters)
+  }
+
+  class KnownCost(val best: Filter, others: Seq[Filter], val cost: Long) extends CostAnalysis(Some(best), others) {
```
anthonyccri: Usually, I make these case classes.

mmatz-ccri: I struggled with this a little.  I could make the parent class a case class or the child classes but not both and it wasn't clear to me which way was best.

anthonyccri: Ah, parent should be a sealed trait

PR #557, opened 2015-04-24
WIP whole world query STII configuration fix

AndrewAnnex:
added more unit tests to make this into a WIP

Build is behaving strangely locally, it looks to be locking up in the accumulodatastoretest.

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
@@ -81,13 +81,13 @@ object FilterHelper {
       case _ => false
     }
 
-  private def isOperationGeomWholeWorld[O <: BinarySpatialOperator](op: O): Boolean = {
+  private def isOperationGeomWholeWorld[Op <: BinarySpatialOperator](op: Op): Boolean = {
     val e2 = op.getExpression2.asInstanceOf[Literal]
     val geom = e2.evaluate(null, classOf[Geometry])
     isWholeWorld(geom)
   }
 
-  def isWholeWorld(g: Geometry): Boolean = g.covers(IndexSchema.everywhere)
+  def isWholeWorld[G <: Geometry](g: G): Boolean = g.covers(IndexSchema.everywhere)
 
   def getGeometryListOf(inMP: Geometry): Seq[Geometry] =
     for( i <- 0 until inMP.getNumGeometries ) yield inMP.getGeometryN(i)
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
@@ -73,6 +73,22 @@ object FilterHelper {
       ff.or(filterList)
   }
 
+  def isFilterWholeWorld(f: Filter): Boolean = f match {
+      case op: BBOX       => isOperationGeomWholeWorld(op)
+      case op: Within     => isOperationGeomWholeWorld(op)
+      case op: Intersects => isOperationGeomWholeWorld(op)
+      case op: Overlaps   => isOperationGeomWholeWorld(op)
+      case _ => false
+    }
+
+  private def isOperationGeomWholeWorld[O <: BinarySpatialOperator](op: O): Boolean = {
+    val e2 = op.getExpression2.asInstanceOf[Literal]
+    val geom = e2.evaluate(null, classOf[Geometry])
+    isWholeWorld(geom)
+  }
+
+  def isWholeWorld(g: Geometry): Boolean = g.covers(IndexSchema.everywhere)
+
   def getGeometryListOf(inMP: Geometry): Seq[Geometry] =
     for( i <- 0 until inMP.getNumGeometries ) yield inMP.getGeometryN(i)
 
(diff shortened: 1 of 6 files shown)


geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
```diff
@@ -73,6 +73,22 @@ object FilterHelper {
...
 
+  def isFilterWholeWorld(f: Filter): Boolean = f match {
+      case op: BBOX       => isOperationGeomWholeWorld(op)
+      case op: Within     => isOperationGeomWholeWorld(op)
+      case op: Intersects => isOperationGeomWholeWorld(op)
+      case op: Overlaps   => isOperationGeomWholeWorld(op)
+      case _ => false
+    }
+
+  private def isOperationGeomWholeWorld[O <: BinarySpatialOperator](op: O): Boolean = {
```
cne1x: [minor style]  I would avoid ```O``` as a single-letter identifier.

mmatz-ccri:
Will this catch or'd geometries whose union is the whole world?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
```diff
@@ -119,7 +119,19 @@ class STIdxStrategy extends Strategy with Logging with IndexFilterHelpers {
...
   }
 
-  def getSTIIIterCfg(iteratorConfig: IteratorConfig,
+  def decomposeToGeometry(f: Filter): Seq[Geometry] = f match {
+    case bbox: BBOX =>
+      val bboxPoly = bbox.getExpression2.asInstanceOf[Literal].evaluate(null, classOf[Geometry])
+      Seq(bboxPoly)
+    case gf: BinarySpatialOperator =>
+      extractGeometry(gf)
+    case or: OrImpl =>
```
cne1x: I thought that, before we could get to this routine, the top-most query had been transformed into DNF, and ```getQueryPlan``` was called on each of the top-level children.  If that's so, how can there be an ```OrImpl``` here?

geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreTest.scala
```diff
@@ -1155,6 +1153,61 @@ class AccumuloDataStoreTest extends Specification {
...
+        o.toString()
+      }
+      explain must contain("No STII Filter")
+      explain must contain("Filter: AcceptEverythingFilter")
+    }
+
+    "create key plan that does not use STII when given an or'd geometry query both of the whole world" in {
+      val sftName = "explainLargeBBOXTest_4"
+      val sft3 = createSchema(sftName)
+      val filter = CQL.toFilter("bbox(geom, -180, -90, 180, 90) OR bbox(geom, -180, -90, 180, 90)")
```
cne1x: I'm not certain this test is the one we want.  Wouldn't it be better if each of the two geometries were smaller than the entire world, but their union covered the (lat, lon) space?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
AndrewAnnex: currently other unit tests fail without this addition. I am not familiar with the STIdxStartegy so I don't know how this process worked in the past. The two unit tests I refer to fail because the previous mutations to the geometry filter fail to split on the "or" or fail to properly union the geometries. This sounds like a regression to me.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/FilterHelper.scala
```diff
@@ -73,6 +73,22 @@ object FilterHelper {
...
+  def isFilterWholeWorld(f: Filter): Boolean = f match {
+      case op: BBOX       => isOperationGeomWholeWorld(op)
+      case op: Within     => isOperationGeomWholeWorld(op)
+      case op: Intersects => isOperationGeomWholeWorld(op)
+      case op: Overlaps   => isOperationGeomWholeWorld(op)
+      case _ => false
+    }
+
+  private def isOperationGeomWholeWorld[Op <: BinarySpatialOperator](op: Op): Boolean = {
+    val e2 = op.getExpression2.asInstanceOf[Literal]
```
elahrvivaz: i don't think expresssion 2 is guaranteed to be the literal... might want to use the checkOrder function

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/IteratorTrigger.scala
```diff
@@ -110,10 +108,10 @@ object IteratorTrigger extends Logging {
       // if the transforms exist, check if the transform is simple enough to be handled by the IndexIterator
       // if it does not exist, then set this variable to false
       val isIndexTransform = transformDefs
-          .map(tDef => isOneToOneIndexTransformation(tDef, sft, indexedAttribute))
-          .orElse(Some(false))
+        .map(tDef => isOneToOneIndexTransformation(tDef, sft, indexedAttribute))
+        .orElse(Some(false))
```
elahrvivaz: in general, i would avoid changing lines just for formatting.

elahrvivaz:
We should also add some logic in the query strategy decider - if the geometry filter is a 'whole world filter', then we don't want to pick the stidx strategy - see org.locationtech.geomesa.core.index.STIdxStrategy#getStrategy

AndrewAnnex:
would the logic for the test need to properly handle Or'd geometry ?

PR #559, opened 2015-04-27
GEOMESA-787 Adding support for UDP stream sources

elahrvivaz:
* Source route must be in the form 'netty4:udp://\<host\>:\<port\>?textline=true'
* Each message must end with a newline character
* Each message must contain exactly one line - extra lines will be dropped
* Packet size can be controlled by the route parameter 'receiveBufferSize'
  * If the message is larger than the packet size then the message will be truncated


anthonyccri:
Merged!

PR #562, opened 2015-04-30
GEOMESA-789 Make SCM and distribution-management properties

cne1x:
Reviewer/committer:  Please also apply this change to the RC-6 release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,10 +65,20 @@
         <log4j.version>1.2.17</log4j.version>
 
         <!-- distribution properties -->
+        <dist.releases.id>repo.locationtech.org</dist.releases.id>
+        <dist.releases.name>GeoMesa Repository - Releases</dist.releases.name>
+        <dist.releases.url>https://repo.locationtech.org/content/repositories/geomesa-releases/</dist.releases.url>
+        <dist.snapshots.id>repo.locationtech.org</dist.snapshots.id>
+        <dist.snapshots.name>GeoMesa Repository - Snapshots</dist.snapshots.name>
+        <dist.snapshots.url>https://repo.locationtech.org/content/repositories/geomesa-snapshots/</dist.snapshots.url>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -65,11 +65,21 @@
...
+        <dist.snapshots.name>GeoMesa Repository - Snapshots</dist.snapshots.name>
+        <dist.snapshots.url>https://repo.locationtech.org/content/repositories/geomesa-snapshots/</dist.snapshots.url>
         <dist.accumulo.lib.ext>accumulo/lib/ext</dist.accumulo.lib.ext>
         <dist.hdfs>hdfs</dist.hdfs>
         <dist.geoserver.lib>geoserver/WEB-INF/lib</dist.geoserver.lib>
         <dist.dev>dev</dist.dev>
 
+        <!-- SCM properties -->
+        <scm.connection>scm:git:git@gitlab.devforce.io:tcri-geospatial/tcri-geospatial-geoserver-authorizations-provider.git</scm.connection>
+        <scm.url>https://gitlab.devforce.io/tcri-geospatial/tcri-geospatial-geoserver-authorizations-provider.git</scm.url>
```
elahrvivaz: this should point to locationtech, right?

anthonyccri:
Can we use the -DaltDeploymentRepository instead of this?

cne1x:
@anthonyccri I assume that we could, but I admit that ```-DaltDeploymentRepository``` feels clunky to me, even more than the proposed solution here (for what little that's worth), and it leaves the SCM hard-wired.  (We could certainly use ```-Dproject.scm.developerConnection``` and its ilk for the SCM on the command-line, too.)  I think the current PR improves readability for folks who want their own private GeoMesa repo/SCM, but I'm easy either way.  Thanks!

anthonyccri:
Seems like ```-DaltDeploymentRepository``` is the standard way to do this sort of thing ... I don't think I've ever seen a project use properties in SCM settings.

cne1x:
Withdrawn as per discussion, above.

PR #568, opened 2015-05-05
GEOMESA-796 Enable user-specific roles in GeoServer

drmathochist:
A trait to mix into a GeoServerRoleService to create a GS role for each user on the fly,
which can be used to secure layers.

### DIFF
--- a/geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
+++ b/geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
@@ -1,0 +1,113 @@
+package org.locationtech.geomesa.plugin.security
+
+import org.geoserver.security.GeoServerRoleService
+import org.geoserver.security.config.SecurityNamedServiceConfig
+import org.geoserver.security.impl.GeoServerRole
+import org.geoserver.security.xml.{XMLSecurityProvider, XMLRoleService}
+
+import java.util.{SortedSet => jSortedSet, TreeSet => jTreeSet, Map => jMap, HashMap => jHashMap, Collections}
+
+import scala.collection.mutable
+import scala.collection.JavaConversions._
+
+object UserNameRoles {
+  private val isUserRoleProp = "IS_USER_ROLE"
+  private def isUserRole(role: GeoServerRole) = role.getProperties.getProperty("IS_USER_ROLE") == "true"
+  private def userRole(username: String) = {
+    val r = new GeoServerRole(username)
+    r.setUserName(username)
+    r.getProperties.setProperty(isUserRoleProp, "true")
+    r
+  }
+}
+
+import UserNameRoles.{userRole, isUserRole}
+
+trait UserNameRoles extends GeoServerRoleService {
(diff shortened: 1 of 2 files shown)


geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+}
+
+class XMLUserNameRoleService extends XMLRoleService with UserNameRoles
+
+// The RoleService seems to be deployed from the SecurityProvider, so this will replace
+// that with one that creates an XMLUserNameRoleService instead of an XMLRoleService
+// In turn, the SecurityProvider is deployed as a bean in the gs-main JAR, specified in
+// applicationSecurityContext.xml.  To replace it
+// 1) unzip the gs-main JAR into a spare directory
```
jahhulbert-ccri: is there a way to do this with SPI or something...I don't believe we can go around zipping/unzipping jar files in production easily.

geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
```diff
@@ -0,0 +1,113 @@
+package org.locationtech.geomesa.plugin.security
```
jahhulbert-ccri: header

geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
```diff
@@ -0,0 +1,113 @@
...
+  // the user's roles ALSO include a new role solely for their use.
+  // this must be created on the fly with the user, and since standard (XML)
+  // RoleServices don't write back to their backing configuration
+  // we need everything to happen as a layer on top of the configured answers
+  abstract override def getRolesForUser(username: String): jSortedSet[GeoServerRole] = {
+    import collection.JavaConversions._
+    val roles = new jTreeSet[GeoServerRole]()
+    roles.addAll(super.getRolesForUser(username))
+    roles.add(storedUserRoles(username))
+    println(s"Retrieved roles for $username: ${roles.map(_.toString).mkString(",")}")
```
elahrvivaz: println

geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
elahrvivaz: However this ends up being configured, the instructions should be in the readme as well.

jnh5y:
Git history needs to be updated to change 'youremail@ccri.com'.

geomesa-plugin/src/main/scala/org/locationtech/geomesa/plugin/security/UserNameRoles.scala
drmathochist: The short answer is "I have no idea, but if someone else DOES I'm glad to hear it."

PR #570, opened 2015-05-07
Added wps-unique functionality.

sivakom:
Signed-off-by: sivakom <sivakom@utexas.edu>

### DIFF
--- a/geomesa-ui/src/main/ui/bower.json
+++ b/geomesa-ui/src/main/ui/bower.json
@@ -14,6 +14,7 @@
     "lodash": "~2.4.1",
     "moment": "~2.7.0",
     "moment-timezone": "~0.1.0",
-    "openlayers" : "https://github.com/openlayers/ol3/releases/download/v3.2.1/v3.2.1.zip"
+    "openlayers" : "https://github.com/openlayers/ol3/releases/download/v3.2.1/v3.2.1.zip",
+    "openlayers2": "https://github.com/openlayers/openlayers/releases/download/release-2.13.1/OpenLayers-2.13.1.zip"
   }
 }
--- a/geomesa-ui/src/main/ui/src/app/unique/unique.js
+++ b/geomesa-ui/src/main/ui/src/app/unique/unique.js
@@ -1,15 +1,16 @@
 angular.module('geomesa.unique', [])
 
-    .controller('UniqueController', ['$scope', function($scope) {
+    .controller('UniqueController', ['$scope', '$http', function($scope, $http) {
         var request = createWPSExecuteRequest('Who', 'true', 'ASC', 'true');
-        OpenLayers.Request.POST({
+        
+        $http({
+            method: 'POST',
             url: 'http://geomesa:8080/geoserver/wps',
-            data: request,
-            success: function(outputs){
-                //For now just log results to console.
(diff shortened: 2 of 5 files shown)


geomesa-ui/src/main/ui/bower.json
```diff
@@ -14,6 +14,7 @@
     "lodash": "~2.4.1",
     "moment": "~2.7.0",
     "moment-timezone": "~0.1.0",
-    "openlayers" : "https://github.com/openlayers/ol3/releases/download/v3.2.1/v3.2.1.zip"
+    "openlayers" : "https://github.com/openlayers/ol3/releases/download/v3.2.1/v3.2.1.zip",
+    "openlayers2": "https://github.com/openlayers/openlayers/releases/download/release-2.13.1/OpenLayers-2.13.1.zip"
```
jdk2pq: Do openlayers 2 and 3 play nicely with each other?

geomesa-ui/src/main/ui/src/app/unique/unique.js
```diff
@@ -0,0 +1,70 @@
+angular.module('geomesa.unique', [])
+
+    .controller('UniqueController', ['$scope', function($scope) {
+        var request = createWPSExecuteRequest('Who', 'true', 'ASC', 'true');
+        OpenLayers.Request.POST({
```
mforkin: Is it necessary to use the OpenLayers post function. Could we use $http or $resource instead? It would be nice to be as decoupled from openlayers as possible.

geomesa-ui/src/main/ui/bower.json
```diff
@@ -14,6 +14,7 @@
     "lodash": "~2.4.1",
     "moment": "~2.7.0",
     "moment-timezone": "~0.1.0",
-    "openlayers" : "https://github.com/openlayers/ol3/releases/download/v3.2.1/v3.2.1.zip"
```
mforkin: Should we include both openlayers 2 & 3? I'm guessing we added both b\c 2 has a helper function that 3 doesn't, but if there was another way to get around this I'd vote for that.

geomesa-ui/src/main/ui/src/app/unique/unique.js
sivakom: Yea I can fix that.

geomesa-ui/src/main/ui/bower.json
kevinwheeler: We haven't ran into any issues yet. We are using them independently. We aren't sending any openlayers2 objects into openlayers3 or vice versa right now. Unless they both try to use the same global variable or something we should be ok.

mforkin: Have you looked into https://github.com/52North/wps-js at all? I haven't yet, but seems like it might be viable and would remove the need for OL2 (is that true or are you using it for something else as well?)

mforkin: doh, maybe it uses OL2, nm

mforkin: NM, looks like this depends on OL2

jdk2pq:
Committed as 7d6ffbbd949a070ba7111c1f6ddbc052a6166592

PR #571, opened 2015-05-07
z index - for comments

elahrvivaz:


geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3Iterator.scala
```diff
@@ -0,0 +1,92 @@
...
+  override def seek(range: AccRange, columnFamilies: java.util.Collection[ByteSequence], inclusive: Boolean): Unit = {
+    source.seek(range, columnFamilies, inclusive)
+    findTop()
+  }
+
+}
+
+object Z3Iterator {
+  def configure(bs: BatchScanner, ll: Z3, ur: Z3) = {
+    val is = new IteratorSetting(1, "z3", "org.locationtech.geomesa.z3iterator.Z3Iterator")
```
elahrvivaz: probably don't want priority 1 - that will go before things like the versioning iterator.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3Iterator.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  override def seek(range: AccRange, columnFamilies: java.util.Collection[ByteSequence], inclusive: Boolean): Unit = {
+    source.seek(range, columnFamilies, inclusive)
+    findTop()
+  }
+
+}
+
+object Z3Iterator {
+  def configure(bs: BatchScanner, ll: Z3, ur: Z3) = {
```
elahrvivaz: It would be better to expose the iterator setting itself - we'll need that for e.g. GeoMesaInputFormat.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
```diff
@@ -0,0 +1,199 @@
...
+  def < (other: Z3) = z < other.z
+  def > (other: Z3) = z > other.z
+  def >= (other: Z3) = z >= other.z
+  def <= (other: Z3) = z <= other.z
+  def + (offset: Long) = new Z3(z + offset)
+  def - (offset: Long) = new Z3(z - offset)
+  def == (other: Z3) = other.z == z
+
+  def decode: (Int, Int, Int) = {
+    ( combine(z), combine(z >> 1), combine(z >> 2) )
```
elahrvivaz: maybe make these lazy vals so that they only get calculated once

geomesa-core/src/main/scala/org/locationtech/geomesa/core/data/tables/Z3Table.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  }
+
+  def configureTable(sft: SimpleFeatureType, z3Table: String, tableOps: TableOperations): Unit = {
+    import scala.collection.JavaConversions._
+
+    val indexedAttributes = getAttributesToIndex(sft)
+    val localityGroups: Map[Text, Text] =
+      indexedAttributes.map { case (name, _) => (name, name) }.toMap.+((BIN_ROW, BIN_ROW))
+    tableOps.setLocalityGroups(z3Table, localityGroups.map { case (k, v) => (k.toString, ImmutableSet.of(v)) } )
```
elahrvivaz: I think it could be useful to have another cf/locality group with the fully encoded simple feature (so that we can pass it back without re-encoding it if we're not transforming), and then break out the attributes into a cf/locality group with each attribute in a different cq that we could use for transformations so we don't have to read the whole feature.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/MergeQueue.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  /** Inserts a single range into the priority queue.
+    *
+    *  @param  range        the element to insert.
+    */
+  @annotation.tailrec
+  final def +=(range: (Long, Long)): Unit = {
+    val res = if(_size == 0) { -1 } else { java.util.Arrays.binarySearch(array, 0, _size, range, ordering) }
+    if(res < 0) {
+      val i = -(res + 1)
```
elahrvivaz: isn't res always going to be -1 here?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
```diff
@@ -0,0 +1,199 @@
...
+  /** Loads either 1000... or 0111... into starting at given bit index of a given dimention */
+  def load(target: Long, p: Long, bits: Int, dim: Int): Long = {
+    val mask = ~(Z3.split(MAX_MASK >> (MAX_BITS-bits)) << dim)
+    val wiped = target & mask
+    wiped | (split(p) << dim)
+  }
+
+  /** Recurse down the oct-tree and report all z-ranges which are contained in the cube defined by the min and max points */
+  def zranges(min: Z3, max: Z3, maxRecurse: Int): Seq[(Long, Long)] = {
+    var mq: MergeQueue = new MergeQueue
```
elahrvivaz: Could you just use a SortedSet, or use  a normal set and sort once at the end?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3Iterator.scala
```diff
@@ -0,0 +1,92 @@
...
+  var zlatmax: Long = -1L
+  var zlonmax: Long = -1L
+  var ztmax: Long = -1L
+
+  var topKey: Key = null
+  var topValue: Value = null
+  val row = new Text()
+  val rowZBytes = Array.ofDim[Byte](8)
+
+  override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] =
```
elahrvivaz: we'll probably need to implement deep copy if we want to stack other iterators on top of this one. There's an example in org.locationtech.geomesa.core.iterators.RowSkippingIterator#deepCopy if you need one.

elahrvivaz:
i think @cne1x is right that we would get better ranges using a z2 + time. Would be interested to see that.

PR #574, opened 2015-05-11
GEOMESA-800 Fixed sorting null values via a FeatureStore

Afterglow375:
* Null values treated as the lowest value when using a sorting query on a FeatureStore

### DIFF
--- a/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
+++ b/geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
@@ -293,11 +293,24 @@ class LazySortedIterator(features: CloseableIterator[SimpleFeature],
       buf.append(features.next())
     }
     features.close()
+    val iter = buf.sortWith(comp).iterator
+
     CloseableIterator(buf.sortWith(comp).iterator)
   }
 
   def attributeToComparable[T <: Comparable[T]](prop: String)(implicit ct: ClassTag[T]): Ordering[SimpleFeature] =
-    Ordering.by[SimpleFeature, T](_.getAttribute(prop).asInstanceOf[T])
+    Ordering.by[SimpleFeature, T](_.getAttribute(prop).asInstanceOf[T])(new Ordering[T] {
+      val evo = implicitly[Ordering[T]]
+
+      override def compare(x: T, y: T): Int = {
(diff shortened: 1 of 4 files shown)


geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/AccumuloDataStoreDefaults.scala
```diff
@@ -0,0 +1,109 @@
...
+  val df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss")
+  df.setTimeZone(TimeZone.getTimeZone("UTC"))
+
+  val defaultSchema = "name:String,geom:Point:srid=4326,dtg:Date"
+  val defaultGeom = WKTUtils.read("POINT(45.0 49.0)")
+  val defaultDtg = new Date(100000)
+  val defaultName = "testType"
+  val defaultFid = "fid-1"
+
+  val defaultTable = "AccumuloDataStoreTest"
```
elahrvivaz: you should make this something distinct like 'getClass.getSimpleName' or something, otherwise different unit tests will use the same mock table and conflicts may arise.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -297,7 +297,18 @@ class LazySortedIterator(features: CloseableIterator[SimpleFeature],
...
-    Ordering.by[SimpleFeature, T](_.getAttribute(prop).asInstanceOf[T])
+    Ordering.by[SimpleFeature, T](_.getAttribute(prop).asInstanceOf[T])(new Ordering[T] {
+      val evo = implicitly[Ordering[T]]
+
+      override def compare(x: T, y: T): Int = {
+        (x, y) match {
+          case (null, null) => 0
+          case (null, _)    => -1
+          case (_, null)    => 1
+          case (_, _)       => evo.compare(x, y)
```
elahrvivaz: i wonder if for something in an inner loop like this it would be more efficient to use if's:
if (x == null) {...} else if (y == null) { ... } else {}
Code would be uglier, but probably faster. Might not matter...

jnh5y:
Merged as a8a34bdf923c66f6f596f2b35ff7556b599191b3 on 1.5.

PR #577, opened 2015-05-11
GEOMESA-802 Allow for short circuiting getCount

jnh5y:
* By default FeatureStore.getCount will return the associated record table size.
 This is done to return quickly.
 Limitations:
  The query/filter is ignored.
  If several features are sharing a table, the record table size will be larger than any one feature.

Since clients may want getCount to return correctly rather than quickly, there are two override methods.

* First, a Java System property, "geomesa.force.count" can be set to "true".
* Second a GeoTools query hint (QueryHints.EXACT_COUNT) can be set to java.lang.Boolean.TRUE.

(If either are true, then getCount will iterate through the result set.)

### DIFF
--- a/geomesa-core/pom.xml
+++ b/geomesa-core/pom.xml
@@ -115,6 +115,10 @@
         </dependency>
         <dependency>
             <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-server</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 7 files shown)


elahrvivaz:
looks good to me

jdk2pq:
Committed as 19cad07a185193108093ca58676d1746d255ca6d

PR #583, opened 2015-05-18
GEOMESA-788 Down-ported New Partitioning Logic Into GeoMesaInputFormat

JakeKurzer:
This tidies up the solution I'd previously written, removing the deprecated function and committing small breaking changes to the GeoSpark API as a result. I believe this is a viable long-term solution, though some further attention should be devoted to understanding the behavior of the AutoAdjustRanges setting, which appears to persist in adjusting ranges even when set to false.

I opted not to use the GeoMesaInputFormat's configure function, as it sets the exact opposite of what we need for spark in a few places. Other than that, all logic for handling the splits is now pushed directly into the GetSplits function of the GeoMesaInputFormat as requested.

### DIFF
--- a/geomesa-compute/pom.xml
+++ b/geomesa-compute/pom.xml
@@ -36,6 +36,10 @@
             <artifactId>geomesa-core-accumulo1.5</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


JakeKurzer:
Forgot sign-off bit...

PR #584, opened 2015-05-18
GEOMESA-788 Down-ported New Partitioning Logic Into GeoMesaInputFormat

JakeKurzer:
Squashed commits covering breaking changes to the GeoSpark API to support controlled splitting in the GeoMesaInputFormat. This allows a user to supply a hint about how many partitions he or she would like to get per executor. 

This could use a second pair of eyes, as there's still sometimes fewer ranges initially than we would expect from a call to the AccumuloInputFormat's getSplits. It looks as though the AutoAdjustRanges behavior is being used even though the flag to disable it is set. @elahrvivaz or @ronq, if you could take a look, it might be good.

### DIFF
--- a/geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
+++ b/geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
@@ -24,18 +24,22 @@ import com.esotericsoftware.kryo.io.{Input, Output}
 import com.google.common.cache.{CacheBuilder, CacheLoader}
 import org.apache.accumulo.core.client.mapreduce.AccumuloInputFormat
 import org.apache.accumulo.core.client.mapreduce.lib.util.{ConfiguratorBase, InputConfigurator}
-import org.apache.accumulo.core.data.{Key, Value}
+import org.apache.accumulo.core.util.{Pair => AccPair}
 import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.io.Text
 import org.apache.spark.rdd.RDD
 import org.apache.spark.serializer.KryoRegistrator
 import org.apache.spark.{SparkConf, SparkContext}
 import org.geotools.data.{DataStore, DataStoreFinder, DefaultTransaction, Query}
 import org.geotools.factory.CommonFactoryFinder
+import org.geotools.filter.text.ecql.ECQL
 import org.locationtech.geomesa.core.data._
-import org.locationtech.geomesa.core.index.{ExplainPrintln, STIdxStrategy, _}
-import org.locationtech.geomesa.feature._
+import org.locationtech.geomesa.core.index._
 import org.locationtech.geomesa.feature.kryo.{KryoFeatureSerializer, SimpleFeatureSerializer}
+import org.locationtech.geomesa.jobs.GeoMesaConfigurator
+import org.locationtech.geomesa.jobs.interop.mapreduce._
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.opengis.filter._
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.JavaConversions._
@@ -43,19 +47,34 @@ import scala.collection.JavaConversions._
 object GeoMesaSpark {
 
   def init(conf: SparkConf, ds: DataStore): SparkConf = {
-    val typeOptions = ds.getTypeNames.map { t => (t, SimpleFeatureTypes.encodeType(ds.getSchema(t))) }
-    typeOptions.foreach { case (k,v) => System.setProperty(typeProp(k), v) }
-    val extraOpts = typeOptions.map { case (k,v) => jOpt(k, v) }.mkString(" ")
-    
+    val typeOptions = ds.getTypeNames.map { t => (t, SimpleFeatureTypes.encodeType(ds.getSchema(t)))}
+    typeOptions.foreach { case (k, v) => System.setProperty(typeProp(k), v)}
+    val extraOpts = typeOptions.map { case (k, v) => jOpt(k, v)}.mkString(" ")
+
     conf.set("spark.executor.extraJavaOptions", extraOpts)
     conf.set("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
     conf.set("spark.kryo.registrator", classOf[GeoMesaSparkKryoRegistrator].getCanonicalName)
   }
 
   def typeProp(typeName: String) = s"geomesa.types.$typeName"
+
   def jOpt(typeName: String, spec: String) = s"-D${typeProp(typeName)}=$spec"
 
-  def rdd(conf: Configuration, sc: SparkContext, ds: AccumuloDataStore, query: Query, useMock: Boolean = false): RDD[SimpleFeature] = {
+  def rdd(conf: Configuration,
+          sc: SparkContext,
+          dsParams: Map[String, String],
+          query: Query,
+          numberOfSplits: Option[Int]): RDD[SimpleFeature] = {
+    rdd(conf, sc, dsParams, query, false, numberOfSplits.getOrElse(-1))
+  }
+
+  def rdd(conf: Configuration,
+          sc: SparkContext,
+          dsParams: Map[String, String],
+          query: Query,
+          useMock: Boolean = false,
+          numberOfSplits: Int = -1): RDD[SimpleFeature] = {
+    val ds = DataStoreFinder.getDataStore(dsParams).asInstanceOf[AccumuloDataStore]
     val typeName = query.getTypeName
     val sft = ds.getSchema(typeName)
     val spec = SimpleFeatureTypes.encodeType(sft)
@@ -68,20 +87,43 @@ object GeoMesaSpark {
 
     ConfiguratorBase.setConnectorInfo(classOf[AccumuloInputFormat], conf, ds.connector.whoami(), ds.authToken)
 
-    if(useMock) ConfiguratorBase.setMockInstance(classOf[AccumuloInputFormat], conf, ds.connector.getInstance().getInstanceName)
-    else ConfiguratorBase.setZooKeeperInstance(classOf[AccumuloInputFormat], conf, ds.connector.getInstance().getInstanceName, ds.connector.getInstance().getZooKeepers)
-
+    if (useMock){
+      ConfiguratorBase.setMockInstance(classOf[AccumuloInputFormat],
+        conf,
+        ds.connector.getInstance().getInstanceName)
+    } else {
+      ConfiguratorBase.setZooKeeperInstance(classOf[AccumuloInputFormat],
+        conf,
+        ds.connector.getInstance().getInstanceName,
+        ds.connector.getInstance().getZooKeepers)
+    }
     InputConfigurator.setInputTableName(classOf[AccumuloInputFormat], conf, ds.getSpatioTemporalTable(sft))
     InputConfigurator.setRanges(classOf[AccumuloInputFormat], conf, qp.ranges)
-    qp.iterators.foreach { is => InputConfigurator.addIterator(classOf[AccumuloInputFormat], conf, is) }
+    qp.iterators.foreach { is => InputConfigurator.addIterator(classOf[AccumuloInputFormat], conf, is)}
 
-    val rdd = sc.newAPIHadoopRDD(conf, classOf[AccumuloInputFormat], classOf[Key], classOf[Value])
-
-    rdd.mapPartitions { iter =>
-      val sft = SimpleFeatureTypes.createType(typeName, spec)
-      val decoder = SimpleFeatureDecoder(sft, featureEncoding)
-      iter.map { case (k: Key, v: Value) => decoder.decode(v.get()) }
+    if (!qp.columnFamilies.isEmpty) {
+      InputConfigurator.fetchColumns(classOf[AccumuloInputFormat],
+        conf,
+        qp.columnFamilies.map(cf => new AccPair[Text, Text](cf, null)))
     }
+
+    if (numberOfSplits != -1) {
+      conf.setInt("SplitsDesired",
+        numberOfSplits * sc.getExecutorStorageStatus.length)
+      InputConfigurator.setAutoAdjustRanges(classOf[AccumuloInputFormat], conf, false)
+      InputConfigurator.setAutoAdjustRanges(classOf[GeoMesaInputFormat], conf, false)
+    }
+    GeoMesaConfigurator.setSerialization(conf)
+    GeoMesaConfigurator.setDataStoreInParams(conf, dsParams)
+    GeoMesaConfigurator.setFeatureType(conf, typeName)
+    if (query.getFilter != Filter.INCLUDE) {
+      GeoMesaConfigurator.setFilter(conf, ECQL.toCQL(query.getFilter))
+    }
+
+    getTransformSchema(query).foreach(GeoMesaConfigurator.setTransformSchema(conf, _))
+
+    sc.newAPIHadoopRDD(conf, classOf[GeoMesaInputFormat], classOf[Text], classOf[SimpleFeature]).map(U => U._2)
+
   }
 
   /**
@@ -113,16 +155,16 @@ object GeoMesaSpark {
     }
   }
 
-  def countByDay(conf: Configuration, sccc: SparkContext, ds: AccumuloDataStore, query: Query, dateField: String = "dtg") = {
-    val d = rdd(conf, sccc, ds, query)
+  def countByDay(conf: Configuration, sccc: SparkContext, dsParams: Map[String, String], query: Query, dateField: String = "dtg") = {
+    val d = rdd(conf, sccc, dsParams, query)
     val dayAndFeature = d.mapPartitions { iter =>
       val df = new SimpleDateFormat("yyyyMMdd")
       val ff = CommonFactoryFinder.getFilterFactory2
       val exp = ff.property(dateField)
-      iter.map { f => (df.format(exp.evaluate(f).asInstanceOf[java.util.Date]), f) }
+      iter.map { f => (df.format(exp.evaluate(f).asInstanceOf[java.util.Date]), f)}
     }
-    val groupedByDay = dayAndFeature.groupBy { case (date, _) => date }
-    groupedByDay.map { case (date, iter) => (date, iter.size) }
+    val groupedByDay = dayAndFeature.groupBy { case (date, _) => date}
+    groupedByDay.map { case (date, iter) => (date, iter.size)}
   }
 
 }
--- a/geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
+++ b/geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
@@ -65,7 +65,7 @@ object GeoMesaSpark {
           dsParams: Map[String, String],
           query: Query,
           numberOfSplits: Option[Int]): RDD[SimpleFeature] = {
-    rdd(conf, sc, dsParams, query, false, numberOfSplits.getOrElse(-1))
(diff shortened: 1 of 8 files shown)


geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -68,20 +87,43 @@ object GeoMesaSpark {
...
+      InputConfigurator.fetchColumns(classOf[AccumuloInputFormat],
+        conf,
+        qp.columnFamilies.map(cf => new AccPair[Text, Text](cf, null)))
+    }
 
-    rdd.mapPartitions { iter =>
-      val sft = SimpleFeatureTypes.createType(typeName, spec)
-      val decoder = SimpleFeatureDecoder(sft, featureEncoding)
-      iter.map { case (k: Key, v: Value) => decoder.decode(v.get()) }
+    if (numberOfSplits != -1) {
```
elahrvivaz: I would keep this an option and do a foreach - it's cleaner than checking for a special value.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -131,20 +131,22 @@ object GeoMesaInputFormat extends Logging {
...
+  var desiredSplitCount: Int = -1
 
   private def init(conf: Configuration) = if (sft == null) {
     val params = GeoMesaConfigurator.getDataStoreInParams(conf)
     val ds = DataStoreFinder.getDataStore(params).asInstanceOf[AccumuloDataStore]
     sft = ds.getSchema(GeoMesaConfigurator.getFeatureType(conf))
     encoding = ds.getFeatureEncoding(sft)
-    numShards = IndexSchema.maxShard(ds.getIndexSchemaFmt(sft.getTypeName))
+    numShards = Math.max(IndexSchema.maxShard(ds.getIndexSchemaFmt(sft.getTypeName)), 1)
+    desiredSplitCount = conf.getInt("SplitsDesired", desiredSplitCount)
```
elahrvivaz: Please put setting and getting the desired splits into the static configurator methods like the other params so the string constants don't have to be messed with.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -155,13 +157,16 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
    * number of shards in the schema as a proxy for number of tservers. We also know that each range will
    * only have a single location, because we always set autoAdjustRanges in the configure method above.
    */
-  override def getSplits(context: JobContext): java.util.List[InputSplit] = {
+  override def getSplits(context: JobContext) : java.util.List[InputSplit] = {
```
elahrvivaz: [picky] remove the extra space that was added

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -155,13 +157,16 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
...
     init(context.getConfiguration)
     val accumuloSplits = delegate.getSplits(context)
-    // try to create 2 mappers per node - account for case where there are less splits than shards
```
elahrvivaz: keep the original comment around

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -155,13 +157,16 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
...
-    val groupSize = Math.max(numShards * 2, accumuloSplits.length / (numShards * 2))
+    var groupSize = Math.max(numShards * 2, accumuloSplits.length / (numShards * 2)) // preserve original behavior.
+    if (desiredSplitCount != -1)
+    {
```
elahrvivaz: [picky] formatting - keep opening bracket on same line as if

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -155,13 +157,16 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
...
+      groupSize = Math.max(1, accumuloSplits.length / desiredSplitCount)
+    }
 
-    // We know each range will only have a single location because of autoAdjustRanges
+   // We know each range will only have a single location because of autoAdjustRanges
```
elahrvivaz: [picky] remove deleted space so comment lines up

JakeKurzer: Actually, we may not have autoadjust set, intentionally, so this comment's gotta go.

elahrvivaz: hmm, why would we not want to autoadjust ranges? if we do run it, then ideally each range will be run on the appropriate local tserver. Otherwise we will get more network traffic/slowness.

JakeKurzer: Auto-adjust ranges merges ranges down, doesn't it? This can lead to very 
large partitions. I'll try it with it on later tonight, if you want.

elahrvivaz: hmm, it might. i believe it also merges ranges up though, e.g. if you have a range that spans 2 tservers it will split it into 2 ranges. my memory on exactly how it works is hazy...

JakeKurzer: The docs aren't terribly clear. I'd appreciate an expert opinion.

elahrvivaz: accumulo docs are pretty bare. i just looked at the source code - but that was a while ago.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
JakeKurzer: Could you explain what you mean a little further? The value would end up in the conf object either way, and would thus need to be gotten in either this way or with a dedicated getter method.

elahrvivaz: It's just good programming - using string constants can't be checked by the compiler. So instead we create methods like: <code>def configureDesiredSplits()... def getDesiredSplits(conf)</code> so that the string doesn't get exposed to users.
Basically copy org.locationtech.geomesa.jobs.GeoMesaConfigurator#(setFilter, getFilter)

JakeKurzer:
Added getter\setter for desiredSplits. I've separated it out from the normal configure method.

JakeKurzer:
I'm testing re-enabling AutoAdjustRanges overnight. So far, it's not doing much of anything to make a difference, but we'll need to do more quantitative tests.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -68,20 +87,43 @@ object GeoMesaSpark {
...
+    } else {
+      ConfiguratorBase.setZooKeeperInstance(classOf[AccumuloInputFormat],
+        conf,
+        ds.connector.getInstance().getInstanceName,
+        ds.connector.getInstance().getZooKeepers)
+    }
     InputConfigurator.setInputTableName(classOf[AccumuloInputFormat], conf, ds.getSpatioTemporalTable(sft))
     InputConfigurator.setRanges(classOf[AccumuloInputFormat], conf, qp.ranges)
-    qp.iterators.foreach { is => InputConfigurator.addIterator(classOf[AccumuloInputFormat], conf, is) }
+    qp.iterators.foreach { is => InputConfigurator.addIterator(classOf[AccumuloInputFormat], conf, is)}
```
elahrvivaz: [picky] not worth changing, but just so you know in the future - we keep the extra space before the ending bracket.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -68,20 +87,43 @@ object GeoMesaSpark {
...
-    rdd.mapPartitions { iter =>
-      val sft = SimpleFeatureTypes.createType(typeName, spec)
-      val decoder = SimpleFeatureDecoder(sft, featureEncoding)
-      iter.map { case (k: Key, v: Value) => decoder.decode(v.get()) }
+    if (numberOfSplits.isDefined) {
+      GeoMesaInputFormat.configureSplits(conf,
+        numberOfSplits.get * sc.getExecutorStorageStatus.length)
+      InputConfigurator.setAutoAdjustRanges(classOf[AccumuloInputFormat], conf, false)
+      InputConfigurator.setAutoAdjustRanges(classOf[GeoMesaInputFormat], conf, false)
+    }
```
elahrvivaz: also not worth changing, but a more idiomatic way to do with is with a foreach:
numberOfSplits.foreach { num => ... }
avoids the explicit .get on an option

geomesa-jobs/src/main/java/org/locationtech/geomesa/jobs/interop/mapreduce/GeoMesaInputFormat.java
```diff
@@ -63,6 +64,16 @@ public static void configure(Job job, Map<String, String> dataStoreParams, Query
...
+    public static void configureSplits(Configuration conf, Integer desiredSplits)
+    {
+        GeoMesaInputFormat$.MODULE$.configureSplits(conf,desiredSplits);
+    }
+
+    public static int getConfiguredSplits(Configuration conf)
+    {
+        return GeoMesaInputFormat$.MODULE$.getConfiguredSplits(conf);
+    }
+
```
elahrvivaz: thanks for remembering the interop :)

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -152,16 +171,18 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
...
     init(context.getConfiguration)
     val accumuloSplits = delegate.getSplits(context)
-    // try to create 2 mappers per node - account for case where there are less splits than shards
-    val groupSize = Math.max(numShards * 2, accumuloSplits.length / (numShards * 2))
+    // fallback on creating 2 mappers per node if desiredSplits is unset.
+    // Account for case where there are less splits than shards
+    var groupSize = Math.max(numShards * 2, accumuloSplits.length / (numShards * 2))
+    if (desiredSplitCount != -1) {
+      groupSize = Math.max(1, accumuloSplits.length / desiredSplitCount)
+    }
```
elahrvivaz: another tidbit for the future: you can avoid using a var (generally discouraged in scala) by something like:
val groupSize = if (desiredSplitCount ...) foo else bar

elahrvivaz:
Should this get added to the mapred classes too?

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
JakeKurzer: While that's more idiomatic, I feel like it makes it less obvious that the option should contain only a single value. Why is this preferred?

elahrvivaz: that's scala for you :)

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
JakeKurzer: Yeah, I use that pattern all over the place normally. Technically, if-else is scala's ? operator equivalent, in fact. I can't remember why I didn't use it here.

JakeKurzer:
Maybe, but I feel like we should probably stop using them. They are deprecated, and have been for a while...

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
JakeKurzer: Increasingly, I feel like very little of scala is actually for me.

elahrvivaz:
Could you add it to the mapred classes too? we still require it for things like scalding.
At the same time, please move the configuration of splits into GeoMesaConfigurator, and namespace the value (e.g. prepend the 'prefix' string).

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
```diff
@@ -159,11 +161,15 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
   override def getSplits(job: JobConf, numSplits: Int) = {
     init(job)
     val accumuloSplits = delegate.getSplits(job, numSplits)
-    // try to create 2 mappers per node - account for case where there are less splits than shards
-    val groupSize = Math.max(numShards * 2, accumuloSplits.length / (numShards * 2))
+    // fallback on creating 2 mappers per node if desiredSplits is unset.
+    // Account for case where there are less splits than shards
+    val groupSize =  if (desiredSplitCount > 1) {
```
elahrvivaz: should this be > 0 ?

elahrvivaz:
i give this my blessing

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
JakeKurzer: I don't think there's a situation where we would want a user asking for one partition. That seems dangerous.

JakeKurzer:
Okay, I can't merge it, so.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -68,20 +87,43 @@ object GeoMesaSpark {
...
+        ds.connector.getInstance().getInstanceName,
+        ds.connector.getInstance().getZooKeepers)
+    }
     InputConfigurator.setInputTableName(classOf[AccumuloInputFormat], conf, ds.getSpatioTemporalTable(sft))
     InputConfigurator.setRanges(classOf[AccumuloInputFormat], conf, qp.ranges)
-    qp.iterators.foreach { is => InputConfigurator.addIterator(classOf[AccumuloInputFormat], conf, is) }
+    qp.iterators.foreach { is => InputConfigurator.addIterator(classOf[AccumuloInputFormat], conf, is)}
 
-    val rdd = sc.newAPIHadoopRDD(conf, classOf[AccumuloInputFormat], classOf[Key], classOf[Value])
+    if (!qp.columnFamilies.isEmpty) {
```
jnh5y: (picky) A for comprehension or a foreach might make this more readable.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/GeoMesaConfigurator.scala
```diff
@@ -43,6 +43,7 @@ object GeoMesaConfigurator {
   private val sftKey           = s"$prefix.sft"
   private val transformsKey    = s"$prefix.transforms.schema"
   private val transformNameKey = s"$prefix.transforms.name"
+  private val desiredSplits = s"$prefix.splitCount.userPreference"
```
jnh5y: (picky) Align '='s?

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
elahrvivaz: cf's is a seq...

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/GeoMesaConfigurator.scala
jnh5y: private val desiredSplits = s"$prefix.splitCount.userPreference"
maybe should be 
 private val desiredSplits = s"$prefix.mapreduce.split.count.hint"
or some such.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
```diff
@@ -171,7 +177,12 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
         split
       }
     }
-    splits.toArray
+    //we've dumped all ranges into groups by location, and tried to get enough splits by breaking up those groups.
+    //we may eventually need to go ahead and do some pretty grim mojo to break each split down small enough
+    //so that we don't blow executor's memory pools. Right now, that's not happening, so I'm not going to push it.
```
jnh5y: Remove this comment.  If you'd like to enter a ticket in JIRA and link to it, that's fine.

In general, I'd discourage opinionated comments.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -170,7 +176,12 @@ class GeoMesaInputFormat extends InputFormat[Text, SimpleFeature] {
         split
       }
     }
-    splits.toList
+    //we've dumped all ranges into groups by location, and tried to get enough splits by breaking up those groups.
+    //we may eventually need to go ahead and do some pretty grim mojo to break each split down small enough
+    //so that we don't blow executor's memory pools. Right now, that's not happening, so I'm not going to push it.
```
jnh5y: Ditto for this file.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
jnh5y: Ah, good point, nevermind.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaInputFormat.scala
JakeKurzer: I'm comfortable removing it, but I'd like to understand why this is an opinionated comment.

elahrvivaz:
merged as 1d0178e399ce5f66f0bcd94b071bf0634fe31e8d

PR #585, opened 2015-05-20
GEOMESA-796 Enable user-specific roles in GeoServer

drmathochist:
A trait to mix into a GeoServerRoleService to create a GS role for each user on the fly,
which can be used to secure layers.


elahrvivaz:
Closing - we need to remove all geoserver-specific code due to IP issues.

PR #592, opened 2015-05-28
Added java.sql.Timestamp to SimpleFeatureTypes

nhambletCCRI:
Follow-on from [551](https://ccristudio.jira.com/browse/GEOMESASUP-551).

I'm surprised that the unit tests fail if you try to read the attributes as Timestamps, but I don't really know much about the internals for GeoMesa. Also, I wasn't really sure where to put the unit test. And am not sure I'm quite ready to tackle an ImportProcess unit tests.

### DIFF
--- a/geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/FeatureWritersTest.scala
+++ b/geomesa-core/src/test/scala/org/locationtech/geomesa/core/data/FeatureWritersTest.scala
@@ -16,9 +16,11 @@
 
 package org.locationtech.geomesa.core.data
 
+import java.sql.Timestamp
 import java.text.SimpleDateFormat
-import java.util.TimeZone
+import java.util.{Date, TimeZone}
 
+import com.vividsolutions.jts.geom.Geometry
 import org.geotools.data._
 import org.geotools.data.simple.SimpleFeatureIterator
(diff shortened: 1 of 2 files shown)


nhambletCCRI:
Mostly this PR is just for comments or suggestions.

elahrvivaz:
I don't think we want to support timestamps... that isn't part of the normal geotools attributes, is it? A better approach might be to create a converter from timestamp to Date. That will keep the type as date but let you do something like feature.setAttribute("dtg", new Timestamp()). Would that work for your use case?

See
org.locationtech.geomesa.utils.geotools.JodaConverterFactory
geomesa/geomesa-utils/src/main/resources/META-INF/services/org.geotools.util.ConverterFactory

nhambletCCRI:
@elahrvivaz I don't know what's normal. My use case is coming from GML getting parsed, where an xsd:datetime becomes a Timestamp with the XSDateTimeBinding from org.geotools.xs.bindings of gt-xsd-core. Which makes me think it's somewhat normal. For somebody. Sometimes.

elahrvivaz:
Just a note: the reason that unit tests are failing is because we'd have to add support for timestamps in our serialization.

elahrvivaz:
closing for now - can revisit later if needed.

PR #594, opened 2015-05-28
Rmorris: QuerySize iterator developed for benchmarking.  Calculates size/count…

drackaer:
… of records scanned and returned.


geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/QueryPlanner.scala
```diff
@@ -166,6 +168,35 @@ case class QueryPlanner(sft: SimpleFeatureType,
...
+    var scannedRecords: Long = 0
+    var resultBytes: Long = 0
+    var resultRecords: Long = 0
+    while(querySizeIterator.hasNext) { // TODO How do I do this in a Scala-y way?
+      val cur = querySizeIterator.next()
+      scannedBytes += cur.getAttribute(QuerySizeIterator.SCAN_BYTES_ATTRIBUTE).asInstanceOf[Long]
+      scannedRecords += cur.getAttribute(QuerySizeIterator.SCAN_RECORDS_ATTRIBUTE).asInstanceOf[Long]
+      resultBytes += cur.getAttribute(QuerySizeIterator.RESULT_BYTES_ATTRIBUTE).asInstanceOf[Long]
+      resultRecords += cur.getAttribute(QuerySizeIterator.RESULT_RECORDS_ATTRIBUTE).asInstanceOf[Long]
+    }
```
elahrvivaz: i think you need to call close() on the query size iterator once you've exhausted it.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
```diff
@@ -112,7 +112,26 @@ class STIdxStrategy extends Strategy with Logging with IndexFilterHelpers {
     val qp = planQuery(filter, iteratorConfig.iterator, output, keyPlanner, cfPlanner)
 
     val table = acc.getSpatioTemporalTable(sft)
-    val iterators = qp.iterators ++ List(Some(stiiIterCfg), densityIterCfg).flatten
+    var iterators: Seq[IteratorSetting] = qp.iterators ++ List(Some(stiiIterCfg), densityIterCfg).flatten
```
jnh5y: We can remove the type annotation 'Seq[IteratorSetting]'.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
```diff
@@ -112,7 +112,26 @@ class STIdxStrategy extends Strategy with Logging with IndexFilterHelpers {
...
+
+    if(query.getHints.containsKey(QUERY_SIZE_KEY)) {
+      val cfg = new IteratorSetting(iteratorPriority_SimpleFeatureFilteringIterator, "query-size-iterator", classOf[QuerySizeIterator])
+      //cfg.addOption(QuerySizeIterator.ORIGINAL_SFT_OPTION,sft) // JNH: This requires the original ECQL filter.
```
jnh5y: (picky) Remove comment.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
elahrvivaz: would be more scala-y to keep this a val and set it as the result of the if

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
```diff
@@ -112,7 +112,26 @@ class STIdxStrategy extends Strategy with Logging with IndexFilterHelpers {
     val qp = planQuery(filter, iteratorConfig.iterator, output, keyPlanner, cfPlanner)
 
     val table = acc.getSpatioTemporalTable(sft)
-    val iterators = qp.iterators ++ List(Some(stiiIterCfg), densityIterCfg).flatten
+    var iterators: Seq[IteratorSetting] = qp.iterators ++ List(Some(stiiIterCfg), densityIterCfg).flatten
+
+    if(query.getHints.containsKey(QUERY_SIZE_KEY)) {
+      val cfg = new IteratorSetting(iteratorPriority_SimpleFeatureFilteringIterator, "query-size-iterator", classOf[QuerySizeIterator])
```
jnh5y: (style) We tend to try and avoid long lines (over 100 characters), so break out the arguments on newlines.  (Look around for examples.)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/index/STIdxStrategy.scala
```diff
@@ -112,7 +112,26 @@ class STIdxStrategy extends Strategy with Logging with IndexFilterHelpers {
...
+        case (None, Some(_))        => ecql
+        case (None, None)           => None
+      }
+      configureFeatureType(cfg, sft)
+      configureFeatureEncoding(cfg, featureEncoding)
+      configureTransforms(cfg, query)  // Do we need transforms?
+      configureEcqlFilter(cfg, combinedFilter.map(ECQL.toCQL))
+
+      iterators = Seq(cfg)
+    }
```
jnh5y: (style) I'd extract this entire function to make the query planner more readable.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
```diff
@@ -0,0 +1,84 @@
...
+class QuerySizeIterator extends GeomesaFilteringIterator with HasFeatureDecoder with HasFilter with HasFeatureType {
+
+  var featureBuilder: SimpleFeatureBuilder = null
+  var querySizeFeatureEncoder: SimpleFeatureEncoder = null
+
+  override def init(source: SortedKeyValueIterator[Key, Value], options: util.Map[String, String], env: IteratorEnvironment): Unit = {
+    super.init(source, options, env)
+    initFeatureType(options)
+    // Need to grab the original SFT in order to properly decode/filter
+    val originalSFT = SimpleFeatureTypes.createType("QuerySizeHijackedIteratorSFT",options.get(core.GEOMESA_ITERATORS_SIMPLE_FEATURE_TYPE))
```
jnh5y: See comment about long lines.

jnh5y: (picky) We tend to have a rather direct style when it comes to naming variables/functions.  QuerySizeSFT is probably a little better.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
```diff
@@ -0,0 +1,84 @@
...
+    super[HasFilter].init(originalSFT, options)
+    super[HasFeatureDecoder].init(originalSFT, options)
```
jnh5y: This may be redundant.  Read about Scala's linearization with mixins.  You may only need 'super.init(originalSFT, options)'.  If the order of execution matters, then you can permute the order of the mixins.

On the other hand, HasFilter and HasFeatureDecoder may each need to reference super for that to work.  (Or at least one of them.)

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
```diff
@@ -0,0 +1,84 @@
...
+    querySizeFeatureEncoder = SimpleFeatureEncoder(querySizeSFT, FeatureEncoding.KRYO)
+  }
+
+  // Could rewrite this to automatically scan to the next top, aggregating results as it goes.
+  override def setTopConditionally() = {
+
+    featureBuilder.reset()
+    topKey = new Key(source.getTopKey)
+
+    val curNumBytes: Long = source.getTopValue.get.length // Could this be wrong?
```
jnh5y: Why the doubt?  If you are unsure, we should probably figure that out.  (And then remove the comment:))

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
```diff
@@ -0,0 +1,84 @@
...
+    if (filter.evaluate(featureDecoder.decode(source.getTopValue.get))) {
+      resultBytes = curNumBytes
+      resultRecords = 1
+    }
+    featureBuilder.set(RESULT_BYTES_ATTRIBUTE, resultBytes)
+    featureBuilder.set(RESULT_RECORDS_ATTRIBUTE, resultRecords)
+    val feature = featureBuilder.buildFeature("feature")
+
+    topValue = new Value(querySizeFeatureEncoder.encode(feature))
+  }
```
jnh5y: Ah!  This is good in that it is mapping features to our 4-tuples.  

We can do better in that we can do a 'map-reduce' server-side.  We shouldn't return from this iterator until the underlying source is exhausted.  When that happens, we should return exactly one SF which contains the aggregate for the whole shebang.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
elahrvivaz: yeah, i think you shouldn't need any of this init code actually... If you look at org.locationtech.geomesa.accumulo.iterators.SpatioTemporalIntersectingIterator#init you should be able to copy the first 3 lines of that. The simple feature type is exposed as a variable in the mixed in trait as 'featureType'

jnh5y: Is it sensible to use QuerySizeIterator.querySizeSFT in place of the originalSFT?  Saves having to recompute it.

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/QuerySizeteratorTest.scala
```diff
@@ -0,0 +1,209 @@
...
+        val geom = "POINT(-101 " + i + ")"
+        ScalaSimpleFeatureFactory.buildFeature(sft, Array(dtg, geom), id)
+      }
+
+      clearFeatures()
+      addFeatures(feats)
+      addFeatures(otherFeats)
+
+
+//      val q = getQuery("(dtg between '2012-01-01T18:00:00.000Z' AND '2012-01-01T23:00:00.000Z') and BBOX(geom, -80, 33, -70, 40)")
```
jnh5y: Remove commented line.

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/QuerySizeteratorTest.scala
```diff
@@ -0,0 +1,209 @@
...
+      val q = getQuery("BBOX(geom, -100, -100, 75, 75)")
+      val results = fs.getFeatures(q).features().toList
+      results must haveLength(1)  // JNH: This still makes sense.
```
jnh5y: As much as I like my comment, it should go;)

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/QuerySizeteratorTest.scala
```diff
@@ -0,0 +1,209 @@
...
+      val currentExpectedSizeInBytes = 30
+      println(scanSizeBytes + "\t" + scanRecords + "\t" + returnSizeBytes + "\t" + returnRecords)
+
+      scanRecords should be greaterThan expectedNumRecords
+      returnSizeBytes should be greaterThan (currentExpectedSizeInBytes-1)*expectedNumRecords
+      returnSizeBytes should be lessThan (currentExpectedSizeInBytes+1)*expectedNumRecords // Give an expanded range on size
+      returnRecords should be equalTo expectedNumRecords
+    }
+
+//    "correctly filter dates" in {
```
jnh5y: Remove commented code.

geomesa-core/src/test/scala/org/locationtech/geomesa/core/iterators/QuerySizeteratorTest.scala
```diff
@@ -0,0 +1,209 @@
...
+      val returnRecords = result.getAttribute(QuerySizeIterator.RESULT_RECORDS_ATTRIBUTE).asInstanceOf[Long]
+
+      val expectedNumRecords = 70
+      val currentExpectedSizeInBytes = 30
+      println(scanSizeBytes + "\t" + scanRecords + "\t" + returnSizeBytes + "\t" + returnRecords)
+
+      scanRecords should be greaterThan expectedNumRecords
+      returnSizeBytes should be greaterThan (currentExpectedSizeInBytes-1)*expectedNumRecords
+      returnSizeBytes should be lessThan (currentExpectedSizeInBytes+1)*expectedNumRecords // Give an expanded range on size
+      returnRecords should be equalTo expectedNumRecords
```
jnh5y: Nice.  Ship it!

jnh5y:
@drackaer I owe you some GeoMesan start-up emails/instructions about our JIRA and Eclipse CLAs.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
elahrvivaz: You probably don't want to extend GeomesaFilteringIterator, as it doesn't really support aggregation like that.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
drackaer: I can get HasFilter implicitly (iirc) but I had to specify super for HasFeatureDecorder, I figured specifying both would improve readability, should I remove the super reference that it finds through linearization?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
jnh5y: Yeah, that's a rough thing to work through.  This can't really extend GFI or the AggregatingIterators...  At least I don't think so... maybe I'm being too quick.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Iterator that returns counts and size-in-bytes of both results filtered and results returned.
+ */
+class QuerySizeIterator extends GeomesaFilteringIterator with HasFeatureDecoder with HasFilter with HasFeatureType {
+
+  var featureBuilder: SimpleFeatureBuilder = null
+  var querySizeFeatureEncoder: SimpleFeatureEncoder = null
+
+  override def init(source: SortedKeyValueIterator[Key, Value], options: util.Map[String, String], env: IteratorEnvironment): Unit = {
```
AndrewAnnex: Personally I would like to see the options be switched to a scala map

elahrvivaz: heh, i don't think accumulo is going to re-write in scala any time soon...

jnh5y: @AndrewAnnex +1.  Let's rewrite Accumulo in Scala.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
drackaer: My one question is whether it is possible to have a padding byte that is empty and we wouldn't want to count it, is there anything wonky about how value is stored?

elahrvivaz: there's actually a getSize method on value - that calls get().length :)

elahrvivaz: It is probably worthwhile to keep track of key bytes read/returned, as I think those are just as important as value bytes.

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
drackaer: Should we circle back and have it be its own thing?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
drackaer: AFAIK querySizeSFT may have a different "schema" from the originalSFT, our input feature types are different from our output feature types.  Or am I understanding this incorrectly?

geomesa-core/src/main/scala/org/locationtech/geomesa/core/iterators/QuerySizeIterator.scala
AndrewAnnex: oops, did not evaluate what I was saying in the comment.

elahrvivaz:
this PR seems like it needs some rebasing...

PR #596, opened 2015-06-01
GEOMESA-780 Kafka Data Store Replay

mmatz-ccri:
* Added replay ability to Kafka Data Store.
* GEOMESA-799 Kafka consumer with ability to start from arbitrary offset


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/filter/package.scala
```diff
@@ -307,33 +306,7 @@ package object filter {
...
-   * a property and a literal, throws an exception.
-   *
-   * @param one
-   * @param two
-   * @return
-   */
-  def checkOrderUnsafe(one: Expression, two: Expression): PropertyLiteral =
-    checkOrder(one, two)
-        .getOrElse(throw new RuntimeException("Expressions did not contain valid property and literal"))
+  val checkOrderUnsafe = org.locationtech.geomesa.filter.checkOrderUnsafe _
```
jnh5y: Do we need to forward these functions or can just relocate them?  (Which I'm totally in favor of.)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
```diff
@@ -35,6 +35,7 @@ import org.locationtech.geomesa.accumulo.index.QueryPlanners.JoinFunction
 import org.locationtech.geomesa.accumulo.index.Strategy._
 import org.locationtech.geomesa.accumulo.iterators._
 import org.locationtech.geomesa.features.SerializationType.SerializationType
+import org.locationtech.geomesa.filter.PropertyLiteral
```
jnh5y: Is this import used?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/PropertyLiteral.scala
```diff
@@ -6,7 +6,7 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  */
 
-package org.locationtech.geomesa.accumulo.filter
+package org.locationtech.geomesa.filter
```
jnh5y: Should this move between sub-modules?  (I haven't checked the full path quite yet.)

geomesa-kafka/geomesa-kafka-datastore/README.md
```diff
@@ -0,0 +1,51 @@
...
+object passed to `createSchema(SimpleFeatureType)`.  These hints are added through the use of the following helper
+methods:
+    
+    
+    KafkaDataStoreHelper.prepareForLive(sft: SimpleFeatureType, zkPath: String) : SimpleFeatureType
+    KafkaDataStoreHelper.prepareForReplay(sft: SimpleFeatureType, rc: ReplayConfig) : SimpleFeatureType
+    
+###Data Producers
+
+Data *producers* should build a SimpleFeatureType object and pass it to `prepareForLive` along with a  zookeeper path that
```
jnh5y: I wonder if 'producers' should use a separate prepareToProduce method.  If the implementation can call between the prepareToProduce and prepareForLive, that's fine.

geomesa-kafka/geomesa-kafka-datastore/README.md
```diff
@@ -0,0 +1,51 @@
...
+Data *producers* should build a SimpleFeatureType object and pass it to `prepareForLive` along with a  zookeeper path that 
+must be unique for a given KafkaDataStore.  The Consumer should pass the returned SimpleFeature type to the call to
+`Datastore.createSchema`.  For example:
+
+    val ds: DataStore = ...
+    val outputSFT: SimpleFeatureType = ...
+    
+    ...
+    
+    val zkPath = "/a/unique/zkpath"
```
jnh5y: What happens if the zkPath is not-unique/reused?

geomesa-kafka/geomesa-kafka-datastore/README.md
```diff
@@ -0,0 +1,51 @@
...
+Data *consumers* that are running in *live* mode access the DataStore through the usual DataStore interface.
+
+###Data Consumers (Replay Mode)
+
+Data *consumers* that are running in *replay* mode will have to first create a new replay schema in the DataStore that 
+is derived from the original schema.  The consumer should get the SimpleFeatureType for original schema and use it (
+along with a `ReplayConfig`) to create the derived schema via a call to `DataStore.prepareForReplay()`.
+
+    val datastore = ...
+    val typename = "FrOut"
```
jnh5y: Let's use a more general sounding name.

geomesa-kafka/geomesa-kafka-datastore/README.md
```diff
@@ -0,0 +1,51 @@
...
+    val sft = ds.getSchema(typeName)
+    val rc = new ReplayConfig(replayStart, replayEnd, readBehind)
+
+    // create the replay schema from the original schema and replay params
+    datastore.createSchema(KafkaDataStoreHelper.prepareForReplay(sft,rc))
+
+It should be noted that the new schema will have a different type name and an additional attribute in the schema (time) 
+from the original SimpleFeatureType
```
jnh5y: (Pedantic) Sentences end with periods

geomesa-kafka/geomesa-kafka-datastore/pom.xml
```diff
@@ -70,6 +79,13 @@
             <artifactId>kafka_2.10</artifactId>
         </dependency>
         <dependency>
+            <!--This is a compile dependency of Kafka and also KafkaGeoMessageConsoleConsumer -->
+            <groupId>net.sf.jopt-simple</groupId>
+            <artifactId>jopt-simple</artifactId>
+            <version>3.2</version>
+            <scope>compile</scope>
```
jnh5y: 1.  Do we have a IP CQ in for this dependency?  If not, I'll need some details.
2.  Let's try and manage the version in the root pom.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
```diff
@@ -0,0 +1,184 @@
...
+  def createOrUpdate(sf: SimpleFeature): CreateOrUpdate = CreateOrUpdate(Instant.now, sf)
+}
+
+/** Create a new simple feature or update an existing [[SimpleFeature]]
+  *
+  * @param feature the [[SimpleFeature]]
+  */
+case class CreateOrUpdate(override val timestamp: Instant, feature: SimpleFeature) extends GeoMessage {
+
+  val id = feature.getID
```
jnh5y: I'd recommend making this a def.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
```diff
@@ -0,0 +1,184 @@
...
+  */
+case class Clear(override val timestamp: Instant) extends GeoMessage
+
+
+/** Encodes [[GeoMessage]]s.  [[Clear]] and [[Delete]] messages are handled directly.  See class
+  * [[GeoMessageEncoder]] for encoding [[CreateOrUpdate]] messages.
+  *
+  * The following encoding is used:
+  *
+  * Key: version (1 byte) type (1 byte) timstamp (8 bytes)
```
jnh5y: Let's not tramp this Tim fellow.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  private val EMPTY = Array.empty[Byte]
+
+  def encodeKey(msg: GeoMessage): Array[Byte] = {
+
+    val msgType: Byte = msg match {
+      case c: CreateOrUpdate => createOrUpdateType
+      case d: Delete => deleteType
+      case x: Clear => clearType
+      case u => throw new IllegalArgumentException(s"Invalid message: '$u'")
```
jnh5y: I'm tempted to suggest using a 'sealed' trait here for GeoMessage.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  val serializer: SimpleFeatureSerializer = new KryoFeatureSerializer(schema, SerializationOptions.withUserData)
+
+  def encodeMessage(msg: GeoMessage): Array[Byte] = msg match {
+    case c: CreateOrUpdate =>
+      encodeCreateOrUpdateMessage(c)
+    case d: Delete =>
+      encodeDeleteMessage(d)
+    case x: Clear =>
+      encodeClearMessage(x)
```
jnh5y: For clarity, I'd suggest removing the 'import GeoMessageEncoder._' line and calling encodeClearMessage and encodeDeleteMessage prefixed by GeoMessageEncoder.  

I searched for these functions and I found a few with the same name and different signatures.  It seems like there's potential for confusion.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
```diff
@@ -0,0 +1,184 @@
...
+  * @param schema the [[SimpleFeatureType]]; required to deserialize [[CreateOrUpdate]] messages
+  */
+class GeoMessageDecoder(schema: SimpleFeatureType) extends Logging {
+
+  case class MsgKey(version: Byte, msgType: Byte, ts: Instant)
+
+  val serializer: SimpleFeatureDeserializer = new KryoFeatureSerializer(schema, SerializationOptions.withUserData)
+
+  def decode(key: Array[Byte], msg: Array[Byte]): GeoMessage = {
+    val MsgKey(version, msgType, ts) = decodeKey(key)
```
jnh5y: Is MsgKey only used to encode/extract for this line?  If so, a Tuple3 should suffice.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
```diff
@@ -0,0 +1,184 @@
...
+    } else {
+      throw new IllegalArgumentException(s"Unknown serialization version: $version")
+    }
+  }
+
+  def decodeKey(key: Array[Byte]): MsgKey = {
+    if (key == null) {
+      throw new IllegalArgumentException("Invalid null key.")
+    }
+    if (key.length != 10) {
```
jnh5y: I'd suggest checking the version and throwing an exception based on that.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
mmatz-ccri: Yes, it was covered by import org.locationtech.geomesa.accumulo.filter._ until I moved PropertyLiteral.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/filter/package.scala
mmatz-ccri: Sure, I'll relocate them.  Or, maybe relocate the whole class to org.locationtech.geomesa.filter (same as it was before zcurve)?  The only problem with that is FilterPackageObjectTest depends on TestData in geomesa-accumulo-datastore/src/test.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/PropertyLiteral.scala
mmatz-ccri: I moved it back where it was before zcurve because geomesa-kafka-datastore needs it.  There's nothing accumulo specific about it.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
```diff
@@ -135,42 +59,55 @@ object KafkaDataStoreFactoryParams {
...
+    if (isProducer) {
+      KafkaProducerFeatureStoreFactory(brokers)
+    } else {
+      KafkaConsumerFeatureSourceFactory(brokers, zk, params)
+    }
   }
 
-  override def createNewDataStore(params: ju.Map[String, Serializable]): DataStore = ???
+  override def createNewDataStore(params: ju.Map[String, Serializable]): DataStore =
+    throw new UnsupportedOperationException
```
jnh5y: Hmm... did the compiler make you add this?  I imagine the override would have taken care of the exception declaration.  

I think it'll be fine either way.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  val TopicKey = "Topic"
+  val ReplayConfigKey = "ReplayConfig"
+
+  /** Creates a copy of the passed SimpleFeatureType, inserting the topic name (derived from zkPath) into user data.*/
+  def prepareForLive(sft: SimpleFeatureType, zkPath: String) : SimpleFeatureType = {
+
+    val builder = FeatureUtils.builder(sft)
+    val preparedSft = builder.buildFeatureType()
+    insertTopic(preparedSft,buildTopicName(zkPath,sft))
```
jnh5y: (Picky) Space after the comma.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
```diff
@@ -0,0 +1,103 @@
...
+
+
+  /** Creates a copy of the passed SimpleFeature type, inserting an encoded ReplayConfig into the user data. */
+  def prepareForReplay(sft: SimpleFeatureType, rConfig: ReplayConfig) : SimpleFeatureType = {
+
+    val builder = FeatureUtils.builder(sft)
+    builder.setName(buildReplayTypeName(sft.getTypeName))
+    ReplayTimeHelper.addReplayTimeAttribute(builder)
+
+    // assumes topic has been prepared, need to check though...
```
jnh5y: If there's a follow-on ticket, feel free to create it and reference it in the comment.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
```diff
@@ -0,0 +1,103 @@
...
+    val preparedSft = builder.buildFeatureType()
+    preparedSft.getUserData.put(ReplayConfigKey,ReplayConfig.encode(rConfig))
+    preparedSft
+  }
+
+  /** modifies the passed SimpleFeature type, adding the topic string to user data */
+  def insertTopic(sft: SimpleFeatureType, topic: String) : SimpleFeatureType = {
```
jnh5y: This function can return 'Unit' since it is mutating state.  (All uses I observed follow that pattern.

jnh5y: This is optional.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
```diff
@@ -0,0 +1,103 @@
...
+      Some(replayName.substring(0, index))
+    } else {
+      None
+    }
+  }
+  
+  def isPreparedForLive(sft: SimpleFeatureType): Boolean =
+    sft.getUserData.containsKey(TopicKey) && !sft.getUserData.containsKey(ReplayConfigKey)
+
+  def isPreparedForReplay(sft: SimpleFeatureType): Boolean = sft.getUserData.containsKey(ReplayConfigKey)
```
jnh5y: Is the TopicKey ever set for 'replay'?

geomesa-kafka/geomesa-kafka-datastore/README.md
mmatz-ccri: The zkPath is not required to be unique.  It's purpose is to provide disambigation if needed.  There may be multiple SFTs under the same zkPath.

geomesa-kafka/geomesa-kafka-datastore/pom.xml
mmatz-ccri: Its was a transitive dependency.  I'm now using it directly but could elminiate it with a little work.

mmatz-ccri: I've moved the version management to the parent where there is already some Kafka related dependency management.  Going forwared should all dependency management be added / moved to the root?

Additionally, I noticed that geomesa-kafka was using a different version of Guava than that managed by the root.  I removed the override.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
mmatz-ccri: It was only used in one place so I removed it.

geomesa-kafka/geomesa-kafka-datastore/README.md
```diff
@@ -0,0 +1,51 @@
...
+from a specified time interval in the past.  
+
+##Usage/Configuration
+Clients of the `KafkaDataStore` select between live mode and replay mode through hints provided by the SimpleFeatureType
+object passed to `createSchema(SimpleFeatureType)`.  These hints are added through the use of the following helper
+methods:
+    
+    
+    KafkaDataStoreHelper.prepareForLive(sft: SimpleFeatureType, zkPath: String) : SimpleFeatureType
+    KafkaDataStoreHelper.prepareForReplay(sft: SimpleFeatureType, rc: ReplayConfig) : SimpleFeatureType
```
elahrvivaz: Could we rename these to something like 'configureForReplay' - or if the original SFT is not modified, something like 'createReplaySft' ?

geomesa-kafka/geomesa-kafka-datastore/pom.xml
elahrvivaz: Should double check that - storm might require a different version of guava.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -16,124 +16,65 @@
...
-
-    override def equals(obj: scala.Any): Boolean = obj match {
-      case other: FeatureHolder => sf.equals(other.sf)
-      case _ => false
-    }
-  }
-
-  val cb = CacheBuilder.newBuilder()
+  override def getBoundsInternal(query: Query) =
+    ReferencedEnvelope.create(new Envelope(-180, 180, -90, 90), DefaultGeographicCRS.WGS84)
```
elahrvivaz: Should use the shared constant for 4326 here - i think it's in simple feature types. It will use the full crs if available.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
elahrvivaz: I guess it's just explicitly unsupported vs not implemented yet.

geomesa-kafka/geomesa-kafka-datastore/pom.xml
mmatz-ccri: Storm is using a 3rd version of guava.  What a mess!  I put the guava version back as it was however it seems to me that using different version of guava within different modules of geomesa is not good.  geomesa-kafka-geoserver-plugin is shading and relocating it so external requirements should not affect what version we use but I'll leave this cleanup for another day.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
```diff
@@ -0,0 +1,103 @@
+package org.locationtech.geomesa.kafka
+
+import java.util.UUID
+
+import org.locationtech.geomesa.utils.geotools.FeatureUtils
+import org.opengis.feature.simple.SimpleFeatureType
+import org.locationtech.geomesa.utils.geotools.Conversions.RichSimpleFeatureType
+
+object KafkaDataStoreHelper {
```
elahrvivaz: I think I would move this all into the KafkaDataStore object - we don't normally have 'helper' classes in scala as the companion object takes that place. @jnh5y thoughts?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+trait KafkaDataStoreSchemaManager extends DataStore  {
+
+  protected def zookeepers: String
+  protected def zkPath: String
+  protected def partitions: Int
+  protected def replication: Int
+
+  //
+  // todo:  consistent exception handling across all methods
+  // todo   - catch and rethrow a known/consistent exception with helpful msg
```
elahrvivaz: should make tickets or delete the comments.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+
+    // build the topic node
+    createZkNode(getTopicPath(typeName), kfc.topic)
+
+    // build the replay config node (optional)
+    // build the topic node
+    kfc.replayConfig.foreach(r => createZkNode(getReplayConfigPath(typeName), ReplayConfig.encode(r)))
+
+    //create the Kafka topic
+    if(!AdminUtils.topicExists(zkClient, kfc.topic))
```
elahrvivaz: use { }

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+    * ``replayType`` is based on.
+    */
+  def getLiveFeatureType(replayType: SimpleFeatureType): Option[SimpleFeatureType] = {
+
+    Try {
+      KafkaDataStoreHelper.extractLiveTypeName(replayType).map(schemaCache.get(_).sft)
+    }.getOrElse(None)
+  }
+
+  override def getNames: util.List[Name] = zkClient.getChildren(zkPath).asScala.map(
```
elahrvivaz: [picky] formatting - maybe move everything after equals down a line so it fits on one. otherwise, split on . like:
foo.bar
  .baz

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+  override def removeSchema(typeName: Name): Unit = removeSchema(typeName.getLocalPart)
+
+  override def removeSchema(typeName: String): Unit = {
+
+    // clean up cache and zookeeper resources
+    schemaCache.invalidate(typeName)
+    zkClient.deleteRecursive(getSchemaPath(typeName))
+
+    // todo: what about other remote caches - do we set up a kafka listener looking for deletes?
+    // todo: what about kafka - who has the responsibility to clean the kafka topic up?
```
elahrvivaz: tickets?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+  def getTopicPath(typeName: String): String = {
+    s"$zkPath/$typeName/Topic"
+  }
+  def getReplayConfigPath(typeName: String): String = {
+    s"$zkPath/$typeName/ReplayConfig"
+  }
+
+  private val zkClient = {
+    // zkStringSerializer is required - otherwise topics won't be created correctly
+    val ret = new ZkClient(zookeepers, Int.MaxValue, Int.MaxValue, ZKStringSerializer)
```
elahrvivaz: we should really be closing this zkClient at some point... maybe override the 'destroy' method in datastore? (might be named something else, but i think there is a lifecycle method)

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+  * (producer, live consumer, replay consumer).
+  *
+  * @constructor
+  *
+  * @param sft the [[SimpleFeatureType]]
+  * @throws IllegalArgumentException if ``sft`` has not been prepared by calling
+  *                                  ``KafkaDataStoreHelper.prepareForLive``
+  */
+@throws[IllegalArgumentException]
+private[kafka] case class KafkaFeatureConfig(sft: SimpleFeatureType) extends AnyRef {
```
elahrvivaz: hmm, is there some significance to a case class extending AnyRef? Just curious.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -0,0 +1,183 @@
...
+    .getOrElse(throw new IllegalArgumentException(
+    s"The SimpleFeatureType '${sft.getTypeName}' may not be used with KafkaDataStore because it has "
+      + "not been 'prepared' via KafkaDataStoreHelper.prepareForLive"))
+
+
+  /** the [[ReplayConfig]], if any */
+  val replayConfig: Option[ReplayConfig] = KafkaDataStoreHelper.extractReplayConfig(sft)
+
+  override def toString: String =
+    s"KafkaSimpleFeatureType: typeName=${sft.getTypeName}; topic=$topic; replayConfig=$replayConfig"
```
elahrvivaz: this should probably be 'KafkaFeatureConfig:' ?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaGeoMessageEncoding.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  def encodeDeleteMessage(topic: String, msg: Delete): MSG =
+    new MSG(topic, encodeKey(msg), GeoMessageEncoder.encodeDeleteMessage(msg))
+}
+
+/** Encodes [[GeoMessage]]s for transport via Kafka.
+  *
+  * @param schema the [[SimpleFeatureType]]; required to serialize [[CreateOrUpdate]] messages
+  */
+class KafkaGeoMessageEncoder(schema: SimpleFeatureType) extends GeoMessageEncoder(schema) {
```
elahrvivaz: Is it overkill to have the separate GeoMessageEncoder at this point? If we only have 1 implementation might be clearer to just have the logic here.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
mmatz-ccri: It already is a sealed trait.  The last case is mainly to handle null although it should not be null...

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaGeoMessageFormatter.scala
```diff
@@ -0,0 +1,176 @@
...
+  *   --formatter org.locationtech.geomesa.kafka.KafkaGeoMessageFormatter
+  *   --property sft.name={sftName}
+  *   --property sft.spec={sftSpec}
+  *
+  * In order to pass the spec via a command argument all "=" characters must be replaced by "%61".  Any
+  * "%" characters that exist prior to replacement of "=" must be replaced by "%37".
+  *
+  * @see KafkaDataStoreLogViewer for an alternative to kafka-console-consumer
+  */
+class KafkaGeoMessageFormatter extends MessageFormatter {
```
elahrvivaz: nice - a formatter for the accumulo shell would be sweet too :)

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaGeoMessageFormatter.scala
```diff
@@ -0,0 +1,176 @@
...
+  private var decoder: KafkaGeoMessageDecoder = null
+
+  override def init(props: Properties): Unit = {
+    if(!props.containsKey(sftNameKey)) {
+      throw new IllegalArgumentException(s"Property '$sftNameKey' is required.")
+    }
+
+    if(!props.containsKey(sftSpecKey)) {
+      throw new IllegalArgumentException(s"Property '$sftSpecKey' is required.")
+    }
```
elahrvivaz: [minor] scala has a 'require' function that does this in a slightly cleaner manner.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaGeoMessageFormatter.scala
```diff
@@ -0,0 +1,176 @@
...
+  }
+}
+
+/** An alternative to the kafka-console-consumer providing options specific to viewing the log messages
+  * internal to the [[KafkaDataStore]].
+  *
+  * To run, first copy the geomesa-kafka-geoserver-plugin.jar to $KAFKA_HOME/libs.  Then create a copy of
+  * $KAFKA_HOME/bin/kafka-console-consumer.sh called "kafka-ds-log-viewer" and in the copy replace the
+  * classname in the exec command at the end of the script with
+  * org.locationtech.geomesa.kafka.KafkaDataStoreLogViewer.
```
elahrvivaz: If these aren't in the readme's already, please copy them there.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
mmatz-ccri: It is used elsewhere.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
mmatz-ccri: Exactly, it not supported.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
mmatz-ccri: Yes, it is.  The SFT must be preparedForLive before it can be preparedForReplay.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
mmatz-ccri: I replaced the TODOs with a reference to this ticket:
https://geomesa.atlassian.net/browse/GEOMESA-818

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
mmatz-ccri: Nope, it was probably an AnyVal at one point and changed rather than removed.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
mmatz-ccri: good catch

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaGeoMessageFormatter.scala
mmatz-ccri: done

geomesa-kafka/geomesa-kafka-datastore/pom.xml
mmatz-ccri: I've marked jopt-simple as provided.  Currently kafka is not a provided dependency.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,163 @@
...
+                       expirationPeriod: Option[Long])(implicit ticker: Ticker)
+  extends KafkaConsumerFeatureCache {
+
+  var qt = new SynchronizedQuadtree
+
+  val cache: Cache[String, FeatureHolder] = {
+
+    val cb = CacheBuilder.newBuilder().ticker(ticker)
+
+    expirationPeriod.map { ep =>
```
elahrvivaz: prefer foreach over map to indicate the side effect

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+        val filter = tfs.filter.getOrElse(Filter.INCLUDE)
+
+        snapshot(time).map(_.getReaderForFilter(filter))
+      }
+    }
+
+    reader.getOrElse(new EmptyFeatureReader[SimpleFeatureType, SimpleFeature](replaySFT))
+  }
+
+  private[kafka] def snapshot(time: Option[Long]): Option[ReplaySnapshotFeatureCache] = {
```
elahrvivaz: private[package] seems an odd combination... might want it to be private[this] ?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+      // reverse ordering for reverse ordered ``messages``
+      val ordering = new Ordering[GeoMessage] {
+        override def compare(x: GeoMessage, y: GeoMessage): Int = y.timestamp.compareTo(x.timestamp)
+      }
+
+      var index = java.util.Arrays.binarySearch(messages, key, ordering)
+
+      if (index < 0) {
+        // no message found at sought time
+        index = -index - 1
```
elahrvivaz: is this equivalent to index = 0? that would be clearer...

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+      }
+
+      // there may be multiple messages at the same time so there may be messages before ``index``
+      // that are at the given ``time``
+      while (index > 0 && messages(index - 1).timestamp.getMillis <= time) index -= 1
```
elahrvivaz: { }

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+    // start 1 ms earlier because there might be multiple messages with the same timestamp
+    val startTime = replayConfig.realStartTime.minus(1L)
+
+    val offsetRequest = FindOffset(msg => {
+      val key = msgDecoder.decodeKey(msg)
+      if (key.ts.isEqual(startTime)) 0 else if (key.ts.isAfter(startTime)) 1 else -1
+    })
+
+    // required: there is only 1 partition;  validate??
+    val stream = kafkaConsumer.createMessageStreams(1, offsetRequest).head
```
elahrvivaz: this is how it's always worked, right? only a single partition.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
mmatz-ccri: I want to be able to unit test the function directly but not make it part of the public api.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+  import ReplayTimeHelper._
+
+  private val replayDate = new java.util.Date(replayTime)
+  private val builder = new SimpleFeatureBuilder(sft)
+  private val attrIndex = sft.indexOf(AttributeName)
+
+  require(attrIndex >= 0, s"Invalid SFT.  The $AttributeName attribute is missing.")
+
+  /** Copy the given ``sf`` and add a value for the replay time attribute. */
+  def addReplayTime(sf: SimpleFeature): SimpleFeature = {
```
elahrvivaz: i would rename this to make clear that a new feature is being created.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+                                                     replayTime: Long,
+                                                     events: Seq[GeoMessage])
+
+  extends KafkaConsumerFeatureCache {
+
+  override lazy val (qt, features) = processMessages
+
+  private def processMessages: (Quadtree, mutable.Map[String, FeatureHolder]) = {
+    val features = new mutable.HashMap[String, FeatureHolder]()
+    val qt = new Quadtree
```
elahrvivaz: does this need to be a synchronized qt?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
mmatz-ccri: It's not equivalent.  The binary search returns the "index of the search key, if it is contained in the array otherwise, <tt>(-(<i>insertion point</i>) - 1)</tt>".  This calculation is to get the correct inseition point.  I added a comment to clarify.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  type SplitCombiner = Seq[TimestampFilterSplit] => Option[TimestampFilterSplit]
+
+  def split(op: BinaryLogicOperator, combiner: SplitCombiner): Option[TimestampFilterSplit] = {
+    val children = op.getChildren.asScala
+    val childSplits = children.flatMap(c => split(c))
+
+    if (childSplits.size != children.size) {
+      // one or more children are invalid
```
elahrvivaz: does this logic work for an OR? couldn't you still get a valid timestamp from one side?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
mmatz-ccri: That what I thought but there is a configuration parameters for the number of partitions.  Why was that parameter added?  Could we support muliple partitions if messages are partitioned by the simple feature ID?  If so, replay would need some more work to handle.  Should the number of partitions parameter be removed or should I create a ticket for reviewing this another time?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/InfiniteIterator.scala
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.locationtech.geomesa.utils
+
+import scala.collection.Iterator.empty
+
+object InfiniteIterator {
```
elahrvivaz: woah

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
mmatz-ccri: No, each query to a replay consumer feature source results in the creation of a new ReplaySnapshotFeatureCache used only to answer that single query and then its thrown away.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -35,6 +38,10 @@ import scala.util.parsing.combinator.JavaTokenParsers
 
 object SimpleFeatureTypes {
 
+  type FR = FeatureReader[SimpleFeatureType, SimpleFeature]
+  type DFR = DelegateFeatureReader[SimpleFeatureType, SimpleFeature]
+  type DFI = DelegateFeatureIterator[SimpleFeature]
```
elahrvivaz: hmm, this doesn't seem like a great place for these... maybe move to a package object somewhere?

elahrvivaz:
looks good to me. i didn't review the kafka consumer stuff, since i wrote it (so it's obviously perfect ;)

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
elahrvivaz: It should be fine to have a single consumer against multiple partitions I think, you just don't get any parallelism so it's kind of pointless. I would create a ticket to revisit the issue.

mmatz-ccri:
Thanks elahrvivaz.  You might want to take a quick look at the consumer stuff since I changed it a bit.  I had an issue with the binary search (NPE referencying a lazy val) so I rewrote it a little to simplify the logic and added a unit tested it.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/GeoMessage.scala
mmatz-ccri: I did some refactoring to clean this up a bit.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreHelper.scala
mmatz-ccri: The KafkaDataStoreHelper is responsible for managing the userdata in SFT required by KafkaDataStore. Its enough of its own thing that I'd like to keep it in its own file.  However, I'm not opposed to renaming it if you can suggest something better.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
mmatz-ccri: added close in dispose()

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/ReplayKafkaConsumerFeatureSource.scala
mmatz-ccri: I could handle each side of the OR and then union the results.  But, you might end up with multiple different versions of the same simple feature at different points in time.  There might be a use case for this but there is also an issues with the different versions having the same ID but not being equal.  Currently we don't need to support this, as far as I know.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
mmatz-ccri: moved

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaGeoMessageEncoding.scala
mmatz-ccri: probably

geomesa-kafka/geomesa-kafka-datastore/README.md
```diff
@@ -0,0 +1,73 @@
...
+    
+###Data Producers
+
+Data *producers* should build a SimpleFeatureType object and pass it to `prepareForLive` along with a zookeeper path that 
+must be unique for a given KafkaDataStore.  The Consumer should pass the returned SimpleFeature type to the call to
```
Afterglow375: Typo, consumer doesn't have anything to do with Datastore.createSchema.

AndrewAnnex:
Merged in: 5595bdaeac207ddfb7903246f3a97bc45b6e0c88

PR #601, opened 2015-06-05
GEOMESA-821 Fix DataStoreFactorySpi entry in GeoMesa plugin META-INF/…

mpfaffenberger:
…services

### DIFF
--- a/geomesa-plugin/src/main/resources/META-INF/services/org.geotools.data.DataStoreFactorySpi
+++ b/geomesa-plugin/src/main/resources/META-INF/services/org.geotools.data.DataStoreFactorySpi
@@ -1,1 +1,0 @@
-org.locationtech.geomesa.accumulo.data.AccumuloDataStoreFactory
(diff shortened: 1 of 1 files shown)


geomesa-plugin/src/main/resources/META-INF/services/org.geotools.data.DataStoreFactorySpi
```diff
@@ -1 +1 @@
-accumulo.data.AccumuloDataStoreFactory
+org.locationtech.geomesa.accumulo.data.AccumuloDataStoreFactory
```
elahrvivaz: i think this file is actually redundant and can be completely removed, as it comes in already from the geomesa-accumulo module.

AndrewAnnex:
@mpfaffenberger is this PR still needed?

mpfaffenberger:
Yep

elahrvivaz:
@mpfaffenberger See my comment re:removing.

geomesa-plugin/src/main/resources/META-INF/services/org.geotools.data.DataStoreFactorySpi
jnh5y: @mpfaffenberger can you try removing the file instead?

mpfaffenberger:
@elahrvivaz @jnh5y updated to remove the file. Not sure what's up with the CLA thing. I definitely signed it and verified that my github email was the same. :octocat:

jnh5y:
https://projects.eclipse.org/user/cla/validate says you're fine Mike, so no worries.

anthonyccri:
Merged.

PR #603, opened 2015-06-08
GEOMESA-795 Write attribute histogram results to accumulo "_stats" table

jdk2pq:
Attribute histogram job results can now be written back to accumulo in a {feature-name}_stats table by setting the "geomesa.hist.write.to.accumulo" to true.

### DIFF
--- a/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
+++ b/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
@@ -19,7 +19,9 @@ package org.locationtech.geomesa.jobs.analytics
 import com.twitter.algebird.Aggregator
 import com.twitter.scalding._
 import com.twitter.scalding.typed.UnsortedGrouped
+import org.apache.accumulo.core.data.Mutation
 import org.apache.hadoop.conf.Configuration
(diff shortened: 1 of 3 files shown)


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/analytics/HistogramJob.scala
```diff
@@ -47,14 +48,18 @@ class HistogramJob(args: Args) extends GeoMesaBaseJob(args) {
   val groupBy    = args.nonStrictList(GROUP_BY)
   val uniqueBy   = args.nonStrictList(UNIQUE_BY)
   val transforms = Option(args.list(TRANSFORM_IN).toArray).filter(_.nonEmpty)
+  val writeToAccumulo  = args.boolean(WRITE_TO_ACCUMULO)
 
   val input = GeoMesaInputOptions(dsParams, feature, filter, transforms)
+  lazy val dsParamsAccOutput = dsParams.updated("tableName", s"${dsParams.get("tableName").get}_stats")
```
AndrewAnnex: not that it will happen, but it may be safer to do a getOrElse to avoid the null.

AndrewAnnex:
I think this can be merged, maybe that one thing I noted above should be addressed first however.

PR #607, opened 2015-06-10
GEOMESA-824 Aggregating bin records in tablet servers

elahrvivaz:
* Adding optional sorting of bin records, since regular feature sorting will no longer work
* BinaryViewerOutputFormat expands aggregated records
* Non-z3 indices do not aggregate yet

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
@@ -105,16 +105,6 @@ object Conversions {
       }
     }
   }
-
-  implicit class RichSimpleFeatureType(val sft: SimpleFeatureType) extends AnyVal {
-
-    def userData[T](key: AnyRef)(implicit ct: ClassTag[T]): Option[T] = {
-      Option(sft.getUserData.get(key)).flatMap {
-        case ct(x) => Some(x)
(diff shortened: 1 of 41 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -180,6 +183,22 @@ object RichAttributeDescriptors {
...
 }
 
+object RichSimpleFeatureType {
+
+  import RichAttributeDescriptors.RichAttributeDescriptor
+  import SimpleFeatureTypes.DEFAULT_DATE_FIELD
+
+  import scala.collection.JavaConversions._
+
+  implicit class RichSimpleFeatureType(val sft: SimpleFeatureType) extends AnyVal {
```
mmatz-ccri: merge with caution: in master there already is an "implicit class RichSimpleFeatureType" as a nested class of "object Conversions"

elahrvivaz: ah, thanks. i need to rebase this on master, actually.

elahrvivaz: rebased and fixed.

anthonyccri:
Merged

PR #614, opened 2015-06-15
GEOMESA-841 Z3 recursion should return something at the bottom

jnh5y:
This is a quick workaround to the fact that Z3's recursion currently returns no indices.


elahrvivaz:
3fb38cffbc4bcd905a879ce1a1d47323d568673d

PR #616, opened 2015-06-15
GEOMESA-623 - GeoMesa Copyright Notice

mm64z:
Changing from old copyright headers to new ones,
per request

Signed-off-by: Michael Matthews <mmatthews@ccri.com>


AndrewAnnex:
it looks like you may be missing several modules like geomesa-raster and geomesa-tools just to name two.

PR #617, opened 2015-06-16
GEOMESA-837 Switching Kafka DS to use bucket index instead of JTS quadtree

elahrvivaz:
* Also fixing potential NPE in BinaryOutputEncoder

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/BucketIndex.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/BucketIndex.scala
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2013-2015 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0 which
+ * accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ */
+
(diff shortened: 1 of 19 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/BucketIndex.scala
```diff
@@ -0,0 +1,72 @@
...
+ */
+class BucketIndex[T](xBuckets: Int = 360,
+                     yBuckets: Int = 180,
+                     extents: Envelope = new Envelope(-180.0, 180.0, -90.0, 90.0)) extends SpatialIndex[T] {
+
+  // create the buckets upfront to avoid having to synchronize the whole array
+  private val gridSnap = new GridSnap(extents, xBuckets, yBuckets)
+  private val buckets = Array.fill(xBuckets, yBuckets)(mutable.HashSet.empty[T])
+
+  override def insert(envelope: Envelope, item: T): Unit = {
```
jnh5y: If we accept Envelopes for insert, how does this not take non-point geometries?

elahrvivaz: I wanted the interface to be general, but we use the center of the envelope to stick it in a bucket. We'd have to dedupe if we stick it in every bucket that intersects the envelope - wanted to speed it up since we're only using points right now.

PR #620, opened 2015-06-16
GEOMESA-843 Enable Zinc compilation

jnh5y:
For information about using Zinc, check out https://github.com/typesafehub/zinc.
You must run a Zinc server separately to see the speed up, but this is optional.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+build/zinc*
+build/scala*
+build/*.jar
 *.iws
 *.ipr
 *.iml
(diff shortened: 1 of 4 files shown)


anthonyccri:
Local build is failing for me.  Seems like a unit test in geomesa-jobs is failing?

elahrvivaz:
Could we make this a profile so that a build doesn't require installation of more dependencies? Some environments might not have zinc available. Although that might introduce subtle inconsistencies if the compiler result differs at all b/t zinc and non-zinc...

elahrvivaz:
@jnh5y any thoughts to making zinc compilation enabled via profile?

elahrvivaz:
merged as a4292a49e69fadf51ab0e2fc0e40b8b9ef76087d

PR #621, opened 2015-06-17
GEOMESA-846  Allowing for appending of attributes to simple features types with kryo encoding

elahrvivaz:

### DIFF
--- a/geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
+++ b/geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
@@ -51,7 +51,7 @@ class KryoBufferSimpleFeature(sft: SimpleFeatureType,
     input.setPosition(input.readInt()) // set to offsets start
     var i = 0
     while (i < offsets.length) {
-      offsets(i) = input.readInt(true)
+      offsets(i) = if (input.position < input.limit) input.readInt(true) else -1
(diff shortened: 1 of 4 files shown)


anthonyccri:
@elahrvivaz  @BrittEvans  Do these appended attributes get persisted into the database or are they just ephemeral?  And if they're ephemeral, why wouldn't we put them in the userData hash map?  Presumably performance?

elahrvivaz:
@anthonyccri They get persisted into the database when new features are written. The old features will have null values for the new attributes.

anthonyccri:
@elahrvivaz Does this work out ok with indexes?

@jnh5y Does this address some of the requests we've had regarding mutable simple feature types?

elahrvivaz:
@anthonyccri what do you mean by indices? Essentially, the kryo deserialization will work on the newly serialized features and the old serialized features.
As an addendum, I think we should deprecate support for avro encoding - we've made a lot of performance improvements that only works with kryo encoding that we could expand to use in more places if we don't have to worry about avro serialization.

jnh5y:
@anthonyccri I haven't had a chance to read the PR yet.  I'd assume that this would make supporting modifySchema a slam-dunk.

jahhulbert-ccri:
I'll add a modify schema command to the tools once it's in

anthonyccri:
@jnh5y It's pretty close . . . though deleting an attribute doesn't exist.

anthonyccri:
Ok, merging.

elahrvivaz:
Actually, deleting attributes might work too... we'd have to test to make sure though. And adding then deleting (or vice versa) would probably screw it up.

PR #622, opened 2015-06-17
GEOMESA-844 Set recommended table configuration options on create

anthonyccri:
* Set table.split.threshold to 128M
* Set table.block.cache.enabled to true
* NOTE: maxShards has been somewhat deprecated

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
@@ -10,8 +10,12 @@ package org.locationtech.geomesa.accumulo.data.tables
 
 import com.typesafe.scalalogging.slf4j.Logging
 import org.apache.accumulo.core.client.BatchDeleter
+import org.apache.accumulo.core.client.admin.TableOperations
+import org.apache.accumulo.core.conf.Property
 import org.apache.accumulo.core.data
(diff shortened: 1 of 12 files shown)


elahrvivaz:
looks good.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
```diff
@@ -86,4 +90,20 @@ object SpatioTemporalTable extends GeoMesaTable with Logging {
     bd.delete()
     bd.close()
   }
+
+
+  override def configureTable(sft: SimpleFeatureType, tableName: String, tableOps: TableOperations): Unit = {
+    // NOTE: since the ST table is likely going away, I'm not inclined to thread maxShards all the way through
+    // the call chain so I just set a default of 40 here
```
jnh5y: I'm fine with this right now.  In order to codify routing extras info, how about widening configureTable to take a Map[String, String]?

anthonyccri: Yeah, I thought about that and possibly just using some hints in the user data of the SFT.  Let's push that off till later though.

jnh5y: Ah, user data, of course.  And yes, future us have lots of to do.

anthonyccri:
Merged.

PR #625, opened 2015-06-24
add basic date time format

cgosse:
This format is needed because there isn't enough string escaping in place to specify it manually


jnh5y:
Committed as 9da0e434bbab3c3a1037b16d3d3e4d4f5b583ff5 on master.

PR #627, opened 2015-06-26
GEOMESA-806 Don't run multiple queries when only secondary filters have OR clauses

elahrvivaz:
* Strategies have been refactored to include the query they are executing - this avoids determining the correct filters twice
* New class - QueryFilterSplitter - handles logic for splitting filters on Boolean clauses and combining them for a particular strategy
* Filters that can't be satisfied by any index now run against the record table instead of the ST table, in anticipation of eventually removing the ST table

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
@@ -14,7 +14,7 @@ import com.typesafe.scalalogging.slf4j.Logging
 import org.apache.accumulo.core.client.IteratorSetting
 import org.apache.accumulo.core.data.{Range => AccRange}
 import org.apache.hadoop.io.Text
-import org.geotools.data.Query
+import org.geotools.factory.Hints
 import org.geotools.filter.text.ecql.ECQL
 import org.geotools.temporal.`object`.DefaultPeriod
 import org.locationtech.geomesa.accumulo._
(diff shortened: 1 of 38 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
```diff
@@ -27,84 +27,75 @@ import org.locationtech.geomesa.accumulo.index.Strategy._
...
-    val ecqlFilter: Option[Filter] = filterListAsAnd(nonSTFilters)
-
     val encoding = queryPlanner.featureEncoding
     val version = acc.getGeomesaVersion(sft)
     val hasDupes = sft.getDescriptor(attributeName).isMultiValued
 
-    val kvsToFeatures = if (query.getHints.isBinQuery) {
+    val attributeIterators = scala.collection.mutable.ArrayBuffer.empty[IteratorSetting]
+
+    val (stFilter, ecqlFilter) = filter.secondary.map { f =>
```
jnh5y: Out of curiosity, why is there special handling for the ST bits of the filter when using the attribute index?  (I'm guessing there is something I haven't had a chance to read through or put together.)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
```diff
@@ -105,63 +106,64 @@ case class QueryPlanner(sft: SimpleFeatureType,
...
     (queryPlans, numClauses)
   }
 
+  // output the query plan for explain logging
+  private def outputPlan(plan: QueryPlan, output: ExplainerOutputType, joinPrefix: String = ""): Unit = {
+    output(s"${joinPrefix}Table: ${plan.table}")
+    output(s"${joinPrefix}Column Families${if (plan.columnFamilies.isEmpty) ": all"
+      else s" (${plan.columnFamilies.size}): ${plan.columnFamilies.take(20)}"} ")
+    output(s"${joinPrefix}Ranges (${plan.ranges.size}): ${plan.ranges.take(5).mkString(", ")}")
+    output(s"${joinPrefix}Iterators (${plan.iterators.size}): ${plan.iterators.mkString("[", "],[", "]")}")
```
jnh5y: Would it make this more readable to call output on a multi-line string here?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
elahrvivaz: Happens in the join case - the attribute value in that case will only be the index value (e.g. date and geom). So we need to know if the filter can be satisfied by that, or if we need to apply it to the record table instead.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
elahrvivaz: ah, possibly... i think i prefer it this way unless you have a strong preference.

jnh5y: Meh, I'm fine either way.

jnh5y:
After one quick read through, this looks reasonable.  I wouldn't mind having more time to double-check logic, but no opposition to this being merged.

elahrvivaz:
@jnh5y I don't think there's a big rush so take your time.

jnh5y:
Committed as 5a29fe791362ea49d8ef93684ed3bfa22b2f3d41 on master.

PR #637, opened 2015-07-10
GEOMESA-851 Using z3 index in feature IDs to provide data locality

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureWriter.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureWriter.scala
@@ -8,7 +8,7 @@
 
 package org.locationtech.geomesa.accumulo.data
 
-import java.util.UUID
+import java.util.concurrent.atomic.AtomicLong
 
 import com.typesafe.scalalogging.slf4j.Logging
 import org.apache.accumulo.core.client.BatchWriter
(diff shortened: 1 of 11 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureWriter.scala
```diff
@@ -71,6 +74,44 @@ object AccumuloFeatureWriter {
...
+  }
+
+  private def withFid(sft: SimpleFeatureType, feature: SimpleFeature, fid: String): SimpleFeature = {
+    try {
+      // FeatureIdImpl appears to be the only implementing class for FeatureId,
+      // but just in case wrap in a try/catch
+      feature.getIdentifier.asInstanceOf[FeatureIdImpl].setID(fid)
+      feature
+    } catch {
+      case e: ClassCastException =>
```
jnh5y: I think it may be better to use a case match rather than a try-catch to handle the FeatureIdImpl vs. other case.

elahrvivaz: yeah, i was trying to speed it up a little, and couldn't find any other implementations of FeatureId, so it seems like a real fringe case that it would be some other implementation - but I will change it.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/Z3Table.scala
```diff
@@ -108,21 +108,25 @@ object Z3Table extends GeoMesaTable {
...
     val z3idx = Longs.toByteArray(z3.z)
+    Bytes.concat(prefix, z3idx)
+  }
 
+  private def getRowKey(ftw: FeatureToWrite, dtgIndex: Int): Array[Byte] = {
+    val geom = ftw.feature.point
+    val x = geom.getX
+    val y = geom.getY
+    val dtg = ftw.feature.getAttribute(dtgIndex).asInstanceOf[Date]
+    val time = if (dtg == null) System.currentTimeMillis() else dtg.getTime
```
jnh5y: Do we actually deal with situations where dtg is null often?  

As sanity check, if the SFT doesn't have a date, do we still have a Z3 table?

elahrvivaz: The SFT is guaranteed to have a dtg field, and normally all features would have a dtg, but this catches the case where a particular feature has a null date. We could just not support that case, but this way it won't throw an NPE and bomb out the whole ingest.

jnh5y:
I've only skimmed, but it looks reasonable.

Are there tests for each of the ID generators?

elahrvivaz:
I didn't add tests for the other generators, as the generator class is loaded in an object by system property, so changing it in one test would affect all other tests. I considered re-loading it dynamically, but didn't want to degrade runtime performance for the sake of unit testing...

elahrvivaz:
General question - I'm sharding the UUID based on a hash of the z3, so we get hex splitting - however, that also splits up the locality. Better to have more locality or nice splits?

anthonyccri:
5789033..af9a208

PR #638, opened 2015-07-10
GEOMESA-819 Including size of query box in z3 range calculation

elahrvivaz:
* Start recursion at the longest common prefix of the min/max bounds
* Depth of recursion depends on the size of our query region
* Compared to fixed recursion depth, increases ranges for small query regions and decreases ranges for large query regions
* Most queries will create less than 10k ranges - usually on the order of 2-4k for moderate queries and 30-500 for small queries

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
@@ -117,31 +117,33 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with 
     // setup Z3 iterator
     val env = geometryToCover.getEnvelopeInternal
     val (lx, ly, ux, uy) = (env.getMinX, env.getMinY, env.getMaxX, env.getMaxY)
+
+    def qp(week: Int, lt: Long, ut: Long, contained: Boolean) =
+      queryPlanForPrefix(week, (lx, ux), (ly, uy), (lt, ut), z3table,
+        kvsToFeatures, iterators, colFamily, numThreads, contained)
+
     val epochWeekStart = Weeks.weeksBetween(Z3Table.EPOCH, interval.getStart)
     val epochWeekEnd = Weeks.weeksBetween(Z3Table.EPOCH, interval.getEnd)
     val weeks = scala.Range.inclusive(epochWeekStart.getWeeks, epochWeekEnd.getWeeks)
     val lt = Z3Table.secondsInCurrentWeek(interval.getStart, epochWeekStart)
     val ut = Z3Table.secondsInCurrentWeek(interval.getEnd, epochWeekEnd)
+
     if (weeks.length == 1) {
-      Seq(queryPlanForPrefix(weeks.head, lt ,ut, lx, ly, ux, uy,
-        z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = false))
+      Seq(qp(weeks.head, lt, ut, contained = false))
     } else {
-      val oneWeekInSeconds = Weeks.ONE.toStandardSeconds.getSeconds
       val head +: xs :+ last = weeks.toList
-      val middleQPs = xs.map { w =>
-        queryPlanForPrefix(w, 0, oneWeekInSeconds, lx, ly, ux, uy,
-          z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = true)
-      }
-      val startQP = queryPlanForPrefix(head, lt, oneWeekInSeconds, lx, ly, ux, uy,
-        z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = false)
-      val endQP = queryPlanForPrefix(last, 0, ut, lx, ly, ux, uy,
-        z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = false)
+      val oneWeekInSeconds = Weeks.ONE.toStandardSeconds.getSeconds
+      val startQP = qp(head, lt, oneWeekInSeconds, contained = false)
+      val endQP = qp(last, 0, ut, contained = false)
+      val middleQPs = xs.map(w => qp(w, 0, oneWeekInSeconds, contained = true))
       Seq(startQP, endQP) ++ middleQPs
     }
   }
 
-  def queryPlanForPrefix(week: Int, lt: Long, ut: Long,
-                         lx: Double, ly: Double, ux: Double, uy: Double,
+  def queryPlanForPrefix(week: Int,
+                         x: (Double, Double),
+                         y: (Double, Double),
+                         t: (Long, Long),
                          table: String,
                          kvsToFeatures: FeatureFunction,
                          is: Seq[IteratorSetting],
@@ -151,9 +153,7 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with 
     val epochWeekStart = Weeks.weeks(week)
     val prefix = Shorts.toByteArray(epochWeekStart.getWeeks.toShort)
 
-    val z3ranges = Z3_CURVE.ranges(lx, ly, ux, uy, lt, ut, 8)
-
-    val accRanges = z3ranges.map { case (s, e) =>
+    val accRanges = Z3_CURVE.ranges(x, y, t).map { case (s, e) =>
       val startRowBytes = Bytes.concat(prefix, Longs.toByteArray(s))
       val endRowBytes = Bytes.concat(prefix, Longs.toByteArray(e))
       val start = new Text(startRowBytes)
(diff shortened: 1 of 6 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
```diff
@@ -117,31 +117,33 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with
...
-        queryPlanForPrefix(w, 0, oneWeekInSeconds, lx, ly, ux, uy,
-          z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = true)
-      }
-      val startQP = queryPlanForPrefix(head, lt, oneWeekInSeconds, lx, ly, ux, uy,
-        z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = false)
-      val endQP = queryPlanForPrefix(last, 0, ut, lx, ly, ux, uy,
-        z3table, kvsToFeatures, iterators, colFamily, numThreads, contained = false)
+      val oneWeekInSeconds = Weeks.ONE.toStandardSeconds.getSeconds
+      val startQP = qp(head, lt, oneWeekInSeconds, contained = false)
+      val endQP = qp(last, 0, ut, contained = false)
```
jnh5y: You don't need the contained = Boolean syntax here, do you?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
```diff
@@ -117,31 +117,33 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with
...
+      val middleQPs = xs.map(w => qp(w, 0, oneWeekInSeconds, contained = true))
```
jnh5y: I know this isn't your code, but I'd consider renaming oneWeekInSeconds to something more helpful...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
elahrvivaz: you don't need it, but intellij recommends 'naming boolean parameters' and flags it as a warning...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
elahrvivaz: I'll rename

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
```diff
@@ -107,52 +108,66 @@ object Z3 {
...
-            mq += (qr.min.z, qr.max.z)
-          }
+  def zranges(min: Z3, max: Z3): Seq[(Long, Long)] = {
+    val (commonPrefix, commonBits) = longestCommonPrefix(min.z, max.z)
+
+    // base our recursion on the depth of the tree that we get 'for free' from the common prefix
+    val maxRecurse = if (commonBits < 30) 7 else if (commonBits < 40) 6 else 5
+
+    val searchRange = Z3Range(min, max)
+    var mq: MergeQueue = new MergeQueue // stores our results
```
jnh5y: Unnecessary type annotation.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
```diff
@@ -107,52 +108,66 @@ object Z3 {
...
-            zranges(min, offset - MAX_DIM, 7, level + 1)
-            //let our children punt on each subrange
-          } else {
-            mq += (qr.min.z, qr.max.z)
-          }
+  def zranges(min: Z3, max: Z3): Seq[(Long, Long)] = {
+    val (commonPrefix, commonBits) = longestCommonPrefix(min.z, max.z)
+
+    // base our recursion on the depth of the tree that we get 'for free' from the common prefix
+    val maxRecurse = if (commonBits < 30) 7 else if (commonBits < 40) 6 else 5
```
jnh5y: What happens if the query size is within 5 of the 63/64-bit limit?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
```diff
@@ -107,52 +108,66 @@ object Z3 {
...
+    // return our aggregated results
     mq.toSeq
   }
 
   /**
+   * Calculates the longest common binary prefix between two z longs
+   *
+   * @return (common prefix, number of bits in common)
+   */
+  def longestCommonPrefix(lower: Long, upper: Long): (Long, Int) = {
```
jnh5y: @cne1x do you have any suggestions / opinions about this method?  I remember you sharing a method for this calculation.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
elahrvivaz: It will get stopped either via the contains clause or via the offset > 0 clause.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
elahrvivaz: removed

geomesa-z3/src/test/scala/org/locationtech/geomesa/curve/Z3Test.scala
```diff
@@ -162,34 +162,46 @@ class Z3Test extends Specification {
...
+        (sfc.index(35, 65, 0), sfc.index(45, 75, day)), // 10^2 degrees, 1 day
+        (sfc.index(35, 55, 0), sfc.index(45, 65, week)), // 10^2 degrees, full week
+        (sfc.index(35, 55, day), sfc.index(45, 75, day * 2)), // 10x20 degrees, 1 day
+        (sfc.index(35, 55, day + hour * 6), sfc.index(45, 75, day * 2)), // 10x20 degrees, 18 hours
+        (sfc.index(35, 65, day + hour), sfc.index(45, 75, day * 6)), // 10^2 degrees, 5 days 23 hours
+        (sfc.index(35, 65, day), sfc.index(37, 68, day + hour * 6)), // 2x3 degrees, 6 hours
+        (sfc.index(35, 65, day), sfc.index(40, 70, day + hour * 6)), // 5^2 degrees, 6 hours
+        (sfc.index(39.999, 60.999, day + 3000), sfc.index(40.001, 61.001, day + 3120)), // small bounds
+        (sfc.index(51.0, 51.0, 6000), sfc.index(51.1, 51.1, 6100)), // small bounds
+        (sfc.index(51.0, 51.0, 30000), sfc.index(51.001, 51.001, 30100)) // small bounds
```
jnh5y: Can we add a test for a query which already has 61/62 bits of precision given?

elahrvivaz: will do.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3.scala
cne1x: @jnh5y I'm fine with this routine as is, especially since it takes advantage of not checking every single bit position.  For fun, if you wanted something less intelligent (but without any conditionals), I suspect that you could use something like this:

```scala
def longestCommonPrefix(lower: Long, upper: Long) = {
  // a bit-mask in which all shared positions in the prefix are 0, and 1 everywhere else
  var v = lower ^ upper
  v = v | (v >> 1L)
  v = v | (v >> 2L)
  v = v | (v >> 4L)
  v = v | (v >> 8L)
  v = v | (v >> 16L)
  v = v | (v >> 32L)

  // count the number of bits not in the shared prefix
  var c = v - ((v >> 1L) & 0x5555555555555555L)
  c = ((c >> 2L) & 0x3333333333333333L) + (c & 0x3333333333333333L)
  c = ((c >> 4L) + c)  & 0x0F0F0F0F0F0F0F0FL
  c = ((c >> 8L) + c)  & 0x00FF00FF00FF00FFL
  c = ((c >> 16L) + c) & 0x0000FFFF0000FFFFL
  c = ((c >> 32L) + c) & 0x00000000FFFFFFFFL

  (~v, 64-c)
}
```

TL/DR:  Stick with what's already there.  It works, and is easy to read.

anthonyccri:
5789033

PR #649, opened 2015-07-22
RFC: Reflection approach.

jnh5y:


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.hadoop.mapred.JobConf
+import org.apache.hadoop.mapreduce.Job
+import org.apache.log4j.Level
+import org.locationtech.geomesa.accumulo.AccumuloVersion
+import AccumuloVersion._
+
+object InputFormatBaseAdapter {
+  def setConnectorInfo(job: JobConf, user: String, token: PasswordToken) = accumuloVersion match {
+    case V15 => setConnectorInfo15(job, user, token)
+    case V16 => setConnectorInfo16(job, user, token)
```
jnh5y: I'm thinking that the methods in this file can be de-duplicated.

anthonyccri: I'm in favor of leaving in the duplicated methods.  It makes it very clear that they are using reflection and could possibly break on us.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  def setConnectorInfo15(job: JobConf, user: String, token: PasswordToken) = {
```
elahrvivaz: we should make the 15/16 methods private so that the external interface is cleaner (assuming they would never be directly called)

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
jnh5y: I think it may make more sense to indicate that the functions are working around a Scala-Java issue and to be explicit about that.

PR #653, opened 2015-07-28
GEOMESA-872 Wait for Kafka topic to be ready for writes

jnh5y:
* Implements a method to wait for a new Kafka topic to be ready for writes.

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
@@ -84,12 +84,15 @@ trait KafkaDataStoreSchemaManager extends DataStore with Logging {
       leader = ZkUtils.getLeaderForPartition(zkClient, topic, partition)
       if(leader.isEmpty) {
         logger.debug(s"Still waiting on a leader for topic: $topic")
+        Thread.sleep(100)
       }
     }
 
     leader match {
       case Some(l) => logger.debug(s"Got a leader for topic: $topic")
-      case _       => throw new Exception(s"Failed to get a leader for topic: $topic")
+      case _       =>
+        throw new Exception(s"Failed to get a leader for partition $partition of " +
+                            s"topic: $topic within $timeToWait milliseconds.")
     }
   }
 
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
@@ -10,8 +10,9 @@ package org.locationtech.geomesa.kafka
 import java.util
 
 import com.google.common.cache.{CacheBuilder, CacheLoader}
+import com.typesafe.scalalogging.slf4j.Logging
 import kafka.admin.AdminUtils
-import kafka.utils.ZKStringSerializer
+import kafka.utils.{ZkUtils, ZKStringSerializer}
 import org.I0Itec.zkclient.ZkClient
 import org.I0Itec.zkclient.exception.ZkNodeExistsException
 import org.geotools.data.DataStore
@@ -28,7 +29,7 @@ import scala.util.control.NonFatal
   *
   * See GEOMESA-818 for additional considerations.
   */
-trait KafkaDataStoreSchemaManager extends DataStore  {
+trait KafkaDataStoreSchemaManager extends DataStore with Logging {
 
   protected def zookeepers: String
   protected def zkPath: String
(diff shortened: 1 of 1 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -61,10 +62,37 @@ trait KafkaDataStoreSchemaManager extends DataStore  {
...
+    waitUntilTopicReady(zkClient, kfc.topic, partitions)
+
     // put it in the cache
     schemaCache.put(typeName, kfc)
   }
 
+  // Wait until a topic is ready for writes.
+  // NB: The assumption here is that waiting for the leader to be elected is sufficient.
+  // Further, this function delegates to a function which waits for a leader for each partition.
+  def waitUntilTopicReady(zkClient: ZkClient, topic: String, partitions: Int, timeToWait: Long = 5000L): Unit = {
```
jnh5y: @mmatz-ccri Do you have an opinion on where this function may go better?  I didn't notice a utils package/area.

mmatz-ccri: I think that this is a good enough place for this function since its only needed by KafkaDataStoreSchemaManager and we don't have a Kafka utils package or project.

mmatz-ccri: Should the timeout be configurable?  Should it be per partition as it is now or per topic?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -61,10 +62,37 @@ trait KafkaDataStoreSchemaManager extends DataStore  {
...
+  }
+
+  def waitUntilTopicReadyForPartition(zkClient: ZkClient, topic: String, partition: Int, timeToWait: Long) {
+    var leader: Option[Int] = None
+    val start = System.currentTimeMillis
+
+    while(leader.isEmpty && System.currentTimeMillis < start + timeToWait) {
+      leader = ZkUtils.getLeaderForPartition(zkClient, topic, partition)
+      if(leader.isEmpty) {
+        logger.debug(s"Still waiting on a leader for topic: $topic")
```
mmatz-ccri: Add a sleep?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -61,10 +62,37 @@ trait KafkaDataStoreSchemaManager extends DataStore  {
...
+      }
+    }
+
+    leader match {
+      case Some(l) => logger.debug(s"Got a leader for topic: $topic")
+      case _       => throw new Exception(s"Failed to get a leader for topic: $topic")
```
mmatz-ccri: Add partition num and timeout to error message.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
jnh5y: How about 100 milliseconds?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
jnh5y: I don't mind it being per partition.  I'm assuming that the partitions will be getting sorted out together, and generally this should get handled within 5 seconds.

There are several assumptions and I'm happy for push back.  If this is sufficient for a first pass, then I'm happy.

mmatz-ccri:
Looks good.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
mteldridge: What is the plan for testing this?  It seems that you need to test
- The looping/sleep/poll logic in a unit test

Would need to wrap the static `ZkUtils.getLeaderForPartition` call in a class and use mockito to mock cases where the method returns normally, never returns, returns late, etc.
- Whether this actually works against zookeeper

This one is harder.  This would be an integration test using zookeeper, perhaps just confirms that it causes no errors.  Ideally, we would want to validate the assumptions, e.g. : 
- Finished leader election is sufficient
- The interactions don't cause other side effects

elahrvivaz:
502986757395b4f5b6678a3a6dae6b5ac425b8a4

PR #654, opened 2015-07-29
GEOMESA-873 ImportProcess should set the default geometry for importe…

cne1x:
…d feature types

Orginally reported by Nick Hamblet.  Added a new routine and unit tests.

### DIFF
--- a/geomesa-process/src/main/scala/org/locationtech/geomesa/process/ImportProcess.scala
+++ b/geomesa-process/src/main/scala/org/locationtech/geomesa/process/ImportProcess.scala
@@ -18,12 +18,15 @@ import org.geotools.process.factory.{DescribeParameter, DescribeProcess, Describ
 import org.locationtech.geomesa.accumulo.data.{AccumuloDataStore, AccumuloFeatureStore}
 
 import scala.collection.JavaConversions._
+import scala.util.Try
 
 @DescribeProcess(
   title = "Geomesa Bulk Import",
   description = "Bulk Import data into Geomesa from another process with no transformations of data"
 )
 class ImportProcess(val catalog: Catalog) extends GeomesaProcess {
+
+  import ImportProcess._
 
   @DescribeResult(name = "layerName", description = "Name of the new featuretype, with workspace")
   def execute(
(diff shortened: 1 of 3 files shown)


geomesa-process/src/main/scala/org/locationtech/geomesa/process/ImportProcess.scala
```diff
@@ -126,3 +130,11 @@ class ImportProcess(val catalog: Catalog) extends GeomesaProcess {
 
 }
 
+object ImportProcess {
+  // the name of this function is almost as long as its body,
+  // but it makes the routine amenable to unit testing
+  def getFeatureCollectionDefaultGeometryNameOption(features: SimpleFeatureCollection): Option[String] =
```
nhambletCCRI: I wonder if this would be a useful helper method somewhere else?

nhambletCCRI:
This PR looks fine to me. It is unfortunate that the corresponding layer, when served up by GeoServer, still uses the first geometry column, which I didn't anticipate.

cne1x:
@nhambletCCRI Was it a layer cooked up with this PR?  If so, then I would say that's sufficient evidence to reject the proposed solution.

nhambletCCRI:
It was a layer built with this PR, but I think it wouldn't matter where the layer came from. At least, that's my read of [this admittedly old post](http://osgeo-org.1560.x6.nabble.com/Two-geometry-columns-GeoServer-ignoring-geometry-columns-td3802701.html)

jnh5y:
Right, I think I've seen the same posts where GeoServer by and large ignores 'default' geometries preferring the first geometry field.

geomesa-process/src/test/scala/org/locationtech/geomesa/process/ImportProcessTest.scala
```diff
@@ -0,0 +1,94 @@
...
+      val sft = builder.buildFeatureType()
+      val features = new DefaultFeatureCollection("id", sft)
+
+      logger.debug(s"[geometry unset] default geometry:  ${sft.getGeometryDescriptor}")
+
+      val expectedGeomName = Try { sft.getGeometryDescriptor.getName.toString }.toOption
+      val actualGeomName = ImportProcess.getFeatureCollectionDefaultGeometryNameOption(features)
+
+      actualGeomName.isEmpty must beTrue
+      expectedGeomName.isEmpty must beTrue
```
jnh5y: Should we be testing if the strings are equal?

jnh5y: Oh!  Nevermind.  I can't read.

PR #656, opened 2015-08-05
GEOMESA-879 Fixing libjars for command-line ingestion

elahrvivaz:
Actual changes were to remove the 3 jars that were generating a warning message for missing jars (which were not actually required), and adding in the curator jars and the typesafe config jar. Also alphabetized the list of jars.
Updated the loading of the CRS to a full try as scala Try does not catch ClassNotFound errors.


jnh5y:
Merged as d2727bc37081b775152884815a929fbb192a9d70 on master.

PR #659, opened 2015-08-07
GEOMESA-881 Fixing incomplete result sets in GeoMesaSpark

elahrvivaz:
* M/r and spark jobs can't operate on multiple query plans, as the input format can only be configured once.
* M/r and spark jobs will now fall back to a full table scan instead of returning incomplete results if there are multiple query plans
* Z3Strategy was tweaked to only use a single query plan
* Only 'OR' queries should create multiple query plans now

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
@@ -34,7 +34,7 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with 
   /**
    * Plans the query - strategy implementations need to define this
    */
-  override def getQueryPlans(queryPlanner: QueryPlanner, hints: Hints, output: ExplainerOutputType) = {
+  override def getQueryPlan(queryPlanner: QueryPlanner, hints: Hints, output: ExplainerOutputType) = {
     val sft = queryPlanner.sft
     val acc = queryPlanner.acc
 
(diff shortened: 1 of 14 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
```diff
@@ -126,57 +126,51 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with
...
       // time range for a chunk is 0 to 1 week (in seconds)
-      val timeMax = Weeks.ONE.toStandardSeconds.getSeconds
-      val startQP = qp(head, lt, timeMax, contained = false)
-      val endQP = qp(last, 0, ut, contained = false)
-      val middleQPs = xs.map(w => qp(w, 0, timeMax, contained = true))
-      Seq(startQP, endQP) ++ middleQPs
+      val tMax = Weeks.ONE.toStandardSeconds.getSeconds
+      val head +: middle :+ last = weeks.toList
+      val headRanges = getRanges(Seq(head), (lx, ux), (ly, uy), (lt, tMax))
+      val lastRanges = getRanges(Seq(last), (lx, ux), (ly, uy), (0, ut))
```
jnh5y: If 'last' is empty, do things work out fine?  (I imagine we have tests, etc.)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
```diff
@@ -126,57 +126,51 @@ class Z3IdxStrategy(val filter: QueryFilter) extends Strategy with Logging with
...
-  def queryPlanForPrefix(week: Int,
-                         x: (Double, Double),
-                         y: (Double, Double),
-                         t: (Long, Long),
-                         table: String,
-                         kvsToFeatures: FeatureFunction,
-                         is: Seq[IteratorSetting],
-                         colFamily: Text,
-                         numThreads: Int,
-                         contained: Boolean = true) = {
```
jnh5y: Hmm.... contained was unused?  Interesting...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
elahrvivaz: last will always not be empty due to the length check just above - middle might be empty and we check for that.
scala> val head +: xs :+ last = List(1, 2)
head: Int = 1
xs: List[Int] = List()
last: Int = 2

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
elahrvivaz: yeah, i went back through the entire history and couldn't find it ever getting used... possibly it would be in one of the original wip branches.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z3Iterator.scala
```diff
@@ -84,24 +81,25 @@ class Z3Iterator extends SortedKeyValueIterator[Key, Value] {
 
   override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] = {
     import scala.collection.JavaConversions._
-
-    val zmin = Z3(zlonmin.toInt, zlatmin.toInt, ztmin.toInt).z.toString
-    val zmax = Z3(zlonmax.toInt, zlatmax.toInt, ztmax.toInt).z.toString
-
     val iter = new Z3Iterator
-    iter.init(source, Map(zminKey -> zmin, zmaxKey -> zmax), env)
+    iter.init(source, Map(zKey -> zMap), env)
```
jnh5y: This change makes two versions of the Z3Iterator incompatible, right?

elahrvivaz: yeah, the client jar will have to match the server jar.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/JobUtils.scala
```diff
@@ -112,4 +117,29 @@ object JobUtils extends Logging {
...
+      case qp: JoinPlan => false
+      case _ => true
+    }
+
+    tryPlan.getOrElse {
+      // this query has a join or requires multiple scans - instead, fall back to the record index
+      logger.warn("Desired query plan requires multiple scans - falling back to full table scan")
+      val qps = ds.getQueryPlan(query, Some(StrategyType.RECORD))
+      if (qps.length > 1) {
+        logger.error("The query being executed requires multiple scans, which is not currently " +
```
jnh5y: Does the record table ever provide more than one Query/ScanPlan?

elahrvivaz: Shouldn't... this is kind of emergency fallback.

jnh5y:
Merged as 0983c073cc8c4098118349ea4c8cccd10355c552 on master.

PR #661, opened 2015-08-10
GEOMESA-882 RemoveSchema not deleting empty z3 tables

AndrewAnnex:
* fixed issue where removeschema left empty z3 tables on accumulo
* added a unit test specific to this case

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -357,6 +357,7 @@ class AccumuloDataStore(val connector: Connector,
       val deleter = connector.createBatchDeleter(name, auths, numThreads, defaultBWConfig)
       table.deleteFeaturesForType(sft, deleter)
       deleter.close()
+      if (table == Z3Table && tableOps.exists(name)) tableOps.delete(name)
     }
   }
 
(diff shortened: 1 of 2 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -357,6 +357,7 @@ class AccumuloDataStore(val connector: Connector,
       val deleter = connector.createBatchDeleter(name, auths, numThreads, defaultBWConfig)
       table.deleteFeaturesForType(sft, deleter)
       deleter.close()
+      if (table == Z3Table && tableOps.exists(name)) tableOps.delete(name)
```
elahrvivaz: you could actually put the deleter code in the else statement for this if, as we don't need to bother deleting all the rows if we're dropping the table. Deleting the rows is fairly slow, so it would be good to avoid.

AndrewAnnex: sounds reasonable to me

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -354,9 +354,13 @@ class AccumuloDataStore(val connector: Connector,
 
     GeoMesaTable.getTables(sft).foreach { table =>
       val name = getTableName(sft.getTypeName, table)
-      val deleter = connector.createBatchDeleter(name, auths, numThreads, defaultBWConfig)
-      table.deleteFeaturesForType(sft, deleter)
-      deleter.close()
+      if (table == Z3Table && tableOps.exists(name)) {
```
elahrvivaz: need to move the 'exists' check down to the next line, otherwise we would create a batch deleter against a table that doesn't exist and it would throw an exception.

jnh5y:
Merged as d749747837a7d6c8332b074a472f4a6eb8c6b00c on master.

PR #664, opened 2015-08-10
GEOMESA-884 Fixing reflective calls to accumulo map/reduce configurat…

elahrvivaz:
…ion methods

### DIFF
--- a/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
+++ b/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
@@ -1,15 +1,14 @@
 package org.locationtech.geomesa.jobs.mapred
 
 import org.apache.accumulo.core.client.mapred.AccumuloInputFormat
-import org.apache.accumulo.core.client.security.tokens.PasswordToken
+import org.apache.accumulo.core.client.security.tokens.{AuthenticationToken, PasswordToken}
 import org.apache.accumulo.core.security.Authorizations
 import org.apache.hadoop.mapred.JobConf
-import org.apache.hadoop.mapreduce.Job
 import org.apache.log4j.Level
-import org.locationtech.geomesa.accumulo.AccumuloVersion
-import AccumuloVersion._
+import org.locationtech.geomesa.accumulo.AccumuloVersion._
 
 object InputFormatBaseAdapter {
+
   def setConnectorInfo(job: JobConf, user: String, token: PasswordToken) = accumuloVersion match {
     case V15 => setConnectorInfo15(job, user, token)
     case V16 => setConnectorInfo16(job, user, token)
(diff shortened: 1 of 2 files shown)


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
```diff
@@ -17,12 +16,14 @@ object InputFormatBaseAdapter {
...
-    val method = classOf[AccumuloInputFormat].getMethod("setConnectorInfo", classOf[Job], classOf[String], classOf[PasswordToken])
+    val method = Class.forName("org.apache.accumulo.core.client.mapred.InputFormatBase")
+        .getMethod("setConnectorInfo", classOf[JobConf], classOf[String], classOf[AuthenticationToken])
     method.invoke(null, job, user, token)
   }
 
   def setConnectorInfo16(job: JobConf, user: String, token: PasswordToken) = {
-    val method = classOf[AccumuloInputFormat].getMethod("setConnectorInfo", classOf[Job], classOf[String], classOf[PasswordToken])
+    val method = classOf[AccumuloInputFormat]
+        .getMethod("setConnectorInfo", classOf[JobConf], classOf[String], classOf[AuthenticationToken])
```
anthonyccri: Do both Acc15 and Acc16 have exactly the same signature?

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
```diff
@@ -1,15 +1,14 @@
 package org.locationtech.geomesa.jobs.mapred
 
 import org.apache.accumulo.core.client.mapred.AccumuloInputFormat
-import org.apache.accumulo.core.client.security.tokens.PasswordToken
+import org.apache.accumulo.core.client.security.tokens.{AuthenticationToken, PasswordToken}
```
anthonyccri: Can we add a unit test for this class?  May not be possible especially since it requires a matrix of deps to run against.  I don't really know how to do that.

elahrvivaz: I'll look into it but yeah, i'm not sure exactly how to do it either.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
elahrvivaz: the args and method name are the same, but the base class is different.
I wasn't able to fully test this against accumulo 1.6 - i was able to verify that at least the setConnector and setZoos worked, but I didn't have an accumulo 1.6 to test against.

elahrvivaz: Seemed better to at least get the stuff we know needs fixing in though.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/InputFormatBaseAdapter.scala
anthonyccri: Yeah, ok.  Don't spend too much time.

jnh5y: The dependencies will be handled by the profiles.

jnh5y:
Merged as e98e76a7f2361861b473a59be1aa007e05c64b6c on master.

PR #665, opened 2015-08-11
For comment - Geomesa 875 - Configurable Tables

jahhulbert-ccri:
@elahrvivaz @jnh5y here's what i was thinking but i'm ok with other ways. basically we want to do it on the "datastore" level so i threw it in the metadata as global metadata. we'd need to add a command line tools option later too to fix things up.

might need a tad more testing lemme know what you think


elahrvivaz:
I think it would be nice if we could store it at the schema level, vs the catalog level. This is kind of a new paradigm: catalog-level configuration.
What do you think about storing it as user data in the sft? That way when someone creates a schema, they can set it as user data beforehand, and then we can re-populate it when we load the sft. It would also mean we don't have to modify any interfaces, since we already pass the sft around everywhere.
The caveat is that it wouldn't be serialized in the sft string form, but that is already true of other stuff we put in it.
I guess in this scenario we would need to modify the config sft parser and the command line tools... not sure about anything else.

jahhulbert-ccri:
@elahrvivaz agreed i'm going to redo it per feature originally i thought we wanted per store and i'll try to put it in user data for the sft

PR #670, opened 2015-08-18
GEOMESA-889 Make exception in getRecommendedBitsResolutionForPolygon …

apc3n:
…a class for case matching

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/GeohashUtils.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/GeohashUtils.scala
@@ -17,7 +17,6 @@ import org.locationtech.geomesa.utils.text.WKTUtils
 import scala.collection.BitSet
 import scala.collection.immutable.HashSet
 import scala.collection.immutable.Range.Inclusive
-import scala.collection.mutable.{HashSet => MutableHashSet}
 import scala.util.control.Exception.catching
 
 /**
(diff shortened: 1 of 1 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/GeohashUtils.scala
```diff
@@ -521,12 +520,14 @@ object GeohashUtils
...
 
       if (numBoxes==0)
-        throw new Exception("Could not satisfy constraints, resolutions " + resolutions + ", constraints " + constraints + ".")
+        throw UnsatisfiedConstraintsException(s"Could not satisfy constraints, resolutions $resolutions, constraints $constraints.")
 
       RecommendedResolution(maxBits, numBoxes)
     }
   }
 
+  case class UnsatisfiedConstraintsException(message: String) extends Exception(message)
```
jnh5y: @apc3n @cne1x Would an IllegalArgumentException here be better/fine?

apc3n: @jnh5y We don't actually need the case class for case matching anymore. So we can either get rid of this PR all together or use it as an opportunity to change this to an illegal argument exception. Let me know what you think is best.

anthonyccri: Go with IllegalArgumentException - it's more standard.

elahrvivaz:
4c5e20d615cd9450b2f630143080ea3e4763f502

PR #674, opened 2015-08-21
GEOMESA-895 Add HashAttributeProcess for styling by any attribute

anthonyccri:
* Added a HashAttributeProcess and a HashAttributeColorProcess
* Takes an attribute name from the environment params and hashes modulo
  a configurable param to an integer.  Can be used to select a color
  per attribute
* Added a Point2Point process that creates line strings via a join on
  an attribute specified by the caller

### DIFF
--- a/geomesa-process/src/main/scala/org/locationtech/geomesa/process/HashAttributeProcess.scala
+++ b/geomesa-process/src/main/scala/org/locationtech/geomesa/process/HashAttributeProcess.scala
@@ -1,0 +1,84 @@
+package org.locationtech.geomesa.process
+
+import com.google.common.hash.Hashing
+import org.geotools.data.DataUtilities
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.geotools.feature.collection.DelegateSimpleFeatureIterator
+import org.geotools.feature.simple.{SimpleFeatureBuilder, SimpleFeatureTypeBuilder}
+import org.geotools.process.ProcessException
+import org.geotools.process.factory.{DescribeParameter, DescribeProcess, DescribeResult}
+import org.geotools.process.vector.VectorProcess
+
+trait HashAttribute {
+  import org.locationtech.geomesa.utils.geotools.Conversions._
+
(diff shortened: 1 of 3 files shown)


geomesa-process/src/main/scala/org/locationtech/geomesa/process/HashAttributeProcess.scala
```diff
@@ -0,0 +1,84 @@
...
+    sftBuilder.init(sft)
+    augmentSft(sftBuilder)
+    val targetSft = sftBuilder.buildFeatureType()
+    val featureBuilder = new SimpleFeatureBuilder(targetSft)
+
+    val results =
+      obsFeatures.features().map { sf =>
+        featureBuilder.reset()
+        featureBuilder.init(sf)
+        val attr = Option(sf.getAttribute(attribute).asInstanceOf[String]).getOrElse("")
```
elahrvivaz: Maybe change to Option(sf.getAttribute(attribute)).map(_.toString).getOrElse("") so that it handles any type?
Could also look up by index instead of name to speed up a little.

geomesa-process/src/main/scala/org/locationtech/geomesa/process/HashAttributeProcess.scala
```diff
@@ -0,0 +1,84 @@
...
+        val hash = math.abs(hashFn.hashString(attr).asInt()) % modulo
+        featureBuilder.set("hash", transformHash(hash))
```
elahrvivaz: Might be good to have a default 'augmentSft' that adds the hash attribute in, since we expect it here.

elahrvivaz: oh, i see you use different types for the attribute - maybe just a comment somewhere that 'hash' is expected.

elahrvivaz:
making two tweaks and merging.

elahrvivaz:
merged as 27449451253ad6781ee8e114628a80fc4c234273

PR #675, opened 2015-08-24
GEOMESA-896 Exception thrown in Point2Point process when a linestring…

anthonyccri:
… is created by multiple points in the exact same location

* If a grouped set of points are all in the exact same location, then an
  exception was being thrown.
* Allow the ability to filter out these singular points.

### DIFF
--- a/geomesa-process/src/main/scala/org/locationtech/geomesa/process/Point2PointProcess.scala
+++ b/geomesa-process/src/main/scala/org/locationtech/geomesa/process/Point2PointProcess.scala
@@ -40,7 +40,10 @@ class Point2PointProcess extends VectorProcess {
                minPoints: Int,
 
                @DescribeParameter(name = "breakOnDay", description = "Break connections on day marks")
-               breakOnDay: Boolean
+               breakOnDay: Boolean,
+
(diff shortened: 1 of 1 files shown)


geomesa-process/src/main/scala/org/locationtech/geomesa/process/Point2PointProcess.scala
```diff
@@ -40,7 +40,10 @@ class Point2PointProcess extends VectorProcess {
                minPoints: Int,
 
                @DescribeParameter(name = "breakOnDay", description = "Break connections on day marks")
-               breakOnDay: Boolean
+               breakOnDay: Boolean,
+
+               @DescribeParameter(name = "filterSingularPoints", description = "Filter on groups that all fall on the same point", defaultValue = "true")
```
jnh5y: Is there any situation where we do not want this filtering?  

The ticket implies that without the filtering, exceptions will be thrown.  If that's the case, we don't need to make it configurable...

anthonyccri: It's a geomesa bug - when you compose this with Import, geomesa geohashing throws an exception on a line with no distance.

jahhulbert-ccri:
f7bedc32a3b040b175b20dde0202a074531c2a6b

PR #677, opened 2015-08-27
GEOMESA-900 Adding xml simple feature parser

elahrvivaz:

### DIFF
--- a/geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
+++ b/geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
@@ -1,0 +1,83 @@
+package org.locationtech.geomesa.convert.xml
+
+import java.io.ByteArrayInputStream
+import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.xpath.{XPathConstants, XPathExpression, XPathFactory}
+
+import com.google.common.base.Charsets
+import com.typesafe.config.Config
+import org.locationtech.geomesa.convert.Transformers.{EvaluationContext, Expr}
+import org.locationtech.geomesa.convert._
+import org.opengis.feature.simple.SimpleFeatureType
+import org.w3c.dom.NodeList
+
+import scala.collection.JavaConversions._
+
+class XMLConverter(val targetSFT: SimpleFeatureType,
+                   val idBuilder: Expr,
+                   val featurePath: Option[XPathExpression],
+                   val inputFields: IndexedSeq[Field]) extends ToSimpleFeatureConverter[String] {
+
+  private val docBuilder = {
+    val factory = DocumentBuilderFactory.newInstance()
+    factory.setNamespaceAware(false)
+    factory.newDocumentBuilder()
+  }
+
+  override def fromInputType(i: String): Seq[Array[Any]] = {
+    // parse the document once, then extract each feature node and operate on it
+    val root = docBuilder.parse(new ByteArrayInputStream(i.getBytes(Charsets.UTF_8))).getDocumentElement
+    featurePath.map { path =>
+      val nodeList = path.evaluate(root, XPathConstants.NODESET).asInstanceOf[NodeList]
+      (0 until nodeList.getLength).map(i => Array[Any](nodeList.item(i)))
+    }.getOrElse(Seq(Array[Any](root)))
+  }
+}
+
+class XMLConverterFactory extends SimpleFeatureConverterFactory[String] {
+
+  private val xpath = XPathFactory.newInstance().newXPath()
+
+  override def canProcess(conf: Config): Boolean = canProcessType(conf, "xml")
+
+  override def buildConverter(sft: SimpleFeatureType, conf: Config): XMLConverter = {
+    val fields    = buildFields(conf.getConfigList("fields"))
+    val idBuilder = buildIdBuilder(conf.getString("id-field"))
+    // feature path can be any xpath that resolves to a node set (or a single node)
+    // it can be absolute, or relative to the root node
+    val featurePath  = if (conf.hasPath("feature-path")) Some(conf.getString("feature-path")) else None
+    new XMLConverter(sft, idBuilder, featurePath.map(xpath.compile), fields)
+  }
+
+  override def buildFields(fields: Seq[Config]): IndexedSeq[Field] = {
+    fields.map { f =>
+      val name = f.getString("name")
+      val transform = if (f.hasPath("transform")) {
+        Transformers.parseTransform(f.getString("transform"))
+      } else {
+        null
+      }
+      if (f.hasPath("path")) {
+        // path can be absolute, or relative to the feature node
+        // it can also include xpath functions to manipulate the result
+        XMLField(name, xpath.compile(f.getString("path")), transform)
+      } else {
+        SimpleField(name, transform)
+      }
+    }.toIndexedSeq
+  }
+}
+
+case class XMLField(name: String, expression: XPathExpression, transform: Expr) extends Field {
+
+  private val mutableArray = Array.ofDim[Any](1)
+
+  override def eval(args: Array[Any])(implicit ec: EvaluationContext): Any = {
+    mutableArray(0) = expression.evaluate(args(0))
+    if (transform == null) {
+      mutableArray(0)
+    } else {
+      super.eval(mutableArray)
+    }
+  }
+}
--- a/geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
+++ b/geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
@@ -1,21 +1,25 @@
 package org.locationtech.geomesa.convert.xml
 
-import java.io.ByteArrayInputStream
+import java.io.StringReader
+import javax.xml.XMLConstants.W3C_XML_SCHEMA_NS_URI
 import javax.xml.parsers.DocumentBuilderFactory
+import javax.xml.transform.stream.StreamSource
+import javax.xml.validation.SchemaFactory
 import javax.xml.xpath.{XPathConstants, XPathExpression, XPathFactory}
 
-import com.google.common.base.Charsets
 import com.typesafe.config.Config
 import org.locationtech.geomesa.convert.Transformers.{EvaluationContext, Expr}
 import org.locationtech.geomesa.convert._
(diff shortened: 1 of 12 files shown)


geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  private val docBuilder = {
+    val factory = DocumentBuilderFactory.newInstance()
+    factory.setNamespaceAware(false)
+    factory.newDocumentBuilder()
+  }
+
+  override def fromInputType(i: String): Seq[Array[Any]] = {
+    // parse the document once, then extract each feature node and operate on it
+    val root = docBuilder.parse(new ByteArrayInputStream(i.getBytes(Charsets.UTF_8))).getDocumentElement
```
BrittEvans: Instead of doing the byte array, how about:

new InputSource(new StringReader(i))

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
```diff
@@ -0,0 +1,83 @@
...
+    featurePath.map { path =>
+      val nodeList = path.evaluate(root, XPathConstants.NODESET).asInstanceOf[NodeList]
+      (0 until nodeList.getLength).map(i => Array[Any](nodeList.item(i)))
```
BrittEvans: Try to include the original string somewhere so we can md5 it for the feature ID.  Or some other solution that prevents us from using the uuid() for the featureId

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XMLConverterTest.scala
```diff
@@ -0,0 +1,171 @@
...
+          | converter = {
+          |   type         = "xml"
+          |   id-field     = "uuid()"
+          |   feature-path = "Feature" // can be any xpath - relative to the root, or absolute
+          |   fields = [
+          |     // paths can be any xpath - relative to the feature-path, or absolute
+          |     { name = "number", path = "number",           transform = "$0::integer" }
+          |     { name = "color",  path = "color",            transform = "trim($0)" }
+          |     { name = "weight", path = "physical/@weight", transform = "trim($0)" }
+          |     { name = "source", path = "/doc/DataSource/name/text()" }
```
BrittEvans: Why does this need text()?

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XMLConverterTest.scala
```diff
@@ -0,0 +1,171 @@
...
+        """
+          | converter = {
+          |   type         = "xml"
+          |   id-field     = "uuid()"
+          |   feature-path = "Feature" // can be any xpath - relative to the root, or absolute
+          |   fields = [
+          |     // paths can be any xpath - relative to the feature-path, or absolute
+          |     { name = "number", path = "number",                   transform = "$0::integer" }
+          |     { name = "color",  path = "color",                    transform = "trim($0)" }
+          |     { name = "weight", path = "floor(physical/@weight)", transform = "trim($0)" }
```
BrittEvans: floor() turns it into a float then turns it back into a string?

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
```diff
@@ -0,0 +1,83 @@
...
+import scala.collection.JavaConversions._
+
+class XMLConverter(val targetSFT: SimpleFeatureType,
+                   val idBuilder: Expr,
+                   val featurePath: Option[XPathExpression],
+                   val inputFields: IndexedSeq[Field]) extends ToSimpleFeatureConverter[String] {
+
+  private val docBuilder = {
+    val factory = DocumentBuilderFactory.newInstance()
+    factory.setNamespaceAware(false)
```
BrittEvans: Would there be any benefit to using an xsd here if we have one?

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
anthonyccri: Are we md5'ing the root node?  I.e. in pseudo code

```scala
val xml = """
   <Doc>
      <Feature>
          <..../>
      </Feature>
      <Feature>
           <..../>
      </Feature>
<Doc>"""

for f in xpath(xml, "/Doc/Feature") {
   md5(f.toString)
}
```

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XMLConverterTest.scala
anthonyccri: Xpath provides a bunch of transformation functions . . . i think this is just demonstrating that they can be used in a path expression.  But it's a good question re: the types . . . we should avoid going back and forth between types to the extent possible for performance.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
anthonyccri: For validation?  I guess it wouldn't be too bad.  Maybe a variable indicating the path to the xsd and a flag saying 'perform-validation'?  I don't think it matters for performance.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
elahrvivaz: Is there a benefit one way or the other? I figured getting the bytes in UTF-8 would be safe, but I'm not entirely sure...

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
elahrvivaz: Yeah, we'd want to hash each feature node, right?

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XMLConverterTest.scala
elahrvivaz: It doesn't - I left it in there just b/c that's how a lot of xpath examples are. I think technically it is more correct, in that it selects the text of the node vs the node itself. I was just trying to provide some different examples of xpath expressions.

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XMLConverterTest.scala
elahrvivaz: yeah, this was just an example of an xpath function. I'll change the type to double.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
elahrvivaz: it's possible we would need it to be able to parse some documents... I'll try to add one in. xsd's can be tricky though - a lot of them link to other xsds, which only works if they are actual files sitting on disk, and not loaded through the classpath (which we would probably need to do for deploying in storm). We might need to revisit this later if we can't get some real-world files to work.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
BrittEvans: I thought it might be faster without creating the byte array, but I don't know if it matters.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
BrittEvans: Yeah, each feature node.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
BrittEvans: Yeah, for validation, just a string in the conf file for an xsd path to turn it on/off.

BrittEvans:
I have a large slug of real data to try this against, maybe wait till we run that through before we merge this in case it needs some tweaks.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
elahrvivaz: I added it as a separate step to validate the document if an xsd is present. I tried adding it to the parser itself, but then the xpaths have to have namespaces registered with them, and it becomes kind of a nightmare... hopefully this will be good enough.

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
elahrvivaz: Changed to input source/string reader

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XMLConverter.scala
BrittEvans: works for me.

elahrvivaz:
All comments so far have been addressed.

anthonyccri:
Merged

PR #679, opened 2015-08-28
GEOMESA-898 Configure Auths on Spark RDD

jahhulbert-ccri:
Configure auths from DS on Spark RDD so that underlying inputformat sets them.

Verified fix works in production with spark 1.5.0-SNAPSHOT and Accumulo 1.6.1


elahrvivaz:
did you decide not to use the GeoMesaInputFormat.configure method?

jahhulbert-ccri:
For now I just added the lines to get it in sync with what we are running currently . The other methods for GeoMesaInputFormat.configure took a JobConf object instead of a Configuration object which I guess may be OK. I can do a bigger refactor for that next and can make another ticket?

anthonyccri:
Sounds reasonable to me.  I'm going to merge.

anthonyccri:
Merged

PR #680, opened 2015-08-28
GEOMESA-903 Reducing polling time in GenericSimpleFeatureStreamSource

elahrvivaz:


geomesa-stream/geomesa-stream-generic/src/main/scala/org/locationtech/geomesa/stream/generic/GenericSimpleFeatureStreamSourceFactory.scala
```diff
@@ -73,7 +73,7 @@ class GenericSimpleFeatureStreamSource(val ctx: CamelContext,
     from(sourceRoute).process { e => inQ.put(e.getIn.getBody.asInstanceOf[String]) }
   }
 
-  override def next: SimpleFeature = outQ.poll(500, TimeUnit.MILLISECONDS)
+  override def next: SimpleFeature = outQ.poll(1, TimeUnit.MILLISECONDS)
```
jnh5y: I wonder if this should be configurable.

elahrvivaz:
I'm going to close this - I think it's a good idea (if a client wants to wait between polling, they can do it explicitly) but I'm not sure if it will affect other projects that rely on the current behavior.

PR #683, opened 2015-09-13
GEOMESA-904 Return individual P2P segments

jahhulbert-ccri:
* Return individual 2-coordinate line segments instead of one long line segment


anthonyccri:
Merged

PR #684, opened 2015-09-14
GEOMESA-911 Disjoint attribute index queries do not work with bin format

anthonyccri:
* Merge disjoint attribute queries with the same secondary filter into
  a single scan plan


anthonyccri:
cc @jnh5y

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
```diff
@@ -217,6 +217,13 @@ class QueryFilterSplitter(sft: SimpleFeatureType) extends Logging {
       // this is a full table scan, we can just append the OR to the secondary filter
       val secondary = orOption(mergeTo.secondary.toSeq ++ andOption(toMerge.primary ++ toMerge.secondary.toSeq))
       mergeTo.copy(secondary = secondary)
+    } else if(toMerge.strategy == StrategyType.ATTRIBUTE && mergeTo.strategy == StrategyType.ATTRIBUTE) {
+      // if we have disjoint attribute queries with the same secondary filter, merge into a multi-range query
```
elahrvivaz: also need to check that the primary filter is on the same attribute.

elahrvivaz:
would be nice to have a new unit test :)

elahrvivaz:
merged as 57c716a9bbe87fd4864dd17098edd40f1b533ce1

PR #685, opened 2015-09-15
GEOMESA-912 Web service for running Spark SQL jobs

elahrvivaz:
* Spark updated to version 1.5


elahrvivaz:
I'll leave this open for comments, but I need to do some more work to avoid having spark as a compile time dependency due to IP restrictions.

elahrvivaz:
Updated to make spark a provided dependency.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/sql/GeoMesaSparkSql.scala
```diff
@@ -0,0 +1,328 @@
...
+object GeoMesaSparkSql extends Logging {
+
+  private val ff = CommonFactoryFinder.getFilterFactory2
+
+  // state to keep track of our sfts and data store connection parameters
+  private val sftsByName = scala.collection.mutable.Map.empty[String, SimpleFeatureType]
+  private val paramsBySft = scala.collection.mutable.Map.empty[SimpleFeatureType, Map[String, String]]
+
+  // singleton spark context
+  private var sc: SparkContext = null
```
jahhulbert-ccri: question...if I am running a spark shell where I already have my SparkContext can I still make use of GeoMesaSparkSql? Perhaps having an apply() method that takes the SC in would work?

elahrvivaz: I'll split it out into an object to manage the spark context (for the rest api), and a class that you can use directly with any spark context.

jahhulbert-ccri: awesome! thanks

elahrvivaz:
@jahhulbert-ccri let me know if it works for you now.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/sql/GeoMesaSparkSql.scala
```diff
@@ -0,0 +1,345 @@
...
+    val sftsWithParams = typeNames.map(sftsByName.apply)
+
+    // extract the cql from the where clause
+    val where = parsedSql.getWhereClause
+    val cql = GeoMesaSparkSql.extractCql(where, dataContext, typeNames)
+    // clear out the cql from the where clause so spark doesn't try to parse it
+    // if it' a sql expression, the expression field will be null
+    // otherwise it has the raw expression, which we assume is cql
+    where.getItems.filter(_.getExpression != null).foreach(where.removeItem)
+    val sqlWithoutCql = parsedSql.toSql
```
anthonyccri: Say that three times fast.

geomesa-web/geomesa-web-data/src/main/scala/org/locationtech/geomesa/web/analytics/AnalyticEndpoint.scala
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  before() {
+    contentType = formats(format)
+  }
+
+  /**
+   * Registers a data store, making it available for spark sql queries
+   */
+  post("/ds/:alias") {
```
anthonyccri: Instead of doing this, should we have the bean take the Geoserver catalog as a constructor arg and process it for GeoMesa data stores?

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/sql/GeoMesaSparkSql.scala
```diff
@@ -0,0 +1,345 @@
...
+  def query(sql: String, splits: Option[Int]): DataFrame = {
+    val parsedSql = dataContext.parseQuery(sql)
+
+    // extract the feature types from the from clause
+    val typeNames = parsedSql.getFromClause.getItems.map(_.getTable.getName)
+    val sftsWithParams = typeNames.map(sftsByName.apply)
+
+    // extract the cql from the where clause
+    val where = parsedSql.getWhereClause
+    val cql = GeoMesaSparkSql.extractCql(where, dataContext, typeNames)
```
anthonyccri: How do we pass in a geometry clause?  Particularly a bbox?

elahrvivaz: There are some sample requests in the readme, but basically you just add it to the where clause like so:
select mySft.myAttr, count(*) as count from mySft where bbox(mySft.geom, -115, 45, -110, 50) AND mySft.dtg during 2015-03-02T10:00:00.000Z/2015-03-02T11:00:00.000Z group by myattr
When we parse out the SQL, if it's not recognized as a normal SQL statement, it doesn't get parsed into an object - so we detect those portions and treat them as CQL.  I'm sure it's not foolproof, but it works for basic cases. In particular, attribute queries will probably end up run by the SQL engine instead of the initial geomesa query.

geomesa-web/geomesa-web-data/src/main/scala/org/locationtech/geomesa/web/analytics/AnalyticEndpoint.scala
elahrvivaz: Yeah, that's a good idea... I was trying to keep geoserver dependencies out of the code, since we have IP issues with it - what do you think?

anthonyccri:
Merged

PR #687, opened 2015-09-22
GEOMESA-915 WMS queries that have viewport bounds and a bbox in the C…

anthonyccri:
…QL clause return viewport results

* Added a tryMergeGeoms helper method to intersect geometries on certain conditions

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
@@ -364,7 +364,7 @@ class FilterHelperTest extends Specification with Mockito with Logging {
       val two = gf.createPolygon(Seq[Coordinate]((1, 1), (2, 1), (2, 2), (1, 2), (1, 1)).toArray)
       val f1 = ff.bbox(ff.property("geom"), ReferencedEnvelope.reference(one.getEnvelopeInternal))
       val f2 = ff.bbox(ff.property("geom"), ReferencedEnvelope.reference(two.getEnvelopeInternal))
-      FilterHelper.tryMergeGeometryFilters(Seq(f1, f2)).head must be equalTo(f2)
+      FilterHelper.tryReduceGeometryFilter(Seq(f1, f2)).head must be equalTo(f2)
     }
   }
 }
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
@@ -9,8 +9,10 @@
 package org.locationtech.geomesa.accumulo.index
 
 import com.typesafe.scalalogging.slf4j.Logging
+import com.vividsolutions.jts.geom.Coordinate
 import org.geotools.factory.CommonFactoryFinder
 import org.geotools.filter.text.ecql.ECQL
+import org.geotools.geometry.jts.{JTSFactoryFinder, ReferencedEnvelope}
 import org.joda.time.{DateTime, DateTimeZone, Interval}
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.accumulo.filter.TestFilters._
@@ -27,6 +29,7 @@ import scala.util.Random
 @RunWith(classOf[JUnitRunner])
 class FilterHelperTest extends Specification with Mockito with Logging {
   val ff = CommonFactoryFinder.getFilterFactory2
+  val gf = JTSFactoryFinder.getGeometryFactory
 
   val min = IndexSchema.minDateTime
   val max = IndexSchema.maxDateTime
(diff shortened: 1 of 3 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -318,5 +320,27 @@ object FilterHelper {
       case f: Filter => Seq(f)
     }
   }
+
+  def tryMergeGeometryFilters(filts: Seq[Filter]): Seq[Filter] = {
```
cne1x: [picky] I would avoid the term "Merge" in this function name, because "merge" suggests "union" more than "intersection" (which is consistent with perusing synonyms).  Perhaps ```tryToReduceGeometryFilters```?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -318,5 +320,27 @@ object FilterHelper {
...
+    import org.geotools.data.DataUtilities._
+    import scala.collection.JavaConversions._
+    val filtFactory = CommonFactoryFinder.getFilterFactory2
+
+    def getAttrName(l: BBOX): String = propertyNames(l).head.getPropertyName
+
+    filts match {
+      // if we have two bbox filters as is common in WMS queries, merge them by intersecting the bounds
+      case Seq(l: BBOX, r: BBOX) if getAttrName(l) == getAttrName(r) =>
```
cne1x: It feels as if it might be useful to extend this to work with longer sequences, even if our initial case is known to have only two elements.  Thoughts?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
```diff
@@ -353,4 +356,15 @@ class FilterHelperTest extends Specification with Mockito with Logging {
       result.asInstanceOf[KnownCost].cost mustEqual 5
     }
   }
+
+  "tryMergeGeoms" should {
+    "intersect BBOX geoms" >> {
```
cne1x: It might be nice to have an additional test that (intentionally) is unable to find an intersection.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
```diff
@@ -353,4 +356,15 @@ class FilterHelperTest extends Specification with Mockito with Logging {
...
+      implicit def tupleToCoord(t: (Int, Int)): Coordinate = new Coordinate(t._1, t._2)
+      val one = gf.createPolygon(Seq[Coordinate]((1, 1), (4, 1), (4, 4), (1, 4), (1, 1)).toArray)
```
jnh5y: Which order (clockwise or counterclockwise) indicates that the defined polygon is on the 'inside'?  

I feel like most times polygons are defined in clockwise order.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
jnh5y: Additionally, there should be a test cases where 1) the intersection is not just one of the boxes, and 2) involving the IDL and a view port spanning the IDL.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
jnh5y: +1.  I'd suggest factoring out this as a function and performing a reduce.

jnh5y: As a Bigger Note, this does not cover situations where someone has drawn a polygon and Intersects is used.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
cne1x: Oh, you mean the antimeridian...  hashtag, "pedantic4ever"

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
anthonyccri: I always define polygons lower-left and counter-clockwise.  Does that matter to the definition of p1 being inside p2?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
anthonyccri: When the viewport spans the IDL and the request has a CQL filter, what happens? Does geoserver produce two Or queries?  i.e. (One side of IDL AND CQL) OR (Other side of IDL AND CQL)?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
anthonyccri: If Intersects is used, will that result in a filter.primary having a Seq(BBOX, BBOX)?  This is intentionally limited.  I feel like there's gotta be code down in geotools that solves this for us.  However, I looked at the FilterToSQL and it looks like it punts to the database.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/FilterHelperTest.scala
anthonyccri: Looks like IDL is handled upstream of this case match.  I am unable to produce an IDL query (viewport spanning the IDL or poly spanning the IDL) that doesn't get handled upstream.

cne1x:
Merged as commit 98a8f6b95c269621ae3d59582d7bfb9d5d95ede8

PR #689, opened 2015-09-28
GEOMESA-917 Reducing deduplication

elahrvivaz:
* OR queries are split into disjoint form, so that we don't have to deduplicate them
* Filter manipulation (handling dwithin units, IDL bounds, etc) is done up front instead of inside the different strategies


anthonyccri:
Can you add a test for a query that is

bbox(whole world) AND dtg during AND (attr = 'foo' OR attr = 'bar' OR attr = 'baz')

where you have an attribute index on 'attr' of high cardinality and ensure that we are not trying to dedupe it?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
```diff
@@ -193,7 +205,25 @@ class QueryFilterSplitter(sft: SimpleFeatureType) extends Logging {
       }
     }
 
-    mergeOverlappedFilters(combineSecondaryFilters(reduceChildOptions(getChildOptions)))
+    // make our queries disjoint ORs - this way we don't have to deduplicate results
+    def makeDisjoint(options: Seq[FilterPlan]): Seq[FilterPlan] = options.map { filterPlan =>
+      if (filterPlan.filters.length < 2) {
+        filterPlan
+      } else {
+        // A OR B OR C becomes... A OR (B NOT A) OR (C NOT A and NOT B)
```
anthonyccri: Does this have a performance impact?

anthonyccri:
Merged.  9b9a290..dd5dcc9

PR #693, opened 2015-10-13
GEOMESA-920 JSON Converter should return null for missing leaf nodes

BrittEvans:
Signed-off-by: Britt <britt@ccri.com>


anthonyccri:
Merged 44d7fe7..9b9a290

PR #702, opened 2015-11-02
GEOMESA-939 Fix Exponential Query Planning Time for High Cardinality …

jahhulbert-ccri:
…OR attribute queries

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
@@ -229,6 +229,11 @@ class QueryFilterSplitter(sft: SimpleFeatureType) extends Logging {
     // where the attr is of high cardinality and is indexed
     // These require special handling to avoid a bug with exponential
     // query planning time.
+    //
+    // This query logic is based on the output of the DNF rewrite
+    //
+    // TODO this should really be evaluated before the DNF logic occurs
+    // and should be handled as part of GEOMESA-941
     def detectSingleAttrOr: Boolean = {
       filter match {
         case or: Or =>
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
@@ -141,6 +141,7 @@ class QueryFilterSplitter(sft: SimpleFeatureType) extends Logging {
 
     // combine the filter plans so that each plan has multiple query filters
     // use the permutations of the different options for each child
+    // TODO GEOMESA-941 Fix algorithmically dangerous (2^N exponential runtime)
     def reduceChildOptions(childOptions: Seq[Seq[FilterPlan]]): Seq[FilterPlan] =
       childOptions.reduce { (left, right) =>
         left.flatMap(l => right.map(r => FilterPlan(l.filters ++ r.filters)))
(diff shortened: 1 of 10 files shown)


elahrvivaz:
looks good to me

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
```diff
@@ -223,7 +224,45 @@ class QueryFilterSplitter(sft: SimpleFeatureType) extends Logging {
       }
     }
 
-    makeDisjoint(mergeOverlappedFilters(combineSecondaryFilters(reduceChildOptions(getChildOptions))))
+    // detect the pattern it the pattern (attr in (1,2,3,4,5,6, etc) AND something else)
```
anthonyccri: [nitpick] comment doesn't make sense

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
```diff
@@ -223,7 +224,45 @@ class QueryFilterSplitter(sft: SimpleFeatureType) extends Logging {
...
+    def detectSingleAttrOr: Boolean = {
+      // each child is expressed as (attr = x(i) AND expr2 AND expr3 ...)
+      val attrs =
+        filter.getChildren.map {
+          case and: And => and.getChildren.flatMap {
```
anthonyccri: Aren't we dealing with ```OR```s here, not ```AND```s.  Shouldn't it be something like:

```sql
bbox(geom, ...) AND dtg between .... AND (attr = '1' OR attr = '2' ....)
```

jahhulbert-ccri: We are...but the ```logicDistributionDNF``` method transforms it out into something like this:
```
[
  [
     [ high = h0 ]
     AND [ geom bbox POLYGON ((40 40, 40 50, 50 50, 50 40, 40 40)) ] 
     AND org.geotools.filter.temporal.DuringImpl@e5a72d93
  ] 
  OR 
  [
    [ high = h1 ]
    AND [ geom bbox POLYGON ((40 40, 40 50, 50 50, 50 40, 40 40)) ]
    AND org.geotools.filter.temporal.DuringImpl@e5a72d93
  ]
  OR 
  [
    [ high = h2 ]
```

So really the  match statement is referring to the "children" of the OR being an AND: 
```
[[ high = h0 ] AND [ geom bbox POLYGON ((40 40, 40 50, 50 50, 50 40, 40 40)) ] AND org.geotools.filter.temporal.DuringImpl@e5a72d93]
```

That being said, you bring up a good point that we should verify it is a top level OR filter whose children are  AND predicates.

jahhulbert-ccri: also note that i think we may want to redesign some of the code anyways to make it easier to follow/inject new optimizations :)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryFilterSplitter.scala
jahhulbert-ccri: added better comment

jahhulbert-ccri:
@anthonyccri @elahrvivaz added an update...lemme know if there's anything else...it's a hairy piece area of the codebase

elahrvivaz:
yeah, trying to figure out OR combinations was tricky - if there's a better way, I'm all for refactoring this code.

jahhulbert-ccri:
Yeah, my gut feels there are some better ways, but they will take much more thought and engineering...that's what I want GEOMESA-941 to be...this was the quick fix for operational purposes :) We'll need to think more about a clean way to do it since its difficult.

anthonyccri:
e976c7c..71aec6d

PR #703, opened 2015-11-02
GEOMESA-922 Adding table prefix to record table splits

jahhulbert-ccri:


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -141,6 +142,28 @@ class AccumuloDataStoreTest extends Specification with AccumuloDataStoreDefaults
...
     }
 
+    "create a schema with custom record splitting options with talbe sharing on" in {
+      val spec = "name:String,dtg:Date,*geom:Point:srid=4326;table.splitter.class=" +
+        s"${classOf[DigitSplitter].getName},table.splitter.options='fmt:%02d,min:0,max:99'"
+      val sft = SimpleFeatureTypes.createType("customsplit2", spec)
+      sft.setTableSharing(true)
+
+      import scala.collection.JavaConversions._
+      val prevsplits = ImmutableSet.copyOf(ds.connector.tableOperations().listSplits("AccumuloDataStoreTest_records").toIterable)
```
jahhulbert-ccri: anybody know how to make it chose the correct copyOf method?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -291,8 +291,14 @@ class AccumuloDataStore(val connector: Connector,
           sft.setSchemaVersion(CURRENT_SCHEMA_VERSION)
           val spatioTemporalSchema = computeSpatioTemporalSchema(sft)
           checkSchemaRequirements(sft, spatioTemporalSchema)
-          createTablesForType(sft)
           writeMetadata(sft, SerializationType.KRYO, spatioTemporalSchema)
+
+          // reload the SFT...but we need to copy over the splits!!!
+          val reloadedSft = getSchema(sft.getTypeName)
+          Seq(SimpleFeatureTypes.TABLE_SPLITTER, SimpleFeatureTypes.TABLE_SPLITTER_OPTIONS)
+            .foreach(k => Option(sft.getUserData.get(k)).foreach(v => reloadedSft.getUserData.put(k, v)))
```
elahrvivaz: maybe we should just copy any key from the original sft that isn't in the reloaded sft - that way we won't have to change this code if we add other config options.

elahrvivaz:
I noticed in the STIDX config, we just use 40 splits - we actually have the st idx schema in the sft after this change - maybe you could fix that too? Doesn't seem worth a separate ticket...
https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala#L95

jahhulbert-ccri:
@elahrvivaz yeah i'll change that too in this one...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
jahhulbert-ccri: sounds reasonable.

anthonyccri:
Is this for when we have ```table.sharing``` turned on?

jahhulbert-ccri:
@anthonyccri it certainly happens when table.sharing is set to true...I believe it does not when table sharing is false (it sticks with the regular hex splits)...but I'll verify 

https://github.com/jahhulbert-ccri/geomesa/blob/fcr_GEOMESA-922/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala#L132

I think I tested that there. but let me create some real tables on pcloud to make sure.

jahhulbert-ccri:
@anthonyccri yeah I verified that the splits include the prefix when the table is shared but do NOT include the prefix when the splits are not shared. Likewise, I verified ingested records in the record table follow the same pattern.

elahrvivaz:
b159be0d97782df60b7c29cdf1b098e987976c4f

PR #705, opened 2015-11-05
GEOMESA-854 Ingesting a Shapefile should not require a SimpleFeatureType specification

jdk2pq:
Made the "--spec" argument optional, but the tools will throw a JCommander ParameterException if the ingest is ingesting a delimited type and the "--spec" is null.

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
@@ -9,7 +9,7 @@ package org.locationtech.geomesa.tools.commands
 
 import java.util
 
-import com.beust.jcommander.{JCommander, Parameter, ParameterException, Parameters}
+import com.beust.jcommander.{JCommander, Parameter, Parameters}
 import com.typesafe.scalalogging.slf4j.Logging
 import org.locationtech.geomesa.tools.DataStoreHelper
 import org.locationtech.geomesa.tools.Utils.Formats._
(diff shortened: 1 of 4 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
```diff
@@ -26,7 +26,11 @@ class IngestCommand(parent: JCommander) extends Command(parent) with Logging {
   override def execute(): Unit = {
     val fmt = Option(params.format).getOrElse(getFileExtension(params.files(0)))
     fmt match {
-      case CSV | TSV => new DelimitedIngest(params).run()
+      case CSV | TSV =>
+        if (Option(params.spec).isEmpty) {
```
elahrvivaz: Not necessary to create an option here if you're just checking it for empty: perfectly fine to do <code>if (params.spec == null)</code>

jahhulbert-ccri:
Does the create command also need a null check? transitive call to https://github.com/locationtech/geomesa/blob/98a8f6b95c269621ae3d59582d7bfb9d5d95ede8/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/FeatureCreator.scala#L27

jdk2pq:
Probably good to have, yes. I'll add that in.

jahhulbert-ccri:
looks good. I assume the the only time we need the spec is when creating a feature so that's a good place to put it.

jahhulbert-ccri:
verified that create feature needs --spec and shapefile ingest works

committed as 23e24e4578c7123300baafad6bafbe000cf5bc76

PR #706, opened 2015-11-09
Blobstore WIP

jnh5y:


geomesa-blobstore/geomesa-blobstore-api/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-blobstore-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-web-core</artifactId>
+        </dependency>
+
+        <!-- provided dependencies -->
+        <dependency>
```
anthonyccri: [future] we should be able to abstract over the underlying store

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.scalatra.{NotFound, BadRequest, Ok}
+
+import scala.collection.JavaConversions._
+
+class BlobstoreServlet extends GeoMesaScalatraServlet {
+  override def root: String = "blob"
+
+  var abs: AccumuloBlobStore = null
+
+  post("/ds/:alias") {
```
anthonyccri: Not sure why we need a post here.  Are we registering new blobstores at runtime via this method?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -0,0 +1,51 @@
...
+    println(s"In ID method, trying to retrieve id $id")
+
+    if (abs == null) {
+      NotFound(reason = "AccumuloBlobStore is not initialized.")
+    } else {
+      val (returnBytes, filename) = abs.get(id)
+      if (returnBytes == null) {
+        NotFound(reason = s"Unknown ID $id")
+      } else {
+        contentType = "application/octet-stream"
```
anthonyccri: Should this be the mime type of the actual file?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.common.io.Files
+import org.locationtech.geomesa.accumulo.data.{AccumuloDataStore, AccumuloDataStoreFactory}
+import org.locationtech.geomesa.blob.core.AccumuloBlobStore
+import org.locationtech.geomesa.web.core.GeoMesaScalatraServlet
+import org.scalatra.{NotFound, BadRequest, Ok}
+
+import scala.collection.JavaConversions._
+
+class BlobstoreServlet extends GeoMesaScalatraServlet {
```
anthonyccri: I'd recommend adding a post method for putting data into the system.  Use the Content-type HTTP header to set the mime-type.  Probably avoid the ```canProcess``` method in that case so the API may have to change to accommodate scenarios in which the user specifies the mime-type.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/handlers/BlobStoreFileHandler.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.locationtech.geomesa.blob.core.AccumuloBlobStore._
+import org.opengis.feature.simple.SimpleFeature
+
+import scala.collection.JavaConversions._
+
+object BlobStoreFileHandler {
+  def buildSF(file: File, params: Map[String, String]): Option[SimpleFeature] = {
+    val handlers = ServiceRegistry.lookupProviders(classOf[FileHandler])
+
+    handlers.find(_.canProcess(file, params)).map(_.buildSF(file, params))
```
anthonyccri: Cache a reference to the mime-type so that we're not walking a list for performance?

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/handlers/WKTFileHandler.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import scala.collection.JavaConversions._
+
+class WKTFileHandler extends AbstractFileHandler {
+  override def canProcess(file: File, params: util.Map[String, String]) = {
+    params.contains("wkt")
+  }
+
+  override def getGeometryFromParams(params: util.Map[String, String]) = {
+    Option(WKTUtils.read(params("wkt")))
```
anthonyccri: Maybe read ```geom``` and ```dtg``` too?  Expose some constants in a java class for users to use.

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-exif-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/exif/MetadataFileHandler.scala
```diff
@@ -0,0 +1,43 @@
...
+
+class MetadataFileHandler extends AbstractFileHandler {
+  override def getGeometryFromFile(file: File): Option[Geometry] = {
+    val gps = ImageMetadataReader.readMetadata(file).getDirectoriesOfType(classOf[GpsDirectory])
+    val gl = gps.head.getGeoLocation
+
+    Option(WKTUtils.read(s"POINT(${gl.getLongitude} ${gl.getLatitude})"))
+  }
+
+  override def canProcess(file: File, params: util.Map[String, String]): Boolean = accept(file)
```
anthonyccri: Why the level of indirection?

geomesa-blobstore/geomesa-blobstore-api/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+            <artifactId>specs2_2.10</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+
+    </dependencies>
+
+    <build>
```
AndrewAnnex: A good thing to do would be to add an optional profile that will build a war containing the servlet along with the scalatra stuff so that it is easy to deploy against wildfly/jboss and other things.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -0,0 +1,112 @@
...
+    val value = entry.getValue
+
+    val filename = key.getColumnQualifier.toString
+
+    (value.get, filename)
+  }
+
+  def putInternal(file: File, id: String) {
+    val localName = file.getName
+    val bytes =  ByteStreams.toByteArray(Files.newInputStreamSupplier(file))
```
AndrewAnnex: if the file is big, this could blow up. I remember seeing a blob store implementation in https://github.com/calrissian/accumulo-recipes that gets around the issue.

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-exif-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/exif/MetadataFileHandler.scala
```diff
@@ -0,0 +1,43 @@
...
+import org.locationtech.geomesa.utils.text.WKTUtils
+
+import scala.collection.JavaConversions._
+
+class MetadataFileHandler extends AbstractFileHandler {
+  override def getGeometryFromFile(file: File): Option[Geometry] = {
+    val gps = ImageMetadataReader.readMetadata(file).getDirectoriesOfType(classOf[GpsDirectory])
+    val gl = gps.head.getGeoLocation
+
+    Option(WKTUtils.read(s"POINT(${gl.getLongitude} ${gl.getLatitude})"))
```
jahhulbert-ccri: maybe just create a Point directly instead of WKT-ing it

geomesa-blobstore/geomesa-blobstore-core/src/test/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStoreTest.scala
```diff
@@ -0,0 +1,89 @@
...
+@RunWith(classOf[JUnitRunner])
+class AccumuloBlobStoreTest extends Specification {
+  val dsParams = Map(
+    "instanceId"        -> "mycloud",
+    "zookeepers"        -> "zoo1:2181,zoo2:2181,zoo3:2181",
+    "user"              -> "myuser",
+    "password"          -> "mypassword",
+    "tableName"         -> "geomesa",
+    "useMock"           -> "true",
+    "featureEncoding"   -> "avro")
```
jahhulbert-ccri: remove feature encoding and default to kryo unless there's some other reason?

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
anthonyccri: Don't they split across multiple cells or something?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -0,0 +1,51 @@
...
+      NotFound(reason = "AccumuloBlobStore is not initialized.")
+    } else {
+      val (returnBytes, filename) = abs.get(id)
+      if (returnBytes == null) {
+        NotFound(reason = s"Unknown ID $id")
+      } else {
+        contentType = "application/octet-stream"
+        response.setHeader("Content-Disposition", "attachment; filename=" + filename)
+
+        Ok(returnBytes)
```
jahhulbert-ccri: should use chunked encoding when sending back large http responses...not sure if this will do that.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -0,0 +1,51 @@
...
+    println("In ds registration method")
+
+    val dsParams = datastoreParams
+    val ds = new AccumuloDataStoreFactory().createDataStore(dsParams).asInstanceOf[AccumuloDataStore]
+
+    if (ds == null) {
+      NotFound(reason = "Could not load data store using the provided parameters.")
+    } else {
+      abs = new AccumuloBlobStore(ds)
+      Ok()
```
jahhulbert-ccri: in reference to @anthonyccri 's comment above...if we are creating a new blob store it would be nice to send back a 201 Created HTTP response with a Location header to be more RESTful and ensure that the resource was registered at that ID...

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
jahhulbert-ccri: +1

AndrewAnnex: yeah that is the part I am remembering, also using some sort of buffered write streaming stuff which is the main issue here. see the following: https://github.com/calrissian/accumulo-recipes/blob/master/store/blob-store/src/test/java/org/calrissian/accumulorecipes/blobstore/impl/AccumuloBlobStoreTest.java

it is a good first pass, but we here would be bringing the whole file into memory here.

jahhulbert-ccri: Agree'd we'd want to not buffer the whole thing in mem in a J2EE server is possible. good use case for a nifi-based ingest or something to put things in a table without going through geoserver

geomesa-blobstore/geomesa-blobstore-api/pom.xml
AndrewAnnex: nvm forgot that we need wfs

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jahhulbert-ccri: if we know it i'd support that...if the user provides one on ingest we should return it

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: Agreed that we should have a POST for uploading a file.  @ellisbjohns will be entering a ticket for that.  Skipping the 'canProcess' may or may not make sense.  We'll have to see how things develop.

geomesa-blobstore/geomesa-blobstore-api/pom.xml
jnh5y: Ayup, agreed.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: As a manner of expediency, I ripped off @elahrvivaz's way of configuring a DS for the SQL Spark.  @ellisbjohns has a ticket for exploring configuration.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: Got point.  @ellisbjohns, can you capture that in a ticket?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: I poked at Scalatra's ability to return something more specific.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: Ditto re: Scalatra's ability to support this.  Ellis will have to dig in, etc.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
jnh5y: @ellisbjohns Can you capture the max-size considerations?

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/handlers/BlobStoreFileHandler.scala
jnh5y: Definitely an interesting optimization.  Accepted mime-types could be exposed by the FileHandler interfaces.  

We may still have to walk plugins which do not announce mime-type.

elahrvivaz:
I thought that storing large blobs in accumulo wasn't very efficient?

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/handlers/WKTFileHandler.scala
jnh5y: Sure.  My main goal was to bang out the AbstractFileHandler and two quick examples.  

As a note, I'm leaving the WKTFileHandler in the core project since it doesn't require any dependencies and can serve a quick example class.

geomesa-blobstore/geomesa-blobstore-core/src/test/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStoreTest.scala
jnh5y: Doh, cut&paste fail.

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-exif-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/exif/MetadataFileHandler.scala
jnh5y: No reason; fixing it.

jnh5y:
@elahrvivaz Do you have a max size suggestion for an Accumulo value?

elahrvivaz:
@jnh5y not really...

jahhulbert-ccri:
@elahrvivaz i don't think is...

jahhulbert-ccri:
@jnh5y here's some starts...split size default is 1G...seems bad to store things that large...I've done things with order 100MB but of course you ruin your data cache quickly with those types of things.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
anthonyccri: Hmm, I suggest getting access to the Geoserver catalog and crawling it for each of its stores and associated connection params.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/handlers/WKTFileHandler.scala
anthonyccri: Yup, agree with that.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
anthonyccri: Capture in a ticket but maybe punt until down the road.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: That could work; that code couldn't be part of the LT project.

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-exif-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/exif/MetadataFileHandler.scala
jnh5y: @jahhulbert-ccri How do you generate just a Point?  I tend to find it cumbersome, and the WKTUtils were quite handy:)

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
anthonyccri: Hmm, is there a workaround a la datastore params?

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-exif-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/exif/MetadataFileHandler.scala
anthonyccri:
```scala
val gf = JTSFactoryFinder.getGeometryFactory
...
// in your method
val pt = gf.createPoint(new Coordinate(gl.getLongitude, gl.getLatitude))
```

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: One could add an applicationContext.xml which provided the params for a DataStore/BlobStore.

PR #708, opened 2015-11-10
GEOMESA-946 AND queries against indexed attributes do not return correct results

Afterglow375:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
@@ -590,6 +590,52 @@ class AttributeIndexStrategyTest extends Specification with TestWithDataStore {
       features must haveLength(3)
       features must contain("alice", "bill", "bob")
     }
+
+    "correctly query on indexed attributes with nonsensical AND queries" >> {
+      "redundant int query" >> {
+        val features = execute("age > 25 AND age > 15")
+        features must haveLength(1)
+        features must contain("bob")
+      }
+
+      "int query that returns nothing" >> {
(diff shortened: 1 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
```diff
@@ -590,6 +590,52 @@ class AttributeIndexStrategyTest extends Specification with TestWithDataStore {
...
     }
+
+    "correctly query on indexed attributes with nonsensical AND queries" >> {
+      "redundant int query" >> {
+        val features = execute("age > 25 AND age > 15")
+        features must haveLength(1)
+        features must contain("bob")
+      }
+
+      "int query that returns nothing" >> {
```
cne1x: What does "int" mean in this context?

elahrvivaz:
heh, so are there still some cases that fail? if not, maybe change the commit message to be something along the lines of 'fixing AND queries against indexed attributes'

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
Afterglow375: int as in "age" is an Integer in the SFT.

cne1x: Sigh.  I could have sworn I saw that label on a test that was doing date filtering.  The longer I stay this afternoon, the more I realize I'm still fogged.  Sorry for the silly question.

elahrvivaz: @cne1x check the commit history ;)

cne1x:
Merged as commit b81c9f810aaae7b720f7a77af9d1fdd40529e166

PR #713, opened 2015-11-13
GEOMESA-967 Users should be able to build GeoMesa with Accumulo 1.5 support

jdk2pq:
Signed-off-by: Jake Kenneally <jdk2pq@ccri.com>

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/AccumuloVersion.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/AccumuloVersion.scala
@@ -72,14 +72,10 @@ object AccumuloVersion extends Enumeration {
     }
   }
 
-  def getNsOps(connector: Connector) = {
-    val nsOps = classOf[Connector].getMethod("namespaceOperations").invoke(connector)
-    val nsOpsClass = Class.forName("org.apache.accumulo.core.client.admin.NamespaceOperations")
-    (nsOps, nsOpsClass)
-  }
-
   def nameSpaceExists(nsOps: AnyRef, nsOpsClass: Class[_], ns: String) = {
-    nsOpsClass.getMethod("exists", classOf[String]).invoke(nsOps, ns).asInstanceOf[Boolean]
+    val existsMethod = nsOpsClass.getMethod("exists", classOf[String])
+    existsMethod.setAccessible(true)
+    existsMethod.invoke(nsOps, ns).asInstanceOf[Boolean]
   }
 
   /**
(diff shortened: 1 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/AccumuloVersion.scala
```diff
@@ -80,9 +90,8 @@ object AccumuloVersion extends Enumeration {
     if (dot > 0) {
       require(accumuloVersion != V15, s"Table namespaces are not supported in Accumulo 1.5 - for table '$table'")
       val ns = table.substring(0, dot)
-      val nsOps = classOf[Connector].getMethod("namespaceOperations").invoke(connector)
-      val nsOpsClass = Class.forName("org.apache.accumulo.core.client.admin.NamespaceOperations")
-      if (!nsOpsClass.getMethod("exists", classOf[String]).invoke(nsOps, ns).asInstanceOf[Boolean]) {
+      val (nsOps, nsOpsClass) = getNsOps(connector)
+      if (!nameSpaceExists(nsOps, nsOpsClass, ns)) {
```
elahrvivaz: could you just do (nsOps, nsOps.getClass, ns) and avoid passing the class back?

geomesa-process/pom.xml
```diff
@@ -22,12 +22,14 @@
         <dependency>
             <groupId>org.apache.accumulo</groupId>
             <artifactId>accumulo-core</artifactId>
-            <scope>provided</scope>
```
elahrvivaz: was this just redundant? we still want it to be provided scope, right?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/AccumuloVersion.scala
jdk2pq: That works, but then I have to use the setAccessible(true) method on a protected class in order to get the "exists" and "create" methods. I'm not sure which is the better approach, but I'll push up the changes to compare.

geomesa-process/pom.xml
jdk2pq: I thought it was redundant since accumulo-core is already marked as provided in GeoMesa.

elahrvivaz: that's fine, just make sure the dependency tree doesn't change.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/AccumuloVersion.scala
elahrvivaz: Eh, well it's all kind of a hack, so either way is fine - i'll leave it up to you

geomesa-process/pom.xml
jdk2pq: ╭─jkenneally@jkenneally  geomesa-process  ‹fcr_make_acc_15_work› 
╰─$ mvn dependency:tree -Dverbose | grep accumulo-core
[INFO] +- org.apache.accumulo:accumulo-core:jar:1.6.2:provided

looks good

anthonyccri:
b655ede..10be6fe

PR #715, opened 2015-11-16
GEOMESA-972 Remove DataStore Edit Panels

jnh5y:
* Removes datastore panels.
* Adds some datastore parameters to the getParametersInfo.


jdk2pq:
In the Accumulo (GeoMesa) data store, the "Password" field no longer hides the text, and the options for writeThreads, queryThreads, and recordThreads no longer exist.

jdk2pq:
Both the Kafka Data Store and Accumulo Coverage Format appear to be the same as before.

jdk2pq:
The Accumulo (GeoMesa) store and layer functionality appears to be working in my testing, other than the visual inconsistencies described above.

jahhulbert-ccri:
@jnh5y are these going anywhere else?

anthonyccri:
@jahhulbert-ccri  No, i think they're actually unnecessary.  Geotools params cover the whole thing.

jnh5y:
@elahrvivaz, I've added back the geoserver-plugin jars.  Since there's no code in the Stream and HBase plugins, I'm wondering if we should bite the bullet and fold the plugin generation into the 'core' modules.

Thoughts?

Potentially, we could make that change right as we transition to 1.2.0.

elahrvivaz:
@jnh5y I'm not sure about that, we still want people to be able to import the modules without getting the uber-jar... it could still be possible in one module but might be trickier.
In general, it might be good to adopt a standard of using 'assemble' or some such profile to build uber-jars - that would speed up the build too. What do you think?

README.md
```diff
@@ -1,11 +1,11 @@
-<p align="center"><a href="http://geomesa.github.io"><img align="center" width="50%" src="https://raw.githubusercontent.com/geomesa/geomesa.github.io/master/img/geomesa-2x.png"></img></a>
-<br/>
-<img align="center" height="150px" src="http://www.geomesa.org/img/geomesa-overview-848x250.png"></img>
-</p>
+<p align="center"><a href="http://geomesa.github.io"><img src="https://raw.githubusercontent.com/geomesa/geomesa.github.io/master/img/geomesa-2x.png"></img></a></p>
```
jdk2pq: Rebase to the latest on master and then run `mvn process-resources -N` to remove these changes.

jnh5y:
@elahrvivaz Good points.  When the shade plugin runs, it creates a different jar with a classifier, so I don't think there'd be any issues with jars to import, etc.

To separate concerns, I've made a ticket for Doing It Right https://geomesa.atlassian.net/browse/GEOMESA-981.

elahrvivaz:
superseded by PR 721

PR #717, opened 2015-11-17
[WIP] GEOMESA-960 Refactored temporal density iterator to range histogram i…

Afterglow375:
…terator

* Comments only PR for now


elahrvivaz:
it would be nice if we didn't lose the ability to do date buckets - and ints, doubles, etc would be nice too. (that said, I don't think anyone uses the time density so maybe it's ok to lose it - is there a project use case that uses this Long iterator?)

Afterglow375:
#739 incorporates and expands upon this.

PR #719, opened 2015-11-17
GEOMESA-937 Use Geomesa-Convert for Ingest Tools

jahhulbert-ccri:
* Use the geomesa-convert framework for ingesting data via the command line tools
* Replace old ingest command line args with args for converter configuration
* Add transforms to converter framework for lists and maps
* Allow SFT to be specified in TypeSafe conf for ingest (as a string or a filename)
* Allow conf files with SFT and converter in the same file
* Configurable paths to SFT and Convert Conf objects in TypeSafe config files


jahhulbert-ccri:
FYI things I want to do before final PR:
1. Handle custom index schema in config or on command line
2. More testing
3. Document converter config and make tutorials (maybe this is a separate ticket)
4. Prepare update for geomesa website to be deployed at release time
5. allow custom shards for ingest via config or command line
6. figure out how to do handle shared tables

Please comment on those as well! thanks

geomesa-tools/README.md
```diff
@@ -498,55 +476,56 @@ The file type is inferred from the extension of the file, so ensure that the for
...
 
-    # file.csv (comma separated)
-    featureId,date,geom
-    1.23623623,01/01/2014 06:30:23,POINT(2 3)
-    290.43234,01/02/2014 08:35:24,POINT(3 3)
-    3.14159,01/03/2014 18:11:56,POINT(4 5)
+    $ cat example1.csv
+    ID,Name,Age,LastSeen,Friends,Lat,Lon
+    23623,Harry,20,2015-05-06,"Will, Mark, Suzan",-100.236523,23
+    26236,Hermoine,25,2015-06-07,"Edward, Bill, Harry",40.232,-53.2356
```
jdk2pq: It's *Hermione*, not Hermoine.

leviosa-gif

geomesa-convert/geomesa-convert-avro/src/test/scala/org/locationtech/geomesa/convert/avro/AvroTransformersTest.scala
```diff
@@ -10,7 +10,7 @@ package org.locationtech.geomesa.convert.avro
 
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.convert.Transformers
-import org.locationtech.geomesa.convert.Transformers.EvaluationContext
+import org.locationtech.geomesa.convert.Transformers.{DefaultCounter, EvaluationContext}
```
elahrvivaz: unnecessary?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -117,16 +146,24 @@ object Transformers extends JavaTokenParsers {
...
+  case class LitFloat(value: java.lang.Float) extends Lit[java.lang.Float]
   case class LitDouble(value: java.lang.Double) extends Lit[java.lang.Double]
+  case class LitBoolean(value: java.lang.Boolean) extends Lit[java.lang.Boolean]
+
   case class Cast2Int(e: Expr) extends Expr {
     override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = e.eval(args).asInstanceOf[String].toInt
   }
-
+  case class Cast2Long(e: Expr) extends Expr {
+    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = e.eval(args).asInstanceOf[String].toLong
```
elahrvivaz: will these always be strings? would something like this be more robust?
```
e.eval(args) collect {
  case s: String => s.toLong
  case l: Long => l
  case l: java.lang.Long => l
  case null => null
}
```

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
```diff
@@ -29,29 +29,41 @@ class DelimitedTextConverterFactory extends SimpleFeatureConverterFactory[String
...
+
     val fields    = buildFields(conf.getConfigList("fields"))
     val idBuilder = buildIdBuilder(conf.getString("id-field"))
-    val pipeSize  = if(conf.hasPath("pipe-size")) conf.getInt("pipe-size") else 16*1024
-    new DelimitedTextConverter(format, targetSFT, idBuilder, fields, pipeSize)
+    new DelimitedTextConverter(baseFmt, targetSFT, idBuilder, fields, opts)
   }
 }
 
+case class DelimitedOptions(skipHeader: Boolean = false,
```
elahrvivaz: it would actually be nice if skipHeader was an int - maybe this doesn't happen much, but for the microsoft tdrive data we have to drop the first 6 lines.

geomesa-tools/examples/ingest/example1.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+geomesa ingest -u user -p password -i inst -z zoo -c catalog -conf example1.conf example1.csv
```
elahrvivaz: convention would be --conf, right? single dash for a single char shortcut?

geomesa-tools/pom.xml
```diff
@@ -53,6 +53,30 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-convert-text</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-convert-xml</artifactId>
+        </dependency>
+        <dependency>
```
elahrvivaz: might be nice to have a geomesa-convert-all module that brings in all the rest.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
```diff
@@ -118,6 +115,86 @@ object Utils {
...
     fs.delete(path, true)
   }
+
+  object Configurator extends Logging {
+    type ConfigParser = String => Option[Config]
+
+    def getConfig(configArg: String): Config =
+      Seq(parseString, parseFile)
+       .view.map(_(configArg))
+       .find(_.nonEmpty)
```
elahrvivaz: is this an option? can't tell, but if so you could flatmap instead

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
```diff
@@ -118,6 +115,86 @@ object Utils {
...
+        .get
+    }
+
+    private[Speculator] def getSpecParser (specArg: String, nameOpt: Option[String]) : SpecParser = () => {
+      nameOpt.map[Option[SimpleFeatureType]] { featureName =>
+        Try { SimpleFeatureTypes.createType(featureName, specArg) }
+        match {
+          case Success(sft) => Some(sft)
+          case Failure(ex)  =>
+            logger.debug(s"Unable to parse sft spec from string $specArg with error ex.getMessage")
```
elahrvivaz: missing ${} on your message

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
```diff
@@ -118,6 +115,86 @@ object Utils {
...
+        }
+      }.getOrElse(Option.empty)
+    }
+
+    private[Speculator] def getConfParser(str: String): SpecParser = () =>
+      Try { SimpleFeatureTypes.createType(ConfigFactory.parseString(str)) }
+      match {
+        case Success(sft) => Some(sft)
+        case Failure(ex)  =>
+          logger.debug(s"Unable to parse sft conf from string $str with error ex.getMessage")
```
elahrvivaz: and here.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
```diff
@@ -118,6 +115,86 @@ object Utils {
...
+          Option.empty
+      }
+
+    def readFile(s: String): Option[String] = {
+      val f = new File(s)
+      if (f.exists && f.canRead) {
```
elahrvivaz: probably want a .isFile too (vs directory)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
```diff
@@ -118,6 +115,86 @@ object Utils {
...
+    private[Configurator] val parseFile: ConfigParser = (configArg: String) =>
+      Try(ConfigFactory.parseFile(new File(configArg))) match {
+        case Success(config) => Some(config)
+        case Failure(ex) =>
+          logger.debug(s"Unable to parse config from file $configArg")
+          None
+      }
+  }
+
+  object Speculator extends Logging {
```
elahrvivaz: name is kind of ambiguous.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
```diff
@@ -25,14 +28,18 @@ class IngestCommand(parent: JCommander) extends Command(parent) with Logging {
...
-          "are csv,tsv,shp")
+    if (fmt == SHP) {
+      val ds = new DataStoreHelper(params).getOrCreateDs()
+      GeneralShapefileIngest.shpToDataStore(params.files(0), ds, params.featureName)
+    } else {
+      val config = Configurator.getConfig(params.config)
+      val sft = Speculator.getSft(params.spec, params.featureName, params.config)
+
+      if (config != null && sft != null) {
+        new DelimitedIngest(params).run()
```
elahrvivaz: should probably pass in the config and sft, or do this check in the delimited ingest class. otherwise the code has to stay in sync between here and there.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/DelimitedIngest.scala
```diff
@@ -17,23 +17,29 @@ import org.apache.commons.codec.binary.Hex
...
 
 class DelimitedIngest(params: IngestParameters) extends AccumuloProperties {
 
+  val sft = Speculator.getSft(params.spec, params.featureName, params.config)
+  val converterConfig = Configurator.getConfig(params.config)
+
   def run(): Unit = {
     // create schema for the feature prior to Ingest job
-    FeatureCreator.createFeature(params)
+//    FeatureCreator.createFeature(params, params.convertSpec)
```
elahrvivaz: commented out code

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/DelimitedIngest.scala
```diff
@@ -83,46 +89,32 @@ class DelimitedIngest(params: IngestParameters) extends AccumuloProperties {
...
       IngestParams.ACCUMULO_USER     -> params.user,
       IngestParams.ACCUMULO_PASSWORD -> getPassword(params.password),
-      IngestParams.DO_HASH           -> params.hash.toString,
-      IngestParams.FORMAT            -> Option(params.format).getOrElse(getFileExtension(params.files(0))),
-      IngestParams.FEATURE_NAME      -> params.featureName,
-      IngestParams.IS_TEST_INGEST    -> false.toString
+      IngestParams.ACCUMULO_MOCK     -> params.useMock.toString,
+      IngestParams.FEATURE_NAME      -> sft.getTypeName,
+      IngestParams.IS_TEST_INGEST    -> false.toString,
+      IngestParams.CONVERTER_CONFIG  -> URLEncoder.encode(converterConfig.root().render, StandardCharsets.UTF_8.displayName)
```
elahrvivaz: you can pass in ConfigRenderOptions.concise() to render to strip out comments and whitespace and save a little space

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.locationtech.geomesa.tools.Utils.IngestParams
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+
+class ScaldingConverterIngestJob(args: Args) extends Job(args) with Logging {
+  import scala.collection.JavaConversions._
+
+  val counter = new DefaultCounter
+
+  lazy val pathList         = DelimitedIngest.decodeFileList(args(IngestParams.FILE_PATH))
+  lazy val sftSpec          = URLDecoder.decode(args(IngestParams.SFT_SPEC), StandardCharsets.UTF_8.displayName)
```
elahrvivaz: hmm, is there a reason we url encode and decode these strings?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
```diff
@@ -0,0 +1,119 @@
...
+      }
+  }
+
+  private def processLine(resources: Resources, s: String) =
+    resources.callback(s)
+      .foreach { sf =>
+        val toWrite = resources.fw.next()
+        toWrite.setAttributes(sf.getAttributes)
+        toWrite.getIdentifier.asInstanceOf[FeatureIdImpl].setID(sf.getID)
+        toWrite.getUserData.putAll(sf.getUserData)
```
elahrvivaz: should set Hints.USE_PROVIDED_FID (or have an option for it at least)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
```diff
@@ -0,0 +1,119 @@
...
+      dsp.instanceIdParam.getName -> args(IngestParams.ACCUMULO_INSTANCE),
+      dsp.tableNameParam.getName  -> args(IngestParams.CATALOG_TABLE),
+      dsp.userParam.getName       -> args(IngestParams.ACCUMULO_USER),
+      dsp.passwordParam.getName   -> args(IngestParams.ACCUMULO_PASSWORD),
+      dsp.authsParam.getName      -> args.optional(IngestParams.AUTHORIZATIONS),
+      dsp.visibilityParam.getName -> args.optional(IngestParams.VISIBILITIES),
+      dsp.mockParam.getName       -> args.optional(IngestParams.ACCUMULO_MOCK)
+    ).collect{ case (key, Some(value)) => (key, value); case (key, value: String) => (key, value) }
+
+  lazy val sft = {
```
elahrvivaz: actually, the sft has already been created in the data store at this point - it would be better to load it from the data store and not pass the string in and parse it, that way it will have all the correct user data set (if needed).

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
```diff
@@ -0,0 +1,119 @@
...
+    resources.callback(s)
+      .foreach { sf =>
+        val toWrite = resources.fw.next()
+        toWrite.setAttributes(sf.getAttributes)
+        toWrite.getIdentifier.asInstanceOf[FeatureIdImpl].setID(sf.getID)
+        toWrite.getUserData.putAll(sf.getUserData)
+        resources.fw.write()
+      }
+
+  def runTestIngest(lines: Iterator[String]) = {
```
elahrvivaz: i really wish we could remove this.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -44,14 +44,17 @@ object SimpleFeatureTypes {
   val USER_DATA_MAP_KEY_TYPE   = "keyclass"
   val USER_DATA_MAP_VALUE_TYPE = "valueclass"
 
-  def createType(conf: Config): SimpleFeatureType = {
-    val nameSpec = conf.getString("type-name")
+  def createType(conf: Config, path: Option[String] = Some("sft")): SimpleFeatureType = {
+    import org.locationtech.geomesa.utils.conf.ConfConversions._
+    val toParse = conf.getConfigOpt(path.get).map(conf.withFallback).getOrElse(conf)
```
elahrvivaz: should do a match on path instead of path.get

elahrvivaz:
looks good

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
jahhulbert-ccri: I think I can probably manage that...I added a test that actually executes the code. So I'll get ride of this thing.
https://github.com/locationtech/geomesa/pull/719/files#diff-ac19bd149f34d32e2a89db3822c0078fR146

geomesa-convert/geomesa-convert-avro/src/test/scala/org/locationtech/geomesa/convert/avro/AvroTransformersTest.scala
jahhulbert-ccri: yeah not sure how that happened :)

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
jahhulbert-ccri: i can imagine string -> number and num -> num like double to float, double to int, or something like that. but realistically i think 99% of the time it's parsing a CSV and casting a string to an int. long term i think the converters should do type inference if they don't already. i.e. if the field is a double, it should be an implicit ::double cast without specifying it.

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
jahhulbert-ccri: +1 or rather... +n :) since we have a use case i'll change it to ```skip-lines = 4``` perhaps?

geomesa-tools/examples/ingest/example1.sh
jahhulbert-ccri: yeah it supports --conf we (esp me) have been bad about single dash args with >1 length...perhaps we should dial that back? we use -c for catalog... we could also support just the single dash arg or a capital -C 

see https://github.com/locationtech/geomesa/blob/master/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala#L43

geomesa-tools/pom.xml
jahhulbert-ccri: would def make it easier

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
jahhulbert-ccri: yup

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
jahhulbert-ccri: yeah...how about SftArgParser and ConvertConfigParser?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
jahhulbert-ccri: yup. in fact those methods throw the same exception anyways instead of returning null...i wrote that...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
jahhulbert-ccri: not sure. i carried that over from previous code. i assume that the scalding stuff does some sort of argparse that required it...since it treats args like --argname value i'll see if i can figure out exactly why and comment it

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
jahhulbert-ccri: yup

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
jahhulbert-ccri: meh actually going to make a follow on ticket maybe

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
jahhulbert-ccri:
```
    val toParse = path match {
      case Some(p) => conf.getConfigOpt(p).map(conf.withFallback).getOrElse(conf)
      case None    => conf
    }
```
look ok?

geomesa-tools/README.md
jahhulbert-ccri: man IntelliJ even caught that. no excuses.

geomesa-tools/examples/ingest/example1.csv
```diff
@@ -0,0 +1,4 @@
+ID,Name,Age,LastSeen,Friends,Lat,Lon
+23623,Harry,20,2015-05-06,"Will, Mark, Suzan",-100.236523,23
+26236,Hermoine,25,2015-06-07,"Edward, Bill, Harry",40.232,-53.2356
```
jdk2pq: If you do make my silly spelling change, make sure it happens here too.

geomesa-tools/src/test/resources/examples/example1.csv
```diff
@@ -0,0 +1,4 @@
+ID,Name,Age,LastSeen,Friends,Lat,Lon
+23623,Harry,20,2015-05-06,"Will, Mark, Suzan",-100.236523,23
+26236,Hermoine,25,2015-06-07,"Edward, Bill, Harry",40.232,-53.2356
```
jdk2pq: And here

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/IngestArgumentsTest.scala
```diff
@@ -0,0 +1,159 @@
...
+        s"-conf  ${confFile.getPath} ${dataFile.getPath}").split("\\s+")
+      args.length mustEqual 14
+
+      Runner.main(args)
+
+      val ds = getDS(id)
+      import Conversions._
+      val features = ds.getFeatureSource("renegades").getFeatures.features().toList
+      features.size mustEqual 3
+      features.map(_.get[String]("name")) must containTheSameElementsAs(Seq("Hermoine", "Harry", "Severus"))
```
jdk2pq: And here

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: yeah, looks good.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
elahrvivaz: also, even with strings, we'd probably want an empty string to equate to null when dealing with numeric types. not sure if that should also apply to string types...

elahrvivaz: I'm getting this issue now with empty strings using the xml parser:
```
{ name = elevation, path = "ele/text()", transform = "$0::double" }
```
```
java.lang.NumberFormatException: empty String
	at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1020)
	at java.lang.Double.parseDouble(Double.java:540)
	at scala.collection.immutable.StringLike$class.toDouble(StringLike.scala:232)
	at scala.collection.immutable.StringOps.toDouble(StringOps.scala:31)
	at org.locationtech.geomesa.convert.Transformers$Cast2Double.eval(Transformers.scala:126)

```

This goes back to the null checks, but slightly different since it's coming from xml. So something that can handle empty strings is what I need here.

elahrvivaz: Maybe the easiest thing short term is in the conversion process, we could try/catch each individual attribute, instead of all attributes. That way if something like this fails it will still load the feature, just without that one attribute.

elahrvivaz:
I might have mentioned this already, but can we update gson?
```
<groupId>com.google.code.gson</groupId>
<artifactId>gson</artifactId>
<version>2.2.4</version>
```
version 2.2.2 caused some parsing i was doing to fail due to it wrapping things with a 'member' key.

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/IngestArgumentsTest.scala
jahhulbert-ccri: whew got em all double checked too. 
$ grep -Ri hermoine *

can i pass my NEWT now?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ScaldingConverterIngestJob.scala
jahhulbert-ccri: ya

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/DelimitedIngest.scala
jahhulbert-ccri: yup

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/DelimitedIngest.scala
jahhulbert-ccri: got it

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
jahhulbert-ccri: yup

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Utils.scala
jahhulbert-ccri: yeah. i think if it wasn't a map and find it would be
```
      Seq(parseString, parseFile)
        .view.flatMap(_(configArg))
        .headOption
        .getOrElse(throw new ParameterException(s"Unable to parse Converter config from argument $configArg"))
```

either works. the view I think makes them fairly equivalent?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
jahhulbert-ccri: would we want to load the feature if one of the attributes for that feature fails? I'm thinking that we're going to need to think/test/write tests/try parsing data/etc before we cut the rc7 with the converters.

elahrvivaz: Yeah, I guess the issue is we don't have any way to handle something like the above empty string (or a null value), so the attribute try/catch would work around that. But ideally we'd have something that handles it explicitly.

jahhulbert-ccri: @elahrvivaz i'm gonna punt this to another ticket and start working on that as an improvement to the converters. But I'm gonna assign it to this release to handle casting and type transformations in a more robust fasion

geomesa-tools/examples/ingest/example1.sh
jahhulbert-ccri: @elahrvivaz @jnh5y @jdk2pq @anthonyccri any opinion

jahhulbert-ccri:
@elahrvivaz we are importing
```
        <dependency>
            <groupId>com.google.code.gson</groupId>
            <artifactId>gson</artifactId>
            <version>2.3.1</version>
        </dependency>
```
is that a https://geomesa.atlassian.net/browse/GEOMESA-982 thing or something else?

geomesa-tools/pom.xml
jahhulbert-ccri: pushing to another ticket cause i have some other places where i want to do this too. https://geomesa.atlassian.net/browse/GEOMESA-983

elahrvivaz:
@jahhulbert-ccri I thought when I checked that 2.2.2 was getting imported (unless you changed it?). If it's 2.3.x that should hopefully be fine too. It seemed a very odd 'bug'/feature that basically broke the json path expressions.

geomesa-tools/examples/ingest/example1.sh
elahrvivaz: Eh, looking at the other commands we're not very consistent anywhere - just go with whatever.

jahhulbert-ccri: yeah i think doing all --long-commands would have been better...or may having just added shorts for really really common things like -u, -p etc

jahhulbert-ccri:
@elahrvivaz yeah i verified we have 2.3.1 with a dep tree...the only place we have 2.2.2 is test scope of geomesa-plugin as a transitive from accumulo-test -> accumulo-monitor -> gson.

lemme know if you see it again

jahhulbert-ccri:
@elahrvivaz @jdk2pq lemme know if there's any other things we want to handle here. I'll do a few more ingests with data from geobench to test out some configs since it is a breaking change. If you know of any command-line ingest users lemme know maybe we can get them to test it.

jahhulbert-ccri:
closing for a new PR

PR #720, opened 2015-11-18
WIP for Stats framework.

jnh5y:


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.opengis.feature.simple.SimpleFeature
+
+import scala.collection.mutable
+
+case class EnumeratedHistogram[T](attribute: String) extends Stat {
+
+  val map: scala.collection.mutable.HashMap[T, Long] = mutable.HashMap[T, Long]()
+
+  override def observe(sf: SimpleFeature): Unit = {
+    val sfval = sf.getAttribute(attribute)
```
elahrvivaz: for performance, it would be better to look up the attribute by index instead of name

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
```diff
@@ -0,0 +1,51 @@
...
+
+    if (sfval != null) {
+      sfval match {
```
elahrvivaz: this is going to throw a match error if it's not a T right? probably would be nicer to just do sfval.asInstanceOf so the exception is clearer.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  override def toJson(): String = {
+    s"[${map.map{ case (key: T, count: Long) => s"$key:$count" }.mkString(",")}]"
+  }
+
+  override def add(other: Stat): Stat = {
+    other match {
+      case eh: EnumeratedHistogram[T] =>
+        combine(eh)
+        this
```
elahrvivaz: should explicitly handle the non-match and throw a decent exception

elahrvivaz:
seems reasonable

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
anthonyccri: Will we ever want to add anything other than another EnumeratedHist?  I suggest getting rid of the generic Stat here and just using EH.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/IteratorStackCounter.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.opengis.feature.simple.SimpleFeature
+
+case class IteratorStackCounter extends Stat {
+  var count: Long = 1
+
+  override def observe(sf: SimpleFeature): Unit = { }
+
+  override def toJson(): String = s"iteratorStackCount:$count"
+
+  override def add(other: Stat): Stat = {
```
anthonyccri: Again, generic matching here is going to be slow.  Make it specific probably with a type param.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
```diff
@@ -0,0 +1,61 @@
...
+
+
+case class MinMax[T <: Comparable[T]](attribute: String) extends Stat {
+
+  var min: T = _
+  var max: T = _
+
+  override def observe(sf: SimpleFeature): Unit = {
+
+    val sfval = sf.getAttribute(attribute)
```
anthonyccri: Use and cache the index of the attribute for performance.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
```diff
@@ -0,0 +1,61 @@
...
+    if (sfval != null) {
+      updateMin(sfval.asInstanceOf[T])
+      updateMax(sfval.asInstanceOf[T])
+    }
+  }
+
+  override def add(other: Stat): Stat = {
+
+    other match {
+      case mm: MinMax[T] =>
```
anthonyccri: Performance again . . . do we need generic behavior here?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
```diff
@@ -0,0 +1,61 @@
...
+    } else {
+      if (min.compareTo(sfval) > 0) {
+        min = sfval
+      }
+    }
+  }
+
+  private def updateMax(sfval : T): Unit = {
+    if (max == null) {
+      max = sfval
```
anthonyccri: set max to Int.MinValue so that we don't have to do a null check here (for performance).

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -0,0 +1,65 @@
...
+trait Stat {
+  def observe(sf: SimpleFeature)
+  def add(other: Stat): Stat
+
+  def toJson(): String
+}
+
+object Stat {
+
+  class StatParser extends RegexParsers {
```
anthonyccri: Rather than making this an external DSL, let's make it an internal DSL using more of a builder pattern.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/IteratorStackCounter.scala
```diff
@@ -0,0 +1,27 @@
...
+  * accompanies this distribution and is available at
+  * http://www.opensource.org/licenses/apache2.0.php.
+  *************************************************************************/
+
+
+package org.locationtech.geomesa.utils.stats
+
+import org.opengis.feature.simple.SimpleFeature
+
+case class IteratorStackCounter extends Stat {
```
anthonyccri: What's the purpose of this class?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
jnh5y: A JavaScript client will not be able to leverage the Builder pattern.  

I'd be happy to have a Builder which creates the DSL.  Does adding that make sense?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/IteratorStackCounter.scala
jnh5y: This is a stat to get inside the query execution.  

Since the 1 is set at instantiation, you get to see how many Iterator Stacks were set up...  Correlating that with query performance may be fun;)

jnh5y: Also, it was dead simple to write.;)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
anthonyccri: A javascript client?

jnh5y: This'll get exposed as WPS, and as someone calls that, I could imagine wanting a little DSL to specify desired stats.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
jnh5y: Ah, and using the SFT at class creation would make that pretty easy...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
jnh5y: @elahrvivaz @anthonyccri I tried to fix this here: https://github.com/locationtech/geomesa/compare/master...jnh5y:wip_stats_hlist

Anthony's point gets at the point that the 'add' method should take the type of the implementing class.  With implementations like 
```
case class MinMax[T <: Comparable[T]](attribute: String) extends Stat[MinMax[T]]
```
we can achieve that.  

The sticky wicket comes in with the implementation of SeqStat.  In order to implement, 'add' there, we'd need to be able to make a List[Monoid] a Monoid again (by delegating to coordinate-wise addition).

I struggled with that for a bit, and I'm happy to hear suggestions.  In my experimentation, I poked at Scalaz and Shapeless, but I haven't seen any clever options.  

One possibility is a SeqStat may have to retain a list of types (somehow), so that each stat can be cast appropriately during add.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
jnh5y: See other note.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
jnh5y: Good suggestion, except this class is meant to be general.  Since T would need to be Numeric, we may be able to wire up a T.min, T.max sort of thing...

elahrvivaz:
closing in favor of #739

PR #723, opened 2015-12-01
GEOMESA-987 Store user names with query stats

elahrvivaz:


elahrvivaz:
This conflicts with another PR, I'll re-open once that is merged.

PR #734, opened 2015-12-10
GEOMESA-999 Adding support for all geometries to z3 index

elahrvivaz:
* Adapted density, temporal-density, map-aggregate iterators to work on z3 indices and use kryo lazy decoding
* Refactored old iterators to extend the kryo iterators with back-compatible decoding
* Added server-side deduplication for aggregate and bin queries
* Modified FilterHelper intervals to account for the fact that joda time intervals are [min, max)


elahrvivaz:
I did some timings tests against old ST index using the tdrive data set as line strings. Results for initial tests were mixed. However, the Z3 only had 2 tablets (on a 10 node cluster), while the ST index had 16. I increased the split size on the Z3 table so that they were roughly equivalent, and included those numbers as well.

query | z3 | z3-with-splits | st
------- | -- |------------ | --
househour | 0.096 | 0.085 | 0.142
blockhour | 0.089 | 0.076 | 0.095
cityhour | 0.083 | 0.082 | 0.175
houseday | 0.527 | 0.461 | 0.177
blockday | 0.695 | 0.301 | 0.392
cityday | 1.128 | 0.476 | 0.733
houseweek | 1.412 | 0.733 | 0.735
blockweek | 1.881 | 0.785 | 1.705
cityweek | 3.116 | 3.042 | 3.926
cityweektaxi | 1.612 | 2.639 | 0.556

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
```diff
@@ -13,11 +13,10 @@ import com.vividsolutions.jts.geom.{Geometry, GeometryCollection}
 import org.apache.accumulo.core.data.Range
```
Afterglow375: When merging in this PR to my PR I found this outdated comment: https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryStrategyDecider.scala#L49

Should probably update the comment given the changes to the z3IdxStrategy, as it would be helpful to people like me.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
```diff
@@ -253,17 +175,61 @@ class BinAggregatingIterator extends SortedKeyValueIterator[Key, Value] with Log
...
    */
-  def writeGeometryWithLabel(sf: KryoBufferSimpleFeature): Unit = {
-    writeGeometry(sf)
-    writeLabelToBuffer(sf)
+  def writeGeometryWithLabel(sf: KryoBufferSimpleFeature, byteBuffer: ByteBuffer): Unit = {
+    writeGeometry(sf, byteBuffer)
+    writeLabelToBuffer(sf, byteBuffer)
   }
 
+  override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] = ???
```
Afterglow375: Is it intentional to leave the implementation of this as ???

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyAggregatingIterator.scala
```diff
@@ -0,0 +1,158 @@
...
+  private var topValue: Value = new Value()
+  private var currentRange: aRange = null
+
+  private var reusableSf: KryoBufferSimpleFeature = null
+
+  // server-side deduplication - not 100% effective, but we can't dedupe client side as we don't send ids
+  private val idsSeen = scala.collection.mutable.HashSet.empty[String]
+  private var maxIdsToTrack = -1
+
+  override def init(src: SortedKeyValueIterator[Key, Value],
```
Afterglow375: Seems like you're forced to call super.init in implementing classes. The problem I had is that I had to initialize the result returned from newResult() from parameters from the jOptions map, so I had to duplicate the result == null check in the beginning of my implemented init and then call super.init afterward. Doesn't seem ideal.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyDensityIterator.scala
```diff
@@ -0,0 +1,251 @@
...
+
+object KryoLazyDensityIterator extends Logging {
+
+  type DensityResult = mutable.Map[(Int, Int), Double]
+  type GridIterator = (SimpleFeature) => Iterator[(Double, Double, Double)]
+
+  // needs to be lazy to avoid class loading issues before init is called
+  lazy val DENSITY_SFT = SimpleFeatureTypes.createType("density", "result:String,*geom:Point:srid=4326")
+
+  val DEFAULT_PRIORITY = 25
```
Afterglow375: The iterators all seem to have their own DEFAULT_PRIORITY defined, which almost look like arbitrary ints to me. Maybe there should be some common place for iterator priorities.

elahrvivaz: Yeah, we originally had these in a common place... I personally think it's easier to keep track of when it's with the iterator. In practice, we almost never combine more than 1 iterator, so the priority is semi-irrelevant. (the caveat to that is that accumulo system iterators run at priority 15 or something, and we don't want to run before those).
It would be nice to have a better way to track them all though...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyAggregatingIterator.scala
elahrvivaz: Let me see if I can make it work a little better...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
elahrvivaz: Yeah, if it ever gets called it would throw a NotImplementedError - but it would only get called if some other iterator is stacked on top of it, which doesn't happen right now... we can always go back and implement it later if we need to.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
elahrvivaz: I added a ticket [here](https://geomesa.atlassian.net/browse/GEOMESA-1014) to update the docs/comments on query planning.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyAggregatingIterator.scala
elahrvivaz: @Afterglow375 I changed the signature to make this work better. Take a look (in fcr_z3_polys) and let me know.

Afterglow375: Yep, looks better to me now.

jahhulbert-ccri:
taking a look at this one next

elahrvivaz:
Need a new release of sfcurve for this PR.

anthonyccri:
6c6873e..70a83df

PR #735, opened 2015-12-14
GEOMESA-1003 Adding 4 splits (shards) to z3

elahrvivaz:

Note: stacks on top of #734, please only consider last commit(s).


elahrvivaz:
Updated timings from z3 polygon testing with this PR (note: number tend to fluctuate a bit):

query | z3 | z3 with manual splits | st | z3 with auto 4 splits (this PR)
--------|-----|-------------------|----|--------------------------------
househour | 0.096 | 0.085 | 0.142 | 0.314
blockhour | 0.089 | 0.076 | 0.095 | 0.084
cityhour | 0.083 | 0.082 | 0.175 | 0.076
houseday | 0.527 | 0.461 | 0.177 | 0.186
blockday | 0.695 | 0.301 | 0.392 | 0.246
cityday | 1.128 | 0.476 | 0.733 | 0.390
houseweek | 1.412 | 0.733 | 0.735 | 0.744
blockweek | 1.881 | 0.785 | 1.705 | 1.017
cityweek | 3.116 | 3.042 | 3.926 | 3.075
cityweektaxi | 1.612 | 0.556 | 2.639 | 0.878

jahhulbert-ccri:
We are holding off till after 1.2.0

jahhulbert-ccri:
did we try any other split levels?

elahrvivaz:
@jahhulbert-ccri I didn't try other shards. You should be able to control it just by changing the constant in Z3Table. Anything that fits in a single byte would be ok, so up to 127. We could also make it dynamic, but that seemed like just more complexity... The thing to keep in mind is that the number of ranges we scan gets multiplied by the number of shards. We should probably include some logic in the z range calculation that limits it based on the shards....

elahrvivaz:
I've rebased this PR, but it depends on an sfcurve merge.

anthonyccri:
70a83df..cf1e720

PR #737, opened 2015-12-14
GEOMESA-982 Geomesa transformers robust

drackaer:
https://geomesa.atlassian.net/browse/GEOMESA-982
Added string2 transformer functions for more robust type casting.

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.locationtech.geomesa.convert.TransformerFunctionFactory
+++ b/geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.locationtech.geomesa.convert.TransformerFunctionFactory
@@ -4,3 +4,4 @@ org.locationtech.geomesa.convert.GeometryFunctionFactory
 org.locationtech.geomesa.convert.IdFunctionFactory
 org.locationtech.geomesa.convert.LineNumberFunctionFactory
 org.locationtech.geomesa.convert.MapListFunctionFactory
+org.locationtech.geomesa.convert.CastFunctionFactory
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
@@ -487,29 +487,18 @@ class CastFunctionFactory extends TransformerFunctionFactory {
   override def functions = Seq(string2double, string2int, string2float, string2long, string2boolean)
 
   val string2double = TransformerFn("string2double") {
-    args => {
-      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toDouble
-    }
+    args => { try { args(0).asInstanceOf[String].toDouble } catch { case e: Exception => args(1) } }
   }
   val string2int = TransformerFn("string2int") {
-    args => {
-      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toInt
-    }
+    args => { try { args(0).asInstanceOf[String].toInt } catch { case e: Exception => args(1) } }
   }
   val string2float = TransformerFn("string2float") {
-    args => {
-      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toFloat
-    }
+    args => { try { args(0).asInstanceOf[String].toFloat } catch { case e: Exception => args(1) } }
   }
   val string2long = TransformerFn("string2long") {
-    args => {
-      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toLong
-    }
+    args => { try { args(0).asInstanceOf[String].toLong } catch { case e: Exception => args(1) } }
   }
   val string2boolean = TransformerFn("string2boolean") {
-    args => {
-      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toBoolean
-    }
+    args => { try { args(0).asInstanceOf[String].toBoolean } catch { case e: Exception => args(1) } }
   }
-  private def isEmptyOrNull(str: Any): Boolean = (str == null) || (str.asInstanceOf[String] == "")
 }
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
@@ -56,7 +56,7 @@ object Transformers extends EnhancedTokenParsers with Logging {
     def cast2int     = expr <~ "::int"     ^^ { e => Cast2Int(e)     }
     def cast2long    = expr <~ "::long"    ^^ { e => Cast2Long(e)    }
     def cast2float   = expr <~ "::float"   ^^ { e => Cast2Float(e)   }
-    def cast2double  = expr <~ "::double"  ^^ { e => Cast2Double(e) }
+    def cast2double  = expr <~ "::double"  ^^ { e => Cast2Double(e)  }
     def cast2boolean = expr <~ "::boolean" ^^ { e => Cast2Boolean(e) }
 
     def fieldLookup = "$" ~> ident ^^ { i => FieldLookup(i) }
@@ -484,14 +484,13 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
 }
 
 class CastFunctionFactory extends TransformerFunctionFactory {
-  override def functions = Seq(string2double, string2int, string2float, string2long, string2boolean)
+  override def functions = Seq(string2double, string2int, string2float, string2long, string2boolean, string2integer)
 
   val string2double = TransformerFn("string2double") {
     args => { try { args(0).asInstanceOf[String].toDouble } catch { case e: Exception => args(1) } }
   }
-  val string2int = TransformerFn("string2int") {
-    args => { try { args(0).asInstanceOf[String].toInt } catch { case e: Exception => args(1) } }
-  }
+  val string2int = TransformerFn("string2int") { str2intHelper() }
+  val string2integer = TransformerFn("string2integer") { str2intHelper() }
   val string2float = TransformerFn("string2float") {
     args => { try { args(0).asInstanceOf[String].toFloat } catch { case e: Exception => args(1) } }
   }
@@ -501,4 +500,6 @@ class CastFunctionFactory extends TransformerFunctionFactory {
   val string2boolean = TransformerFn("string2boolean") {
     args => { try { args(0).asInstanceOf[String].toBoolean } catch { case e: Exception => args(1) } }
   }
+
+  def str2intHelper() = (args: Seq[Any]) => { try { args(0).asInstanceOf[String].toInt } catch { case e: Exception => args(1) } }
 }
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
@@ -56,7 +56,7 @@ object Transformers extends EnhancedTokenParsers with Logging {
     def cast2int     = expr <~ "::int"     ^^ { e => Cast2Int(e)     }
     def cast2long    = expr <~ "::long"    ^^ { e => Cast2Long(e)    }
     def cast2float   = expr <~ "::float"   ^^ { e => Cast2Float(e)   }
-    def cast2double  = expr <~ "::double"  ^^ { e => Cast2Double(e)  }
+    def cast2double  = expr <~ "::double"  ^^ { e => Cast2Double(e) }
     def cast2boolean = expr <~ "::boolean" ^^ { e => Cast2Boolean(e) }
 
     def fieldLookup = "$" ~> ident ^^ { i => FieldLookup(i) }
@@ -157,7 +157,6 @@ object Transformers extends EnhancedTokenParsers with Logging {
   case class LitDouble(value: java.lang.Double) extends Lit[java.lang.Double]
   case class LitBoolean(value: java.lang.Boolean) extends Lit[java.lang.Boolean]
 
-  // TODO Better handling of casts including null handling, etc GEOMESA-982
   case class Cast2Int(e: Expr) extends Expr {
     override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = e.eval(args).asInstanceOf[String].toInt
   }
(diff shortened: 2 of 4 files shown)


jahhulbert-ccri:
@drackaer please merge-squash to a single commit

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,43 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
...
   }
-
+}
+class CastFunctionFactory extends TransformerFunctionFactory {
+  override def functions = Seq(string2double,string2int,string2float,string2long,string2boolean)
+
+  val string2double = TransformerFn("string2double") {
+    args => {
+      if (args(0) == null || args(0).asInstanceOf[String] == "") {
+        if (args(1) == null || args(1).asInstanceOf[String] == "") args(1) else args(1).asInstanceOf[String].toDouble
```
elahrvivaz: probably want to make this an explicit null here in the first part of the if - otherwise you might expect a double but actually have an empty string.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,43 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
...
+      } else args(0).asInstanceOf[String].toDouble
+    }
+  }
+  val string2int = TransformerFn("string2int") {
+    args => {
+      if (args(0) == null || args(0).asInstanceOf[String] == "") {
```
elahrvivaz: I might pull out this check into a function, or just use something like https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence)

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
drackaer: Not sure what you mean by that, can you clarify?  The only way you should be getting a non-double returned is if you explicitly provide one as a default value [i.e. args(1)](this should be clearer when I add the documentation, my bad).  Am I misunderstanding what you meant?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
drackaer: That would certainly make it look nicer, will do

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
elahrvivaz: that may be true... i just saw this possibility:
```if (args(1).asInstanceOf[String] == "") args(1)```
which would return an empty string instead of a double. Actually, would args(1) even be a string?
```string2int($1, 0)``` vs ```string2int($1, 0::int)``` vs ```string2int($1, '0')``` - which is the syntax we are expecting now?

Comment (drackaer) [inline geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala]: One point is that the Transformers class as built will not parse string2int($1, 0).  But that being said you are right, I am not sure if args(1) needs to ever be a string, my assumption was that we may have cases were a default to empty string was intentional, but if not then I can simplify the code a bit.

File: geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.locationtech.geomesa.convert.TransformerFunctionFactory [hunk 3]
```diff
@@ -3,4 +3,5 @@ org.locationtech.geomesa.convert.DateFunctionFactory
 org.locationtech.geomesa.convert.GeometryFunctionFactory
 org.locationtech.geomesa.convert.IdFunctionFactory
 org.locationtech.geomesa.convert.LineNumberFunctionFactory
-org.locationtech.geomesa.convert.MapListFunctionFactory
\ No newline at end of file
+org.locationtech.geomesa.convert.MapListFunctionFactory
+org.locationtech.geomesa.convert.CastFunctionFactory
```
jahhulbert-ccri: run organize imports in intellij?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -56,7 +56,7 @@ object Transformers extends EnhancedTokenParsers with Logging {
     def cast2int     = expr <~ "::int"     ^^ { e => Cast2Int(e)     }
     def cast2long    = expr <~ "::long"    ^^ { e => Cast2Long(e)    }
     def cast2float   = expr <~ "::float"   ^^ { e => Cast2Float(e)   }
-    def cast2double  = expr <~ "::double"  ^^ { e => Cast2Double(e)  }
+    def cast2double = expr <~ "::double"  ^^ { e => Cast2Double(e) }
```
jahhulbert-ccri: undo whitespace chage

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,43 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
         }.toMap.asJava
     }
   }
-
+}
+class CastFunctionFactory extends TransformerFunctionFactory {
```
jahhulbert-ccri: [style] newline before class

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,43 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
...
+  override def functions = Seq(string2double,string2int,string2float,string2long,string2boolean)
```
jahhulbert-ccri: [style] spaces after commas

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,43 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
...
+
+  val string2double = TransformerFn("string2double") {
```
jahhulbert-ccri: @drackaer will the empty/null handling apply to converters with casting? I think that's what we originally wanted to do here instead of explicitly calling string2double when i have something like:

```
 converter = {
   type         = "delimited-text",
   format       = "DEFAULT",
   id-field     = "md5(string2bytes($0))",
   fields = [
     { name = "oneup",    transform = "$1" },
     { name = "phrase",   transform = "$2" },
     { name = "somedub",   transform = "$3::double"     },
     { name = "geom",     transform = "geometry($4)"   }
   ]
 }
```

data:
```
1,hello,3.0,Point(46.0 45.0)
2,world,,Point(90.0 90.0)
```

If it handles that can you add explicit tests for that?

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
```diff
@@ -16,7 +16,7 @@ import org.apache.commons.codec.binary.Base64
 import org.joda.time.DateTime
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.convert.Transformers
-import org.locationtech.geomesa.convert.Transformers.EvaluationContext
+import org.locationtech.geomesa.convert.Transformers.{Predicate, EvaluationContext}
```
jahhulbert-ccri: optimize imports

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
```diff
@@ -274,7 +274,160 @@ class TransformersTest extends Specification {
...
+            exp.eval(Array("", null, "false")) mustEqual false
+          }
+          "boolean string2boolean null default" >> {
+            val exp = Transformers.parseTransform("string2boolean($1,$2)")
+            exp.eval(Array("", "true", null)) mustEqual true
+            exp.eval(Array("", "false", null)) mustEqual false
+            exp.eval(Array("", "", null)) mustEqual null
+            exp.eval(Array("", null, null)) mustEqual null
+          }
+        }
```
jahhulbert-ccri: lets add some unit tests tests with full converter config to see what the CSV/json/etc data would look like for these cases...but see above comment about type casting first

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,43 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
...
+    }
+  }
+  val string2int = TransformerFn("string2int") {
+    args => {
+      if (args(0) == null || args(0).asInstanceOf[String] == "") {
+        if (args(1) == null || args(1).asInstanceOf[String] == "") args(1) else args(1).asInstanceOf[String].toInt
+      } else args(0).asInstanceOf[String].toInt
+    }
+  }
+  val string2float = TransformerFn("string2float") {
```
jahhulbert-ccri: another thought... @elahrvivaz @anthonyccri might chime in...I'd almost name this short form using ```str``` instead of ```string``` but maybe thats too short.

I feel very strongly that we need to do a final design review/follow tickets for all the convert stuff to standardize/review functions, syntax, etc before the rc7 release.

elahrvivaz: +1 for overall review. I'm not really in favor of names like '2whatever' - I would advocate for fully spelled out camel-case names for consistency. E.g. ```stringLength``` (or just ```length```) instead of ```strlen```, etc. We could always keep abbreviated synonyms around too.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
elahrvivaz: We decided to make the conversions functions, so that you have to be explicit about the default value. It was a lot easier to have a function with 2 args than try to change the regex parsers to add a default to the :: syntax. We left the original conversions the same, so they can be used in the case where there is no appropriate default and you want the record to fail.
We could obviously do something else, but that was our thinking.

geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.locationtech.geomesa.convert.TransformerFunctionFactory
drackaer: Nothing should have changed on line 6, not sure what is going on there.  This is a (text file?) list of factories and not imports, so not sure if there is any cleanup I can do here.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
cne1x: Thank you, @elahrvivaz !

drackaer: I am with @elahrvivaz on style, I usually prefer longer english names over abbreviations, but I was trying to match the style already used by Transformers.scala.  I will hold off on any changes here until that overall review happens.

elahrvivaz: @drackaer we can go ahead and get this PR merged in, then do a holistic review of everything. So yeah, don't worry about the names for now.

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
elahrvivaz: @drackaer you'd have to put the tests in e.g. geomesa-convert-text, there are examples there you can copy/extend.

elahrvivaz: need to make sure it works from the config, e.g. ```{ name = myInt, transform = "string2int($1, 0::int)" }``` vs calling it with an already parsed array.

drackaer: Alright, I will check that out

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
```diff
@@ -274,7 +274,160 @@ class TransformersTest extends Specification {
...
+          }
+          "double null string string2double null default" >> {
+            val exp = Transformers.parsePred("doubleEq(string2double($1,$3), string2double($2,$3))")
+            exp.eval(Array("", null, "1.0", null)) must beFalse
+            exp.eval(Array("", null, null, null)) must beTrue
+          }
+        }
+        "string2int" >> {
+          "int string2int zero default" >> {
+            val exp: Predicate = Transformers.parsePred("intEq(string2int($1, $3), string2int($2, $3))")
```
elahrvivaz: It might be better to just evaluate the string2int for a couple different inputs, and not use the intEq. That way you're checking the direct output. Would also require less tests.
e.g. for string to int, really you just need: "string2int($1, '0'::int)" for valid value, null string, empty string
Ditto other tests.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -482,5 +481,35 @@ class MapListFunctionFactory extends TransformerFunctionFactory {
...
+    }
+  }
+  val string2int = TransformerFn("string2int") {
+    args => {
+      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toInt
+    }
+  }
+  val string2float = TransformerFn("string2float") {
+    args => {
+      if (isEmptyOrNull(args(0))) { args(1) } else args(0).asInstanceOf[String].toFloat
```
elahrvivaz: I guess another issue is what if the string has something, but it's not a valid e.g. float? We should probably wrap the else statement in a try/catch, and return the default if it fails.

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
drackaer: Agreed, it was part of a learning process (where I wasn't sure what was going on at first) and didn't go back and fix it after I understood what was going on.  Reducing the number of tests now.

elahrvivaz:
@drackaer looks good

geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.locationtech.geomesa.convert.TransformerFunctionFactory
jahhulbert-ccri: :) sorry bout that one!

jahhulbert-ccri:
looks good...picky bracket align here: https://github.com/drackaer/geomesa/blob/geomesa_transformers_robust/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala#L59

@elahrvivaz @drackaer I'd also be cool with adding ```string2integer``` in addition to ```string2int```. in recent geomesa training folks were confused by having a mix of ```int``` and ```integer``` so it may be useful for a follow-on ticket to always support both int and integer.

drackaer:
@jahhulbert-ccri
Bracket has been given a talking to and brought back into line.
The whole int thing was something I noticed as well, since the other type names were written out (i.e. string vs. str, boolean vs. bool).  I can easily add an the 2integer function.
Do we have any idea on timing for that style review?

elahrvivaz:
@drackaer go ahead and add the string2integer. I think once that's done we can merge this, then go back and do a holistic overview.

elahrvivaz:
c3c6d9ed233f23d474f743a93cd719f4d4090c3b

PR #739, opened 2015-12-15
GEOMESA-960, GEOMESA 1011 Added stat iterator

Afterglow375:
* Removed TemporalDensityIterator
* Added Stats Iterator
* Added MinMax, IteratorStackCounter, EnumeratedHistogram, and RangeHistogram statistics


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -0,0 +1,219 @@
...
+        null
+      }
+    }
+
+    override def getBinKey(binSize: Date, bucketIndex: Int, lowerEndpoint: Date): Date = {
+      new Date(lowerEndpoint.getTime + (binSize.getTime * bucketIndex))
+    }
+  }
+
+  implicit object BinAbleLong extends BinAble[lang.Long] {
```
Afterglow375: Does not seem ideal to have type classes for individual numeric types whose implementations are the same. Wasn't sure how to make a type class for all numeric types though.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -0,0 +1,173 @@
...
+      "IteratorStackCounter" ^^ { case _ => IteratorStackCounter() }
+    }
+
+    def enumeratedHistogramParser: Parser[EnumeratedHistogram[_]] = {
+      "EnumeratedHistogram(" ~> attributeNameRegex <~ ")" ^^ {
+        case attribute =>
+          val attrIndex = getAttrIndex(attribute)
+          val attrType = sft.getType(attribute).getBinding
+          val attrTypeString = attrType.getName
+          attrType match {
```
Afterglow375: Need to restrict types here to known types for when deserializing the stat from a String. Without keeping track of the type, I don't see how it's possible to go from the String-serialized stat to the stat.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/StatsIterator.scala
```diff
@@ -0,0 +1,145 @@
...
+        namespace
+      }
+    SimpleFeatureTypes.createType(outNamespace, name, StatsIterator.STATS_ITERATOR_SFT_STRING)
+  }
+
+  def combineStats(stat1: Stat, stat2: Stat): Stat = {
+    stat1.add(stat2)
+  }
+
+  def encodeStat(stat: Stat): String = {
```
Afterglow375: This is a holdover from the TemporalDensityIterator that I maintained, but do not really understand the purpose of. There were two ways to serialize the TemporalDensityIterator as a String:

1) Encode to byte array and stringify
2) Encode to JSON

I suppose #1 is faster and cheaper in terms of memory, but was that the only reason two methods of serializing to a string were used?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -0,0 +1,219 @@
...
+
+import org.locationtech.geomesa.utils.stats.BinHelper._
+
+/**
+ * The range histogram's state is stored in a hashmap, where the keys are the bins and the values are the counts.
+ * A bin is a range of values but is treated as the lower endpoint of the bin to persist in the hashmap.
+ * A value is in a particular bin if it is equal to the bin value or less than the subsequent bin value.
+ *
+ * e.g. a range of 0 to 3 with 3 bins will result in these bins: [0, 1), [1, 2), [2, 3) and the hashmap will contain
+ * keys 0, 1, and 2.
```
Afterglow375: Had some trouble thinking through dividing a range of discrete values into approximately equal sub-ranges. I settled on leaving it to the user to use sensible values for the number of bins and lower/upper endpoints of the histogram range. If the number of bins doesn't evenly divide the range, the last bin will be larger than the rest.

e.g. 7 bins, range of ints from 0 to 10, will result in a bin size of 1, where the bins will be: [0, 1), [1, 2), [2, 3), [3, 4), [4, 5), [5, 6), [7, 10)

This isn't an issue for doubles/floats.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
```diff
@@ -376,12 +379,13 @@ object QueryPlanner extends Logging {
...
       BinAggregatingIterator.BIN_SFT
     } else if (query.getHints.isDensityQuery) {
-      Z3DensityIterator.DENSITY_SFT
-    } else if (query.getHints.containsKey(TEMPORAL_DENSITY_KEY)) {
+      KryoLazyDensityIterator.DENSITY_SFT
+    } else if (query.getHints.isStatsIteratorQuery) {
+      KryoLazyStatsIterator.createFeatureType(baseSft)
+    } else if (query.getHints.isTemporalDensityQuery) {
       TemporalDensityIterator.createFeatureType(baseSft)
     } else if (query.getHints.containsKey(MAP_AGGREGATION_KEY)) {
```
elahrvivaz: Should use query.getHints.isMapAggregatingQuery - I had changed it in the base PR.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/index.scala
```diff
@@ -40,29 +40,31 @@ package object index {
...
+    val QUERY_STRATEGY_KEY            = new ClassKey(classOf[StrategyType])
 
-    val DENSITY_BBOX_KEY     = new ClassKey(classOf[ReferencedEnvelope])
-    val DENSITY_WEIGHT       = new ClassKey(classOf[java.lang.String])
-    val WIDTH_KEY            = new IntegerKey(256)
-    val HEIGHT_KEY           = new IntegerKey(256)
+    val DENSITY_BBOX_KEY              = new ClassKey(classOf[ReferencedEnvelope])
+    val DENSITY_WEIGHT                = new ClassKey(classOf[java.lang.String])
+    val WIDTH_KEY                     = new IntegerKey(256)
+    val HEIGHT_KEY                    = new IntegerKey(256)
```
elahrvivaz: please fix the alignment/whitespace.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/stats/StatsIteratorProcess.scala
```diff
@@ -0,0 +1,81 @@
...
+                 name = "statString",
+                 description = "The string indicating what stats to instantiate")
+               statString: String
+
+               ): SimpleFeatureCollection = {
+
+    logger.debug("Attempting Geomesa stats iterator process on type " + features.getClass.getName)
+
+    if (features.isInstanceOf[ReTypingFeatureCollection]) {
+      logger.warn("WARNING: layer name in geoserver must match feature type name in geomesa")
```
elahrvivaz: should this be a failure condition? or it's just a lot slower?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/stats/StatsIteratorProcess.scala
```diff
@@ -0,0 +1,81 @@
...
+    val visitor = new StatsVisitor(features, statString)
+    features.accepts(visitor, new NullProgressListener)
+    visitor.getResult.asInstanceOf[StatsIteratorResult].results
+  }
+}
+
+class StatsVisitor(features: SimpleFeatureCollection, statString: String)
+  extends FeatureCalc with Logging {
+
+  val retType = TemporalDensityIterator.createFeatureType(features.getSchema)
```
elahrvivaz: Should this be StatsIterator.createFeatureType?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/stats/QueryStat.scala
```diff
@@ -124,17 +122,15 @@ object QueryStatTransform extends StatTransform[QueryStat] {
...
-      case BIN_TRACK_KEY        => "BIN_TRACK_KEY"
-      case TEMPORAL_DENSITY_KEY => "TEMPORAL_DENSITY_KEY"
-      case TIME_INTERVAL_KEY    => "TIME_INTERVAL_KEY"
-      case RETURN_ENCODED       => "RETURN_ENCODED"
-      case TIME_BUCKETS_KEY     => "TIME_BUCKETS_KEY"
-      case DENSITY_BBOX_KEY     => "DENSITY_BBOX_KEY"
-      case WIDTH_KEY            => "WIDTH_KEY"
-      case HEIGHT_KEY           => "HEIGHT_KEY"
-      case _                    => "unknown_hint"
+      case TRANSFORMS                       => "TRANSFORMS"
```
elahrvivaz: please fix whitespace

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
```diff
@@ -0,0 +1,65 @@
...
+class EnumeratedHistogram[T](val attrIndex: Int,
+                             val attrType: String) extends Stat {
+  val frequencyMap: mutable.Map[T, Long] = new mutable.HashMap[T, Long]().withDefaultValue(0)
+
+  override def observe(sf: SimpleFeature): Unit = {
+    val sfval = sf.getAttribute(attrIndex)
+    if (sfval != null) {
+      sfval match {
+        case tval: T =>
+          frequencyMap(tval) += 1
```
elahrvivaz: Maybe do ```frequencyMap(sfval.asInstanceOf[T]) += 1``` instead? Seems like that would be faster than a match, and since the match doesn't have any other cases, it is functionally equivalent.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  }
+
+  override def add(other: Stat): Stat = {
+    other match {
```
elahrvivaz: same thing here with .asInstanceOf vs match. (probably applies in the other stat impls as well)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumeratedHistogram.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  override def toJson(): String = {
+    val jsonMap = frequencyMap.toMap.map { case (k, v) => k.toString -> v }
+    new JSONObject(jsonMap).toString()
+  }
+
+  override def clear(): Unit = frequencyMap.clear()
+
+  override def equals(obj: Any): Boolean = {
+    obj.isInstanceOf[EnumeratedHistogram[T]] && {
```
elahrvivaz: [style] this might be a good case for a match, since you're doing a check and cast. It's a little cleaner like:
obj match {
  case eh: EnumeratedHistogram => // rest of comparison
  case _ => false
}

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/IteratorStackCounter.scala
```diff
@@ -0,0 +1,39 @@
...
+  override def observe(sf: SimpleFeature): Unit = { }
+
+  override def add(other: Stat): Stat = {
+    other match {
+      case o: IteratorStackCounter => count += o.count
+    }
+    this
+  }
+
+  override def toJson(): String = s"""{ "count": $count }"""
```
elahrvivaz: would we want some indication of the stat type in the json?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/IteratorStackCounter.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  override def clear(): Unit = count = 1L
```
elahrvivaz: wouldn't the count default to 0? although maybe since we only would hit this in an iterator, there is always 1?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
```diff
@@ -0,0 +1,115 @@
...
+      case mm: MinMax[T] =>
+        updateMin(mm.min)
+        updateMax(mm.max)
+    }
+
+    this
+  }
+
+  private def updateMin(sfval: T): Unit = {
+    if (min.compareTo(sfval) > 0) {
```
elahrvivaz: so being able to compare without a null check is nice, but it seems like we need some way to return an 'empty' min/max - otherwise if you don't get any values you'd get *all* values. is that handled anywhere?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -0,0 +1,238 @@
...
+    override def getBinSize(numBins: Int, lowerEndpoint: Date, upperEndpoint: Date): Date = {
+      new Date((upperEndpoint.getTime - lowerEndpoint.getTime) / numBins)
+    }
+
+    override def getBinIndex(attributeValue: Date,
+                             binSize: Date,
+                             numBins: Int,
+                             lowerEndpoint: Date,
+                             upperEndpoint: Date): Date = {
+      if (attributeValue.getTime >= lowerEndpoint.getTime && attributeValue.getTime < upperEndpoint.getTime) {
```
elahrvivaz: [style] dates have a .before and .after that would be a little cleaner than comparing times.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: hmm, scala has Numeric, but I'm not sure if that would work for java types - you could try though.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: Could you convert the e.g. ints into doubles for binning, then when unbinning convert back to ints and sum collisions?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -0,0 +1,197 @@
...
+   */
+  def clear(): Unit
+
+  /**
+   * Equals method to simplify testing
+   *
+   * @param obj object to compare
+   * @return true if equal to obj
+   */
+  override def equals(obj: Any): Boolean
```
elahrvivaz: wouldn't this be part of any class anyway without explicitly declaring it here?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -0,0 +1,197 @@
...
+ * This class contains parsers which dictate how to instantiate a particular Stat.
+ * Stats are created by passing a stats string as a query hint (QueryHints.STATS_STRING).
+ *
+ * A valid stats string should adhere to the parsers here:
+ * e.g. "MinMax(attributeName);IteratorCount" or "RangeHistogram(attributeName,10,0,100)"
+ * (see tests for more use cases)
+ */
+object Stat {
+  class StatParser(sft: SimpleFeatureType) extends RegexParsers {
+    val attributeNameRegex = """\w+""".r
```
elahrvivaz: this regex might need to be looser - I can't find the restrictions now, but i think dashes at least are allowed in attribute names, might be others too. @jahhulbert-ccri @cne1x do you happen to know?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
elahrvivaz: maybe encode the names like we do in the sft spec? e.g. 'foo:Integer'

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    def minMaxParser: Parser[MinMax[_]] = {
+      "MinMax(" ~> attributeNameRegex <~ ")" ^^ {
+        case attribute =>
+          val attrIndex = getAttrIndex(attribute)
+          val attrType = sft.getType(attribute).getBinding
+          val attrTypeString = attrType.getName
+          attrType match {
+            case _ if attrType == classOf[Date] =>
```
elahrvivaz: [style] this might read a little cleaner as an if() else()

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatHelpers.scala
```diff
@@ -0,0 +1,10 @@
+package org.locationtech.geomesa.utils.stats
+
+import java.text.SimpleDateFormat
+
+import org.joda.time.format.DateTimeFormat
+
+object StatHelpers {
+  val javaDateFormat = new SimpleDateFormat("EEE MMM dd HH:mm:ss z yyyy")
```
elahrvivaz: you probably don't want to have this in an object - java data formats aren't thread safe. (joda ones are, so the below line is ok)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerialization.scala
```diff
@@ -0,0 +1,265 @@
...
+    require(split.size == 4)
+
+    val attrIndex = split(0).toInt
+    val attrTypeString = split(1)
+    val min = split(2)
+    val max = split(3)
+
+    val attrType = Class.forName(attrTypeString)
+    attrType match {
+      case _ if attrType == classOf[Date] =>
```
elahrvivaz: [style] same as noted above, in general you don't want to use a match if you're not actually extracting anything. (sometimes it can look cleaner, but not really in this case...)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerialization.scala
```diff
@@ -0,0 +1,265 @@
...
+    val keyValues = split(2).split(",")
+
+    val attrType = Class.forName(attrTypeString)
+    attrType match {
+      case _ if attrType == classOf[Date] =>
+        val eh = new EnumeratedHistogram[Date](attrIndex, attrTypeString)
+        keyValues.foreach {
+          case (keyValuePair) =>
+            val splitKeyValuePair = keyValuePair.split("->")
+            eh.frequencyMap.put(StatHelpers.javaDateFormat.parse(splitKeyValuePair(0)), splitKeyValuePair(1).toLong)
```
elahrvivaz: as noted above, not thread safe.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerialization.scala
```diff
@@ -0,0 +1,265 @@
...
+          returnStats += stat
+        case EH_BYTE =>
+          val stat = unpackEnumeratedHistogram(statBytes)
+          returnStats += stat
+        case RH_BYTE =>
+          val stat = unpackRangeHistogram(statBytes)
+          returnStats += stat
+      }
+
+      bytePointer += statSize + 5
```
elahrvivaz: [style] might be a little cleaner to use the native byte buffer position, length, etc methods and avoid manually tracking the byte pointer

elahrvivaz:
@Afterglow375 Looks good. I had mainly stylistic comments. I'll see if I can refactor the iterator setup a bit...

elahrvivaz:
So seems like the temporal density iterator is still kind of partially around? We should probably just remove it completely...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
Afterglow375: Tried doing that, but you still end up with uneven bins if the bin key themselves are ints. e.g. in the case of the 7 bins from 0 to 10, most bins will be of size 1, while some will be size 2. Are you suggesting just making all bin keys doubles?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
Afterglow375: Not sure what you mean here. Encode the Stat's type within the SFT? The issue is highlighted here (where I need conditional logic to go from String values to whatever type the Stat is):
https://github.com/Afterglow375/geomesa/blob/fcr_stats_iterator/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerialization.scala#L106

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/stats/StatsIteratorProcess.scala
Afterglow375: Don't really know. Copied that in from the TemporalDensityProcess.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
elahrvivaz: I just meant instead of ```MinMax("foo")``` use ```MinMax("foo:String")```. You'd still have to convert from a string to a class, but you wouldn't need the sft for it.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: Yeah, the bins can't always be the same size if they're discrete values. don't think there's any way around that. But is that a problem? Seems like it would be better to have some size 1, some size 2, than to have all size 1 except one of size 3.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
Afterglow375: Ah true, though that requires more effort on part of the user. Not sure which would be ideal.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/IteratorStackCounter.scala
Afterglow375: Perhaps. I don't know how the stats will be used in actual use cases.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
Afterglow375: I don't think so. Not too clear on what you mean.

elahrvivaz: I mean AnyRef (or Object) has an equals method already defined - so there isn't any point I can see to defining it here unless you're actually implementing it.

Afterglow375: Ah right, misread your comment. Yeah, I put the signature for equals in the trait in an attempt to force stats to have to implement equals, but that didn't work.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
Afterglow375: Handled in the serialization/deserialization. An "empty" MinMax will have its min and max as "null".

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
Afterglow375: But then there's no real control over which bins are larger or smaller. Jim suggested allowing the stat parser DSL to accept all the bins a user would explicitly provide. e.g. if a user wants 3 bins from 0-18, 18-30, and 30-45, they would convey that information in the query hint to obtain that.

I'm of the opinion this change should be accomplished as smaller ticket in the future.

elahrvivaz:
I've rebased this, but can't push to @Afterglow375's repo, so opening new PR #816

PR #740, opened 2015-12-15
GEOMESA-1010 Required intermediate variables are not always calculated in input converters

elahrvivaz:


elahrvivaz:
I removed the checks for only calculating the required fields - trying to trace dependencies was complicated. In practice, I would assume that people would not put extraneous fields in their converters...

anthonyccri:
This feature was built to support query projections in file based data stores.  Not sure we can

anthonyccri:
db8b083..ee41c8f

PR #743, opened 2015-12-18
GEOMESA-1013 Converter library cleanup and standardization

elahrvivaz:
Note: stacks on top of #740, please consider only the subsequent commits.

* Standardized converter function names
* Added converter support for native ints, longs, booleans, floats, doubles
* Added converter support for json longs and booleans
* Simplified SimpleFeatureConverter interface
* Improved erorr logging for converters
* Updated converter docs

Signed-off-by: Emilio Lahr-Vivaz elahrvivaz@ccri.com

PR #744, opened 2015-12-21
GEOMESA-977,GEOMESA-1002 Discoverable simple feature types and converters

elahrvivaz:
Note: stacks on top of #743 - please consider only last commit(s)

* Ingest command will now accept named values for simple feature types and converters that will be loaded from the classpath
* Simple feature types can be added by placing an application.conf file on the classpath with the types under 'geomesa.sfts' as an array
* Simple feature type converters can be added by placing an application.conf file on the classpath with the converters under 'geomesa.co
* Simple feature type converters should add a 'name' attribute in order to be identifiable.
* New command line method 'env' will list available simple feature types and converters.

### DIFF
--- a/geomesa-web/geomesa-web-data/pom.xml
+++ b/geomesa-web/geomesa-web-data/pom.xml
@@ -80,7 +80,7 @@
             <dependency>
                 <groupId>com.beust</groupId>
                 <artifactId>jcommander</artifactId>
-                <version>1.32</version>
+                <version>1.48</version>
                 <scope>provided</scope>
             </dependency>
             <dependency>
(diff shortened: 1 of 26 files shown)


geomesa-web/geomesa-web-data/pom.xml
```diff
@@ -80,7 +80,7 @@
             <dependency>
                 <groupId>com.beust</groupId>
                 <artifactId>jcommander</artifactId>
-                <version>1.32</version>
+                <version>1.48</version>
```
elahrvivaz: note: 1.48 is already approved by locationtech: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10140

elahrvivaz:
9d6d22122c529c0c4941bd2758d672408c68f235

PR #750, opened 2015-12-24
GEOMESA-1019 Updating scala to 2.11.7

elahrvivaz:
Note: stacks on top of #749

* Updating typesafe scala logging
* Removing command-line scalding repl (not yet available in scalding with scala 2.11)

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #751, opened 2015-12-24
GEOMESA-1020 Removing all geoserver code for IP

elahrvivaz:
Note: stacks on top of #750.

* Code will be ported to a geoserver-specific project with a difference license


anthonyccri:
1d1d414..622c8f0

PR #752, opened 2016-01-08
GEOMESA-1022 Added scala xml to lib, example xml ingest

jahhulbert-ccri:
Other things:
* renamed tools tar.gz to include -tools so it won't untar to same dir as the dist jar will...
* changed file permissions to not be x on jars in lib
* included example application.conf in conf dir


elahrvivaz:
this looks good - merging.

elahrvivaz:
57df1c1b5766d2a414fe5948caaea4ed0578d3d2

PR #760, opened 2016-01-13
GEOMESA-957,959,974,985 BlobStore improvements

AndrewAnnex:
* first pass at README for blobstore
* implemented deletes
* got other RESTful routes working in blobstore
* synchronized blobstore creation


AndrewAnnex:
note: I am not sure I am correctly removing the geotools feature from accumulo yet...

geomesa-blobstore/README.md
```diff
@@ -17,3 +17,40 @@ The class `WKTFileHandler` serves as an example of that latter approach.
 
 `FileHandler` implementations are loaded at runtime by the Java ServiceLoader.  As such, available 
 implementations should have their classname registered in `META-INF/services/org.locationtech.geomesa.blob.core.handlers.FileHandler`.
+
+## REST api
+
+The RESTful api for the Blobstore can be easily utilized via cURL as so:
+
+Before we can do anything we must first register a blob store. This typically only needs to be run once in the lifetime
+of the webserver, so if the server is shutdown or restarted the registration call will need to be repeated.
```
jahhulbert-ccri: confused...if it only needs to be run once why do we need to repeat it?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,43 +8,53 @@
...
-  var abs: AccumuloBlobStore = null
-
-  // TODO: Revisit configuration and persistence of configuration.
-  // https://geomesa.atlassian.net/browse/GEOMESA-958
-  // https://geomesa.atlassian.net/browse/GEOMESA-984
-  post("/ds") {
+  // caps blob size at 10MB
+  configureMultipartHandling(
+    MultipartConfig(
+      maxFileSize = Some(50*1024*1024),
```
jahhulbert-ccri: ideally this would be configurable in the UI for geoserver...or at least a system prop to override it

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,43 +8,53 @@
...
+      maxRequestSize = Some(100*1024*1024)
+    )
+  )
+  error {
+    case e: SizeConstraintExceededException => RequestEntityTooLarge("Uploaded file too large!")
+    case e: IOException => halt(500, "IO Exception on server")
```
jahhulbert-ccri: make sure that this logs the error + stacktrace! else add logger.error()

geomesa-blobstore/README.md
AndrewAnnex: hadn't gotten to it yet, basically it is only persistent so long as the webserver is running. ie, if i call it and then 5 minutes later kill my tomcat or jetty, then you need to call it again. I think there is a ticket/ pr for something doing something else, unless I was reading too much into it in which case it depends on how you define 'persistent'

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: can do

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: oh I need to update that comment there... maybe I could also expose those properties via PUTs?

elahrvivaz: don't expose them via api... kind of defeats the whole purpose.

geomesa-blobstore/README.md
elahrvivaz: There is a ticket for it. The slight problem is that ideally it would use the geoserver_data_dir, which could introduce license issues. However, the solution has already been implemented in e.g. geomesa/geomesa-web/geomesa-web-data/src/main/resources/applicationContext.xml.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,43 +8,60 @@
...
+      maxRequestSize = Some(maxRequestSize * 1024 * 1024)
+    )
+  )
+  error {
+    case e: SizeConstraintExceededException =>
+      logger.error("IO Exception in BlobstoreServlet, Error message: {} \n Stacktrace: {}", e.getMessage, e.getStackTrace)
+      RequestEntityTooLarge("Uploaded file too large!")
+    case e: IOException =>
+      logger.error("IO Exception in BlobstoreServlet, Error message: {} \n Stacktrace: {}", e.getMessage, e.getStackTrace)
+      halt(500, s"IO Exception on server: ${e.getMessage}")
```
elahrvivaz: should use GeoMesaScalatrServlet.handleError - common place to handle exposing internal messages. Although that doesn't use halt...

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,43 +8,60 @@
...
-    val ds = new AccumuloDataStoreFactory().createDataStore(dsParams).asInstanceOf[AccumuloDataStore]
+  val tempDir = Files.createTempDirectory("blobs", BlobstoreServlet.permissions)
 
-    if (ds == null) {
-      NotFound(reason = "Could not load data store using the provided parameters.")
+  delete("/:id/?") {
+    val id = params("id")
+    logger.debug("In DELETE method for blobstore, attempting to delete: {}", id)
+    if (abs == null) {
+      NotFound(reason = "AccumuloBlobStore is not initialized.")
```
elahrvivaz: maybe BadRequest instead of NotFound?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,43 +8,60 @@
...
     } else {
-      // TODO: Synchronize Blobstore creation
-      // https://geomesa.atlassian.net/browse/GEOMESA-985
-      abs = new AccumuloBlobStore(ds)
-      Ok()
+      abs.delete(id)
+      NoContent(reason = s"deleted feature: $id")
```
elahrvivaz: Ok instead of NoContent?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -53,10 +70,68 @@ class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
...
+  // https://geomesa.atlassian.net/browse/GEOMESA-958
+  // https://geomesa.atlassian.net/browse/GEOMESA-984
+  post("/ds/?") {
+    logger.debug("In ds registration method")
+
+    val dsParams = datastoreParams
+    val ds = new AccumuloDataStoreFactory().createDataStore(dsParams).asInstanceOf[AccumuloDataStore]
+
+    if (ds == null) {
+      NotFound(reason = "Could not load data store using the provided parameters.")
```
elahrvivaz: BadRequest instead?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -53,10 +70,68 @@ class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
...
+    } else {
+      this.synchronized {
+        abs = new AccumuloBlobStore(ds)
```
elahrvivaz: in order for this synchronize block to make a difference, you would have to synchronize every access to the accumulo blob store in this class...

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -53,10 +70,68 @@ class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
...
+                UnprocessableEntity(reason = s"Unable to process file: ${file.name}")
+            }
+            tempFile.delete()
+            actRes
+        }
+      }
+    } catch {
+      case NonFatal(ex) =>
+        logger.error("Error uploading file", ex)
+        UnprocessableEntity(reason = ex.getMessage)
```
elahrvivaz: should use handleError, as above. The general idea is returning messages gives an attacker information about the system.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -53,10 +70,68 @@ class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
...
+  // scalatra routes bottom up, so we want the ds post to be checked first
+  post("/") {
+    try {
+      logger.debug("In file upload post method")
+      if (abs == null) {
+        NotFound(reason = "AccumuloBlobStore is not initialized.")
+      } else {
+        fileParams.get("file") match {
+          case None =>
+            halt(400, reason = "no file parameter in request")
```
elahrvivaz: BadRequest? halt seems kind of like return... should we be avoiding that?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -53,10 +70,68 @@ class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
...
+          case Some(file) =>
+            val otherParams = multiParams.toMap.map { case (s, p) => s -> p.head }
+            val tempFile = new File(tempDir.toString + '/' + file.getName)
```
elahrvivaz: chance of collisions here... should probably use a random string in there somewhere.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -53,10 +70,68 @@ class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
...
+            file.write(tempFile)
+            val actRes: ActionResult = abs.put(tempFile, otherParams) match {
+              case Some(id) =>
+                Created(body = id, headers = Map("Location" -> s"${request.getRequestURL append id}"))
```
elahrvivaz: i think we discourage the 'space' notation, e.g. should be request.getRequestURL.append(id)

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -78,6 +79,32 @@ class AccumuloBlobStore(ds: AccumuloDataStore) {
...
+    // https://geomesa.atlassian.net/browse/GEOMESA-986
+    val bd = connector.createBatchDeleter(blobTableName, new Authorizations(), bwc.getMaxWriteThreads, bwc)
+    bd.setRanges(List(new Range(new Text(id))))
+    bd.delete()
+    bd.close()
+    deleteFeature(id)
+  }
+
+  private def deleteFeature(id: String): Unit = {
+    val transaction = new DefaultTransaction("removeBlobFeature")
```
elahrvivaz: This is an accumuloFeatureStore, right? we don't support transactions... can just use Transaction.AUTO_COMMIT

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: what

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: we need to preserve the file name to return, I wanted to avoid changing the accumuloblobstore as much as possible but I can look into it.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: I think I followed a unit test in accumulofeaturewriter, would switching to autocommit make this just work?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: I could just make it an atomic reference: https://twitter.github.io/scala_school/concurrency.html

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: I don't think it is a true BadRequest, as it is more a state issue, less a client error (unless you think by not init'ing the datastore the client made an error), see; https://httpstatuses.com/400

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: didn't know about that part of the trait, sounds good to me

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: oh.. guess I get it

jahhulbert-ccri: agreed...make sure no stack traces ever go back to the client

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jahhulbert-ccri: and catch fire?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: note: I am going to re-write this so that I am not writing out temp files.

geomesa-blobstore/README.md
AndrewAnnex: ok I get it now, hmm yeah I might need to take down the pr to chew on it for a while as we probably want the layer to be published in geoserver

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: Actually, using the file name as provided is probably a security risk itself... but maybe it's acceptable if we document it.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: @jahhulbert-ccri @elahrvivaz can you guys write up some best practices and put them somewhere? Also in my next commit can you both be more explicit in what changes you would like to see, at the moment I feel that I am guessing.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: perhaps I am wrong, but I would expect as a user to get out what I put in.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: done

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: Here's a brief overview of some of the security risks:
http://security.stackexchange.com/questions/503/what-are-the-security-risks-of-letting-the-users-upload-content-to-my-site
In addition to that, if you e.g. knew that someone was uploading a file named 'myreport', you could maliciously replace that file by uploading something with the same name. You have to remember that web services are multi-threaded and serve concurrent users. Thus, ensuring users can't impact each other is important.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: why? is it a status code thing?

elahrvivaz: dunno, not found seems un-intuitive...

geomesa-blobstore/README.md
AndrewAnnex: I have pushed changes that allow the data store connection parameters to be stored in the geoserver data dir like some of the other geomesa-web code.

jahhulbert-ccri:
@AndrewAnnex wondering if maybe we could replace this string https://github.com/ccri/geomesa/blob/fcr_blobstore_rest_improvements/geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala#L135 with an SftBuilder in case we change it in the future.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,107 @@
...
+  // scalatra routes bottom up, so we want the ds post to be checked first
+  post("/") {
+    try {
+      logger.debug("In file upload post method")
+      if (abs == null) {
+        handleError("AccumuloBlobStore is not initialized.", null)
+      } else {
+        fileParams.get("file") match {
+          case None =>
+            handleError("no file parameter in request", null)
```
elahrvivaz: did you verify passing null is ok here? i haven't tried that...

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,107 @@
...
+      handleError("Uploaded file too large!", e)
+    case e: IOException =>
+      handleError("IO exception in BlobstoreServlet", e)
+  }
 
-    if (ds == null) {
-      NotFound(reason = "Could not load data store using the provided parameters.")
+  delete("/:id/?") {
+    val id = params("id")
+    logger.debug("In DELETE method for blobstore, attempting to delete: {}", id)
```
jahhulbert-ccri: picky can we just redo the comments to say "Attemping to delete: {}" and drop the "In DELETE method" part...same for other "In ID method" etc.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
```diff
@@ -0,0 +1,84 @@
...
+    * Registers a data store, making it available for later use
+    */
+  post("/ds/:alias") {
+    logger.debug("In ds registration method")
+    val dsParams = datastoreParams
+    val ds = new AccumuloDataStoreFactory().createDataStore(dsParams).asInstanceOf[AccumuloDataStore]
+    if (ds == null) {
+      BadRequest(reason = "Could not load data store using the provided parameters.")
+    } else {
+      abs = new AccumuloBlobStore(ds)
```
elahrvivaz: should keep the todos about synchronizing access to the blob store

elahrvivaz: or synchronize it...

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -78,6 +80,29 @@ class AccumuloBlobStore(ds: AccumuloDataStore) {
...
+  }
+
+  private def deleteFeature(id: String): Unit = {
+    val removalFilter = Filters.ff.id(Filters.ff.featureId(id))
+    val fd = ds.getFeatureWriter(blobFeatureTypeName, removalFilter, Transaction.AUTO_COMMIT)
+    try {
+      fd.remove()
+    } catch {
+      case e: Exception =>
+        logger.error(e.getMessage)
```
elahrvivaz: it would be better to put a meaningful message and pass the exception like:
`logger.error("Couldn't remove feature from blobstore", e)`

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,107 @@
...
+            handleError("no file parameter in request", null)
+          case Some(file) =>
+            val otherParams: mutable.Map[String, String] = mutable.Map()
+            multiParams.foreach{case (s, p) => otherParams.add(s, p.head)}
+            if (!otherParams.contains(filenameFieldName)) {
+              // we put the true filename in here so that we can preserve it in the blob table
+              otherParams.put(filenameFieldName, file.getName)
+            }
+            val tempFile = File.createTempFile(UUID.randomUUID().toString, FilenameUtils.getExtension(file.getName))
+            file.write(tempFile)
```
jahhulbert-ccri: perhaps change to 
```
val tempFile = ...
try {
  // other stuff
} finally {
  tempFile.delete()
}
```

geomesa-blobstore/geomesa-blobstore-core/src/test/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStoreTest.scala
```diff
@@ -38,10 +38,29 @@ class AccumuloBlobStoreTest extends Specification {
 
   val testfile1 = "testFile.txt"
   val testfile2 = "testFile2.txt"
+  val testfile3 = "testFile3.txt"
   var testFile1Id = ""
 
   "AccumuloBlobStore" should {
-    "be able able to store and retrieve a file" in {
+
+    "be able to store and delete a file" in pending("Deleting from accumulodatastore may not be working in blobstore") {
```
elahrvivaz: not familiar with 'in pending' - does that mean it doesn't work? that's the feature you just added, right?

jahhulbert-ccri: deleting fails...debugger shoes that the modify feature writer is pulling out a feature with the feature id of ```1``` which probably indicates a user data problem or something like that...

jahhulbert-ccri: should make this test work before merge

jahhulbert-ccri: hit this with the debug:

https://github.com/ccri/geomesa/blob/fcr_blobstore_rest_improvements/geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala#L51

call ```sf.getFeatureID```

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: oh, no idea. I need to check that.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: sure

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
AndrewAnnex: the blob store is now a `@volatile` should be synched

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: can do

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: like it

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
elahrvivaz: huh, I didn't realize volatile implied synchronization. nice.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.locationtech.geomesa.accumulo.data.{AccumuloDataStore, AccumuloDataStoreFactory}
+import org.locationtech.geomesa.blob.core.AccumuloBlobStore
+import org.locationtech.geomesa.web.core.GeoMesaBaseDataStoreServlet
+import org.scalatra.{Ok, BadRequest}
+
+import scala.collection.JavaConversions._
+
+trait GeoMesaPersistentBlobStoreServlet extends GeoMesaBaseDataStoreServlet {
+
+  @volatile var abs: AccumuloBlobStore = null
```
elahrvivaz: so you're persisting aliases now, but you aren't initializing the blobstore on startup, right? and the blob store isn't associated with any particular alias, it is just the last alias registered? seems like the workflow is still a little confused.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: yeah that would not end well

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
AndrewAnnex: well to be fair this was all out of scope of the tickets in the pr title

elahrvivaz: well you added the persistence stuff :)
If you're going to add the persistence, you should make it work, otherwise there doesn't appear to be any point to having the persistence here....

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,114 @@
...
-  // TODO: Revisit configuration and persistence of configuration.
-  // https://geomesa.atlassian.net/browse/GEOMESA-958
-  // https://geomesa.atlassian.net/browse/GEOMESA-984
-  post("/ds") {
-
-    logger.debug("In ds registration method")
-
-    val dsParams = datastoreParams
-    val ds = new AccumuloDataStoreFactory().createDataStore(dsParams).asInstanceOf[AccumuloDataStore]
+  val maxFileSize: Int = Try(System.getProperty(BlobstoreServlet.maxFileSizeSysProp, "50").toInt).getOrElse(50)
```
jahhulbert-ccri: I was thinking this should throw an exception if it is not an integer and fail. perhaps use `Option(System.getProperty()).map(_.toInt).getOrElse()` instead? @elahrvivaz what do you think.

jahhulbert-ccri: also, possibly create a follow-on ticket for using SI suffixes like 50m or 2g or refactor this: https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/GeoMesaBatchWriterConfig.scala#L28

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
jahhulbert-ccri: yeah that's cool i'm re-reading this: https://twitter.github.io/scala_school/concurrency.html

jahhulbert-ccri: nm too simple. back to java concurrency book.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -78,6 +79,32 @@ class AccumuloBlobStore(ds: AccumuloDataStore) {
...
+    deleteFeature(id)
+  }
+
+  private def deleteFeature(id: String): Unit = {
+    val removalFilter = Filters.ff.id(Filters.ff.featureId(id))
+    val fd = ds.getFeatureWriter(blobFeatureTypeName, removalFilter, Transaction.AUTO_COMMIT)
+    try {
+      while (fd.hasNext) {
+        fd.next()
+        fd.remove()
```
jahhulbert-ccri: @elahrvivaz can you verify this while loop is the correct way to delete? I instructed @AndrewAnnex to do it this way

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -87,9 +114,9 @@ class AccumuloBlobStore(ds: AccumuloDataStore) {
     (value.get, filename)
   }
 
-  private def putInternal(file: File, id: String) {
-    val localName = file.getName
-    val bytes =  ByteStreams.toByteArray(Files.newInputStreamSupplier(file))
+  private def putInternal(file: File, id: String, params: Map[String, String]) {
+    val localName = getFileName(file, params)
+    val bytes = ByteStreams.toByteArray(Files.asByteSource(file).openBufferedStream())
```
jahhulbert-ccri: This probably doesn't close the input stream and makes 2 copies of the byte array...plus this is simpler:
`val bytes = Files.toByteArray(file)`

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -87,9 +114,9 @@ class AccumuloBlobStore(ds: AccumuloDataStore) {
...
 
     val m = new Mutation(id)
```
jahhulbert-ccri: sorry to go outside scope of PR but there needs to be a bw.flush() after the mutation is added in order to actually return  a 200 series response to the user. currently this could result in a mutation rejected exception or other err and never get persisted.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
elahrvivaz: Yeah, I think this works. In typical geotools fashion, I think you can also do ds.getFeatureSource(sftName).removeFeatures(filter)

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: yeah, i thought the same thing about the try - probably should be try {} catch { throw meaningful exception }

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: sure

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: ok

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: I think I can work this in

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
AndrewAnnex: I was reading something last night that may counter my claim, @jahhulbert-ccri should I just switch everything to be synchronized calls?

elahrvivaz: @AndrewAnnex what were you reading? from what i saw, sounds like volatile basically synchronizes every read/write

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: unless the feature source needs to be cast to something else,  both SimpleFeatureSource and FeatureSource don't implement `removeFeatures` at least according to the ide.

elahrvivaz: you probably need to cast to SimpleFeatureStore - but I would leave it as it is with the loop, anyway.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
AndrewAnnex: http://www.javamex.com/tutorials/synchronization_volatile.shtml

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: I am going to keep it as it is now for now

geomesa-blobstore/README.md
anthonyccri: This is probably a bit of a security hole.  I wonder if there's a way to abstract over the geoserver catalog in such a way as to not bring it into the dependency list.  That way we can avoid this problem in our REST services.

elahrvivaz: we don't pull it in directly - we actually do have an abstraction over it. We invoke some geoserver classes through spring beans to get the folder to persist things to. not sure if that is allowed or not... see e.g.
https://github.com/locationtech/geomesa/blob/master/geomesa-web/geomesa-web-data/src/main/resources/applicationContext.xml#L18-L49

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/GeoMesaPersistentBlobStoreServlet.scala
AndrewAnnex: I have a branch I have been working on that addresses this, not yet ready to merge into this pr.

jahhulbert-ccri:
i'll try to look through this again today

AndrewAnnex:
I haven't updated it yet

AndrewAnnex:
I need to update the readme one last time

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+import org.locationtech.geomesa.web.core.GeoMesaBaseDataStoreServlet
+import org.scalatra._
+import org.scalatra.servlet.{FileUploadSupport, MultipartConfig, SizeConstraintExceededException}
 
 import scala.collection.JavaConversions._
+import scala.collection.{Map, concurrent}
 
-class BlobstoreServlet extends GeoMesaScalatraServlet with LazyLogging {
-  override def root: String = "blob"
+class BlobstoreServlet(val persistence: FilePersistence) extends GeoMesaBaseDataStoreServlet with FileUploadSupport with GZipSupport {
```
elahrvivaz: long line

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+      handleError("Uploaded file too large!", e)
+    case e: IOException =>
+      handleError("IO exception in BlobstoreServlet", e)
+  }
 
-    logger.debug("In ds registration method")
+  // TODO: may need to be switched to a synchronized map of some kind
+  val blobStores: concurrent.Map[String, AccumuloBlobStore] = new ConcurrentHashMap[String, AccumuloBlobStore]
+  getPersistedDataStores.foreach {
+    store => connectToBlobStore(store._2).map(abs => blobStores.putIfAbsent(store._1, abs))
```
elahrvivaz: [minor] pattern matching would be nice instead of _1, _2

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+  /**
+    * Override of the requestPath function, may need to match on type of request if wrapped
+    * depending on deployment environment
+    *
+    * @param request
+    * @return
+    */
+  override def requestPath(implicit request: HttpServletRequest): String = request match {
+    case r: HttpServletRequestWrapper =>
+      r.getMethod.toLowerCase(Locale.ENGLISH) match {
```
elahrvivaz: Locale.US, not ENGLISH

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+  }
+
+  /**
+    * Remove the reference to an existing data store
+    */
+  delete("/ds/:alias/?") {
+    val alias = params("alias")
+    val prefix = keyFor(alias)
+    try {
+      persistence.removeAll(persistence.keys(prefix).toSeq)
```
elahrvivaz: should remove from blobstore map too

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+        BadRequest(reason = "AccumuloBlobStore is not initialized.")
+      case Some(abs) =>
+        val id = params("id")
+        logger.debug("Attempting to get blob for id: {} from store: {}", id, alias)
+        val (returnBytes, filename) = abs.get(id)
+        if (returnBytes == null) {
+          NotFound(reason = s"Unknown ID $id")
+        } else {
+          contentType = "application/octet-stream"
+          response.setHeader("Content-Disposition", "attachment;filename=" + filename)
```
elahrvivaz: maybe as a follow up ticket, we could add support for return types instead of downloads, so you could use this to display things on a web page. Might not be a use case though.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+      case Some(abs) =>
+        logger.debug("Attempting to ingest file to BlobStore")
+        try {
+          fileParams.get("file") match {
+            case None =>
+              logger.error("no file parameter in request")
+              BadRequest()
+            case Some(file) =>
+              val otherParams = collection.mutable.Map[String, String]()
+              multiParams.foreach{ case (s, p) => otherParams.update(s, p.head) }
```
elahrvivaz: [picky] space before {

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+                    Created(body = id, headers = scala.collection.immutable.Map("Location" -> request.getRequestURL.append(id).toString))
+                  case None =>
+                    BadRequest(reason = "Unable to process file")
+                }
+              } catch {
+                case e: Exception => handleError("", e)
+              } finally {
+                tempFile.delete()
+              }
+              actRes
```
elahrvivaz: i think you want to remove actRes - handle error will be the return type if the exception block gets triggered, but this would override that. So should be try { ... created or badrequest } catch { handle error } finally {} /end of method

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+              if (!otherParams.contains(filenameFieldName)) {
+                // we put the true filename in here so that we can preserve it in the blob table
+                otherParams.put(filenameFieldName, file.getName)
+              }
+              val tempFile = File.createTempFile(UUID.randomUUID().toString, FilenameUtils.getExtension(file.getName))
+              val actRes = try {
+                file.write(tempFile)
+                abs.put(tempFile, otherParams.toMap) match {
+                  case Some(id) =>
+                    Created(body = id, headers = scala.collection.immutable.Map("Location" -> request.getRequestURL.append(id).toString))
```
elahrvivaz: long lone - note don't think you need scala.collection.immutable here, that is the default import.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+                  case None =>
+                    BadRequest(reason = "Unable to process file")
+                }
+              } catch {
+                case e: Exception => handleError("", e)
```
elahrvivaz: put some kind of message here instead of empty string

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,234 @@
...
+  )
+  error {
+    case e: SizeConstraintExceededException =>
+      handleError("Uploaded file too large!", e)
+    case e: IOException =>
+      handleError("IO exception in BlobstoreServlet", e)
+  }
 
-    logger.debug("In ds registration method")
+  // TODO: may need to be switched to a synchronized map of some kind
```
elahrvivaz: i think you can remove this todo, concurrent hashmap looks ok to me

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/handlers/BlobStoreFileHandler.scala
```diff
@@ -29,9 +31,9 @@ object BlobStoreFileHandler {
   }
 }
 
-trait AbstractFileHandler extends FileHandler {
+trait AbstractFileHandler extends FileHandler with BlobStoreFileName {
   val builderLocal: ThreadLocal[SimpleFeatureBuilder] = new ThreadLocal[SimpleFeatureBuilder] {
-    override def initialValue(): SimpleFeatureBuilder = new SimpleFeatureBuilder(sft)
+    override def initialValue(): SimpleFeatureBuilder = new SimpleFeatureBuilder(sft).featureUserData(Hints.USE_PROVIDED_FID, java.lang.Boolean.valueOf("true"))
```
elahrvivaz: long line. can also use the java.lang.Boolean.TRUE constant instead of valueOf

geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/GeoMesaBaseDataStoreServlet.scala
```diff
@@ -0,0 +1,43 @@
...
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.web.core
+
+import org.locationtech.geomesa.utils.cache.FilePersistence
+
+trait GeoMesaBaseDataStoreServlet extends GeoMesaScalatraServlet {
```
elahrvivaz: maybe rename this class to be more descriptive - like PersistentDataStoreServlet

elahrvivaz:
@AndrewAnnex I looked though it again, had a bunch of comments :)

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: heh

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: you mean delete the blobstore / geomesa tables right?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: yeah I think there is a ticket for something like this, definitely a cool thing to do down the road.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: no - this method doesn't really delete anything - just removes the reference to it in the 'registered' aliases. So in your case, you also need to remove the reference in the blobstore map.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jahhulbert-ccri: USA! USA! USA!

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: gotcha

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,240 @@
...
+  }
+
+  /**
+    * Remove the reference to an existing data store and removes tables
+    */
+  delete("/ds/:alias/?") {
+    val alias = params("alias")
+    val prefix = keyFor(alias)
+    try {
+      blobStores.get(alias).foreach(_.delete())
```
elahrvivaz: is this deleting actual accumulo data? i think you just want to remove reference from the map.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
```diff
@@ -8,55 +8,240 @@
...
+          r.getPathInfo.replace(s"/$root", "")
+        case _        =>
+          r.getPathInfo
+      }
+    case _ => super.requestPath
+  }
+
+  private def handleBadRequest(msg: String): ActionResult = {
+    logger.error(msg)
+    if (debug) {
```
elahrvivaz: don't think we need to hide bad request messages - it's probably helpful to know what you did wrong.

elahrvivaz:
@AndrewAnnex had a couple more comments

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: that is what you asked for in the comment

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: this contradicts previous comments. I am keeping the style as it is now

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: nah, bruh, you misunderstood.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: which previous comments? you don't want to expose inner workings, but telling someone they forgot a parameter or something is ok. tying this into the same debug var we use to expose everything isn't good.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: I am calling ```persistence.removeAll(persistence.keys(prefix).toSeq)```, does this not do exactly that (remove reference from the map)?

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: this is a style used in all of the other servlets. I am happy to reference previous comments.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
elahrvivaz: yeah, i'm talking about the blobstore map.

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
jnh5y: NB: The FeatureSource interfaces are read-only.  The FeatureStores are read-write.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: I think I see...

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
jnh5y: Here, I'm fine with either way.  Generally, I'd agree with @elahrvivaz, but we can leave it as now.

AndrewAnnex: I decided to make things consistent in the pr as it was not clear where no output was preferred over some output. maybe I over-adjusted.

geomesa-blobstore/geomesa-blobstore-api/src/main/scala/org/locationtech/geomesa/blob/api/BlobstoreServlet.scala
AndrewAnnex: this is now fixed

elahrvivaz:
last comment, number 114: merging
i'm going to add some try/catches in for ya

elahrvivaz:
ae45474cd973469e72e0ec7a606ae8f7bb45e36f

PR #762, opened 2016-01-14
GEOMESA-1031 Allowing for schemas without geometries

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
@@ -31,7 +31,7 @@ object SpatioTemporalTable extends GeoMesaTable with LazyLogging {
   val INDEX_CHECK = s"~$INDEX_FLAG~"
   val DATA_CHECK = s"~$DATA_FLAG~"
 
-  override def supports(sft: SimpleFeatureType): Boolean = true
+  override def supports(sft: SimpleFeatureType): Boolean = sft.getGeometryDescriptor != null
 
   override val suffix: String = "st_idx"
 
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/Z3FeatureIdGenerator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/Z3FeatureIdGenerator.scala
@@ -13,6 +13,7 @@ import java.util.{Date, UUID}
 import com.google.common.primitives.Longs
 import com.vividsolutions.jts.geom.{Geometry, Point}
 import org.locationtech.geomesa.accumulo.data.tables.Z3Table
+import org.locationtech.geomesa.utils.geotools.GeometryUtils
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.uuid.{FeatureIdGenerator, RandomLsbUuidGenerator}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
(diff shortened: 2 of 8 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/Z3FeatureIdGenerator.scala
```diff
@@ -54,12 +55,19 @@ object Z3UuidGenerator extends RandomLsbUuidGenerator {
...
-
-    val pt = sf.getAttribute(sft.getGeomIndex) match {
-      case p: Point => p
-      case g: Geometry => g.getCentroid
+        .getOrElse(System.currentTimeMillis())
+
+    val pt = {
+      val i = sft.getGeomIndex
+      if (i == -1) {
+        GeometryUtils.zeroPoint
```
jahhulbert-ccri: hahaha spatial hotspot!

jahhulbert-ccri:
looks good. i was hoping you'd implement a timeseries database instead but i'll take it

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
```diff
@@ -31,7 +31,7 @@ object SpatioTemporalTable extends GeoMesaTable with LazyLogging {
   val INDEX_CHECK = s"~$INDEX_FLAG~"
   val DATA_CHECK = s"~$DATA_FLAG~"
 
-  override def supports(sft: SimpleFeatureType): Boolean = true
+  override def supports(sft: SimpleFeatureType): Boolean = sft.getGeometryDescriptor != null
```
jahhulbert-ccri: wonder if we should also have sft.getDtgField.isDefined ?not sure if we need to check that or not...

elahrvivaz: no, STIdx works without a dtg.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/Z3FeatureIdGenerator.scala
elahrvivaz: hmm, yeah maybe i should make the point random... it's kind of a fringe case though.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/SpatioTemporalTable.scala
jahhulbert-ccri: ah yeah

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/Z3FeatureIdGenerator.scala
jahhulbert-ccri: i was thinking just for ingest purposes since the id doesn't really matter...although if its random it's not deterministic...

jahhulbert-ccri: prob doesn't matter

elahrvivaz: yeah, the id doesn't matter - but the geom is approximately 5 bytes of the UUID that would always be the same.

jahhulbert-ccri:
verified ingest and export of a file like this:
```
a,b,c
d,e,f
h,i,j
```

@elahrvivaz anything else you want to change here? else ready to merge

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -201,8 +201,11 @@ object RichSimpleFeatureType {
   // in general we store everything as strings so that it's easy to pass to accumulo iterators
   implicit class RichSimpleFeatureType(val sft: SimpleFeatureType) extends AnyVal {
 
-    def getGeomField: String = sft.getGeometryDescriptor.getLocalName
-    def getGeomIndex: Int = sft.indexOf(sft.getGeometryDescriptor.getLocalName)
+    def getGeomField: String = {
```
jahhulbert-ccri: should these be Option[String] and Option[Int] like getDtgField is for consistency?

elahrvivaz: Ideally, yes, but in practice I don't think so... almost everywhere we use it, we're in code that is excluded up front by the table.supports mechanism. So basically we would be doing getGeomField.getOrElse(throw new Exception). The places we aren't protected I added checks.
Yeah, it might lead to some subtle bugs down the road, but I think sfts without geoms is kind of a fringe case anyway...

elahrvivaz: @jahhulbert-ccri I'm good with it to be merged.

jahhulbert-ccri:
2bc17c780633fc343a68caa96f206901c48aae74

PR #777, opened 2016-01-21
GEOMESA-1060 Adding try/catch in GeoMesaOutputFormat to prevent job f…

elahrvivaz:
…ailures from invalid features, fixing tools ingestion counters to include features that failed to write

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ConverterIngest.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ConverterIngest.scala
@@ -22,6 +22,7 @@ import org.joda.time.Period
 import org.joda.time.format.PeriodFormatterBuilder
 import org.locationtech.geomesa.accumulo.data.AccumuloDataStoreFactory
 import org.locationtech.geomesa.convert.SimpleFeatureConverters
+import org.locationtech.geomesa.convert.Transformers.DefaultCounter
 import org.locationtech.geomesa.utils.classpath.PathUtils
 import org.locationtech.geomesa.utils.stats.CountingInputStream
 import org.opengis.feature.simple.SimpleFeatureType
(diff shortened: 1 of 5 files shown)


vpipkt:
I have identified a sample of data which causes the previously identified issue - mapreduce job fails out... Note also the contradictory number of reported records ingested. I will attempt job on dcloud so we can all see what's going on.

    [                                                            ]   0% complete 7108 ingested 1 failed in 00:00:33
    [============================================================] 100% complete 0 ingested 0 failed in 00:00:48 :46 
    Job failed with state FAILED due to: Task failed task_1453124677016_0027_m_000001
    Job failed as tasks failed. failedMaps:1 failedReduces:0

    Distributed ingestion complete in 00:00:48
    Ingested 0 features with no failures.

vpipkt:
The plot thickens. I ran the same sample on `dcloud` and ingest was fine. I will double-check things regarding my prior comment.

vpipkt:
Seems my previous issue was due to local configuration issues. After sanity checking things, pulling, and rebuilding - ingest of same sample of data works as expected on both dcloud and cloud-local

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ConverterIngest.scala
```diff
@@ -104,8 +98,11 @@ class ConverterIngest(dsParams: Map[String, String],
               toWrite.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.TRUE)
               try {
                 fw.write()
+                written.incrementAndGet()
```
jahhulbert-ccri: I guess technically we have to wait unless the flush happens on the BW to claim its truly done...but I feel like we don't really want to deal with that unless we have to....you have to keep a list of features that are added to the BW and then clear the list when you call flush(). Similar to we we've done in storm to ack tuples.

elahrvivaz: ah, good point... but yeah, that would be hard to track, b/c you don't even know when it flushes, right?

jahhulbert-ccri: you can call ```AccumuloFeatureWriter.flush()``` and it will pause until the mutations are sent to the tablet servers (and else throw a MutationsRejectedException) (as far as I remember)

jahhulbert-ccri: but of course...you'd have to have a timer...that would flush every X features or mb, etc.

jahhulbert-ccri: Unfortunately I don't know that the FW in it's current implementation super supports that. It would be nice to have a transaction that did that for us...exposed the flush time, flush ack, callback something in the API

anthonyccri:
b608809..791ceab

PR #780, opened 2016-01-23
GEOMESA-1063 API for discovering available feature types and converters

anthonyccri:
* Added some methods for accessing feature types and converters that
  are available on the classpath
* All files name "geomesa-data.conf" on the classpath will be loaded.
* Each file is expected to have a root element ```geomesa.converters```
  with one key corresponding to the name of the data type.  Under the
  data type name should be ```sft``` and ```converter``` definitions.

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
@@ -16,6 +16,7 @@ import com.typesafe.config.{Config, ConfigFactory}
 import com.typesafe.scalalogging.LazyLogging
 import org.locationtech.geomesa.convert.Transformers._
 import org.locationtech.geomesa.features.ScalaSimpleFeature
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.JavaConversions._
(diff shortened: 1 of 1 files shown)


jahhulbert-ccri:
FYI example confs here; https://github.com/locationtech/geomesa/blob/master/geomesa-tools/conf/application.conf

anthonyccri:
@jahhulbert-ccri good catch.  Ok, seems like we need to do a bit of disambiguation.  @elahrvivaz check out https://github.com/geomesa/gm-data

I like the idea of having definitions in modular jar files available on the classpath.  Especially if we have the wrapper classes generated and dropped into those jars.  I think we also need the ability to load and modify definitions at runtime - not from the classpath.  Anyone have thoughts around this?

jahhulbert-ccri:
We 100% need to load and modify while running in addition to the classpath so that we can support interactive type/converter creation without restarting the containers...

Currently we can do it via a separate jar file like this and reference them by name...which i'm going to probably change to include dropdown of known types.

https://github.com/jahhulbert-ccri/geomesa-nifi/tree/master/geomesa-nifi-resources/src/main/resources

nifi-geomesa1

elahrvivaz:
We should be able to dynamically drop in jars on the classpath right now - you just have to have a file named application.conf in your jar, and we will pick it up with ConfigFactory.load:
https://github.com/typesafehub/config#standard-behavior

elahrvivaz:
Also, I added a 'name' to converters, so we can match them up with the sft using that.

elahrvivaz:
for runtime changes, we could use zookeeper distributed caching. we already have that mostly done in whiptail, hooked up to a rest api, but we could make it part of the command line tools instead.

jahhulbert-ccri:
closing for additional work...copied branch here: https://github.com/jahhulbert-ccri/geomesa/tree/fcr_discover_api

PR #787, opened 2016-01-27
GEOMESA-1066,GEOMESA-1067 KafkaConsumer should handle brokers going down

ddseapy:
* Underlying consumer is not set to null when disconnected, preventing NPE in reCreateConsumer method
* Exceptions thrown in the consumer's fetch method (such as a broker being down) are caught with and recovered from without killing the thread


elahrvivaz:
looks good, @ddseapy did you test it out?
Could you add the sign-off footer? `commit --amend -s` should do it

elahrvivaz:
022698002572e99bfd6bc77a78eca78762dcb26e

PR #793, opened 2016-02-04
GEOMESA-1079 - VisibilityFilter's accept method should not assume second arg is a filter after gt version upgrade

ddseapy:
* Creates a Function instead of the filter


mmatz-ccri:
I didn't test it but looks good.

elahrvivaz:
61ba534bd928ed8b7e34842070db29da1d7aa297

PR #797, opened 2016-02-11
GEOMESA-1087 Updating docs to Java 8

elahrvivaz:


jahhulbert-ccri:
@elahrvivaz are we gonna bump to 8 for compile or keep with 7 for a little while? I'd probably vote to keep compiling with 7 for now if possible since centos 6 is still that by default.

elahrvivaz:
@jahhulbert-ccri the cassandra pr that @anthonyccri is working on doesn't build unless you use java 8. We're still compiling for a target of 7 though, I think... if not we should probably fix that.

elahrvivaz:
I added <source> and <target> tags to the scala compiler - the java compiler already had it.

jahhulbert-ccri:
merging

jahhulbert-ccri:
78550cbd3a98a7e267173572077699a17043370d

PR #803, opened 2016-02-15
GEOMESA-1089 Updating poms to add profile for building on LocationTech

elahrvivaz:
* Updating hadoop version to 2.5.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
         <!-- environment-dependent versions -->
         <zookeeper.version>3.4.5</zookeeper.version>
         <thrift.version>0.9.1</thrift.version>
-        <hadoop.version>2.2.0</hadoop.version>
+        <hadoop.version>2.5.2</hadoop.version>
         <spark.version>1.5.0</spark.version>
 
         <!-- logging properties - we expect logging jars to be provided -->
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -62,7 +62,7 @@
         <!-- environment-dependent versions -->
         <zookeeper.version>3.4.5</zookeeper.version>
         <thrift.version>0.9.1</thrift.version>
-        <hadoop.version>2.2.0</hadoop.version>
+        <hadoop.version>2.5.2</hadoop.version>
```
jahhulbert-ccri: may not matter but we might want to build against 2.5.0 but suggest 2.5.2 for deploys. I would assume though that all 2.5.x are semi compat.

elahrvivaz: why 2.5.0? I just looked for the most recent 2.5.x...

pom.xml
```diff
@@ -132,35 +132,91 @@
...
-                <maven.test.jvmargs>-Duser.timezone=UTC -Xms512m -Xmx3g -XX:-UseGCOverheadLimit -Djava.net.preferIPv4Stack=true</maven.test.jvmargs>
+                <maven.test.jvmargs>-Duser.timezone=UTC -Xms512m -Xmx3g -XX:-UseGCOverheadLimit -Djava.net.preferIPv4Stack=true -Djava.awt.headless=true</maven.test.jvmargs>
                 <test.fork.count>1</test.fork.count>
                 <test.fork.reuse>false</test.fork.reuse>
             </properties>
         </profile>
+
+        <profile>
+            <!-- disables all non-LT repositories for a build with only CQ'd jars -->
+            <id>locationtech</id>
```
jahhulbert-ccri: thought I wonder if we could do a <mirrorOf>* or something just in case some transitive pom defines its own repos not sure if those would get picked up

elahrvivaz: hmm, would maven pick up repos defined in dependencies? i feel like not unless you did something unusual like maybe a pom with <scope>import</scope>

jahhulbert-ccri:
mergin

jahhulbert-ccri:
0986392a527bd2c05ae8a7b4b953e671a8a12f48

PR #810, opened 2016-02-19
GEOMESA-1104 Adding docs to default build

elahrvivaz:
* Having docs be an included module will handle pom versions during releases
* Docs will not actually be built unless you use the 'docs' profile


jnh5y:
Committed as ccf70d35a6120e83808a71b9954dd3372d913ba5.

PR #815, opened 2016-02-24
GEOMESA-1110 Implement KDS Feature Listener

jnh5y:
* Adds event firing when KDS receives updates.
* Add KafkaFeatureEvent class which contains the feature updated.

Run the unit test with
> mvn test -Dtest=LiveKafkaConsumerFeatureSourceTest -Dorg.locationtech.geomesa.test-kafka=true

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
@@ -37,8 +37,7 @@ abstract class KafkaConsumerFeatureSource(entry: ContentEntry,
   with ContentFeatureSourceSecuritySupport
   with ContentFeatureSourceReTypingSupport {
 
-  override def getBoundsInternal(query: Query) =
-    ReferencedEnvelope.create(new Envelope(-180, 180, -90, 90), CRS_EPSG_4326)
+  override def getBoundsInternal(query: Query) = KafkaConsumerFeatureSource.wholeWorldBounds
 
   override def buildFeatureType(): SimpleFeatureType = schema
 
@@ -49,6 +48,10 @@ abstract class KafkaConsumerFeatureSource(entry: ContentEntry,
   override def getReaderInternal(query: Query): FR = addSupport(query, getReaderForFilter(query.getFilter))
 
   def getReaderForFilter(f: Filter): FR
+}
+
+object KafkaConsumerFeatureSource {
+  lazy val wholeWorldBounds = ReferencedEnvelope.create(new Envelope(-180, 180, -90, 90), CRS_EPSG_4326)
 }
 
 case class FeatureHolder(sf: SimpleFeature, env: Envelope) {
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
@@ -15,6 +15,7 @@ import java.util.concurrent.{Executors, LinkedBlockingQueue, ScheduledThreadPool
 import com.google.common.base.Ticker
 import com.google.common.cache._
 import com.typesafe.scalalogging.LazyLogging
+import com.vividsolutions.jts.geom.{Point, Envelope}
 import org.geotools.data.FeatureEvent.Type
 import org.geotools.data.store.ContentEntry
 import org.geotools.data.{FeatureEvent, Query}
@@ -23,7 +24,7 @@ import org.geotools.filter.identity.FeatureIdImpl
 import org.geotools.geometry.jts.ReferencedEnvelope
 import org.locationtech.geomesa.kafka.consumer.KafkaConsumerFactory
 import org.locationtech.geomesa.utils.geotools.Conversions._
-import org.locationtech.geomesa.utils.geotools.FR
+import org.locationtech.geomesa.utils.geotools._
 import org.locationtech.geomesa.utils.index.{BucketIndex, SpatialIndex}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import org.opengis.filter.Filter
@@ -44,7 +45,7 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
 
   private[kafka] val featureCache = new LiveFeatureCache(sft, expirationPeriod)
 
-  private val contentState = entry.getState(getTransaction)
+  private lazy val contentState = entry.getState(getTransaction)
 
   private val msgDecoder = new KafkaGeoMessageDecoder(sft)
   private val queue = new LinkedBlockingQueue[GeoMessage]()
@@ -127,10 +128,24 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
       queue.take() match {
         case update: CreateOrUpdate =>
           featureCache.createOrUpdateFeature(update)
-          contentState.fireFeatureEvent(new KafkaFeatureEvent(this, Type.CHANGED, null, update.feature))
-        case del: Delete            => featureCache.removeFeature(del)
-          contentState.fireFeatureEvent(new FeatureEvent(this, Type.REMOVED, null, KafkaFeatureEvent.buildId(del.id)))
-        case clr: Clear             => featureCache.clear()
+          fireEvent(new KafkaFeatureEvent(this,
+                                          Type.CHANGED,
+                                          KafkaFeatureEvent.buildBounds(update.feature),
+                                          update.feature))
+
+        case del: Delete            =>
+          featureCache.removeFeature(del)
+          fireEvent(new FeatureEvent(this,
+                                     Type.REMOVED,
+                                     KafkaFeatureEvent.buildBounds(featureCache.features(del.id).sf),
+                                     KafkaFeatureEvent.buildId(del.id)))
+
+        case clr: Clear             =>
+          featureCache.clear()
+          fireEvent(new FeatureEvent(this,
+                                     Type.REMOVED,
+                                     KafkaConsumerFeatureSource.wholeWorldBounds,
+                                     Filter.INCLUDE))
         case m                      => throw new IllegalArgumentException(s"Unknown message: $m")
       }
     }
@@ -139,6 +154,15 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
   override def getCountInternal(query: Query): Int = featureCache.size(query.getFilter)
 
   override def getReaderForFilter(f: Filter): FR = featureCache.getReaderForFilter(f)
+
+  override def canEvent: Boolean = true
+
+  // Lazily fires events.
+  def fireEvent(event: => FeatureEvent) = {
+    if (contentState.hasListener) {
+      contentState.fireFeatureEvent(event)
+    }
+  }
 }
 
 import org.locationtech.geomesa.kafka.KafkaFeatureEvent._
@@ -158,6 +182,19 @@ object KafkaFeatureEvent {
     set.add(fid)
 
     ff.id(set)
+  }
+
+  def buildBounds(feature: SimpleFeature): ReferencedEnvelope = {
+    try {
+      val geom = feature.getDefaultGeometry.asInstanceOf[Point]
+      val lon = geom.getX
+      val lat = geom.getY
+
+      ReferencedEnvelope.create(new Envelope(lon, lon, lat, lat), CRS_EPSG_4326)
+    } catch {
+      case t: Throwable =>
+        KafkaConsumerFeatureSource.wholeWorldBounds
+    }
   }
 }
 
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
@@ -128,14 +128,14 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
     while (running.get) {
       queue.take() match {
         case update: CreateOrUpdate =>
-          featureCache.createOrUpdateFeature(update)
           fireEvent(KafkaFeatureEvent.changed(this, update.feature))
+          featureCache.createOrUpdateFeature(update)
         case del: Delete            =>
-          featureCache.removeFeature(del)
           fireEvent(KafkaFeatureEvent.removed(this, featureCache.features(del.id).sf))
+          featureCache.removeFeature(del)
         case clr: Clear             =>
-          featureCache.clear()
           fireEvent(KafkaFeatureEvent.cleared(this))
+          featureCache.clear()
         case m                      => throw new IllegalArgumentException(s"Unknown message: $m")
       }
     }
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
@@ -15,10 +15,11 @@ import java.util.concurrent.{Executors, LinkedBlockingQueue, ScheduledThreadPool
 import com.google.common.base.Ticker
 import com.google.common.cache._
 import com.typesafe.scalalogging.LazyLogging
-import com.vividsolutions.jts.geom.{Point, Envelope}
+import com.vividsolutions.jts.geom.{Envelope, Point}
 import org.geotools.data.FeatureEvent.Type
+import org.geotools.data.simple.SimpleFeatureSource
 import org.geotools.data.store.ContentEntry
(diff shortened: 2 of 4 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -114,14 +122,18 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
...
-      case update: CreateOrUpdate => featureCache.createOrUpdateFeature(update)
-      case del: Delete            => featureCache.removeFeature(del)
-      case clr: Clear             => featureCache.clear()
-      case m                      => throw new IllegalArgumentException(s"Unknown message: $m")
+  override def run(): Unit =
+    while (running.get) {
+      queue.take() match {
+        case update: CreateOrUpdate =>
+          featureCache.createOrUpdateFeature(update)
+          contentState.fireFeatureEvent(new KafkaFeatureEvent(this, Type.CHANGED, null, update.feature))
```
anthonyccri: Can you make these lazy?  Basically, we shouldn't create new objects unless there are in fact some listeners.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -114,14 +122,18 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
...
+        case del: Delete            => featureCache.removeFeature(del)
+          contentState.fireFeatureEvent(new FeatureEvent(this, Type.REMOVED, null, KafkaFeatureEvent.buildId(del.id)))
```
anthonyccri: Ditto above.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSourceTest.scala
```diff
@@ -1,16 +1,20 @@
 /***********************************************************************
-* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
-* All rights reserved. This program and the accompanying materials
-* are made available under the terms of the Apache License, Version 2.0
-* which accompanies this distribution and is available at
-* http://www.opensource.org/licenses/apache2.0.php.
-*************************************************************************/
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
```
anthonyccri: I think these spaces will break the maven enforcer.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
anthonyccri: Also, does ```fireFeatureEvent``` block?  Either we fire events asynchronously on a thread pool or we document it in such a way so that listeners ensure async.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -114,14 +122,18 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
...
+        case clr: Clear             => featureCache.clear()
```
elahrvivaz: should we fire events for this case?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
elahrvivaz: I guess we don't care about bounds either? Seems like it could be useful.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSourceTest.scala
```diff
@@ -105,6 +109,60 @@ class LiveKafkaConsumerFeatureSourceTest extends Specification with HasEmbeddedK
...
+        val ll = maxLon - maxLon/i
+
+        val sf = fw.next()
+        sf.getIdentifier.asInstanceOf[FeatureIdImpl].setID("testlistener")
+        sf.setAttributes(Array("smith", 30, DateTime.now().toDate).asInstanceOf[Array[AnyRef]])
+        sf.setDefaultGeometry(gf.createPoint(new Coordinate(ll, ll)))
+        fw.write()
+      }
+
+      logger.info("Wrote feature")
```
elahrvivaz: can we change log to debug, so we don't spam the console during tests? ditto below.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSourceTest.scala
```diff
@@ -105,6 +109,60 @@ class LiveKafkaConsumerFeatureSourceTest extends Specification with HasEmbeddedK
...
+      while(latch.getCount > 0) {
```
elahrvivaz: might be good to have an exit check in case of failure - otherwise this will just hang. but maybe doesn't matter...

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/TestLambdaFeatureListener.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import org.geotools.data.{FeatureEvent, FeatureListener}
+
+class TestLambdaFeatureListener(lambda: KafkaFeatureEvent => Unit) extends FeatureListener {
+  override def changed(featureEvent: FeatureEvent): Unit = {
+    featureEvent match {
+      case kfe: KafkaFeatureEvent =>
+        lambda(kfe)
+      case _ =>
+        println(s"Received event ${featureEvent.getType}: ${featureEvent.toString}.")
```
elahrvivaz: change to logger

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
anthonyccri: Geo bounds?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
anthonyccri: Hmm, I think so, but it should be an event per feature, right?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
elahrvivaz: yeah, looks like the base feature event has a bounds referenced envelope - presumably for filtering by listeners.

jnh5y: In order, 
1.  lazy - I'll try and think about it.
2.  blocking:  it looks like it does: https://github.com/geotools/geotools/blob/master/modules/library/data/src/main/java/org/geotools/data/store/ContentState.java#L372-L394

I'd argue that should be handled in the ContentDataStore.  I'll fiddle with some options and bug the GeoTools Devel list.

3.  Geo bounds.  Yeah, fair enough, I'll see about sorting that out.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
jnh5y: I think
```
new FeatureEvent(this, Type.REMOVED, WHOLE_WORLD_BOUNDS, Filter.INCLUDE)
```
would be fair.  Thoughts?

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSourceTest.scala
jnh5y: Good point.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSourceTest.scala
jnh5y: Hmm... yeah... have to think about it.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -51,6 +50,10 @@ abstract class KafkaConsumerFeatureSource(entry: ContentEntry,
   def getReaderForFilter(f: Filter): FR
 }
 
+object KafkaConsumerFeatureSource {
+  lazy val wholeWorldBounds = ReferencedEnvelope.create(new Envelope(-180, 180, -90, 90), CRS_EPSG_4326)
```
mmatz-ccri: This might be useful in other places too.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -114,19 +123,79 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
...
+  def fireEvent(event: => FeatureEvent) = {
+    if (contentState.hasListener) {
+      contentState.fireFeatureEvent(event)
+    }
+  }
+}
+
+import org.locationtech.geomesa.kafka.KafkaFeatureEvent._
+
+class KafkaFeatureEvent(source: AnyRef,
```
mmatz-ccri: Is this class really needed or only for testing?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -114,19 +123,79 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
...
+                        eventType: FeatureEvent.Type,
+                        bounds: ReferencedEnvelope,
+                        val feature: SimpleFeature)
+  extends FeatureEvent(source, eventType, bounds, buildId(feature.getID)) { }
+
+object KafkaFeatureEvent {
```
mmatz-ccri: Suggestion: add 3 factory methods: 
```
KafkaFeatureEvent.changed(src, feature): FeatureEvent
KafkaFeatureEvent.removed(src, feature): FeatureEvent
KafkaFeatureEvent.cleared(src): FeatureEvent
```
Or something like that.  That will make run() more readable.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSourceTest.scala
```diff
@@ -105,6 +109,60 @@ class LiveKafkaConsumerFeatureSourceTest extends Specification with HasEmbeddedK
...
+      }
+      val producerDS = DataStoreFinder.getDataStore(producerParams)
+      producerDS.createSchema(sft)
+
+      val listenerConsumerDS = DataStoreFinder.getDataStore(consumerParams)
+      val consumerFC = listenerConsumerDS.getFeatureSource("listeners")
+
+      val latch = new CountDownLatch(numUpdates)
+
+      val fl = new TestLambdaFeatureListener((fe: KafkaFeatureEvent) => {
```
cne1x: [style]  As a reader, it would help me if we replaced variables such as ```fl``` with something more descriptive like, ```featureListener```.  I'm happy to be overridden.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
jnh5y: Yes!  In order to minimize depending on random jars in crazy places, this one-liner is repeated a few places.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
jnh5y: I was close to doing that previously, so yeah, went ahead and did that.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -114,19 +124,86 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
...
-    queue.take() match {
-      case update: CreateOrUpdate => featureCache.createOrUpdateFeature(update)
-      case del: Delete            => featureCache.removeFeature(del)
-      case clr: Clear             => featureCache.clear()
-      case m                      => throw new IllegalArgumentException(s"Unknown message: $m")
+  override def run(): Unit =
+    while (running.get) {
+      queue.take() match {
+        case update: CreateOrUpdate =>
+          fireEvent(KafkaFeatureEvent.changed(this, update.feature))
```
elahrvivaz: does it matter if we fire the event before updating or vice/versa? seems like maybe it should be the other way

jnh5y: For updates and clears, no.

For deletes, since we decided on setting the bounds, and KDS delete control messages do not contain the feature info, we have to retrieve it from the featureCache.

Initially, when I had things in the other order, I couldn't get deletes to work in the listener, and I was confused.  After a little debugging, I found that the ordering matters.  Due to that, I decided to make the order consistent.

jnh5y: Incidentally, if *somehow* a delete message for a 'missing' ID gets sent to Kafka, the listener thread will die (and not restart).  
For that, I'm created: https://geomesa.atlassian.net/browse/GEOMESA-1112

elahrvivaz: yeah, wasn't sure if the listeners would expect the event to have already happened, or that it was about to happen.

jnh5y: I think it is fair to say that the event is happening.  I haven't seen the API care about order....

mmatz-ccri: The JavaDoc says

> Typically fired to signify that a change has occurred in the DataStore backing the FeatureSource. 

So I would think that the order should be reversed.  But, whatever you think...
If you wanted to reverse the order, just save the feature first (val f = ...; delete; fire(this, f))

jnh5y: @anthonyccri correctly noted that we ought to minimize object creation for the more common, no subscribed feature listeners.

I think the most sensible way to support exactly what you are describing would be to push the listeners down into the FeatureCache.

ContentFeatureSouce.addFeatureListeners is sadly final, so doing that would require a bit of hack.

anthonyccri:
Merged.

PR #817, opened 2016-02-25
GEOMESA-1111 Return null for non-existant field names

ddseapy:

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
@@ -221,13 +221,14 @@ object Transformers extends EnhancedTokenParsers with LazyLogging {
   }
 
   case class FieldLookup(n: String) extends Expr {
-    var idx = -1
-    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = {
-      if (idx == -1) {
-        idx = ctx.indexOf(n)
-      }
-      if (idx < 0) null else ctx.get(idx)
+    var doEval: EvaluationContext => Any = { ctx =>
+      val idx = ctx.indexOf(n)
+      doEval =
+        if(idx < 0)  _  => null
+        else         ec => ec.get(idx)
+      doEval(ctx)
     }
+    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = doEval(ctx)
 
     override def dependenciesOf(fieldNameMap: Map[String, Field]): Seq[String] =
       Seq(n) ++ fieldNameMap.get(n).flatMap { f => Option(f.transform).map(_.dependenciesOf(fieldNameMap)) }.getOrElse(Seq.empty)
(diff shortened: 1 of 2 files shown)


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -226,7 +226,7 @@ object Transformers extends EnhancedTokenParsers with LazyLogging {
       if (idx == -1) {
         idx = ctx.indexOf(n)
       }
-      ctx.get(idx)
+      if (idx < 0) null else ctx.get(idx)
```
elahrvivaz: i wonder if we could code this a little smarter - we're going to be calling ctx.indexOf for every eval if the idx is -1. Nothing is springing to mind though...

elahrvivaz: well actually we could just make it default to -2 - but we're still doing 2 checks every time. probably doesn't really matter...

elahrvivaz: we could wrap the get in a try/catch, which would be basically free for the normal case, but make the null case way more expensive...

anthonyccri: How about this?

```scala
     var doEval: EvaluationContext => Any = null
     override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = {
       if (idx == -1) {
         idx = ctx.indexOf(n)
         doEval = 
           if(idx < 0)  _  => null
           else         ec => ec.get(idx)
       }
       doEval(ctx)
     }
```

ddseapy: I can see how this would work if default and the "idx ==" check where both -2 instead of -1.  Then it would only do ctx.indexOf on the first eval() execution.

ddseapy: or if "idx == -1" was "doEval == null", which is a bit clearer, basically saying "if we don't have a doEval defined then define it")

ddseapy: would this work to avoid the check altogether?:
```scala
var doEval: EvaluationContext => Any = { ctx =>
      val idx = ctx.indexOf(n)
      doEval =
        if(idx < 0)  _  => null
        else         ec => ec.get(idx)
      doEval(ctx)
    }
    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = doEval(ctx)
```

elahrvivaz: looks reasonable

elahrvivaz:
hehe - @ddseapy you used master for your working branch?
merged as e5ecf9933a4d46e50d30572ae9ee43399f8caf81

PR #820, opened 2016-03-04
GEOMESA-1118 - Remove geomesa-utils-gs-plugin

ddseapy:
* Update to commons-pool2 to avoid relocating commons-pool in plugin
* Remove geomesa-utils-gs-plugin module


build/cqs.tsv
```diff
@@ -86,6 +86,7 @@ org.apache.commons:commons-compress	1.4.1	compile
 org.apache.commons:commons-csv	1.0	compile
 org.apache.commons:commons-lang3	3.3.2	compile
 org.apache.commons:commons-math	2.1	compile
+org.apache.commons:commons-pool2	2.2	compile
```
ddseapy: CQ here: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10369

pom.xml
```diff
@@ -1300,7 +1300,12 @@
             <dependency>
                 <groupId>commons-pool</groupId>
                 <artifactId>commons-pool</artifactId>
-                <version>1.6</version>
```
ddseapy: This change isn't technically necessary as dependencyManagement in gs-plugin and web modules takes care of making sure we use the provided one (1.5.3) in geoserver WEB-INF/lib.  If issue for cq or preferred, can set back to 1.6.

geomesa-web/geomesa-web-data/README.md
```diff
@@ -290,7 +290,7 @@ curl --header 'Accept: text/plain' --get --data-urlencode 'q=select mySft.myAttr
 | geomesa-jobs-1.2.0-SNAPSHOT.jar | 543025 |
 | geomesa-raster-1.2.0-SNAPSHOT.jar | 165265 |
 | geomesa-security-1.2.0-SNAPSHOT.jar | 39567 |
-| geomesa-utils-gs-plugin-1.2.0-SNAPSHOT.jar | 979554 |
+| geomesa-utils-1.2.0-SNAPSHOT.jar | 979554 |
```
elahrvivaz: can you update the jar size too?

pom.xml
elahrvivaz: there is a cq already for 1.5.4, and 1.6 - we could open one for 1.5.3. Are we actually still using this, or could we mark it provided scope?

ddseapy: it comes in as a transitive dependency in several modules from:
org.geotools:gt-referencing:jar:14.1:compile
org.geotools:gt-opengis:jar:14.1:compile
org.apache.camel:camel-netty4:jar:2.15.1:test

elahrvivaz: let's just set it back to 1.6, as it was working that way and we can avoid another CQ.

elahrvivaz:
ce3df4c4031560914710395413a3bdf2e0590799

PR #830, opened 2016-03-10
Clarification for distribution.

timothymschier:
Had to go looking for the assemble.sh file; this will make it easier.


elahrvivaz:
@timothymschier thanks for the PR. Did the assemble.sh file work for you? When I just ran it, it build the project but didn't find any tar files. Might be better to just remove that file and references to it.

elahrvivaz:
Another note: this is also mentioned in the main README.md and should be changed there too.

timothymschier:
Hi Emilio,

Yeah you’re right I did have issues even with that file. It’s close but doesn’t do a full distribution.
I wrote my own update script for the geomesa-tools which does:

-        mvn install –DskipTests
(this creates the geomesa-tools/target/geomesa-tools-1.2.1-SNAPSHOT-bin.tar.gz file

-        tar –xzvf of that –bin.tar.gz file
(assemble.sh didn’t find this file)

-        copy the unzipped tar to $GEOMESA_HOME to overwrite the previous copy
(assemble.sh doesn’t do this step)

So yes I should retract this PR and that process should go up.

Cheers,

Tim

elahrvivaz:
@timothymschier thanks, we're going to just remove the references to that script, since it doesn't work. Let us know if you find any more issues!

PR #832, opened 2016-03-10
GEOMESA-1008 Fixing dwithin queries

elahrvivaz:


elahrvivaz:
Note: there is a pending question about a needing sign-off from an outside contributor who located the issue.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
```diff
@@ -216,17 +223,25 @@ object QueryPlanner extends LazyLogging {
...
+   * @param sft simple feature type to be queried
+   * @return new query with filters updated
+   */
+  def updateFilter(base: Query, sft: SimpleFeatureType): Query = {
+    val query = new Query(base)
     // add the bbox from the density query to the filter
-    // if configure has been called already, don't re-add - sometimes the bbox is split by
-    // IDL handling, causing us to not detect it as a duplicate
-    if (query.getHints.isDensityQuery && !query.getHints.isConfigured) {
+    if (query.getHints.isDensityQuery) {
```
jahhulbert-ccri: just checking if the comment from before about IDL handling is no longer relevant?

elahrvivaz: Yeah - I split the configure query stuff into two methods - the old configureQuery which can get called multiple times but is now idempotent, and then this method updateFilter, which will only get called once in the query planner just before calculating ranges, etc. It doesn't modify the existing query, it returns a new query - so technically the comment would still be valid, but it would never get called like that anymore. I'll add a comment to the method signature to make it clearer.

elahrvivaz: updated.

jahhulbert-ccri: cool. thnx.

jahhulbert-ccri:
looks good. just gotta check the change from stats itr vs temporal density itr since I'm not familiar with that yet.

elahrvivaz:
@jahhulbert-ccri i merged in master, should be good to go now

jahhulbert-ccri:
k

jahhulbert-ccri:
70deaacb8fedd1f02c55232cb0aae90d04bdb7f3

PR #833, opened 2016-03-11
GEOMESA-1130 Users of KafkaDataStore should be able to specify a filt…

anthonyccri:
…er when consuming features

* Added a method to the data store to subscribe to a feature source with
  a filter.

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
@@ -59,7 +59,9 @@ class KafkaDataStore(override val zookeepers: String,
 
   override def createFeatureSource(entry: ContentEntry) = createFeatureSource(entry, Query.ALL)
   private def createFeatureSource(entry: ContentEntry, query: Query) = {
-    if(query.getStartIndex == null) query.setStartIndex(0)
+    // Query has an issue when start index is not set - hashCode throws an exception
+    // However, Query.ALL has an overridden hashCode that ignores startIndex
+    if(query.getStartIndex == null && !query.equals(Query.ALL)) query.setStartIndex(0)
     featureSourceCache.get(FeatureSourceCacheKey(entry, query))
   }
 
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
@@ -16,10 +16,11 @@ import com.google.common.cache.{CacheBuilder, CacheLoader}
 import com.typesafe.scalalogging.LazyLogging
 import org.geotools.data.DataAccessFactory.Param
 import org.geotools.data.store.{ContentDataStore, ContentEntry, ContentFeatureSource}
-import org.geotools.data.{DataStore, DataStoreFactorySpi}
+import org.geotools.data.{DataStore, DataStoreFactorySpi, Query}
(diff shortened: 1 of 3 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
```diff
@@ -128,8 +128,10 @@ class LiveKafkaConsumerFeatureSource(entry: ContentEntry,
     while (running.get) {
       queue.take() match {
         case update: CreateOrUpdate =>
-          fireEvent(KafkaFeatureEvent.changed(this, update.feature))
-          featureCache.createOrUpdateFeature(update)
+          if(query.getFilter.evaluate(update.feature)) {
```
cne1x: Would this same filter check also apply to deletes?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
```diff
@@ -39,15 +40,28 @@ class KafkaDataStore(override val zookeepers: String,
...
+  def getFeatureSource(typeName: String, filt: Filter): ContentFeatureSource = {
+    val entry = ensureEntry(new NameImpl(typeName))
+    val query = new Query(typeName, filt)
+    query.setStartIndex(0)
+    createFeatureSource(entry, query)
+  }
+
+  override def createFeatureSource(entry: ContentEntry) = createFeatureSource(entry, Query.ALL)
+  private def createFeatureSource(entry: ContentEntry, query: Query) = {
+    if(query.getStartIndex == null) query.setStartIndex(0)
```
elahrvivaz: probably shouldn't use Query.ALL - query.getStartIndex will always be null, and then query.setStartIndex will throw an unsupported operation exception:
https://github.com/geotools/geotools/blob/master/modules/library/api/src/main/java/org/geotools/data/ALLQuery.java#L59
https://github.com/geotools/geotools/blob/master/modules/library/api/src/main/java/org/geotools/data/ALLQuery.java#L224

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
anthonyccri: Deletes do not have the feature . . . only the feature id.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
anthonyccri: Ok, i'll change it to be a local 'All' and check for 'All'

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/LiveKafkaConsumerFeatureSource.scala
cne1x: That's right.  I forget that we only get the DELETE filter back.

From the end-users' point of view, then...
1.  an INSERT/UPDATE will send me an event iff it satisfies my filter;
2.  a DELETE will always sent me an event, even if what was removed does not satisfy my filter

Am I understanding this correctly?

cne1x:
Building this now; if it succeeds, I'll merge this PR.

cne1x:
Merge-squashed as commit b64db18b2fd9430e710874dff29252b38a135e0c

PR #834, opened 2016-03-14
GEOMESA-1132 Exporting avro files through tools

elahrvivaz:

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/ExportCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/ExportCommand.scala
@@ -31,17 +31,17 @@ class ExportCommand(parent: JCommander) extends CommandWithCatalog(parent) with 
   override val params = new ExportParameters
 
   override def execute() = {
-
-    val fmt = params.format.toLowerCase
+    val fmt = Formats.fromString(params.format)
     val features = getFeatureCollection(fmt)
     val exporter: FeatureExporter = fmt match {
-      case CSV | TSV       => DelimitedExport(getWriter(), params)
-      case SHP             => new ShapefileExport(getFile())
-      case GeoJson | JSON  => new GeoJsonExport(getWriter())
-      case GML             => new GmlExport(getOutputStream())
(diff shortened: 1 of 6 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/ExportCommand.scala
```diff
@@ -120,11 +120,11 @@ class ExportCommand(parent: JCommander) extends CommandWithCatalog(parent) with
 object ExportCommand {
   @Parameters(commandDescription = "Export a GeoMesa feature")
   class ExportParameters extends OptionalCqlFilterParameters {
-    @Parameter(names = Array("-F", "--format"), description = "Format to export (csv|tsv|gml|json|shp|bin)")
+    @Parameter(names = Array("-F", "--format"), description = "Format to export (csv|tsv|gml|json|shp|bin|avro)")
     var format: String = "csv"
 
-    @Parameter(names = Array("-m", "--max-features"), description = "Maximum number of features to return. default: Long.MaxValue")
-    var maxFeatures: Integer = Int.MaxValue
+    @Parameter(names = Array("-m", "--max-features"), description = "Maximum number of features to return. default: Unlimited")
```
jahhulbert-ccri: Is it actually Int.MaxValue because of geotools?

jahhulbert-ccri:
looks good. i'll verify one on the command line.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/ExportCommand.scala
elahrvivaz: I think that if max features isn't set, we'll just keep returning values forever... if it is set, then it has to fit inside an Int. I could be wrong about that...

jahhulbert-ccri: I really hope its unlimited...we should PR geotools for using a long #bigdata

jahhulbert-ccri:
7c295d68bad92291e4260273134219dd0f938faf

PR #835, opened 2016-03-15
GEOMESA-1133 Auto detection of simple feature types and conversions for Avro and delimited text input files

elahrvivaz:
* Moved ConverterInputFormat into geomesa-jobs
* Added AvroInputFormat for reading avro data files

~~Note: stacks on top of #834~~


jahhulbert-ccri:
@elahrvivaz can you check out the build fail?

elahrvivaz:
@jahhulbert-ccri there's an intermittent failure in one of the metrics tests (not part of this PR). I noticed this in a local build a couple days ago... I'll file an issue for it.

jahhulbert-ccri:
@elahrvivaz XML ingest is failing. I believe that ```org.locationtech.geomesa.tools.Utils.Formats``` needs to be extended to use XML for command. However, that brings up the issue that if a user puts a converter in the lib/ directory we should be able to ingest it without it being in Formats

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
```diff
@@ -27,19 +27,32 @@ class IngestCommand(parent: JCommander) extends Command(parent) with LazyLogging
...
+      tryDs.dispose()
 
-      new ConverterIngest(dsParams, sft, converterConfig, params.files, params.threads).run()
+      if (params.spec == null && params.config == null && Seq(TSV, CSV, AVRO).contains(fmt)) {
+        if (params.featureName == null) {
+          throw new ParameterException("Feature name is required when a schema is not specified")
+        }
+        // auto-detect the import schema
+        logger.info("No schema or converter defined - will attempt to detect schema from input files")
+        new AutoIngest(dsParams, params.featureName, params.files, params.threads, fmt).run()
```
jahhulbert-ccri: britt will be so happy

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
```diff
@@ -27,19 +27,32 @@ class IngestCommand(parent: JCommander) extends Command(parent) with LazyLogging
...
       val dsParams = new DataStoreHelper(params).paramMap
-      require(DataStoreFinder.getDataStore(dsParams) != null, "Could not load a data store with the provided parameters")
-      val sft = CLArgResolver.getSft(params.spec, params.featureName)
-      val converterConfig = CLArgResolver.getConfig(params.config)
+      val tryDs = DataStoreFinder.getDataStore(dsParams)
+      require(tryDs != null, "Could not load a data store with the provided parameters")
+      tryDs.dispose()
 
-      new ConverterIngest(dsParams, sft, converterConfig, params.files, params.threads).run()
+      if (params.spec == null && params.config == null && Seq(TSV, CSV, AVRO).contains(fmt)) {
```
jahhulbert-ccri: picky - Might be useful to have a comment here explaining AutoIngest when spec/config are null

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
```diff
@@ -27,19 +27,32 @@ class IngestCommand(parent: JCommander) extends Command(parent) with LazyLogging
   override val params = new IngestParameters()
 
   override def execute(): Unit = {
-    val fmt = Formats.fromString(Option(params.format).getOrElse(getFileExtension(params.files(0))))
+    val extensions = params.files.map(getFileExtension)
+    require(extensions.tail.forall(_ == extensions.head), "Input files must all be of the same file type")
```
jahhulbert-ccri: may want to rethink this one as well in relation to non-standard converters

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AutoIngestDelimited.scala
```diff
@@ -0,0 +1,171 @@
...
+ * geomesa tools export, for convenience.
+ *
+ * The first line of each file is expected to be a delimited (tab or comma) list
+ * of attributes for a simple feature type, starting with the reserved word 'id'.
+ * For example:
+ *
+ * id,name:String,dtg:Date,geom:Point:srid=4326
+ *
+ * The first line of the file will be used to create a simple feature type.
+ * All subsequent lines must be type convertible to the appropriate simple feature attribute.
```
jahhulbert-ccri: What do you think about having the user provide the spec as before on the command line to --spec or -s and then that just lines up with the existing csv row (I guess it would be a little annoying if you need to skip header). Thought pattern being the user has a CSV file and does need to edit the CSV file (it may be in S3 or HDFS already for example) in order to AutoIngest it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/FeatureExporter.scala
```diff
@@ -171,19 +155,18 @@ class DelimitedExport(writer: Writer, format: Formats, attributes: Option[String
...
   }
 
-  def stringify(o: Object): String = o match {
-    case null    => ""
-    case d: Date => dateFormat.format(d)
-    case _       => o.toString
+  def stringify(o: Object): Object = o match {
+    case g: Geometry => WKTUtils.write(g)
+    case d: Date     => GeoToolsDateFormat.print(d.getTime)
+    case _           => o
```
jahhulbert-ccri: Not specifically this PR but I guess we should probably deal with List and Map at some point safely here. Haven't looked to see what they do recently but I assume that java.util.Map might not be friendly :)

jahhulbert-ccri:
I'm wondering if our Converters aren't setting the USE_PROVIDED_FID field...Local mode vs MapReduce gives different action on the FeatureWriter. Local uses the FID (as evidenced in the LocalIngest coping the FID in) and GeoMesaOutputFormat seems to require the USE_PROVIDED_FID Hint be set properly. I'll recompile master to verify but here's what I got from exporting a couple of rounds of ingest...After MapReduce it has more UUID features:

```
0093d2fc-f827-41f9-985d-88d6b3f96efd,23623,Harry,20,2015-05-06T00:00:00.000Z,"[Will, Mark, Suzan]",POINT (-100.236523 23)
0093d2fc-f827-41f9-b55e-590914217083,23623,Harry,20,2015-05-06T00:00:00.000Z,"[Will, Mark, Suzan]",POINT (-100.236523 23)
0093d2fc-f827-41f9-b9fe-a6b218500ba8,23623,Harry,20,2015-05-06T00:00:00.000Z,"[Will, Mark, Suzan]",POINT (-100.236523 23)
0093d2fc-f827-41f9-bc0f-4622a2cf8732,23623,Harry,20,2015-05-06T00:00:00.000Z,"[Will, Mark, Suzan]",POINT (-100.236523 23)
3233,3233,Severus,30,2015-10-23T00:00:00.000Z,"[Tom, Riddle, Voldemort]",POINT (3 -62.23)
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
```diff
@@ -0,0 +1,235 @@
...
+                written.incrementAndGet()
+              } catch {
+                case e: Exception =>
+                  logger.error(s"Failed to write '${DataUtilities.encodeFeature(toWrite)}'", e)
+                  failed.incrementAndGet()
+              }
+              bytesRead.addAndGet(countingStream.getCount)
+              countingStream.resetCount()
+            }
+          } finally {
```
jahhulbert-ccri: It's a little gross but I got an exception here in this try from the AvroDataFile reader and it was swallowed bc of IOUtils.closeQuietly(fw)...the lazy eval hits in the finally block causing an exception which gets caught and the original is lost!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
elahrvivaz: added

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/commands/IngestCommand.scala
elahrvivaz: k, i'll take it out.

elahrvivaz: Would it be valid to say that if they don't specify a format, then all the files have to be the same type? Otherwise we base the ingest off the extension of the first file, which seems like it could lead to some issues and/or confusion.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AutoIngestDelimited.scala
elahrvivaz: Ah, I'm not sure it would handle that original use case - you would still have to modify the file to add the header row.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/FeatureExporter.scala
elahrvivaz: Did you try the auto-import with lists and maps?

elahrvivaz:
It makes sense that the ID isn't getting set in the m/r job, but I can't figure out how it was working before... anyway I'll fix it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
elahrvivaz: Yeah, that makes sense... if the avro file is messed up, then the schema will never get created, and the call to get feature writer will fail. I'll make this a var or something.

elahrvivaz: fixed

elahrvivaz:
removed check for all files being the same - note we could have some inconsistencies if you try to ingest different file types as we use the first file to detect the type if it's not specified.

elahrvivaz:
@jahhulbert-ccri Thanks for checking this out. I think everything is fixed with the following questions still left open:
1. Should we handle lists/maps in this PR? I haven't tried auto-ingesting them, at least with csv it might not work.
2. Each file has to have a header row that defines the SFT. Should we allow this to be overridden/replaced with a command line arg?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AutoIngestDelimited.scala
jahhulbert-ccri: hmmm. yeah. We can make another ticket maybe to ingest a CSV file or TSV while passing in the --spec option.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/FeatureExporter.scala
jahhulbert-ccri: Not yet. I'll try it

jahhulbert-ccri:
@elahrvivaz 
1. We can do Lists/Maps as a follow-on PR using the defaults that we use for the converters. I can get to that tomorrow unless you want to look at it.
2. I'd prefer the idea of using just the --spec command line arg since that way the user doesn't have to edit the file.

elahrvivaz:
Created follow up tickets here:
https://geomesa.atlassian.net/browse/GEOMESA-1138
https://geomesa.atlassian.net/browse/GEOMESA-1139

jahhulbert-ccri:
passed cloud local litmus tests https://github.com/jahhulbert-ccri/geomesa-test/blob/master/bin/test-geomesa.sh

jahhulbert-ccri:
gonna merge tonight

jahhulbert-ccri:
c6babb89d5fa0a9a4f21adac56ab41ea92b3b2c7

PR #837, opened 2016-03-20
WIP native API

anthonyccri:
For comments


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftBuilder.scala
```diff
@@ -58,6 +58,7 @@ abstract class InitBuilder[T] {
   def floatType  (name: String, opts: Opts = Opts()) = append(name, opts, "Float")
   def doubleType (name: String, opts: Opts = Opts()) = append(name, opts, "Double")
   def booleanType(name: String, opts: Opts = Opts()) = append(name, opts, "Boolean")
+  def binaryType (name: String, opts: Opts = Opts()) = append(name, opts, "Array[Byte]")
```
anthonyccri: Oops, that should go way.

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaQuery.java
```diff
@@ -0,0 +1,51 @@
...
+        public GeoMesaQuery build() {
+            GeoMesaQuery query = new GeoMesaQuery();
+            query.filt = ff.and(
+                    ff.bbox("geom", minx, maxx, miny, maxy, "EPSG:4326"),
+                    ff.between(ff.property("dtg"), ff.literal(start), ff.literal(end)));
+            return query;
+        }
+    }
+
+    public Filter getFilt() {
```
jnh5y: Can we say Filter instead of Filt?

geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala
```diff
@@ -0,0 +1,82 @@
...
+    val ds =
+      DataStoreFinder.getDataStore(Map(
+        AccumuloDataStoreParams.tableNameParam.key -> name,
+        AccumuloDataStoreParams.zookeepersParam.key -> zk,
+        AccumuloDataStoreParams.instanceIdParam.key -> instanceId,
+        AccumuloDataStoreParams.userParam.key -> user,
+        AccumuloDataStoreParams.passwordParam.key -> pass,
+        AccumuloDataStoreParams.mockParam.key -> (if(mock) "TRUE" else "FALSE")
+      )).asInstanceOf[AccumuloDataStore]
+    new AccumuloGeoMesaIndex[T](ds, name, valueSerializer)
```
jnh5y: Hmm...  I'm having a moment of trying to decide if the valueSerializer should be fixed for the DataStore in some concrete way...

geomesa-native-api/src/test/java/org/locationtech/geomesa/api/GeoMesaIndexTest.java
```diff
@@ -0,0 +1,81 @@
...
+            return gson.fromJson(new String(bytes), DomainObject.class);
+        }
+    }
+
+    @Test
+    public void testNativeAPI() {
+        DomainObject one = new DomainObject("1", 1, 1.0);
+        DomainObject two = new DomainObject("2", 2, 2.0);
+
+        GeoMesaIndex<DomainObject> index = AccumuloGeoMesaIndex.build("hello", "zoo1:2181", "mycloud", "myuser", "mypass", true, new DomainObjectValueSerializer());
```
jnh5y: (picky) Long lines;)

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    Iterable<T> query(GeoMesaQuery query);
+
+    /**
+     * Put a value in the GeoMesa index
+     * @param t
+     * @param geometry
+     * @param dtg
+     */
+    void put(T t, Geometry geometry, Date dtg);
```
cne1x: From the perspective of designing the API, is there any value to sub-dividing ```put``` into ```insert``` and ```update```?  It feels as if we might want to do so, in the event that we would like there ever to be special handling for how to update existing records (beyond the GeoMesa default behavior).  Just a thought.

geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala
```diff
@@ -0,0 +1,82 @@
...
+import org.locationtech.geomesa.accumulo.data.tables.Z3Table
+import org.locationtech.geomesa.accumulo.data.{AccumuloDataStore, AccumuloDataStoreParams}
+import org.locationtech.geomesa.utils.geotools.{Conversions, SftBuilder}
+
+class AccumuloGeoMesaIndex[T](ds: AccumuloDataStore,
+                              fname: String,
+                              serde: ValueSerializer[T]) extends GeoMesaIndex[T] {
+
+  // TODO: figure out how to use Array[Byte] for payload
+  val base32Encoder = BaseEncoding.base32()
```
elahrvivaz: you can reference this issue that we already filed: GEOMESA-823

geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  override def query(query: GeoMesaQuery): Iterable[T] = {
+    import Conversions._
+
+    import scala.collection.JavaConverters._
+
+    fs.getFeatures(query.getFilt)
+      .features()
+      .map { f => serde.fromBytes(base32Encoder.decode(f.getAttribute(1).asInstanceOf[String])) }
+      .toIterable.asJava
```
elahrvivaz: I think the self-closing behavior persists through the call to .toIterable... would it make sense to have the query method return Iterator with Closeable?

elahrvivaz:
Nice. It would be useful (but not necessary) to have some default serializer implementations that are just light wrappers around common frameworks, like java serialization, kryo, avro, etc.

geomesa-native-api/src/test/java/org/locationtech/geomesa/api/GeoMesaIndexTest.java
anthonyccri: That's java for ya!

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java
anthonyccri: Good point.

geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala
anthonyccri: Yes, definitely.

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java
elahrvivaz: It is a good idea. Update might be tricky in this case as feature ID is not explicitly defined anywhere...

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaQuery.java
anthonyccri: Yes

geomesa-native-api/src/test/java/org/locationtech/geomesa/api/GeoMesaIndexTest.java
jahhulbert-ccri: prob should put final in front of every variable

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaQuery.java
```diff
@@ -0,0 +1,51 @@
...
+public class GeoMesaQuery {
+
+    private Filter filt = Filter.INCLUDE;
+
+    public static class GeoMesaQueryBuilder {
+        private static FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
+        private Double minx, miny, maxx, maxy;
+        private Date start;
+        private Date end;
+        public static GeoMesaQueryBuilder builder() {
```
jahhulbert-ccri: perhaps shorten to name the nested class Builder with a public no arg constructore to shorten life...though i've forgotten my java style.

```
GeomesaQuery query = new GeomesaQuery.Builder().within(1,2,3,4).during(foo).build()
```

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/ValueSerializer.java
```diff
@@ -0,0 +1,13 @@
+package org.locationtech.geomesa.api;
+
+/**
+ * Required by GeoMesaIndex in order to serialize
+ * a value object into the index
+ * @param <T>
+ */
+public interface ValueSerializer<T> {
```
jahhulbert-ccri: later on down the road might be nice to provide ```reuse``` objects in the api but this is OK for now

geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala
jahhulbert-ccri: yeah its probably best to make the Serializer association by classname instead of by value? Or some other loading mechanism.

geomesa-native-api/src/test/java/org/locationtech/geomesa/api/GeoMesaIndexTest.java
jahhulbert-ccri: you can always create the classic Java solution of making an Opts Class or a Config class to avoid all the parameter passing in the method :)

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java
jahhulbert-ccri: Yeah I guess the question is do you want an explicit feature ID type API and if so you'll want to have ```get``` ```insert``` and ```update``` methods.  I've always thought a ```put``` implies some sort of key or identifier anyways so I'd change this to ```insert``` even if we don't have an explicit id (though in English put this thing at this time and geometry sounds appealing)

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Delete a value from the GeoMesa index
+     * @param t
+     */
+    void delete(T t);
```
jahhulbert-ccri: delete gets easier if you go with the feature ID :) since geomesa can support that. However, if you only have the Z3 index the delete of a named object is harder and the better syntax would be something like ```delete where <query>``` like you do in SQL. Else you'd have to keep a ref to the object identifier somehow since you'd have no way to find it in the Z3 index alone and we certainly don't want to store a records table in this case I don't think.

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java
jahhulbert-ccri: One idea to help with deletes below is to make the signature something like

```byte[] insert(T t, Geometry geometry, Date dtg)``` where the return ```byte[]``` or of course ```base64 String``` could be a returned ID which is compound from the Z3 bytes followed by a feature ID that you enforce the user provides. This also provides a distinct lookup for a record for delete or update.

**Comment by jahhulbert-ccri** [comment]
@elahrvivaz is right about the serializers...also we'll probably have to support Polygon queries out of the box so we can draw on maps.

File: geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java (hunk 5 above)
Comment (jahhulbert-ccri) [inline geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaIndex.java]: if we only use the Z3 index it's a little rough to actually do updates and deletes without knowing the geometry beforehand

**Comment by jnh5y** [comment]
I think the native API could be generalized to allow for three more features:

1.  The ability to extract indexing information from the Native object.
2.  The ability to configure additional secondary indices.
3.  The ability to store the serialized thing in table, in another table, or flat out elsewhere.

With 1 & 3, the blobstore stores to look like a 'special case' for the Native API.  With 2, we can potentially have support more interesting queries, etc.

File: geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala [hunk 11]
```diff
@@ -0,0 +1,103 @@
...
+
+class AccumuloGeoMesaIndex[T](ds: AccumuloDataStore,
+                              fname: String,
+                              serde: ValueSerializer[T]) extends GeoMesaIndex[T] {
+  private val ff = CommonFactoryFinder.getFilterFactory2
+
+  if(!ds.getTypeNames.contains(fname)) {
+    val sft =
+      new SftBuilder()
+        .date("dtg", default = true, index = true)
```
cne1x: Do we ever want just a Z2 on location?

geomesa-native-api/src/main/scala/org/locationtech/geomesa/api/AccumuloGeoMesaIndex.scala
```diff
@@ -0,0 +1,103 @@
...
+      .features()
+      .map { f => serde.fromBytes(f.getAttribute(1).asInstanceOf[Array[Byte]]) }
+      .toIterable.asJava
+  }
+
+  override def insert(value: T, geom: Geometry, dtg: Date): String = {
+    val bytes = serde.toBytes(value)
+    val fw = writers.get(fname)
+    val sf = fw.next()
+    val id = Z3UuidGenerator.createUuid(geom, dtg.getTime).toString
```
cne1x: Do we intend to support user-provided IDs?  Clearly, the serialized native object will contain its own ID, but aren't there use-cases in which the consumer wants to query based on the native IDs rather than our synthetic ones?

jnh5y:
Merged as faab277af1c2b92bc2d860b607b89cd1edb7ce7b.

PR #838, opened 2016-03-21
GEOMESA-1140 Use Attr Idx Keys for Query

jahhulbert-ccri:
* Fulfill queries using the keys AND values of the attribute index
  instead of just values

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
@@ -71,12 +71,21 @@ class AttributeIdxStrategy(val filter: QueryFilter) extends Strategy with LazyLo
     val priority = FILTERING_ITER_PRIORITY
 
     // query against the attribute table
-    val singleTableScanPlan: ScanPlanFn = (schema, filter, transform) => {
+    val singleAttrValueOnlyPlan: ScanPlanFn = (schema, filter, transform) => {
       val iterators = if (filter.isDefined || transform.isDefined) {
         Seq(KryoLazyFilterTransformIterator.configure(schema, filter, transform, priority))
       } else {
         Seq.empty
       }
+      // need to use transform to convert key/values if it's defined
+      val kvsToFeatures = queryPlanner.kvsToFeatures(transform.map(_._2).getOrElse(schema))
+      BatchScanPlan(attrTable, ranges, iterators, Seq.empty, kvsToFeatures, attrThreads, hasDupes)
+    }
+
+    // query against the attribute table
+    val singleAttrPlusValuePlan: ScanPlanFn = (schema, filter, transform) => {
+      val iterators = Seq(AttrKeyPlusValueIterator.configure(sft, schema, attributeSftIndex, filter, transform, priority))
+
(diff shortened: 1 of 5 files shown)


jahhulbert-ccri:
FYI need to backport to 1.1.x will do once its finalized

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
```diff
@@ -116,19 +125,22 @@ class AttributeIdxStrategy(val filter: QueryFilter) extends Strategy with LazyLo
...
         }
       }
     } else if (descriptor.getIndexCoverage() == IndexCoverage.FULL) {
       // we have a fully encoded value - can satisfy any query against it
-      singleTableScanPlan(sft, filter.secondary, hints.getTransform)
-    } else if (IteratorTrigger.canUseIndexValues(sft, filter.secondary, transform)) {
-      // we can use the index value
+      singleAttrValueOnlyPlan(sft, filter.secondary, hints.getTransform)
+    } else if (IteratorTrigger.canUseAttrIdxValues(sft, filter.secondary, transform)) {
+      // we can use the index value. Transform can only include
```
elahrvivaz: partial comment?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/AttrKeyPlusValueIterator.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    // Setter functions to pull values from the index SF into the projected trans SF
+    import scala.collection.JavaConversions._
+    val setters: Seq[(KryoBufferSimpleFeature, ScalaSimpleFeature) => Unit] =
+      transSft.getAttributeDescriptors
+        .map(_.getLocalName)
+        .filter(n => idxSft.getAttributeDescriptors.exists(_.getLocalName == n))
+        .map { name =>
+          val transIdx = transSft.getAttributeDescriptors.indexWhere(_.getLocalName == name)
+          val indexIdx = idxSft.getAttributeDescriptors.indexWhere(_.getLocalName == name)
```
elahrvivaz: [minor] you can just do idxSft.indexOf(name)

elahrvivaz:
looks good - do we care about bin queries or stat aggregating queries?

elahrvivaz:
[idly] i wonder if it would be worthwhile to use the regular kryo iterator, and stack another iterator on top that adds the attribute value to the return feature. would be slower (has to transform twice), but would keep the codebase simpler, and might be ok for the relatively small-ish queries this use case is targetting.

jahhulbert-ccri:
@elahrvivaz yeah i was wondering the same thing about stacking itrs. I wonder how bad the perf hit would be. Expected result sizes range generally from 50 to 800 to upwards of 50k entries returned. Usually most are are expected to be 1k-10k. If anyone else reads this maybe they have an opinion. Either is OK for me. I'll look at what it takes to stack...It would definitely make it simpler.

We do care about bin and stat aggregating queries in the long term for sure. I'll prob make a ticket for that though. We can always refactor that in.

elahrvivaz:
b9afce00075274441cae7744e0cbe1816cd6b0d9

PR #846, opened 2016-03-29
GEOMESA-1123 Adding max features for bin queries

elahrvivaz:
Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

~~Note: stacks on top of #845~~

PR #850, opened 2016-03-30
GEOMESA-1152 Adding Count iterator

elahrvivaz:
Note: stacks on top of #846

* Added geometry and string implementations for most Stats
* Added kryo encoding for stats
* Refactored aggregating iterators


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -32,35 +38,47 @@ trait Stat {
   def observe(sf: SimpleFeature): Unit
 
   /**
-   * Meant to be used to combine two Stats of the same subtype.
-   * Used in the "reduce" step client-side.
+   * Add another stat to this stat. Avoids allocating another object.
```
jnh5y: Good call.

PR #853, opened 2016-04-09
GEOMESA-951,GEOMESA-1023,GEOMESA-1082 Auths GDALFileHandler and Java API

AndrewAnnex:
* Added GDALFileHander to BlobStore, supports GDAL raster formats
* Added Java Interfaces
* Fixed deletes


geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/interop/GeoMesaBlobStore.java
```diff
@@ -0,0 +1,72 @@
...
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+
+package org.locationtech.geomesa.blob.core.interop;
+
+import org.geotools.data.Query;
+import org.opengis.filter.Filter;
+import scala.Option;
```
anthonyccri: It would be preferable to not use Scala classes in the Java API.

build/cqs.tsv
```diff
@@ -1,3 +1,4 @@
+asm:asm	3.1	compile
```
jnh5y: I know you may have just re-run a script to generate this.  Did you need the change in version of ASM?

build/cqs.tsv
```diff
@@ -113,6 +114,7 @@ org.codehaus.jackson:jackson-mapper-asl	1.9.3	compile
 org.eclipse.emf:common	2.6.0	compile
 org.eclipse.emf:ecore	2.6.1	compile
 org.eclipse.xsd:xsd	2.6.0	compile
+org.gdal:gdal	2.0.0	compile
```
jnh5y: Do we have a CQ for this?

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.io.File;
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+public class AccumuloGeoMesaBlobStore implements GeoMesaBlobStore {
+
+    public AccumuloBlobStore accumuloBlobStore;
```
elahrvivaz: this should probably be private or at least protected/package protected

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    public AccumuloGeoMesaBlobStore(String instanceId, String tableName, String zookeepers, String user, String password) throws Exception {
+        Map<String, Serializable> dataStoreParams = new HashMap<>();
+        dataStoreParams.put("instanceId", instanceId);
+        dataStoreParams.put("tableName", tableName);
+        dataStoreParams.put("zookeepers", zookeepers);
+        dataStoreParams.put("user", user);
+        dataStoreParams.put("password", password);
```
elahrvivaz: prefer using the constants from AccumuloDataStoreFactory

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Fetches Blob by id
+     *
+     * @param id String feature Id of the Blob, from getIds functions
+     * @return Tuple2 of (blob, filename)
+     */
+    @Override
+    public Tuple2<byte[], String> get(String id) {
```
elahrvivaz: Like @anthonyccri said, keep scala classes out of the public method signatures

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -45,29 +48,30 @@ class AccumuloBlobStore(ds: AccumuloDataStore) extends LazyLogging with BlobStor
   val bw = connector.createBatchWriter(blobTableName, bwc)
```
elahrvivaz: we should close this batch writer during cleanup somehow... add a ticket at least

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/README.md
```diff
@@ -0,0 +1,8 @@
+#GeoMesa GDALFileHandler
```
elahrvivaz: Nice - please add some docs to the geomesa documentation as well.

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
```diff
@@ -0,0 +1,107 @@
...
+      false
+  }
+
+  override def getGeometryFromFile(file: File): Option[Geometry] = {
+    if (!initialized) {
+      None
+    } else {
+      val nitfFile = gdal.Open(file.getAbsolutePath, gdalconstConstants.GA_ReadOnly)
+      val geom = getImageBoundsGeometry(nitfFile)
+      nitfFile.delete()
```
elahrvivaz: this should probably be in a try/finally

elahrvivaz:
Food for thought - we can support byte arrays natively in simple features now - maybe the concept of a blob-store is outdated...

build/cqs.tsv
AndrewAnnex: I am not depending on it directly as far as I know, I will look into why this was added

build/cqs.tsv
AndrewAnnex: yes and it was approved

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
AndrewAnnex: sure

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
AndrewAnnex: ahh yep I should do that

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
AndrewAnnex: I didn't have a good idea for using something other than tuple, maybe just return a java map?

AndrewAnnex:
I think there may still be a need for a pure binary blob table of some kind (lookups are very fast) but that depends on how the binary arrays were implemented. Otherwise the FileHandler SPI is still useful so maybe a refactor is in order to further simplify things.

jnh5y:
@elahrvivaz @AndrewAnnex Considering that blobs could be kilobytes or megabytes, yeah, the blobstore is a sensible idea.

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
anthonyccri: How about a simple struct class like ```BlobEntry```.

anthonyccri:
Agree with @jnh5y .  Blobstore is useful for large blobs where we want an index of the blobs independent of the actual blobs.

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/interop/GeoMesaBlobStore.java
jahhulbert-ccri: yeah. actually isn't the convention in Java usually to return a null and doc it in the javadoc ? Yeah. follow this convention: https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#put(K,%20V) perhaps

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/interop/GeoMesaBlobStore.java
```diff
@@ -0,0 +1,72 @@
...
+ * An interface to define how users may ingest and query a GeoMesa BlobStore
+ */
+public interface GeoMesaBlobStore {
+
+    /**
+     * Add a File to the blobstore, relying on available FileHandlers to determine ingest
+     * @param file File to ingest
+     * @param params Map String to String, see AccumuloBlobStore for keys
+     */
+    Option<String> put(File file, Map<String, String> params);
```
jahhulbert-ccri: see above about put

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/interop/GeoMesaBlobStore.java
AndrewAnnex: well we followed a slightly different pattern by returning an option of the blob's key (which is a simplefeature id). I can switch over to that pattern, but I will have to change a number of unit tests. I would prefer to return the blob's key as a string (or null) instead of the value

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
AndrewAnnex: the call to delete doesn't actually delete the file from the system but frees native resources (http://gdal.org/java/org/gdal/gdal/Dataset.html#delete()). I probably will copy the null check from the canProcess function but a try/finally is likely not needed here as the geom is already an Option.

build/cqs.tsv
AndrewAnnex: it looks like spark-core is bringing this in via jersey-server. So it doesn't look like I did anything...

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/interop/GeoMesaBlobStore.java
AndrewAnnex: I went ahead and decided to return a Map.Entry<String, byte[]>.

AndrewAnnex: that is for the 'get' functions so far

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
AndrewAnnex: added a ticket with a todo the line above

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
elahrvivaz: it sounds just like a close(). Typical pattern is:
open resource
try { stuff } finall { close resource }
please update

AndrewAnnex:
@jahhulbert-ccri @jnh5y can we get this merged in please

jahhulbert-ccri:
i'll look through it

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class AccumuloGeoMesaBlobStore implements GeoMesaBlobStore {
+
+    protected AccumuloBlobStore accumuloBlobStore;
+
+    public AccumuloGeoMesaBlobStore(Map<String, Serializable> dataStoreParams) throws Exception {
+        AccumuloDataStoreFactory accumuloDataStoreFactory = new AccumuloDataStoreFactory();
+        AccumuloDataStore ds = (AccumuloDataStore) accumuloDataStoreFactory.createDataStore(dataStoreParams);
+        if (ds == null) {
+            throw new Exception("Error initializing AccumuloGeoMesaBlobStore");
```
jahhulbert-ccri: @elahrvivaz any opinion on making this an illegal argument exception and removing ```Exception``` from the signature? @AndrewAnnex picky comment - maybe change comment to "Unable to locate Accumulo DataStore with given parameters" instead of just can't initialize since it'll tell us that via the stacktrace

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.io.File;
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+public class AccumuloGeoMesaBlobStore implements GeoMesaBlobStore {
+
+    protected AccumuloBlobStore accumuloBlobStore;
```
jahhulbert-ccri: make it ```protected final``` to force initialization via constructor

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
elahrvivaz: the method should probably throw IOException as that's what DataStoreFinder throws...
Better error messages are always good

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public AccumuloGeoMesaBlobStore(Map<String, Serializable> dataStoreParams) throws Exception {
+        AccumuloDataStoreFactory accumuloDataStoreFactory = new AccumuloDataStoreFactory();
+        AccumuloDataStore ds = (AccumuloDataStore) accumuloDataStoreFactory.createDataStore(dataStoreParams);
```
jahhulbert-ccri: don't need to cast here

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+public class AccumuloGeoMesaBlobStore implements GeoMesaBlobStore {
+
+    protected AccumuloBlobStore accumuloBlobStore;
+
+    public AccumuloGeoMesaBlobStore(Map<String, Serializable> dataStoreParams) throws Exception {
+        AccumuloDataStoreFactory accumuloDataStoreFactory = new AccumuloDataStoreFactory();
```
jahhulbert-ccri: preference add final keywords :)

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -9,31 +9,36 @@
...
 import org.opengis.filter.Filter
 
 import scala.collection.JavaConversions._
 import scala.util.control.NonFatal
 
-class AccumuloBlobStore(ds: AccumuloDataStore) extends LazyLogging with BlobStoreFileName {
+class AccumuloBlobStore(gds: DataStore) extends GeoMesaBlobStore
+  with BlobStoreFileName with LazyLogging {
 
+  private val ds = gds.asInstanceOf[AccumuloDataStore]
```
elahrvivaz: was there any reason to make this a private val instead of leaving it as the constructor signature? I would change it back

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
```diff
@@ -0,0 +1,115 @@
...
+class GDALFileHandler extends AbstractFileHandler with LazyLogging {
+  import GDALFileHandler._
+
+  val geomFactory = new GeometryFactory()
+  val initialized = Try(gdal.AllRegister()) match {
+    case Success(v) =>
+      logger.info("Successfully loaded GDAL native interface bindings")
+      true
+    case Failure(e) =>
+      logger.error("Failed to load GDAL native libraries in GeoMesa GDAL FileHandler, Ensure that the path of the native GDAL libraries and JNI libraries are in the LD_LIBRARY_PATH environment variable", e)
```
elahrvivaz: excessively long line, even for you @AndrewAnnex

elahrvivaz: Also, 'Ensure' should be lower case

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  override def getGeometryFromFile(file: File): Option[Geometry] = {
+    if (!initialized) {
+      None
+    } else {
+      val gdalFile = gdal.Open(file.getAbsolutePath, gdalconstConstants.GA_ReadOnly)
+      if (gdalFile == null) {
+        None
+      } else {
+        var geom: Option[Geometry] = None
```
elahrvivaz: bad scala practice - don't need the var at all. rewrite as:
```
try {
  getImageBoundsGeom(dgalFile)
finally {
  gdalFile.delete()
}
```

geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/GeoMesaBaseDataStoreServlet.scala
```diff
@@ -0,0 +1,43 @@
...
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.web.core
+
+import org.locationtech.geomesa.utils.cache.FilePersistence
+
+trait GeoMesaBaseDataStoreServlet extends GeoMesaScalatraServlet {
```
elahrvivaz: is this supposed to be part of this PR? Seems like a duplicate of org.locationtech.geomesa.web.core.PersistentDataStoreServlet

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,129 @@
...
+                                    String password,
+                                    String auths) throws Exception {
+        Map<String, Serializable> dataStoreParams = new HashMap<>();
+        dataStoreParams.put(AccumuloDataStoreParams.instanceIdParam().key, instanceId);
+        dataStoreParams.put(AccumuloDataStoreParams.tableNameParam().key, tableName);
+        dataStoreParams.put(AccumuloDataStoreParams.zookeepersParam().key, zookeepers);
+        dataStoreParams.put(AccumuloDataStoreParams.userParam().key, user);
+        dataStoreParams.put(AccumuloDataStoreParams.passwordParam().key, password);
+        dataStoreParams.put(AccumuloDataStoreParams.authsParam().key, auths);
+        new AccumuloGeoMesaBlobStore(dataStoreParams);
```
jahhulbert-ccri: does this actually work in Java ?

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
jahhulbert-ccri: yeah either remove the asInstanceOf or put it in the signature

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -42,32 +47,39 @@ class AccumuloBlobStore(ds: AccumuloDataStore) extends LazyLogging with BlobStor
...
-  def get(id: String): (Array[Byte], String) = {
-    // TODO: Get Authorizations using AuthorizationsProvider interface
-    // https://geomesa.atlassian.net/browse/GEOMESA-986
-    val scanner = connector.createScanner(blobTableName, new Authorizations())
+  def get(id: String): JMap.Entry[String, Array[Byte]] = {
+    val ret = getInternal(id)
+    Maps.immutableEntry(ret._1, ret._2)
+  }
+
+  def getInternal(id: String): (String, Array[Byte]) = {
```
jahhulbert-ccri: make private?

geomesa-blobstore/geomesa-blobstore-core/src/main/scala/org/locationtech/geomesa/blob/core/AccumuloBlobStore.scala
```diff
@@ -42,32 +47,39 @@ class AccumuloBlobStore(ds: AccumuloDataStore) extends LazyLogging with BlobStor
...
-  def getIds(query: Query): Iterator[String] = {
+  def getIds(query: Query): JIterator[String] = {
     fs.getFeatures(query).features.map(_.getAttribute(idFieldName).asInstanceOf[String])
   }
 
-  def get(id: String): (Array[Byte], String) = {
-    // TODO: Get Authorizations using AuthorizationsProvider interface
-    // https://geomesa.atlassian.net/browse/GEOMESA-986
-    val scanner = connector.createScanner(blobTableName, new Authorizations())
+  def get(id: String): JMap.Entry[String, Array[Byte]] = {
```
jahhulbert-ccri: add override to all the defs that implement interface?

geomesa-blobstore/geomesa-blobstore-core/src/test/java/org/locationtech/geomesa/blob/core/AccumuloGeoMesaBlobStoreTest.java
```diff
@@ -0,0 +1,104 @@
...
+        testParams.put("instanceId", "mycloud");
+        testParams.put("zookeepers", "zoo1:2181,zoo2:2181,zoo3:2181");
+        testParams.put("user", "myuser");
+        testParams.put("password", "mypassword");
+        testParams.put("tableName", "geomesaJava");
+        testParams.put("useMock", "true");
+        try {
+            agbs = new AccumuloGeoMesaBlobStore(testParams);
+        } catch(Exception e) {
+            System.out.println("Error initializing test geomesa blob store in AccumuloGeoMesaBlobStoreTest.java");
```
jahhulbert-ccri: use a logger instead and log the stack trace
```public static final logger = Logger.getLogger(.....)```

File: geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala [hunk 23]
```diff
@@ -0,0 +1,115 @@
...
+
+}
+
+object GDALFileHandler {
+  val dst_srs =
+    """DST_SRS=GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",
+      |SPHEROID["WGS_1984",6378137,298.257223563]],
+      |PRIMEM["Greenwich",0],
+      |UNIT["Degree",0.017453292519943295]]""".stripMargin
+  val opts = new util.Vector[String](List(dst_srs).asJava)
```
jahhulbert-ccri: ```Vector````.... #java1.4 :laughing:

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Fetches Blob by id
+     *
+     * @param id String feature Id of the Blob, from getIds functions
+     * @return Tuple2 of (blob, filename)
+     */
+    @Override
+    public Map.Entry<String, byte[]> get(String id) {
```
anthonyccri: Javadoc doesn't quite match anymore

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/interop/GeoMesaBlobStore.java
```diff
@@ -0,0 +1,72 @@
...
+     * @return
+     */
+    Iterator<String> getIds(Query query);
+
+    /**
+     * Fetches Blob by id
+     * @param id String feature Id of the Blob, from getIds functions
+     * @return Tuple2 of (blob, filename)
+     */
+    Map.Entry<String, byte[]> get(String id);
```
anthonyccri: Javadoc doesn't quite match.

geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/GeoMesaBaseDataStoreServlet.scala
AndrewAnnex: huh, not sure how that happened or why

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
AndrewAnnex: that is down to the GDAL java api, I can use something else with an implicit conversion if you feel strongly about it.

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
AndrewAnnex: type erasure

AndrewAnnex:
apparently things are working? not sure why locally I was having issues

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
jahhulbert-ccri: yeah i was just laughing at the API its just funny bc it was like 2003 and the world was young

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
jahhulbert-ccri: yup. too bad.

geomesa-blobstore/geomesa-blobstore-core/src/test/java/org/locationtech/geomesa/blob/core/AccumuloGeoMesaBlobStoreTest.java
```diff
@@ -31,6 +33,8 @@
     AccumuloGeoMesaBlobStore agbs;
     Random rand = new Random();
 
+    public static final Logger logger = Logger.getLogger("AccumuloGeoMesaBlobStoreTest");
```
jahhulbert-ccri: one of the last things...make it ```private```

geomesa-blobstore/geomesa-blobstore-handlers/geomesa-blobstore-gdal-handler/src/main/scala/org/locationtech/geomesa/blob/handlers/gdal/GDALFileHandler.scala
```diff
@@ -47,13 +49,11 @@ class GDALFileHandler extends AbstractFileHandler with LazyLogging {
       if (gdalFile == null) {
         None
       } else {
-        var geom: Option[Geometry] = None
         try {
-          geom = getImageBoundsGeometry(gdalFile)
+           getImageBoundsGeometry(gdalFile)
```
jahhulbert-ccri: picky - extra space

geomesa-blobstore/geomesa-blobstore-core/src/test/java/org/locationtech/geomesa/blob/core/AccumuloGeoMesaBlobStoreTest.java
```diff
@@ -9,9 +9,11 @@
 
 package org.locationtech.geomesa.blob.core;
 
+import org.apache.log4j.Logger;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+
```
jahhulbert-ccri: extra lines in import?

jahhulbert-ccri:
@AndrewAnnex thanks for the updates. a few little picky things. Question for you - do you have a sample GDAL file or whatnot that we can use for local testing? Other than that we'll try to get it merged tomorrow if @elahrvivaz is ok.

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -42,15 +42,26 @@ public AccumuloGeoMesaBlobStore(String instanceId,
...
         dataStoreParams.put(AccumuloDataStoreParams.userParam().key, user);
         dataStoreParams.put(AccumuloDataStoreParams.passwordParam().key, password);
         dataStoreParams.put(AccumuloDataStoreParams.authsParam().key, auths);
-        new AccumuloGeoMesaBlobStore(dataStoreParams);
+        AccumuloDataStore ds = genNewADS(dataStoreParams);
+        if (ds == null) {
+            throw new IOException("Error initializing AccumuloGeoMesaBlobStore");
+        } else {
+            accumuloBlobStore = new AccumuloBlobStore(ds);
+        }
```
elahrvivaz: this works - i think the original intent was that you just call
``` 
this(dataStoreParams)
```

geomesa-blobstore/geomesa-blobstore-core/src/main/java/org/locationtech/geomesa/blob/core/impl/AccumuloGeoMesaBlobStore.java
```diff
@@ -0,0 +1,140 @@
...
+                                    String password,
+                                    String auths) throws IOException {
+        Map<String, Serializable> dataStoreParams = new HashMap<>();
+        dataStoreParams.put(AccumuloDataStoreParams.instanceIdParam().key, instanceId);
+        dataStoreParams.put(AccumuloDataStoreParams.tableNameParam().key, tableName);
+        dataStoreParams.put(AccumuloDataStoreParams.zookeepersParam().key, zookeepers);
+        dataStoreParams.put(AccumuloDataStoreParams.userParam().key, user);
+        dataStoreParams.put(AccumuloDataStoreParams.passwordParam().key, password);
+        dataStoreParams.put(AccumuloDataStoreParams.authsParam().key, auths);
+        AccumuloDataStore ds = genNewADS(dataStoreParams);
```
elahrvivaz: this works - i think the original intent was that you just call
``` 
this(dataStoreParams)
```
(if you're confused, this comment got posted in the wrong spot initially)

AndrewAnnex: I tried that, but intellij was not happy with that unless I removed the `final` from the accumuloblobstore further up the file.

elahrvivaz: oh, interesting

jahhulbert-ccri: the call to this() has to be the first line in the java constructor is that what the problem was?

AndrewAnnex: yep

jahhulbert-ccri:
[WARNING] Missing header in: /home/travis/build/locationtech/geomesa/geomesa-blobstore/geomesa-blobstore-core/src/test/java/org/locationtech/geomesa/blob/core/AccumuloGeoMesaBlobStoreTest.java

AndrewAnnex:
whamp, I think my ide does that

jahhulbert-ccri:
FYI I'll run through this with a local setup to test it out (and I want to know how it works) then merge.

jahhulbert-ccri:
gonna move some classes around before merge

jahhulbert-ccri:
Reference #886 instead

PR #854, opened 2016-04-11
[WIP] Kafka DataStore command line tools

Afterglow375:
* Separated geomesa-tools into accumulo, kafka, and common packages
* Shared code goes in common package
* Made two separate scripts for the command line tools: geomesa-accumulo and geomesa-kafka
* Added list, create, describe, and removeschema commands for kafka


Afterglow375:
Here's some example output for some of the commands:

**describe**
$ geomesa-kafka describe -f atat_test6 -z zoo1,zoo2,zoo3 -b r01sv03:9092,r01sv04:9092 -p /geomesa/ds/kafka
Using GEOMESA_HOME = /opt/devel/src/geomesa/geomesa-tools/target/geomesa-tools-1.2.2-SNAPSHOT
Describing attributes of feature type 'atat_test6' at zkPath '/geomesa/ds/kafka'...
id: Integer
attr: Double
dtg: Date
geom: Geometry (ST-Geo-index)

Fetching Kafka topic metadata...
Number of partitions: 5
{TopicMetadata for topic geomesa-ds-kafka-atat_test6 -> 
Metadata for partition [geomesa-ds-kafka-atat_test6,0] is       partition 0     leader: 0 (r01sv03:9092)        replicas: 0 (r01sv03:9092)      isr: 0 (r01sv03:9092)   isUnderReplicated: false
Metadata for partition [geomesa-ds-kafka-atat_test6,1] is       partition 1     leader: 1 (r01sv04:9092)        replicas: 1 (r01sv04:9092)      isr: 1 (r01sv04:9092)   isUnderReplicated: false
Metadata for partition [geomesa-ds-kafka-atat_test6,2] is       partition 2     leader: 0 (r01sv03:9092)        replicas: 0 (r01sv03:9092)      isr: 0 (r01sv03:9092)   isUnderReplicated: false
Metadata for partition [geomesa-ds-kafka-atat_test6,3] is       partition 3     leader: 1 (r01sv04:9092)        replicas: 1 (r01sv04:9092)      isr: 1 (r01sv04:9092)   isUnderReplicated: false
Metadata for partition [geomesa-ds-kafka-atat_test6,4] is       partition 4     leader: 0 (r01sv03:9092)        replicas: 0 (r01sv03:9092)      isr: 0 (r01sv03:9092)   isUnderReplicated: false}

**list without zkPath**
$ geomesa-kafka list -z zoo1,zoo2,zoo3 -b r01sv03:9092,r01sv04:9092                                                                     
Using GEOMESA_HOME = /opt/devel/src/geomesa/geomesa-tools/target/geomesa-tools-1.2.2-SNAPSHOT
Running List Features without zkPath...
zkPath - schema
/geomesa/ds/kafka - NTOC-CAR-REPORT
/geomesa/ds/kafka - VMF-CSV
/geomesa/ds/kafka - AIR-FORCE-CTA-REPORT
/geomesa/ds/kafka - AIS-OTH-XML
/geomesa/ds/kafka - IBS-CMF-BFT-XML
/geomesa/ds/kafka/test - atat_test6
/geomesa/ds/kafka - laisic_track_report_so
/geomesa/ds/kafka - atat_test6
/geomesa/ds/kafka - atat_test1
/peter/pan - atat_test
/geomesa/ds/kafka - atat_test
/geomesa/ds/kafka - ais_oth_so
/geomesa/ds/kafka - SILK-ACROPOLIS

**list with zkPath**
$ geomesa-kafka list -z zoo1,zoo2,zoo3 -b r01sv03:9092,r01sv04:9092 -p /geomesa/ds/kafka
Using GEOMESA_HOME = /opt/devel/src/geomesa/geomesa-tools/target/geomesa-tools-1.2.2-SNAPSHOT
Running List Features using zkPath /geomesa/ds/kafka...
AIR-FORCE-CTA-REPORT
NTOC-CAR-REPORT
atat_test6
IBS-CMF-BFT-XML
SILK-ACROPOLIS
VMF-CSV
test
ais_oth_so
AIS-OTH-XML
laisic_track_report_so
atat_test
atat_test1

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -45,7 +45,7 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
     val typeName = featureType.getTypeName
     val schemaPath: String = getSchemaPath(typeName)
     if (zkClient.exists(schemaPath)) {
-      throw new IllegalArgumentException(s"Type $typeName already exists")
+      throw new IllegalArgumentException(s"Type $typeName already exists at $zkPath")
```
jnh5y: (grammar)  I'd go ahead and end this sentence with a period

geomesa-tools/bin/geomesa-accumulo
```diff
@@ -0,0 +1,251 @@
...
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License, Version 2.0 which
+# accompanies this distribution and is available at
+# http://www.opensource.org/licenses/apache2.0.php.
+#
+
+setGeoHome ()
+{
+    SOURCE="${BASH_SOURCE[0]}"
+    # resolve $SOURCE until the file is no longer a symlink
```
cne1x: This looks like code I've seen in a number of other places...  If we copied this from somewhere, we should probably give attribution.

geomesa-tools/bin/geomesa-kafka
```diff
@@ -247,5 +247,5 @@ else
         shift 1
     fi
     CLASSPATH="$GEOMESA_CP:$ACCUMULO_CP:$HADOOP_CP"
-    java ${GEOMESA_OPTS} -cp ${CLASSPATH} org.locationtech.geomesa.tools.Runner "$@"
+    java ${GEOMESA_OPTS} -cp ${CLASSPATH} org.locationtech.geomesa.tools.kafka.KafkaRunner "$@"
```
jnh5y: I'd suggest keeping 'geomesa' as an Accumulo-based command for the duration of the GeoMesa 1.2 series.  

As such, copying bin/geomesa to geomesa-kafka makes sense, but a copy of bin/geomesa would remain.

@jahhulbert-ccri  @elahrvivaz @cne1x thoughts?

geomesa-tools/bin/geomesa-accumulo
```diff
@@ -0,0 +1,251 @@
...
+        [[ "${SOURCE}" != /* ]] && SOURCE="${bin}/${SOURCE}"
+    done
+    bin="$( cd -P "$( dirname "${SOURCE}" )" && cd ../ && pwd )"
+    export GEOMESA_HOME="$bin"
+    export PATH=${GEOMESA_HOME}/bin:$PATH
+    echo >&2 "Warning: GEOMESA_HOME is not set, using $GEOMESA_HOME"
+}
+
+GEOMESA_OPTS="-Duser.timezone=UTC -DEPSG-HSQL.directory=/tmp/$(whoami)"
+GEOMESA_DEBUG_OPTS="-Xmx8192m -XX:MaxPermSize=512m -XX:-UseGCOverheadLimit -Xdebug -Xnoagent -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=9898"
```
cne1x: Are we targeting Java 8?  If so, then ```MaxPermSize``` is no longer a supported option.  In any event, 8 GB for the maximum heap size seems rather aggressive.  Do we need that much space?

geomesa-tools/pom.xml
```diff
@@ -66,6 +66,28 @@
...
+        <!-- Kafka CLI dependencies -->
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-kafka-datastore</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.11</artifactId>
+            <version>${kafka.version}</version>
```
jnh5y: Let's manage these two dependencies in the root pom.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/AccumuloRunner.scala
```diff
@@ -0,0 +1,35 @@
...
+  *************************************************************************/
+
+package org.locationtech.geomesa.tools.accumulo
+
+import org.locationtech.geomesa.tools.accumulo.commands._
+import org.locationtech.geomesa.tools.common.Runner
+import org.locationtech.geomesa.tools.common.commands.{Command, HelpCommand, VersionCommand}
+
+object AccumuloRunner extends Runner {
+  override val scriptName: String = "geomesa-accumulo"
```
jnh5y: ScriptName?  Hmm... I wonder if this'll mess with my suggestion 'geomesa' and 'geomesa-accumulo' both work the same...

geomesa-tools/bin/geomesa-accumulo
```diff
@@ -0,0 +1,251 @@
...
+        echo >&2 ""
+        echo "Now set to ${GEOMESA_CONF_DIR}"
+    fi
+    echo >&2 ""
+fi
+
+GEOMESA_OPTS="${GEOMESA_OPTS} -Dlog4j.configuration=file://${GEOMESA_CONF_DIR}/log4j.properties"
+
+
+# ACCUMULO paths, user can hard set these, or rely on this script to find set them via
```
cne1x: [picky] I think there's an ```ACCUMULO_HOME``` intended at the end of this comment line, if I'm reading it correctly.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/ExplainCommand.scala
```diff
@@ -32,5 +34,7 @@ class ExplainCommand(parent: JCommander) extends CommandWithCatalog(parent) with
 
 object ExplainCommand {
   @Parameters(commandDescription = "Explain how a GeoMesa query will be executed")
-  class ExplainParameters extends RequiredCqlFilterParameters {}
+  class ExplainParameters extends GeoMesaConnectionParams
```
jnh5y: Ok, I'm probably not seeing the big picture.  For each of these sets of parameters, why are the GeoMesaConnectionParams + others getting mixed in?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/IngestRasterCommand.scala
```diff
@@ -81,7 +82,18 @@ class PathValidator extends IParameterValidator {
 
 object IngestRasterCommand {
   @Parameters(commandDescription = "Ingest a raster file or raster files in a directory into GeoMesa")
-  class IngestRasterParameters extends CreateRasterParams {
+  class IngestRasterParameters extends AccumuloConnectionParams
+    with AccumuloRasterTableParam {
+
+    @Parameter(names = Array("--write-memory"), description = "Memory allocation for ingestion operation")
```
jnh5y: Nice.  It might be worth checking to see if wiring this was already in as a ticket.

geomesa-tools/bin/geomesa-accumulo
```diff
@@ -0,0 +1,251 @@
...
+    echo >&2 ""
+fi
+
+function findJars() {
+    home="$1"
+    CP=""
+    if [ -n "$home" -a -d "$home" ]; then
+        for jar in $(find ${home} -name "*.jar"); do
+            if [[ ("$2" != "true") || (("$jar" != *"test"*) && ("$jar" != *"slf4j"*)) ]]; then
+                if [ "$jar" = "${jar%-sources.jar}" -a "$jar" = "${jar%-test.jar}" ]; then
```
cne1x: This condition confuses me; what is it intended to test?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/AutoIngest.scala
```diff
@@ -6,14 +6,15 @@
...
 
 import java.io.File
 import java.util.concurrent.atomic.AtomicLong
 
 import com.typesafe.scalalogging.LazyLogging
-import org.locationtech.geomesa.tools.Utils.Formats
-import org.locationtech.geomesa.tools.Utils.Formats.Formats
+import org.locationtech.geomesa.tools.accumulo.Utils
+import Utils.Formats
+import Utils.Formats.Formats
```
jnh5y: (picky)  I prefer fully qualified imports.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/AutoIngestDelimited.scala
```diff
@@ -19,8 +19,9 @@ import org.geotools.factory.GeoTools
 import org.geotools.util.Converters
 import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.locationtech.geomesa.jobs.mapreduce.{FileStreamInputFormat, FileStreamRecordReader}
-import org.locationtech.geomesa.tools.Utils.Formats
-import org.locationtech.geomesa.tools.Utils.Formats.Formats
+import org.locationtech.geomesa.tools.accumulo.Utils
+import Utils.Formats
+import Utils.Formats.Formats
```
jnh5y: Ditto re: FQCNs.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/LocalRasterIngest.scala
```diff
@@ -17,8 +17,9 @@ import org.locationtech.geomesa.accumulo.index.DecodedIndexValue
 import org.locationtech.geomesa.raster.data.Raster
 import org.locationtech.geomesa.raster.util.RasterUtils
 import org.locationtech.geomesa.raster.util.RasterUtils.IngestRasterParams
+import org.locationtech.geomesa.tools.accumulo.Utils
 import org.locationtech.geomesa.utils.geohash.BoundingBox
-import org.locationtech.geomesa.tools.Utils.Formats._
+import Utils.Formats._
```
jnh5y: Maybe the optimize imports commands disagrees with me?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/common/CLArgResolver.scala
```diff
@@ -1,12 +1,12 @@
...
-* are made available under the terms of the Apache License, Version 2.0
-* which accompanies this distribution and is available at
-* http://www.opensource.org/licenses/apache2.0.php.
-*************************************************************************/
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+  * All rights reserved. This program and the accompanying materials
+  * are made available under the terms of the Apache License, Version 2.0
+  * which accompanies this distribution and is available at
+  * http://www.opensource.org/licenses/apache2.0.php.
+  *************************************************************************/
```
jnh5y: (picky) Bump back to the other formatting.  (I think @elahrvivaz's script to check headers prefers the other way.)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/CommandWithKDS.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.locationtech.geomesa.tools.common.commands.Command
+import org.locationtech.geomesa.tools.kafka.{KafkaConnectionParams, DataStoreHelper}
+
+/**
+  * Abstract class for commands that require a KafkaDataStore
+  */
+abstract class CommandWithKDS(parent: JCommander) extends Command(parent) {
+  override val params: KafkaConnectionParams
+  lazy val ds = new DataStoreHelper(params).getDataStore
+  lazy val zkPath = params.zkPath
```
jnh5y: What happens when the zkPath isn't given?  Does it default to /geomesa/kafka/ds or whatever the default is?

jnh5y:
A first-pass looks reasonable to me.  @elahrvivaz @jahhulbert-ccri @mmatz-ccri input?

geomesa-tools/bin/geomesa-kafka
cne1x: Agreed.  Let's keep "geomesa" (laster to be named "geomesa-accumulo"?), and simply add "geomesa-kafka" for now.

cne1x:
Two quick items:
1.  We typically squash-merge our development commits into a single PR commit.
2.  You should sign -- via ```git commit --sign``` your PR commit for LocationTech's sake.

geomesa-tools/bin/geomesa-kafka
elahrvivaz: yep

geomesa-tools/bin/geomesa-accumulo
elahrvivaz: i think we're building with java 8 but still targeting java 7?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/ExplainCommand.scala
elahrvivaz: I haven't looked closely, but the idea is that parameters get re-used all over, so having them as mixins maximizes re-use of the code. Previously we extended classes, but that didn't allow for mix-n-match.

geomesa-tools/bin/geomesa-accumulo
Afterglow375: Don't know. This is the original geomesa script unchanged. git blame reveals Emilio wrote it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/LocalRasterIngest.scala
Afterglow375: Looks like it.

geomesa-tools/bin/geomesa-accumulo
elahrvivaz: This is to verify that it's not a -test or -sources jar. `${foo%suffix}` will extract the suffix from foo, so:
```bash
$ foo="test.jar"
$ echo ${foo%-sources.jar}
test.jar
$ foo="test-sources.jar"
$ echo ${foo%-sources.jar}
test
```
if foo == foo with the -sources (and -test) stripped off, then it doesn't end in -sources and it's a legit jar.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/CommandWithKDS.scala
Afterglow375: Yes, that is the default. That's configured by the KDS, so no zkPath parameter even needs to be present in the map of connection parameters.

Afterglow375:
@cne1x Yeah for some reason I thought I would just close this then open up a new PR with a merge squashed branch.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/ExplainCommand.scala
jahhulbert-ccri: Another fun trick is Parameters delegate! http://jcommander.org/#Parameter_delegates

jahhulbert-ccri:
So we probably don't want to change the default bin/geomesa script to be bin/geomesa-accumulo for 1.2.x I would probably leave that as bin/geomesa for now and add new ones like geomesa-kafka then for 1.3.x put in a ticket to rename it to bin/geomesa-accumulo

geomesa-tools/bin/geomesa-kafka
```diff
@@ -0,0 +1,251 @@
...
+        CP="${CP:1}"
+    fi
+    echo $CP
+}
+
+# Start constructing GEOMESA_CP (classpath)
+# include geomesa first so that the correct log4j.properties is picked up
+GEOMESA_CP="${GEOMESA_CONF_DIR}:$(findJars $GEOMESA_LIB)"
+GEOMESA_CP="${GEOMESA_CP}:${ACCUMULO_CONF_DIR}:${HADOOP_CONF_DIR}"
+ACCUMULO_CP="$(findJars $ACCUMULO_LIB)"
```
jahhulbert-ccri: do we need accumulo in the kafka script?

geomesa-tools/pom.xml
```diff
@@ -66,6 +66,26 @@
...
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-kafka-datastore</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.11</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>slf4j-log4j12</artifactId>
+                    <groupId>org.slf4j</groupId>
```
jahhulbert-ccri: switch order to make groupID first and maybe consider making it exclude ```org.slf4j:*```

geomesa-tools/bin/geomesa-kafka
```diff
@@ -0,0 +1,251 @@
+#! /usr/bin/env bash
+#
+# Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License, Version 2.0 which
+# accompanies this distribution and is available at
+# http://www.opensource.org/licenses/apache2.0.php.
+#
+
+setGeoHome ()
```
elahrvivaz: it would be nice if we had a common script that the other two (geomesa-accumulo and geomesa-kafka) could `source` to pick up the common functions and code.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/DataStoreHelper.scala
```diff
@@ -30,7 +30,8 @@ class DataStoreHelper(params: GeoMesaParams) extends AccumuloProperties {
 
   /**
    * Get a handle to a datastore for a pre-existing catalog table
-   * @throws Exception if the catalog table does not exist in accumulo
+    *
+    * @throws Exception if the catalog table does not exist in accumulo
```
elahrvivaz: [picky] either roll back this change or align the *s in the comment

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/CreateCommand.scala
```diff
@@ -6,39 +6,27 @@
...
+import org.locationtech.geomesa.accumulo.index.Constants
+import org.locationtech.geomesa.tools.accumulo.commands.CreateCommand.CreateParameters
+import org.locationtech.geomesa.tools.accumulo.{GeoMesaConnectionParams, OptionalAccumuloSharedTablesParam}
+import org.locationtech.geomesa.tools.common.{CLArgResolver, FeatureTypeNameParam, FeatureTypeSpecParam, OptionalDTGParam}
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
-import org.opengis.feature.simple.SimpleFeatureType
 
-object FeatureCreator extends LazyLogging {
+class CreateCommand(parent: JCommander) extends CommandWithCatalog(parent) with LazyLogging {
```
elahrvivaz: i like the renaming

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/QueryStatsCommand.scala
```diff
@@ -6,16 +6,17 @@
...
 import com.beust.jcommander.{JCommander, Parameter, Parameters}
 import org.locationtech.geomesa.accumulo.stats.RasterQueryStatTransform
 import org.locationtech.geomesa.raster.data.AccumuloRasterStore
-import org.locationtech.geomesa.tools.AccumuloProperties
-import org.locationtech.geomesa.tools.commands.QueryStatsCommand.QueryStatsParameters
+import org.locationtech.geomesa.tools.accumulo.commands.QueryStatsCommand.QueryStatsParameters
+import org.locationtech.geomesa.tools.accumulo.{AccumuloConnectionParams, AccumuloProperties, AccumuloRasterTableParam}
+import org.locationtech.geomesa.tools.common.commands.Command
 
 class QueryStatsCommand(parent: JCommander) extends Command(parent) with AccumuloProperties {
```
elahrvivaz: could rename this class too while you're in there...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/CreateCommand.scala
```diff
@@ -0,0 +1,60 @@
...
+        println(s"Created feature $featureName")
+      } else {
+        logger.error(s"There was an error creating feature '$featureName' with spec " +
+          s"'$sftString'. Please check that all arguments are correct " +
+          "in the previous command.")
+      }
+    } catch {
+      // if the schema already exists at the specified zkPath
+      case e: IllegalArgumentException => {
+        logger.warn(e.getMessage)
```
elahrvivaz: Does the message provide enough info back to the user? or should we explicitly say what's in your comment there.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/DescribeCommand.scala
```diff
@@ -0,0 +1,69 @@
...
+      sft.getAttributeDescriptors.foreach { attr =>
+        sb.clear()
+        val name = attr.getLocalName
+
+        // TypeName
+        sb.append(name)
+        sb.append(": ")
+        sb.append(attr.getType.getBinding.getSimpleName)
+
+        if (sft.getGeometryDescriptor == attr) sb.append(" (ST-Geo-index)")
```
elahrvivaz: Maybe change 'ST-Geo-index' to something like 'default geometry'

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/DescribeCommand.scala
```diff
@@ -0,0 +1,69 @@
...
+        if (attr.getDefaultValue != null)      sb.append("- Default Value: ", attr.getDefaultValue)
+
+        println(sb.toString())
+      }
+
+      val zkClient = new ZkClient(params.zookeepers, Int.MaxValue, Int.MaxValue, ZKStringSerializer)
+      val topicName = zkClient.readData[String](ds.asInstanceOf[KafkaDataStore].getTopicPath(params.featureName))
+      val topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topicName, zkClient)
+
+      println("\nFetching Kafka topic metadata...")
```
elahrvivaz: @jahhulbert-ccri this should go out under system.err, right? Data on system.out, info on system.err

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.locationtech.geomesa.tools.kafka.{DataStoreHelper, OptionalZkPathParams}
+
+class ListCommand(parent: JCommander) extends Command(parent) with LazyLogging {
+  override val command = "list"
+  override val params = new ListParameters()
+  lazy val zkClient = new ZkClient(params.zookeepers, Int.MaxValue, Int.MaxValue, ZKStringSerializer)
+
+  override def execute() = {
+    if (params.zkPath == null) {
+      logger.info(s"Running List Features without zkPath...")
```
elahrvivaz: change this to show the default zkpath

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
```diff
@@ -0,0 +1,80 @@
...
+      logger.info(s"zkPath - schema")
+      ZkUtils.getAllTopics(zkClient).filter(_.contains('-')).foreach {
```
elahrvivaz: is there a more robust way to determine if a topic is actually an sft?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
```diff
@@ -0,0 +1,80 @@
...
+        printZkPathAndTopicString
+      }
+    } else {
+      logger.info(s"Running List Features using zkPath ${params.zkPath}...")
```
elahrvivaz: seems like the if and the else here should have the same logic. It probably makes sense to move the filtering of the topics into the kafka data store.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+class RemoveSchemaCommand(parent: JCommander) extends CommandWithKDS(parent) with LazyLogging {
+  override val command = "removeschema"
+  override val params = new RemoveSchemaParams
+
+  override def execute() = {
+    if (Option(params.pattern).isEmpty && Option(params.featureName).isEmpty) {
+      throw new IllegalArgumentException("Please provide either featureName or pattern to removing.")
```
elahrvivaz: grammar

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+    val typeNamesToRemove = getTypeNamesFromParams()
+    validate(typeNamesToRemove) match {
+      case Success(_) =>
+        if (params.force || promptConfirm(typeNamesToRemove)) {
+          removeAll(typeNamesToRemove)
+        } else {
+          logger.info(s"Cancelled schema removal.")
+        }
+      case Failure(ex) =>
+        println(s"Feature validation failed on error: ${ex.getMessage}.")
```
elahrvivaz: don't need println and log message do we?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+        ds.removeSchema(tname)
+        if (ds.getNames.contains(tname)) {
+          throw new Exception(s"Error removing feature type '$zkPath:$tname'.")
+        }
+      } match {
+        case Success(_) =>
+          println(s"Removed $tname")
+        case Failure(ex) =>
+          println(s"Failure removing type $tname")
+          logger.error(s"Failure removing type $tname.", ex)
```
elahrvivaz: ditto about logger/println

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+        logger.error(ex.getMessage)
+    }
+  }
+
+  protected def removeAll(typeNames: List[String]) = {
+    typeNames.foreach { tname =>
+      Try {
+        ds.removeSchema(tname)
+        if (ds.getNames.contains(tname)) {
+          throw new Exception(s"Error removing feature type '$zkPath:$tname'.")
```
elahrvivaz: i would loop through everything and remove it, then go back and compile a list of types that failed to remove. That way it will remove as many as possible even if the first one fails.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+          println(s"Removed $tname")
+        case Failure(ex) =>
+          println(s"Failure removing type $tname")
+          logger.error(s"Failure removing type $tname.", ex)
+      }
+    }
+  }
+
+  protected def getTypeNamesFromParams() =
+    Option(params.featureName).toList ++ Option(params.pattern).map { p =>
```
elahrvivaz: seems like you can specify both a feature name and a pattern - is that intentional?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  protected def validate(typeNames: List[String]) = Try {
+    if (typeNames.isEmpty) {
+      throw new IllegalArgumentException("No feature type names found from pattern or provided feature type name.")
+    }
+
+    val validFeatures = ds.getTypeNames
+    typeNames.foreach { f =>
+      if (!validFeatures.contains(f)) {
+        throw new IllegalArgumentException(s"Feature type name $f does not exist at zkPath $zkPath")
```
elahrvivaz: aren't these exceptions going to get caught by your Try? (try does let some through so not sure)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+  }
+
+  protected def promptConfirm(featureNames: List[String]) =
+    PromptConfirm.confirm(s"Remove schema ${featureNames.mkString(",")} at zkPath $zkPath? (yes/no): ")
```
elahrvivaz: [picky] make schema plural if there are more than 1

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/DataStoreHelper.scala
```diff
@@ -0,0 +1,37 @@
...
+  /**
+    * Gets a KDS
+    *
+    * @throws Exception if the KDS cannot be created with the given parameters
+    */
+  def getDataStore() = {
+    val kdsFactory = new KafkaDataStoreFactory()
+    Option(kdsFactory.createDataStore(paramMap)).getOrElse {
+      throw new Exception("Could not load a data store with the provided parameters: " +
+        s"${paramMap.map { case (k,v) => s"$k=$v" }.mkString(",")}")
```
mmatz-ccri: unwrap s"{...}"

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/KafkaParams.scala
```diff
@@ -0,0 +1,63 @@
...
+
+package org.locationtech.geomesa.tools.kafka
+
+import com.beust.jcommander.Parameter
+
+/**
+  * Shared Kafka-specific command line parameters
+  */
+
+trait KafkaConnectionParams {
```
mmatz-ccri: extends ZookeepersParam

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/DescribeCommand.scala
```diff
@@ -0,0 +1,69 @@
...
+      println(s"Number of partitions: ${topicMetadata.partitionsMetadata.size}")
+      println(topicMetadata)
+    } catch {
+      case npe: NullPointerException =>
+        logger.error(s"Error: feature '${params.featureName}' not found. Check arguments...", npe)
+      case e: Exception =>
```
mmatz-ccri: maybe `case NonFatal(e)`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
```diff
@@ -0,0 +1,80 @@
...
+import com.beust.jcommander.{JCommander, Parameters}
+import com.typesafe.scalalogging.LazyLogging
+import kafka.utils.{ZKStringSerializer, ZkUtils}
+import org.I0Itec.zkclient.ZkClient
+import org.I0Itec.zkclient.exception.ZkNoNodeException
+import org.locationtech.geomesa.tools.common.commands.Command
+import org.locationtech.geomesa.tools.kafka.commands.ListCommand._
+import org.locationtech.geomesa.tools.kafka.{DataStoreHelper, OptionalZkPathParams}
+
+class ListCommand(parent: JCommander) extends Command(parent) with LazyLogging {
```
mmatz-ccri: any reason this does not extend CommandWithKDS?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
```diff
@@ -0,0 +1,80 @@
...
+  override val command = "list"
+  override val params = new ListParameters()
+  lazy val zkClient = new ZkClient(params.zookeepers, Int.MaxValue, Int.MaxValue, ZKStringSerializer)
```
mmatz-ccri: maybe move this to CommandWithKDS

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
jnh5y: No...  The only thing common to all topic names is that the zkPath is concatenated with the SFT name with a '-'.

The next method should go through the yoga to figure out zkPath + SFT checking...

geomesa-tools/bin/geomesa-kafka
Afterglow375: No. I just copied over the geomesa accumulo script to start out with.

geomesa-tools/bin/geomesa-kafka
Afterglow375: Yeah, these bash scripts need some maintenance... I may leave it for a future PR though, as I don't know what's ideal.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/QueryStatsCommand.scala
Afterglow375: Rename it to what?

elahrvivaz: to match the command string - QueryRasterStatsCommmad

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
Afterglow375: Nope, and that's why I think there should be a restriction on the zkPath. To me, it doesn't seem very safe to let the user use any arbitrary zkPath. IMO, zkPaths should be prefixed with geomesa or something like that.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
Afterglow375: I don't really know... I saw that the geomesa accumulo commands interchanged use of println vs logger statements. Don't know the reasoning behind that.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
Afterglow375: Can you clarify? to me it seems like the schemas will still get removed even if the first fails due to the foreach.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
Afterglow375: Nope... that was copied from the accumulo removeschema command.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
jnh5y: @anthonyccri @mmatz-ccri Can you speak to the idea of arbitrary zkPaths?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
elahrvivaz: oops, you're right - never mind

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
Afterglow375: Not sure. Again, this bit of code was copied from the accumulo removeSchema command.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListCommand.scala
Afterglow375: zkPath is optional, in which case a KDS isn't used. If no zkPath is specified, the list command searches the entirety of zookeeper for potential geomesa schemas.

geomesa-tools/bin/geomesa-kafka
jahhulbert-ccri: I played around with this in cloud local. What I believe I found was this:
1. You have to copy/paste the code to find your base dir, because you will need that in order to source the first env script
2. a geomesa-env would be a great idea. or a geomesa.conf type thing to set some basic versions.

jahhulbert-ccri: https://geomesa.atlassian.net/browse/GEOMESA-1092

fyi

geomesa-tools/bin/geomesa-kafka
```diff
@@ -0,0 +1,173 @@
...
+    GEOMESA_LIB=${GEOMESA_HOME}/lib
+elif [[ $1 = configure ]]; then
+    message="Warning: GEOMESA_LIB was already set, probably by a prior configuration."
+    message="${message}\n The current value is ${GEOMESA_LIB}."
+    echo >&2 ""
+    echo -e >&2 "$message"
+    echo >&2 ""
+    read -p "Do you want to reset this to ${GEOMESA_HOME}/lib? Y\n " -n 1 -r
+    if [[  $REPLY =~ ^[Yy]$ ]]; then
+        GEOMESA_LIB=${GEOMESA_HOME}/lib
```
jahhulbert-ccri: are we gonana change this a lib-common, lib-accumulo, lib-kafka ?

geomesa-tools/pom.xml
```diff
@@ -66,6 +66,30 @@
...
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-kafka-datastore</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.11</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j:*</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
```
jahhulbert-ccri: @Afterglow375 Actually just make the artifact id a ```*``` and leave the group id as ```org.slf4j``` the other syntax of ```org.slf4j:*``` is a shorthand syntax for ```groupId:artifactId:version``` i believe

geomesa-tools/pom.xml
```diff
@@ -66,6 +66,30 @@
...
+                </exclusion>
+                <exclusion>
+                    <groupId>org.xerial.snappy</groupId>
+                    <artifactId>snappy-java</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
```
jahhulbert-ccri: Is this needed for geomesa tools or kafka? Note that hadoop needs version 11.0.2 on the classpath to work properly and mixing in a 2nd version can break things depending on what this is.

geomesa-tools/pom.xml
```diff
@@ -108,17 +132,13 @@
...
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
             <groupId>org.apache.accumulo</groupId>
             <artifactId>accumulo-start</artifactId>
         </dependency>
         <dependency>
             <groupId>org.apache.zookeeper</groupId>
             <artifactId>zookeeper</artifactId>
+            <scope>compile</scope>
```
jahhulbert-ccri: do you need it at compile in order to do the component.xml ?

geomesa-tools/src/main/assemblies/component.xml
```diff
@@ -15,8 +15,23 @@
...
+            <directoryMode>755</directoryMode>
+            <fileMode>0644</fileMode>
+            <useProjectArtifact>false</useProjectArtifact>
+            <useTransitiveDependencies>true</useTransitiveDependencies>
+            <useTransitiveFiltering>true</useTransitiveFiltering>
+            <includes>
+                <include>com.google.guava:guava</include>
+                <include>org.apache.kafka:kafka_2.11</include>
+                <include>org.locationtech.geomesa:geomesa-kafka-datastore</include>
+                <include>org.apache.zookeeper:zookeeper</include>
```
jahhulbert-ccri: so i believe for accumulo scripts we decided to have this is a separate install instead of putting things in the lib dir because we wanted to use ACCUMULO_HOME or HADOOP_DIR or whatever or ZOOKEEPER_HOME if they were set. We should try to be consistent here. We usually end up installing the tools anyway. Something to at least think about

pom.xml
```diff
@@ -1438,6 +1438,26 @@
...
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.kafka</groupId>
+                <artifactId>kafka_2.11</artifactId>
+                <version>${kafka.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <artifactId>slf4j-log4j12</artifactId>
+                        <groupId>org.slf4j</groupId>
```
jahhulbert-ccri: same as before change to: 
```
<groupId>org.slf4j</groupId>
<artifactId>*</artifactId>
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
jahhulbert-ccri: FYI we tweaked the println vs log messages so that we could have things to go stdout in a standard way and errors to go elsewhere if people wanted to pipe the results into something. we log to stderr by default so thats why we differentiate from std out.

jnh5y:
I tried to run 'geomesa-kafka help' and I hit this exception:

jnh5y@jnh-pc:tools> geomesa-kafka help
Using GEOMESA_HOME = /opt/geomesa/geomesa-1.2.2-SNAPSHOT/dist/tools/geomesa-tools-1.2.2-SNAPSHOT
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/accumulo/core/client/TableNotFoundException
	at org.locationtech.geomesa.tools.accumulo.commands.TableConfCommand.<init>(TableConfCommand.scala:33)
	at org.locationtech.geomesa.tools.accumulo.AccumuloRunner$.<init>(AccumuloRunner.scala:30)
	at org.locationtech.geomesa.tools.accumulo.AccumuloRunner$.<clinit>(AccumuloRunner.scala)
	at org.locationtech.geomesa.tools.common.commands.HelpCommand.execute(HelpCommand.scala:27)
	at org.locationtech.geomesa.tools.common.Runner$class.main(Runner.scala:26)
	at org.locationtech.geomesa.tools.kafka.KafkaRunner$.main(KafkaRunner.scala:15)
	at org.locationtech.geomesa.tools.kafka.KafkaRunner.main(KafkaRunner.scala)
Caused by: java.lang.ClassNotFoundException: org.apache.accumulo.core.client.TableNotFoundException
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 7 more

jnh5y:
I'd argue that 'listen' should print the SimpleFeatures being written instead of the KDS command messages.

My output was:

jnh5y@jnh-pc:tools> geomesa-kafka listen -f KafkaQuickStart -z localhost -b localhost:9092 -p /geomesa/ds/kafka
Using GEOMESA_HOME = /opt/geomesa/geomesa-1.2.2-SNAPSHOT/dist/tools/geomesa-tools-1.2.2-SNAPSHOT
Listening to KafkaQuickStart...


CreateOrUpdate(2016-04-14T15:30:41.522Z,ScalaSimpleFeature:1)
CreateOrUpdate(2016-04-14T15:30:41.562Z,ScalaSimpleFeature:2)

jnh5y:
As a separate ticket, I think that typing just a command without arguments should *not* print the entire list of commands.

To wit, this output
"""
jnh5y@jnh-pc:tools> geomesa-kafka removeschema
Using GEOMESA_HOME = /opt/geomesa/geomesa-1.2.2-SNAPSHOT/dist/tools/geomesa-tools-1.2.2-SNAPSHOT
Error parsing arguments: The following options are required: -b, --brokers -z, --zookeepers -p, --zkpath 
Usage: geomesa-kafka [command] [command options]
  Commands:
    create          Create a feature definition in GeoMesa
    describe        Describe the attributes of a given feature in GeoMesa
    help            Show help
    list            List GeoMesa features for a given zkPath
    listen          Listen to a GeoMesa Kafka topic
    removeschema    Remove a schema and associated features from GeoMesa
    version         GeoMesa Version
"""

should only have the first two line of output.  

@elahrvivaz  @jahhulbert-ccri thoughts?

jnh5y:
For 'removeschema', the -f flag should be required, but the output for the command doesn't make that clear.

"jnh5y@jnh-pc:tools> geomesa-kafka removeschema
Using GEOMESA_HOME = /opt/geomesa/geomesa-1.2.2-SNAPSHOT/dist/tools/geomesa-tools-1.2.2-SNAPSHOT
Error parsing arguments: The following options are required: -b, --brokers -z, --zookeepers -p, --zkpath"

elahrvivaz:
re: printing entire list of commands - i agree, but yeah lets make that a separate ticket (as it's the current behavior)

elahrvivaz:
re: listen - we do want to indicate the message type though (add/remove/etc). Looks like the output currently is just a toString - to make the simple feature a string you can use DataUtilities.encodeFeature()

jnh5y:
re:listen.  My opinion is that create/updates should have the fields.  Deletes may only be able to provide the ID which is enough.

Being able to see the C/U feature fields will help with debugging quite a bit.

elahrvivaz:
Yeah, something like (example output is what DataUtilities would give):
```
Add feature: fid-1=myname|POINT (45 49)
Delete feature: fid-1
```

geomesa-tools/pom.xml
Afterglow375: The maven assembly plugin excludes guava for the accumulo command line tools. I need it for the KDS cache.

geomesa-tools/pom.xml
Afterglow375: I need zookeeper jars for the kafka commands.

geomesa-tools/bin/geomesa-kafka
```diff
@@ -0,0 +1,173 @@
...
+    fi
+    if [ "${CP:0:1}" = ":" ]; then
+        CP="${CP:1}"
+    fi
+    echo $CP
+}
+
+# Start constructing GEOMESA_CP (classpath)
+# include geomesa first so that the correct log4j.properties is picked up
+GEOMESA_CP="${GEOMESA_CONF_DIR}:$(findJars $GEOMESA_LIB):$(findJars $GEOMESA_HOME/kafka-lib)"
```
jahhulbert-ccri: @Afterglow375 @elahrvivaz what do you think about making the dir layout be:
```
lib/common
lib/kafka
lib/accumulo
```
like that. so we don't cram the top level dir. For now the current things going into lib would all go into either ```lib/accumulo``` or ```lib/common``` or whatever. But nothing would be in plan ```lib``` except subdirs. Similar to how hadoop does it.

elahrvivaz: i like it

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/common/commands/HelpCommand.scala
```diff
@@ -6,13 +6,13 @@
...
-package org.locationtech.geomesa.tools.commands
+package org.locationtech.geomesa.tools.common.commands
 
 import java.util
 
 import com.beust.jcommander.{JCommander, Parameter, Parameters}
-import org.locationtech.geomesa.tools.Runner.commandUsage
-import org.locationtech.geomesa.tools.commands.HelpCommand.HelpParameters
+import org.locationtech.geomesa.tools.accumulo.AccumuloRunner.commandUsage
+import org.locationtech.geomesa.tools.common.commands.HelpCommand.HelpParameters
```
jahhulbert-ccri: FYI just ran this:
```
ahulbert@doubletrouble:/tmp/geomesa-tools-1.2.2-SNAPSHOT$ bin/geomesa-kafka help
Warning: GEOMESA_HOME is not set, using /tmp/geomesa-tools-1.2.2-SNAPSHOT
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/accumulo/core/client/TableNotFoundException
        at org.locationtech.geomesa.tools.accumulo.commands.TableConfCommand.<init>(TableConfCommand.scala:33)
        at org.locationtech.geomesa.tools.accumulo.AccumuloRunner$.<init>(AccumuloRunner.scala:30)
        at org.locationtech.geomesa.tools.accumulo.AccumuloRunner$.<clinit>(AccumuloRunner.scala)
        at org.locationtech.geomesa.tools.common.commands.HelpCommand.execute(HelpCommand.scala:27)
        at org.locationtech.geomesa.tools.common.Runner$class.main(Runner.scala:26)
        at org.locationtech.geomesa.tools.kafka.KafkaRunner$.main(KafkaRunner.scala:15)
        at org.locationtech.geomesa.tools.kafka.KafkaRunner.main(KafkaRunner.scala)
Caused by: java.lang.ClassNotFoundException: org.apache.accumulo.core.client.TableNotFoundException
        at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
```

can you check the imports, etc. and make sure we don't need accumulo installed to run the kafka tool?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/RemoveSchemaCommand.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  protected def promptConfirm(featureNames: List[String]) =
+    PromptConfirm.confirm(s"Remove schema(s) ${featureNames.mkString(",")} at zkPath $zkPath? (yes/no): ")
+}
+
+object RemoveSchemaCommand {
+  @Parameters(commandDescription = "Remove a schema and associated features from GeoMesa")
+  class RemoveSchemaParams extends SimpleProducerKDSConnectionParams
+    with OptionalFeatureTypeNameParam
```
jahhulbert-ccri: Should this be a required feature type name?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/ListenCommand.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.locationtech.geomesa.tools.kafka.ConsumerKDSConnectionParams
+import org.locationtech.geomesa.tools.kafka.commands.ListenCommand.ListenParameters
+
+class ListenCommand(parent: JCommander) extends Command(parent) with LazyLogging {
+  override val command = "listen"
+  override val params = new ListenParameters()
+
+  override def execute(): Unit = {
+    println(s"Listening to ${params.featureName}...")
+    KafkaDataStoreLogViewer.run(params.zookeepers, params.zkPath, params.featureName, params.fromBeginning)
```
jahhulbert-ccri: wondering if we should check for existance first of the ```-f``` arg in order to give a better error message? unless it's designed to work otherwise.

jahhulbert-ccri:
trying out the tools. @Afterglow375 did you update the README.md and the docs/tools part? We can probably go ahead and add a kafka -tools subsection here: http://www.geomesa.org/documentation/user/commandline_tools.html
  and we'll need to figure out if we want a KafkaTools.md or just add it to this: https://github.com/locationtech/geomesa/blob/master/geomesa-tools/README.md

Afterglow375:
@jnh5y I fixed the exception when running geomesa-kafka help.

The -f flag is not required for the removeschema command because you can alternatively provide a --pattern flag. If you're missing the -f or --pattern, the command will throw an exception.

Afterglow375:
New PR created to replace this one: https://github.com/locationtech/geomesa/pull/862

PR #865, opened 2016-04-22
GEOMESA-1000 adding loose bbox config

elahrvivaz:


vpipkt:
Giving this a look now

vpipkt:
I had a set of queries known to bring back features outside the filter. Using `looseBoundingBox=false` in fact returns a smaller set of features that are all inside the filter. Checked this for 18 such queries with bounding boxes ranging from 0.005 degrees a side to 15 degrees. All of these queries do hit the z3 index; none hit z2.

anthonyccri:
Merged

PR #866, opened 2016-04-22
GEOMESA-1182,GEOMESA-538 Stats-based query planning

elahrvivaz:
* Stats are gathered during ingestion, and can be run manually through the command-line tools
* There are now two options for query planning - stats based or the old index priority based
* Query planning can be controlled by:
  * The query hint org.locationtech.geomesa.accumulo.index.QueryHints#COST_EVALUATION_KEY
  * System property 'geomesa.query.cost.type' - either 'Stats' or 'Index'
  * Hint takes priority, then system property, otherwise uses stats based planning


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
```diff
@@ -0,0 +1,479 @@
...
+    // reload the tracker - for long-held updaters, this will refresh the histogram ranges
+    stat = statFunction
+  }
+}
+
+object GeoMesaMetadataStats {
+
+  // how many buckets to sort each attribute into
+  // max space on disk = 8 bytes * size - although we use optimized serialization so likely 1-3 bytes * size
+  val DefaultHistogramSize  = 1000
```
elahrvivaz: I need to re-evaluate this - once we get up to 100's of millions of points, the histograms will be mostly worthless. Probably need to grow them dynamically or split them up by date range or something.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -621,43 +628,19 @@ class AccumuloDataStore(val connector: Connector,
...
-      val lockPath =  s"/org.locationtech.geomesa/accumulo/ds/$catalogTable"
-      val lock = new InterProcessSemaphoreMutex(client, lockPath)
-      if (!lock.acquire(120, TimeUnit.SECONDS)) {
-        throw new RuntimeException(s"Could not acquire distributed lock at '$lockPath'")
-      }
-      // delegate lock that will close the curator client upon release
-      new Releasable {
-        override def release(): Unit = try lock.release() finally client.close()
-      }
+    val path = s"/org.locationtech.geomesa/accumulo/ds/$catalogTable"
```
jahhulbert-ccri: path and timeout like good zookeeper/site-based properties for the future :tiger:

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -10,13 +10,13 @@
...
 import java.io.Serializable
-import java.util.{Map => JMap, Collections}
+import java.util.{Collections, Map => JMap}
 
 import org.apache.accumulo.core.client.mock.{MockConnector, MockInstance}
 import org.apache.accumulo.core.client.security.tokens.PasswordToken
 import org.apache.accumulo.core.client.{Connector, ZooKeeperInstance}
 import org.geotools.data.DataAccessFactory.Param
-import org.geotools.data.{Parameter, DataStoreFactorySpi}
+import org.geotools.data.{DataStoreFactorySpi, Parameter}
```
jahhulbert-ccri: :abc:

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
```diff
@@ -8,254 +8,307 @@
...
+
+  /**
+   * Deletes all values associated with a given feature type
+   *
+   * @param typeName simple feature type name
+   */
+  def delete(typeName: String)
+}
+
+object GeoMesaMetadata {
```
jahhulbert-ccri: good idea to move out of that package object

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/BoundsFilterVisitor.scala
```diff
@@ -0,0 +1,83 @@
...
+    *
+    * @param filter fitler to evaluate
+    * @param envelope an existing envelope that will be included in the final bounds
+    * @return the bounds
+    */
+  def visit(filter: Filter, envelope: Envelope = null): ReferencedEnvelope = {
+
+    val visitor = ExtractBoundsFilterVisitor.BOUNDS_VISITOR
+
+    val result = filter match {
```
jahhulbert-ccri: i guess this requires scala 2.11 :100:

jahhulbert-ccri:
Do we have a plan to deal with large tables...Running an MR job would still be rough against it. I guess we could just rely on ingest or sampling.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/StatsCombiner.scala
```diff
@@ -0,0 +1,60 @@
...
+        (typeName, StatSerializer(SimpleFeatureTypes.createType(typeName, v)))
+    }
+  }
+
+  override def reduce(key: Key, iter: java.util.Iterator[Value]): Value = {
+    val head = iter.next()
+    if (!iter.hasNext) {
+      head
+    } else {
+      val sftName = AccumuloBackedMetadata.getTypeNameFromMetadataRowKey(key.getRow.toString)
```
jahhulbert-ccri: I wonder if we could push this in via iterator config instead of running a scanner here. It should be cached but it may be better to push it in up front.

elahrvivaz:
yeah, need to figure out something for large tables... it took about 20 seconds to re-generate the whole stats for 8.85M twitter records on tcloud using the tools (so some of that was setup overhead). But once we get above a certain threshold, the histograms will become sort of meaningless as every bucket will have millions of features.
The stats iterator doesn't support sampling yet, but we could add that pretty easily.
I like the iterator vs a m/r job b/c it's easy to configure, run, etc. But having a m/r job would be nice if it could run against the rfiles directly (like you said).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/StatsCombiner.scala
elahrvivaz: Yeah, there's just one combiner for all the feature types in a table. I played around with having one per feature type, but there isn't any way to restrict the rows that it gets applied to, and it got a little messy, so I went with this.

geomesa-tools/README.md
```diff
@@ -230,8 +232,81 @@ To remove a feature type and it's associated data from a catalog table, use the
...
+    $ geomesa help stats-list
+    View statistics on a GeoMesa feature type
+    Usage: list-stats [options]
+      Options:
+        -a, --attributes
+           Comma-separated list of attribute names
+        --auths
+           Accumulo authorizations
+        -b, --bins
+           How many buckets the data will be divided into
```
jahhulbert-ccri: might be useful to throw out a starting number to try (e.g. 10)

jahhulbert-ccri:
@elahrvivaz yeah I think running a scan against x billion entries (even with a combiner) from a single box could be bad. I think that's the other motivation for the job, so that when it fails, freezes, etc. it will work. By default, I assume it hits the record table? really seems like a good motivation for the 2 table idea :) the compacted, static history table and the small ingest table with recent data.

jahhulbert-ccri:
overall looks pretty good. I will give it a whirl locally and see what happens and see if anyone else wants to look through. @elahrvivaz may need to rebase since I think we moved the tools packages around with the KDS tools PR

elahrvivaz:
@jahhulbert-ccri from a single box should be fine - all the aggregation is done in the tservers, so only a single record ever comes back to the client. By default we run Filter.INCLUDE against the z2 table now.

jahhulbert-ccri:
sounds good.

elahrvivaz:
going to close this for now, needs more work.

PR #879, opened 2016-05-05
GEOMESA-1200 An AccumuloDataStore should not default to the full Accu…

cne1x:
…mulo user's authorizations when an empty list of authorizations has been passed to the constructor

This PR should be *discussed* before being merged.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
@@ -45,7 +45,8 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
     val masterAuthsStrings = masterAuths.map(b => new String(b))
 
     // get the auth params passed in as a comma-delimited string
-    val configuredAuths = authsParam.lookupOpt[String](params).getOrElse("").split(",").filter(s => !s.isEmpty)
+    val rawConfiguredAuths = authsParam.lookupOpt[String](params)
+    val configuredAuths = rawConfiguredAuths.getOrElse("").split(",").filter(s => !s.isEmpty)
 
     // verify that the configured auths are valid for the connector we are using (fail-fast)
     if (!connector.isInstanceOf[MockConnector]) {
@@ -56,8 +57,11 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
       }
     }
 
-    // if no auths are specified, only data without visibilities can be returned
-    val auths: List[String] = configuredAuths.toList
+    // if the caller provided any non-null string for authorizations, use it;
+    // otherwise, grab all authorizations to which the Accumulo user is entitled
+    val auths: List[String] =
+      if (rawConfiguredAuths.isDefined) configuredAuths.toList  // may be empty!
+      else masterAuthsStrings.toList
 
     val authProvider = security.getAuthorizationsProvider(params, auths)
     val auditProvider = security.getAuditProvider(params).getOrElse {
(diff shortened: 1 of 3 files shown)


amihalik:
Thanks Chris.  This looks good.

jnh5y:
I would be interested in having a way to use all the Accumulo user's auths without having to enumerate them.  

I can see that being difficult to expose.

amihalik:
Can you differentiate between a null vs empty string?  an empty string meaning "no auths" and a null being "all auths"?

cne1x:
@amihalik @jnh5y @anthonyccri:  I've pushed up a commit that represents an attempt to differentiate between auths not given at all and empty auths.  Please take a look, and see if this is moving in the right direction.  Thanks!

elahrvivaz:
@cne1x is the proposed differentiation between null and empty possible to configure through the geoserver datastore page? I would guess not... maybe we should add a boolean parameter to 'use connector auths' (could probably word it better than that).

cne1x:
@elahrvivaz I like that idea.  A simple toggle for whether to grab the provided auths or the full set of user auths sounds reasonable.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -56,9 +59,14 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
       }
     }
 
-    // if no auths are specified we default to the connector auths
-    // TODO would it be safer to default to no auths?
-    val auths: List[String] = if (configuredAuths.nonEmpty) configuredAuths.toList else masterAuthsStrings.toList
+    // if the caller provided any non-null string for authorizations, use it;
+    // otherwise, grab all authorizations to which the Accumulo user is entitled
+    if (configuredAuths.length != 0 && forceEmptyAuths) {
+      throw new IllegalArgumentException("Forcing empty auths is checked, but expclicit auths are provided")
```
elahrvivaz: typo 'expclicit' :)

mm64z: (picky/sp) expclicit? -> explicit

elahrvivaz:
looks good. just or completeness, could we add a test that verified forceEmptyAuths works if the connector has some underlying auths? (not sure that's possible with mock accumulo though...)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithDataStore.scala
```diff
@@ -28,6 +28,7 @@ import org.specs2.mutable.Specification
 import org.specs2.specification.{Fragments, Step}
 
 import scala.collection.JavaConverters._
+import scala.collection.JavaConverters._
```
mdzimmerman: Auto-added by IntelliJ?

elahrvivaz:
looks good

cne1x:
@amihalik What do you think of the recent refactoring?  We've preserved the way users are accustomed to working with our GeoServer layers, but added an explicit flag for the one case where there are no explicit authorizations provided, and the user does *not* want the default behavior of assuming all of the Accumulo user's authorizations.  In any other case in which explicit authorizations are provided, they will be used; the flag only covers the single case in which no auths are given, and is meant as a compromise between ease-of-use for people in GeoServer and correctness.  Your thoughts?

elahrvivaz:
merging

elahrvivaz:
49445fcf363fb564223f8c0c81ca95217a554a75

PR #880, opened 2016-05-06
GEOMESA-1202 Frequency stat

elahrvivaz:
* Added Frequency and Z3Frequency stats, backed by CountMinSketch
* Added Z3RangeHistogram stat
* Switched geometry stats from geohashes to z-indices
* Added cardinality to MinMax stat, backed by HyperLogLog
* Added 'immutable' flag for deserializing stats that you don't want modified
* Added unobserve method to handle delete of features
* Fixed some bugs in stat bounds handling
* Removed mutable equals and hash code in stats (can cause issues with hashmaps), replaced with 'isEquivalent' method
* Improved stat serialization


elahrvivaz:
I'm going to update this to use the zValue for the geometry instead of using geohashes

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
```diff
@@ -0,0 +1,212 @@
...
+            new RichCountMinSketch(sketch).isEquivalent(s.sketch)
+      case _ => false
+    }
+  }
+}
+
+object Frequency {
+
+  // the seed for our frequencies - frequencies can only be combined if they have the same seed.
+  val Seed = -27
```
jnh5y: Ok, can you say more about the frequency;)?  Is -27 completely arbitrary?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
```diff
@@ -23,114 +24,162 @@ import scala.reflect.ClassTag
...
-    override val max: java.lang.Long = java.lang.Long.MIN_VALUE
+    override val min: java.lang.Long = java.lang.Long.MIN_VALUE
+    override val max: java.lang.Long = java.lang.Long.MAX_VALUE
   }
 
   implicit object MinMaxFloat extends ComparableMinMax[java.lang.Float] {
-    override val min: java.lang.Float = java.lang.Float.MAX_VALUE
-    override val max: java.lang.Float = java.lang.Float.MIN_VALUE
+    override val min: java.lang.Float = java.lang.Float.MIN_VALUE
+    override val max: java.lang.Float = java.lang.Float.MAX_VALUE
```
jnh5y: Ok, so were these just wrong before?  Or were they being used to indicate that no value had been set?

elahrvivaz: It was just confusing... they were being set to these values as initial markers. i switched the logic to set the min to the max and check for that value in the MinMax itself. E.g. lines 34, 103, 106 etc

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
elahrvivaz: Yeah, arbitrary. It's like a random number seed. It just has to be the same between frequencies or we can't combine them.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -125,10 +127,10 @@ object BinnedArray {
 
 class BinnedIntegerArray(length: Int, bounds: (Integer, Integer)) extends BinnedArray[Integer](length, bounds) {
 
-  require(bounds._1 < bounds._2,
+  require(bounds._1 <= bounds._2,
     s"Upper bound must be greater than lower bound: lower=${bounds._1} upper=${bounds._2}")
```
cne1x: Missing update on the comment, yes?  "Upper bound must be no less than lower bound" or something equivalent?

elahrvivaz: good catch, thanks

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -142,23 +144,26 @@ class BinnedIntegerArray(length: Int, bounds: (Integer, Integer)) extends Binned
...
+    val hi = math.max(lo, bounds._1 + math.floor(binSize * (index + 1)).toInt)
+    (if (lo > bounds._2) bounds._2 else lo, if (hi > bounds._2) bounds._2 else hi)
   }
 }
 
 class BinnedLongArray(length: Int, bounds: (jLong, jLong)) extends BinnedArray[jLong](length, bounds) {
 
-  require(bounds._1 < bounds._2,
+  require(bounds._1 <= bounds._2,
     s"Upper bound must be greater than lower bound: lower=${bounds._1} upper=${bounds._2}")
```
cne1x: Similar comment update.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -172,21 +177,24 @@ class BinnedLongArray(length: Int, bounds: (jLong, jLong)) extends BinnedArray[j
...
+    (if (lo > bounds._2) bounds._2 else lo, if (hi > bounds._2) bounds._2 else hi)
   }
 }
 
 
 class BinnedFloatArray(length: Int, bounds: (jFloat, jFloat)) extends BinnedArray[jFloat](length, bounds) {
 
-  require(bounds._1 < bounds._2,
+  require(bounds._1 <= bounds._2,
     s"Upper bound must be greater than lower bound: lower=${bounds._1} upper=${bounds._2}")
```
cne1x: Update comment, please.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -216,7 +224,7 @@ class BinnedFloatArray(length: Int, bounds: (jFloat, jFloat)) extends BinnedArra
 
 class BinnedDoubleArray(length: Int, bounds: (jDouble, jDouble)) extends BinnedArray[jDouble](length, bounds) {
 
-  require(bounds._1 < bounds._2,
+  require(bounds._1 <= bounds._2,
     s"Upper bound must be greater than lower bound: lower=${bounds._1} upper=${bounds._2}")
```
cne1x: Update comment, please.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: actually i'm going to change it back to strictly less than - for a while i thought i needed equals but not any more.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -254,47 +262,45 @@ class BinnedDoubleArray(length: Int, bounds: (jDouble, jDouble)) extends BinnedA
...
     }
   }
 
-  require(start < end, s"Upper bound must be greater than lower bound: lower=${bounds._1} upper=${bounds._2}")
-
-  private val firstDiff = start.zip(end).indexWhere { case (l, r) => l != r }
-  private val zeroInt = '0'.toInt
-  private val nineInt = '9'.toInt - zeroInt
-  private val alphaOffset = 'a'.toInt - 10
+  require(start <= end, s"Upper bound must be greater than lower bound: lower=${bounds._1} upper=${bounds._2}")
```
cne1x: Update comment, please.  Do we care to report ```bounds._1``` and ```bounds._2``` or ```start``` and ```end```?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -254,47 +262,45 @@ class BinnedDoubleArray(length: Int, bounds: (jDouble, jDouble)) extends BinnedA
...
-  def stringToLong(s: String): Long = {
-    val base36 = s.slice(firstDiff, start.length).replaceAll("[^0-9a-z]", "0")
-    if (base36.length > 12) {
-      // won't fit in a long...
-      jLong.parseLong(base36.substring(0, 12), 36)
-    } else {
-      jLong.parseLong(base36, 36)
-    }
+  // # of base 36 numbers we can consider
+  private val sigDigits = math.max(1, math.round(math.log(length) / math.log(36)).toInt)
```
cne1x: Should ```math.round``` here by ```math.ceil```?  Consider the case where ```length = 37```, in which ```Math.log(37) / Math.log(36) = 1.007645829325495```.  Don't we mean there to be 2 digits required, and not 1?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -254,47 +262,45 @@ class BinnedDoubleArray(length: Int, bounds: (jDouble, jDouble)) extends BinnedA
...
-  private val binSize = math.max(1, (max - min).toFloat / length)
+  def stringToLong(s: String, padding: Char = '0'): Long = {
+    val normalized = if (s.length >= prefixLength + sigDigits) {
+      s.substring(prefixLength, prefixLength + sigDigits)
+    } else {
+      s.substring(prefixLength).padTo(sigDigits, padding)
+    }
+    jLong.parseLong(normalized, 36)
+  }
+  def longToString(l: Long, padding: Char = '0'): String =
```
cne1x: [picky]  It's probably true that nobody else cares, but using ```l``` in a context where it can easily be mistaken for a ```1``` is potentially confusing.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -387,16 +409,17 @@ class BinnedGeometryArray(length: Int, bounds: (Geometry, Geometry)) extends Bin
...
 
   override def bounds(index: Int): (Geometry, Geometry) = {
     if (index < 0 || index > length) {
       throw new ArrayIndexOutOfBoundsException(index)
     }
-    val loInt = min + math.ceil(binSize * index).toInt
-    val hiInt = min + math.floor(binSize * (index + 1)).toInt - 1
-
-    (intToGeom(loInt), intToGeom(hiInt))
+    // TODO the way geohashes work, we want to use floor instead of round otherwise we can jump around
```
cne1x: Is this comment relevant here?  We already appear to be using ```floor```.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
```diff
@@ -0,0 +1,212 @@
...
+  /**
+    * Number of observations in the frequency map
+    *
+    * @return number of observations
+    */
+  def size: Long = sketch.size()
+
+  override def observe(sf: SimpleFeature): Unit = {
+    val value = sf.getAttribute(attribute)
+    if (value != null) {
```
cne1x: Do we really want to discard all information about nulls?  Wouldn't it be more useful to maintain this as a separate count?  I understand that the sketch should only use the non-null entries, but knowing how many nulls seems as if it could also be extremely useful.  Just a thought.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
```diff
@@ -0,0 +1,212 @@
...
+    plus
+  }
+
+  override def +=(other: Frequency[T]): Unit = new RichCountMinSketch(sketch).add(other.sketch)
+
+  override def clear(): Unit = new RichCountMinSketch(sketch).clear()
+
+  override def isEmpty: Boolean = sketch.size == 0
+
+  override def toJson: String = {
```
cne1x: [picky]  This is really only converting a subset of the state to JSON, right, not the entire object?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: I'll add them both

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
elahrvivaz: yeah - do you think displaying the table values would be useful? I guess you could re-create the count min sketch using that info, but it seems like a stretch.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/FrequencyZ3.scala
```diff
@@ -0,0 +1,144 @@
...
+                  val dtgIndex: Int,
+                  val precision: Int,
+                  val eps: Double = 0.005,
+                  val confidence: Double = 0.95) extends Stat {
+
+  override type S = FrequencyZ3
+
+  private val mask = Frequency.getMask(precision)
+
+  private [stats] val sketches = scala.collection.mutable.Map.empty[Short, CountMinSketch]
```
cne1x: One ```CountMinSketch``` per week?  That seems like a lot.  (Thinking aloud.)  GDELT has data from 1970 through today, so that's 46 years (rounding up), so that's about 52 * 46 = 2392 sketches for all of GDELT (through December 2016).  How much memory would that take?  Or am I misunderstanding how this is being stored?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
elahrvivaz: yeah, that is a good point. how about converting them to an empty string? i don't think the countMinSketch will accept nulls as it uses hashcodes to insert things.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: I'll fix

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
```diff
@@ -23,114 +24,162 @@ import scala.reflect.ClassTag
...
 
+  abstract class ComparableMinMax[T <: Comparable[T]] extends MinMaxDefaults[T] {
+    override def min(left: T, right: T): T = if (left.compareTo(right) > 0) right else left
+    override def max(left: T, right: T): T = if (left.compareTo(right) < 0) right else left
+  }
+
   implicit object MinMaxString extends ComparableMinMax[String] {
-    override val min: String = "~~~"
-    override val max: String = ""
+    override val min: String = ""
```
cne1x: Do we need to worry about encoding here?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/FrequencyZ3.scala
elahrvivaz: This class is not being used at present... I had numbers on the sizes, let me see if I can find them again.
The z3 histogram will be used however - i'll post those numbers too.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
elahrvivaz: how so? strings are comparable directly without any encoding/decoding.

cne1x: If "~" is ASCII 127, is it a reasonable upper-bound when strings might come in in Unicode?  That's all I was thinking, but if it's obtuse, please ignore me.

elahrvivaz: oh, good point. I thought ~ was pretty much the last char, as we use it in accumulo ranges. I may be incorrect about that...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -25,61 +29,102 @@ import scala.reflect.ClassTag
...
-      val plus = new RangeHistogram(attribute, length, endpoints)
-      var i = 0
-      while (i < plus.bins.length) {
-        plus.bins.counts(i) += (bins.counts(i) + other.bins.counts(i))
-        i += 1
+  override def unobserve(sf: SimpleFeature): Unit = {
+    val value = sf.getAttribute(attribute)
+    if (value != null) {
+      val i = bins.indexOf(value.asInstanceOf[T])
+      if (i == -1) {
```
cne1x: For this to happen, wouldn't we have to unobserve a feature that contains a value we've never seen before?  Is that reasonable?  (I'm uncertain in precisely what context we would call ```unobserve```, clearly.)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -91,43 +136,91 @@ class RangeHistogram[T](val attribute: Int, initialBins: Int, initialEndpoints:
...
+      case v: Date   => (v, new Date(v.getTime + 60000))
+      case v: Geometry =>
+        val env = v.getCentroid.buffer(1.0).getEnvelopeInternal
+        val min = GeometryUtils.geoFactory.createPoint(new Coordinate(env.getMinX, env.getMinY))
+        val max = GeometryUtils.geoFactory.createPoint(new Coordinate(env.getMaxX, env.getMaxY))
+        (min, max)
+    }
+    buf.asInstanceOf[(T, T)]
+  }
+  /**
```
cne1x: [style] Blank line between methods?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/FrequencyZ3.scala
elahrvivaz: Max histogram size = 8 bytes * number of bins
We use variable length encoding, so counts up to ~2M will take 3 bytes or less (instead of 8).
I have been using a max bin size of 10000, which equates to ~29kb on disk for 1B records.
When storing, I'm using the 'splitByWeek' method - that way each entry will only be 29kb (assuming 1B records per week...), and when using them we only need to retrieve the weeks that we care about.
It looks like gdelt is 40M records - that would be ~16722 records per week, which would end up as ~10kb on disk - so 24MB...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -91,43 +136,91 @@ class RangeHistogram[T](val attribute: Int, initialBins: Int, initialEndpoints:
...
   /**
-    * Creates expanded endpoints that cover both input endpoints
+    * Gets the bounds of the array that actually contain values.
     */
-  def expandEndpoints[T](one: (T, T), two: (T, T))(implicit defaults: MinMax.MinMaxDefaults[T]): (T, T) = {
-    val leftExpand = defaults.minmax(two._1, one._1, one._2)
-    defaults.minmax(two._2, leftExpand._1, leftExpand._2)
+  private def getActualBounds[T](bins: BinnedArray[T])(implicit defaults: MinMax.MinMaxDefaults[T]): (T, T) = {
+    val minIndex = bins.counts.indexWhere(_ != 0)
+    val maxIndex = bins.counts.length - bins.counts.reverse.indexWhere(_ != 0) - 1
```
cne1x: If there are 10 bin counts, all of which are empty, then ```maxIndex = 10 - -1 - 1 = 10```, which should cause an overflow on the line computing ```max```, shouldn't it?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/FrequencyZ3.scala
elahrvivaz: The frequency precision we use is 2000 longs per frequency - so a single frequency would encode to (1-8) bytes * 2000, according to the total number of records stored.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: Yeah, you are correct. It's not entirely clear b/c the code that's using this hasn't been put up for a pr yet :)
We call unobserve when deleting features. It is probably unlikely to happen, but I don't think it's impossible given the way we create and write a new stat that only sees the features it's dealing with, and then we combine then on read using an accumulo combining iterator.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: you're right, good catch.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/FrequencyZ3.scala
jnh5y: Off-the-wall question:  with what visibility are stats stored?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: oh, i guess this only gets called when we know that there are some non-zero values - but still, it might not get called that way in the future.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/FrequencyZ3.scala
elahrvivaz: no visibility. storing different stats for different visibility features might be needed in the future, but the first cut isn't going to deal with that.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
elahrvivaz: hmm, actually using an empty string would be equivalent to 0 for numeric values - maybe that is ok? I think i'll create an issue to track this and not change anything yet.

geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/stats/HistogramTest.scala
```diff
@@ -70,6 +80,15 @@ class HistogramTest extends Specification with StatTestHelper {
         unpacked.toJson mustEqual stat.toJson
       }
 
+      "deserialize as immutable value" >> {
+        val stat = newStat[String]("strAttr")
+        val packed = StatSerializer(sft).serialize(stat)
+        val unpacked = StatSerializer(sft).deserialize(packed)
+        unpacked.toJson mustEqual stat.toJson
+
```
cne1x: [style] Extraneous blank lines.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
elahrvivaz: https://geomesa.atlassian.net/browse/GEOMESA-1210

geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/stats/HistogramTest.scala
elahrvivaz: copy-paste fail... fixed.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: I'll note that this whole class is kind of a hack... let me think about this one.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Frequency.scala
cne1x: Nah.  I was just curious what the purpose was of the JSON format.  It's no big deal.

elahrvivaz: Ideally it would show the whole stat - when calling through WPS for example, you generally get back the json. But for stats like this one it is hard to represent as a string...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
cne1x: Confirmed that this is an issue.  Consider:
```scala
val max = "~~~"
val s1 = "The quick brown fox jumped over the lazy coder."
val s2 = "Œ"  // a single Unicode character
println(s1 > max, s2 > max)
println(s2 > max)
```
Yields ```(false, true)```.

cne1x: Looks like we may want a constant that looks more like:   ```val s3 = "\uFFFF\uFFFF\uFFFF"```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: Guess I'm not sure - I think that if we used `ceil` we would end up with a lot of empty buckets. If you have any better ideas for putting strings into buckets I'd appreciate it. I also wanted to handle common prefixes, so that e.g. if all your data starts with 'user_id_xxx' everything doesn't end up in the same bucket.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/MinMax.scala
elahrvivaz: I'm going to change it to `s"${Char.MaxValue}${Char.MaxValue}${Char.MaxValue}"`. Seem better? (technically I think we would need an array of Int.MaxLength filled with Char.MaxValue to be the final final sorted string...)

elahrvivaz: oh, didn't see your note - yeah, i think those are equivalent.

build/cqs.tsv
```diff
@@ -3,6 +3,7 @@ asm:asm	3.3.1	compile
 com.101tec:zkclient	0.3	compile
 com.adobe.xmp:xmpcore	5.1.2	compile
 com.beust:jcommander	1.48	compile
+com.clearspring.analytics:stream	2.9.2	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11377

build/cqs.tsv
```diff
@@ -74,6 +75,7 @@ it.geosolutions.jaiext.utilities:jt-utilities	1.0.8	compile
 it.geosolutions.jaiext.vectorbin:jt-vectorbin	1.0.8	compile
 it.geosolutions.jaiext.warp:jt-warp	1.0.8	compile
 it.geosolutions.jaiext.zonal:jt-zonal	1.0.8	compile
+it.unimi.dsi:fastutil	7.0.6	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11378

PR #881, opened 2016-05-06
GEOMESA-1203 VisibilityFilter's cache should be thread safe

ddseapy:


elahrvivaz:
looks good to me

jahhulbert-ccri:
looks good...lets hope they don't find more of these: https://issues.scala-lang.org/browse/SI-7943

#java.util.concurrent.ConcurrentHashMap :coffee:

ddseapy:
Yeah, originally I tried going with a ConcurrentHashMap, but I'd have to write some sort of wrapper around it since id have to use "putIfAbsent" instead of "getOrElseUpdate" because it doesn't allow the lazy evaluation of the passed value (defeating the purpose of using the vizcache). This approach seemed easier.

elahrvivaz:
merging

elahrvivaz:
f4e84a0fc66bb2f892d2ecbeb0c0071296dfda58

PR #882, opened 2016-05-07
GEOMESA-1204 Use LocationTech/Eclipse Jar signing

jnh5y:
* Integrates with the Eclipse signing service.
  This service is (and can) only used by the LocationTech build server.

* Adds documentation for the LocationTech downloads page.

### DIFF
--- a/docs/common.py
+++ b/docs/common.py
@@ -74,6 +74,12 @@ rst_epilog = """
 
 .. |release_source_tarball| replace:: https://github.com/locationtech/geomesa/archive/geomesa-%(release)s.tar.gz
 
+.. |eclipse_release| replace:: 1.2.0 
+
+.. |eclipse_release_tarball| replace:: http://download.locationtech.org/geomesa/1.2.0/geomesa-dist-1.2.0-bin.tar.gz
+
+.. |eclipse_release_source_tarball| replace:: http://download.locationtech.org/geomesa/1.2.0/geomesa-source-1.2.0.tar.gz 
+
 .. |development| replace:: %(version_devel)s
 
 .. |release_1_1| replace:: %(release_1_1)s
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -44,7 +44,13 @@ Versions and Downloads
 
 * Source: https://github.com/locationtech/geomesa/archive/master.tar.gz
 
-**Latest Eclipse IP Reviewed release**: |eclipse_release|
+**Fully Eclipse IP reviewed release**: |eclipse_release|
+
+GeoMesa is part of the Locationtech working group at Eclipse. Eclipse fully reviews each major release for IP concerns.
+
+.. warning::
+
+    Eclipse releases may not contain all the bug fixes and improvements from the latest release.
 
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -191,6 +191,37 @@
                 </repository>
             </repositories>
         </profile>
+   <profile>
```
elahrvivaz: [picky] indenting

docs/common.py
```diff
@@ -74,6 +74,12 @@
 
 .. |release_source_tarball| replace:: https://github.com/locationtech/geomesa/archive/geomesa-%(release)s.tar.gz
 
+.. |eclipse_release| replace:: 1.2.0 
+
+.. |eclipse_release_tarball| replace:: http://download.locationtech.org/geomesa/1.2.0/geomesa-dist-1.2.0-bin.tar.gz
+
+.. |eclipse_release_source_tarball| replace:: http://download.locationtech.org/geomesa/1.2.0/geomesa-source-1.2.0.tar.gz
```
jahhulbert-ccri: should these be dynamic instead of hardcoded? and is 1.2.0 right?

jahhulbert-ccri: Also, I'd like to have people not download 1.2.0 since 1.2.1 has some fixes :)

jnh5y: 1.2.1 has not been fully Eclipse reviewed:(

jnh5y: And incidentally, that's why I hardcoded it...  Additionally, we have to upload the files manually, so yeah...

elahrvivaz: yeah, i don't like pointing people to older versions either... hasn't 1.2.1 been implicitly reviewed? We don't go through the release process for dot releases, so none of them would ever be approved... or am I missing something?

jnh5y: This is where there's going to be a mismatch for fully reviewed versions.  I was trying to find a way to point to the last reviewed version....

docs/user/installation_and_configuration.rst
```diff
@@ -44,6 +44,11 @@ Versions and Downloads
 
 * Source: https://github.com/locationtech/geomesa/archive/master.tar.gz
 
+**Latest Eclipse IP Reviewed release**: |eclipse_release|
```
elahrvivaz: Can we change this to 'Fully Eclipse IP Reviewed'? And maybe also add a note that you should only use this if you are very worried about licensing issues...

docs/common.py
jahhulbert-ccri: so in general...if we are doing "bug fix" releases they aren't "reviewed"?

docs/user/installation_and_configuration.rst
```diff
@@ -44,6 +44,17 @@ Versions and Downloads
 
 * Source: https://github.com/locationtech/geomesa/archive/master.tar.gz
 
+**Fully Eclipse IP reviewed release**: |eclipse_release|
```
cne1x: [picky] The wording here seems a bit funky, even though I know what you're going for.  Perhaps something like, "Most recent release for which Eclipse has fully vetted IP"?  Or, "Most recent release whose IP has been fully reviewed by Eclipse"?  I dunno'.

PR #885, opened 2016-05-11
avro2sft jahhulbert

jahhulbert-ccri:
A few things that we probably still need to do:
* backwards compat testing

For avro generically:
* instead of using bytes for lists/maps do array, map as avro schema items so that its more avro friendly and not an opaque byte array

Also, didn't automatically change ```-``` to ```_``` but we can consider that I guess. These don't easily encode/decode back and forth anyways since this impl doesn't always escape underscores but that would be the proper thing to do like to do with table names :computer:


geomesa-features/geomesa-feature-avro/src/main/scala/org/locationtech/geomesa/features/avro/FieldNameEncoder.scala
```diff
@@ -0,0 +1,71 @@
...
+  def deHexEscape(s: String): String = {
+    val chars = s.toCharArray
+    var idx = 0
+    val sb = new StringBuilder
+    while (idx < chars.length) {
+      if (chars(idx).isLetterOrDigit) {
+        sb.append(chars(idx))
+        idx += 1
+      } else {
+        val hex = chars(idx + 1).toString + chars(idx + 2).toString
```
elahrvivaz: could this run into issues if there were literal underscores in the original string?

geomesa-features/geomesa-feature-avro/src/test/scala/org/locationtech/geomesa/features/avro/AvroSimpleFeatureWriterTest.scala
```diff
@@ -131,6 +133,26 @@ class AvroSimpleFeatureWriterTest extends Specification with Mockito with Abstra
 
       there was one(encoder).writeArrayEnd()
     }
+
+    "use unmangled names when requested" >> {
+      import org.locationtech.geomesa.security._
+      val sf = createSimpleFeature
+
+      val vis = "test&usa"
```
elahrvivaz: might not want to have this as a visibility in our code...

geomesa-tools/README.md
```diff
@@ -825,6 +827,27 @@ Export queries and statistics logged for raster tables by using the `querystats`
 
 #### Example command:
     geomesa queryrasterstats -u username -p password -t somerastertable -num 10
+
+### sft2avro
```
elahrvivaz: maybe we could rename this to something like 'generateAvroSchema' so we can get away from the x2x syntax that we tried to remove from the converters.

geomesa-features/geomesa-feature-avro/src/test/scala/org/locationtech/geomesa/features/avro/AvroSimpleFeatureWriterTest.scala
jahhulbert-ccri: yup

geomesa-features/geomesa-feature-avro/src/main/scala/org/locationtech/geomesa/features/avro/FieldNameEncoder.scala
jahhulbert-ccri: totally. i'll just escape them too like we do in the table names

geomesa-tools/README.md
jahhulbert-ccri: cool

elahrvivaz:
e49df1d9696edaca15cc2697665beb2bb711cab1

PR #891, opened 2016-05-17
GEOMESA-1028 GeoMesa ingest commands should warn when restricted fiel…

cne1x:
…d names are used

This affects both the KafkaDataStore and the AccumuloDataStore.
A significant number of unit tests had to be amended to pass
with this change.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -39,6 +39,7 @@ import org.locationtech.geomesa.security.{AuditProvider, AuthorizationsProvider}
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.{FeatureSpec, NonGeomAttributeSpec}
+import org.locationtech.geomesa.utils.index.GeoMesaSchemaValidator
 import org.opengis.feature.`type`.Name
 import org.opengis.feature.simple.SimpleFeatureType
 import org.opengis.filter.Filter
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/TemporalIndexCheckTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/TemporalIndexCheckTest.scala
@@ -13,6 +13,7 @@ import org.junit.runner.RunWith
 import org.locationtech.geomesa.accumulo._
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.locationtech.geomesa.utils.index.TemporalIndexCheck
 import org.opengis.feature.simple.SimpleFeatureType
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/TestData.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/TestData.scala
@@ -55,7 +55,7 @@ object TestData extends LazyLogging {
       .build()
 
   def getTypeSpec(suffix: String = "2") = {
-    s"POINT:String,LINESTRING:String,POLYGON:String,attr$suffix:String:index=true," +
+    s"A_POINT:String,A_LINESTRING:String,A_POLYGON:String,attr$suffix:String:index=true," +
         s"geom:Geometry:srid=4326,dtg:Date,dtg_end_time:Date;geomesa.mixed.geometries=true"
   }
 
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
@@ -6,12 +6,12 @@
 * http://www.opensource.org/licenses/apache2.0.php.
 *************************************************************************/
 
-package org.locationtech.geomesa.accumulo.index
+package org.locationtech.geomesa.utils.index
 
 import com.typesafe.scalalogging.LazyLogging
 import com.vividsolutions.jts.geom.Geometry
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.{DEFAULT_DATE_KEY, RichSimpleFeatureType}
-import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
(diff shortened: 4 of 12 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/TestData.scala
```diff
@@ -55,7 +55,7 @@ object TestData extends LazyLogging {
       .build()
 
   def getTypeSpec(suffix: String = "2") = {
-    s"POINT:String,LINESTRING:String,POLYGON:String,attr$suffix:String:index=true," +
+    s"A_POINT:String,A_LINESTRING:String,A_POLYGON:String,attr$suffix:String:index=true," +
```
elahrvivaz: ooh, are POINT etc also reserved?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -39,6 +39,7 @@ import org.locationtech.geomesa.security.{AuditProvider, AuthorizationsProvider}
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.{FeatureSpec, NonGeomAttributeSpec}
+import org.locationtech.geomesa.utils.index.GeoMesaSchemaValidator
```
jnh5y: Does this import get used?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/TemporalIndexCheckTest.scala
```diff
@@ -13,6 +13,7 @@ import org.junit.runner.RunWith
 import org.locationtech.geomesa.accumulo._
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.locationtech.geomesa.utils.index.TemporalIndexCheck
```
jnh5y: Ditto?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
```diff
@@ -20,10 +20,31 @@ object GeoMesaSchemaValidator {
...
+  */
+object ReservedWordCheck extends LazyLogging {
+
+  // ensure that no attribute names are reserved words within GeoTools that will cause query problems
+  def validateAttributeNames(sft: SimpleFeatureType): Unit = {
+    val reservedWords = FeatureUtils.sftReservedWords(sft)
+    if (reservedWords.nonEmpty) {
+      // TODO:  Make this an exception after GeoMesa 1.2.2
+      logger.warn(
+        "The simple feature type contains one or more attributes whose names are reserved words:  " +
```
elahrvivaz: extra space after colon (unless that is from the chicago manual of style :)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/TemporalIndexCheckTest.scala
elahrvivaz: they used the be in the same package, thus no import - now they've been moved to utils so import needed.

elahrvivaz:
looks good

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
```diff
@@ -20,10 +20,31 @@ object GeoMesaSchemaValidator {
...
+  * names is a reserved word in ECQL.  Using those reserved words in a simple
+  * feature type will cause queries to fail.
+  */
+object ReservedWordCheck extends LazyLogging {
+
+  // ensure that no attribute names are reserved words within GeoTools that will cause query problems
+  def validateAttributeNames(sft: SimpleFeatureType): Unit = {
+    val reservedWords = FeatureUtils.sftReservedWords(sft)
+    if (reservedWords.nonEmpty) {
+      // TODO:  Make this an exception after GeoMesa 1.2.2
```
elahrvivaz: any reason to wait?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
cne1x: hehe...  And yes, it was formerly in the CMoS.  With the advent of WYSIWYG, it's dropped out of fashion, but code remains a fixed-width-font format, so I would *always* use two spaces after the colon.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
cne1x: Ooh... @jnh5y and I discussed this one.  If we make it an exception now -- which is how it was during debugging, so that we could find all of the unit tests that broke -- then we will necessarily break code that may have been working for folks.  (For example, if you ingested with a field named "between", but never queried against it.)  We decided that being less invasive up front was warranted; we can always tighten up more easily than we can re-release to relax a constraint.  Your thoughts?

jnh5y: For more waffling, I'm tempted to suggest having --strict / --lax options;).

jnh5y: Let's stick to a warning for now.

elahrvivaz: this would only affect new schemas, right? doesn't really matter to me...
so ingesting with the fields itself isn't problematic, it's just if you try to query them, correct?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
```diff
@@ -61,6 +82,7 @@ object TemporalIndexCheck extends LazyLogging {
 object MixedGeometryCheck extends LazyLogging {
 
   import java.lang.{Boolean => jBoolean}
+
```
jnh5y: (Picky) extra newline?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
cne1x: I believe you are correct on both points, Emilio.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
cne1x: More interesting than that, I think:  I organized imports here, and that likes to separate java.* imports from other packages.  If you like, I can remove the blank line.

PR #896, opened 2016-05-19
GEOMESA-1216 Fixing binned array bounds/merge

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
@@ -92,6 +92,14 @@ abstract class BinnedArray[T](val length: Int, val bounds: (T, T)) {
     * @return bounds for the bin
     */
   def bounds(index: Int): (T, T)
+
+  /**
+    * Indicates if the value is below the range of this array
+    *
+    * @param value value
+    * @return true if below, false otherwise (implies above if indexOf == -1)
+    */
+  def isBelow(value: T): Boolean
 }
 
 object BinnedArray {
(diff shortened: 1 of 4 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -274,83 +290,52 @@ class BinnedDoubleArray(length: Int, bounds: (jDouble, jDouble)) extends BinnedA
...
+
+  private lazy val (start, end): (String, String) = bounds
+  private lazy val prefixLength = start.zip(end).indexWhere { case (l, r) => l != r }
+  private lazy val prefix = start.substring(0, prefixLength)
+
+  override protected def convertToLong(value: String): Long = {
+    val normalized = BinnedStringArray.normalize(value)
+    if (normalized < start) { 0L } else if (normalized > end) { Long.MaxValue } else {
+      // note: 12 is the most base-36 numbers we can fit in Long.MaxValue
+      val sigDigits = normalized.substring(prefixLength).padTo(12, '0').substring(0, 12)
```
cne1x: Does ```padTo``` add zeros to the left or the right side?  Don't we want the zeros added to the left?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -155,14 +155,14 @@ object RangeHistogram {
 
   def buffer[T](value: T): (T, T) = {
     val buf = value match {
-      case v: String => (v, v + "z")
-      case v: Int    => (v - 1, v + 1)
-      case v: Long   => (v - 1, v + 1)
-      case v: Float  => (v - 1, v + 1)
-      case v: Double => (v - 1, v + 1)
-      case v: Date   => (v, new Date(v.getTime + 60000))
+      case v: String => (v + "0", v + "z")
```
cne1x: Do we have a pre-defined constant for the alphabet for this base conversion?  If so, can we reference it here?  Otherwise, it might takes folks a minute to realize where the "0" and "z" characters come from / what they represent.  Just a suggestion.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
elahrvivaz: yeah, padTo appends on the right side. But in this case, that's what we want, as otherwise almost every string that's not 12 digits will end up being on the very low end of the range and probably sort into the first bucket.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
elahrvivaz: i'll add one, or see if there's something out there already i can re-use.

elahrvivaz: added

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RangeHistogram.scala
```diff
@@ -154,15 +154,16 @@ class RangeHistogram[T](val attribute: Int, initialBins: Int, initialEndpoints:
 object RangeHistogram {
 
   def buffer[T](value: T): (T, T) = {
```
Afterglow375: What's the purpose of this function?

elahrvivaz: I'll add some scala docs - it's meant to take a single value and create a range from it.
When we build up stats, we take the current bounds to create the range histogram. But, if for a given attribute all the features have the same value, we wouldn't have a valid range to create a histogram with (because lower bound has to be less than the upper bound). So we take the single values and 'buffer' it when creating the histogram.
The histogram will expand as needed, so the initial bounds aren't too important. But the logic gets a lot simpler if you know that the lower bound will always be less than the upper bound.

elahrvivaz: added scaladoc

PR #895, opened 2016-05-19
GEOMESA-1155,GEOMESA-1214 Spatial/temporal filter handling doesn't co…

elahrvivaz:
…nsider attribute name or exression order

* Merged SafeTopologicalFilterVisitor and LocalNameVisitor and made them idempotent
* Removed some old methods that are no longer relevant
* Merged attibute index table sharing tests

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #900, opened 2016-05-24
GEOMESA-1182,GEOMESA-538 Stats-based query planning

elahrvivaz:
~~Note: stacks on top of #895 and #896, please consider only the last commit(s).~~

* Stats are gathered during ingestion, and can be run manually through the command-line tools
* There are now two options for query planning - stats based or the old index priority based
* Query planning can be controlled by:
  * The query hint org.locationtech.geomesa.accumulo.index.QueryHints#COST_EVALUATION_KEY
  * System property 'geomesa.query.cost.type' - either 'Stats' or 'Index'
  * Hint takes priority, then system property, otherwise uses stats based planning


anthonyccri:
What?!?  This is amazing.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
```diff
@@ -8,254 +8,280 @@
...
-   * @param numThreads the number of concurrent threads to spawn for querying
+   * @param typeName the name of the table to query and delete from
    */
-  override def delete(featureName: String, numThreads: Int): Unit =
-    if (tableLock.synchronized(tableExists)) {
-      val range = new Range(getMetadataRowKey(featureName))
-      val auths = authorizationsProvider.getAuthorizations
-      val deleter = connector.createBatchDeleter(catalogTable, auths, numThreads, metadataBWConfig)
+  override def delete(typeName: String): Unit = {
+    if (synchronized(tableExists)) {
```
anthonyccri: Does this need to wrap the whole method?  Does it have an affect on performance if it does?  Maybe we don't care that much because it's a `delete`.

elahrvivaz: the deletion should be thread safe, although obviously you wouldn't normally be calling it multiple times. The synchronization is just for the check to see if the table actually exists - we don't create the metadata table until we write something to it (so that e.g. if you fat-finger a catalog table in geoserver you don't create a random accumulo table).

jahhulbert-ccri:
so far looks good. gonna give it another lookover and try it out.

docs/user/commandline_tools.rst
```diff
@@ -420,6 +422,114 @@ Explain how a given GeoMesa query will be executed::
...
+      text [ unavailable ]
+      dtg [ 2016-02-01T00:09:12.000Z to 2016-03-01T00:21:02.000Z ] cardinality: 2161132
+      geom [ -171.75, -45.5903996, 157.7302, 89.99997102 ] cardinality: 2119237
+
+    $ geomesa stats-bounds -u username -p password -i instance -z zoo1,zoo2,zoo3 \
+        -c geomesa.data -f twitter --exact \
+        -q 'BBOX(geom,-70,45,-60,55) AND dtg DURING 2016-02-02T00:00:00.000Z/2016-02-03T00:00:00.000Z'
+      Running stat query...
+        user_id [ 1011811424 to 99124417 ] cardinality: 115
+        user_name [ foo_user to bar_user ] cardinality: 113
```
jnh5y: Is the sorting backwards here?  Does 'bar_user' come first in the dictionary?

docs/user/commandline_tools.rst
```diff
@@ -420,6 +422,114 @@ Explain how a given GeoMesa query will be executed::
...
+    $ geomesa stats-count -u username -p password -i instance -z zoo1,zoo2,zoo3 \
+        -c geomesa.data -f twitter \
+        -q 'BBOX(geom,-70,45,-60,55) AND dtg DURING 2016-02-02T00:00:00.000Z/2016-02-03T00:00:00.000Z'
+      Estimated count: 2681
+
+    $ geomesa stats-count -u username -p password -i instance -z zoo1,zoo2,zoo3 \
+        -c geomesa.data -f twitter --exact \
+        -q 'BBOX(geom,-70,45,-60,55) AND dtg DURING 2016-02-02T00:00:00.000Z/2016-02-03T00:00:00.000Z'
+      Running stat query...
+      Count: 182
```
jnh5y: If we can *always* get an estimated count, we should update geomesa.force.count to allow an estimated count hint...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -156,13 +155,12 @@ class AccumuloDataStore(val connector: Connector,
...
       // back compatible check if user data wasn't encoded with the sft
-      // noinspection ScalaDeprecation
       if (!sft.getUserData.containsKey(SCHEMA_VERSION_KEY)) {
-        metadata.read(typeName, DTGFIELD_KEY).foreach(sft.setDtgField)
+        metadata.read(typeName, "dtgfield").foreach(sft.setDtgField)
         sft.setSchemaVersion(metadata.readRequired(typeName, VERSION_KEY).toInt)
 
         // If no data is written, we default to 'false' in order to support old tables.
-        if (metadata.read(typeName, SHARED_TABLES_KEY).exists(_.toBoolean)) {
+        if (metadata.read(typeName, "tables.sharing").exists(_.toBoolean)) {
```
jnh5y: Is there a reason to ditch the globally defined variables here?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -267,8 +287,8 @@ class AccumuloDataStore(val connector: Connector,
    * @return feature reader
    */
   override def getFeatureReader(query: Query, transaction: Transaction): AccumuloFeatureReader = {
-    val stats = Some(this).collect { case w: StatWriter => w }.map((_, auditProvider))
     val qp = getQueryPlanner(query.getTypeName)
+    val stats = Some(this).collect { case w: StatWriter => w }.map((_, auditProvider))
```
jnh5y: What does this re-ordering do for us?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureSource.scala
```diff
@@ -32,54 +32,54 @@ import org.locationtech.geomesa.accumulo.process.query.QueryVisitor
...
-  //  This approach gives a rough upper count for the size of the query results.
-  //  For Filter.INCLUDE, this is likely pretty close; all others, it is a lie.
-
-  // Since users may want *actual* counts, there are two ways to force exact counts.
-  //  First, one can set the System property "geomesa.force.count".
-  //  Second, there is an EXACT_COUNT query hint.
+  override lazy val getSchema: SimpleFeatureType = dataStore.getSchema(featureName)
+
+  /**
+   * The default behavior for getCount is to use estimated statistics.
```
jnh5y: NICE!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureSource.scala
```diff
@@ -32,54 +32,54 @@ import org.locationtech.geomesa.accumulo.process.query.QueryVisitor
...
 
-  // The default behavior for getCount is to use Accumulo to look up the number of entries in
-  //  the record table for a feature.
-  //  This approach gives a rough upper count for the size of the query results.
-  //  For Filter.INCLUDE, this is likely pretty close; all others, it is a lie.
-
-  // Since users may want *actual* counts, there are two ways to force exact counts.
-  //  First, one can set the System property "geomesa.force.count".
-  //  Second, there is an EXACT_COUNT query hint.
+  override lazy val getSchema: SimpleFeatureType = dataStore.getSchema(featureName)
```
jnh5y: Hmm...  I don't think this needs to be lazy.  Are we seeing situations where getSchema isn't being called?

elahrvivaz: ah, no it used to be a def but I figured we only need to compute it once so I made it a lazy val.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: nothing, i was just trying to be more consistent across methods.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: The variables are deprecated now - instead of leaving them in there and marking them so I just copied the string values here, since they're not used anywhere else.

docs/user/commandline_tools.rst
elahrvivaz: We might not always be able to get an estimated count, if e.g. it is an older data-store and stats haven't been run on it. That may be an edge case though...
I think I updated the force.count hint for queries, did you mean something else?

docs/user/commandline_tools.rst
elahrvivaz: oops, good catch - I scrubbed out the actual user names and of course you always use foo then bar :)

elahrvivaz: fixed here and the line below and the same lines in the tools readme - let me know if you spot any other places.

docs/user/commandline_tools.rst
```diff
@@ -424,6 +426,114 @@ Explain how a given GeoMesa query will be executed::
...
+
+Example usage::
+
+    $ geomesa stats-analyze -u username -p password -c geomesa.data -f twitter
+      Running stat analysis for feature type twitter...
+      Stats analyzed:
+        Total features: 8852601
+        Bounds for geom: [ -171.75, -45.5903996, 157.7302, 89.99997102 ]
+        Bounds for dtg: [ '2016-02-01T00:09:12.000Z' to '2016-03-01T00:21:02.000Z' ]
+        Bounds for user_id: [ '100000215' to '99999502' ]
```
Afterglow375: When I run the stats-analyze command I see the cardinality appended at the end of each line. I assume that is the count of unique attribute values. Maybe document it here?

elahrvivaz: thanks, i added that after i had written the docs. I'll update.

docs/user/commandline_tools.rst
```diff
@@ -32,6 +31,10 @@ Run ``geomesa`` without any arguments to produce the following usage text::
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
         list                List GeoMesa features for a given catalog
         queryrasterstats    Export queries and statistics about the last X number of queries to a CSV file.
+        stats-analyze       Analyze statistics on a GeoMesa feature type
+        stats-bounds        View bounds on attributes in a GeoMesa schema
+        stats-count         View feature counts in a GeoMesa schema
+        stats-histogram     View statistics on a GeoMesa feature type
```
Afterglow375: [picky] improve description for stats-histogram. Also make the descriptions more consistent (i.e. "on a GeoMesa feature type" vs "in a GeoMesa schema")

docs/user/commandline_tools.rst
```diff
@@ -424,6 +426,114 @@ Explain how a given GeoMesa query will be executed::
...
+      Count: 182
+
+
+stats-histogram
+~~~~~~~~~~~~~~~
+
+Counts the features in your data set, grouped into sorted bins. You may specify the number of bins to group
+attribute into. You can count total features, or features that match a CQL filter. You can use
+pre-calculated stats for a quick estimation, or get the definitive result by querying the
+data set using the '--exact' flag.
```
Afterglow375: What does the bins flag default to?

elahrvivaz: depends... for precalculated stats, it is calculated based on the cardinality and the type of attribute. For exact queries, I believe it defaults to 1000.

Afterglow375: I find the string histogram a bit strange to look at in that it includes numbers.

e.g.
Binned histogram for 'actor2Name':
  [ 0 to 2x ] 4188
  [ 2x to 5u ] 3574
  [ 5u to 8r ] 3664
  [ 8r to bo ] 4354
  [ bo to el ] 5457
  [ el to hi ] 6158
  [ hi to kf ] 6928
  [ kf to nc ] 7402
  [ nc to q9 ] 8093
  [ q9 to t6 ] 11196
  [ t6 to w3 ] 11765
  [ w3 to z ] 2573114

In this case, actor2Name is the names of countries so there's no numbers involved. Also, where does the empty string go in this histogram? I assume after z? If so, that's inconsistent with the stats-analyze command which lists the empty string first: actor2Name [  to ZURICH ] cardinality: 6032

elahrvivaz: So sorting strings into bins is not trivial... the approach I took is to translate the string into a base36 number - hence the digits. An empty string should get converted into zeros, so it would go in the first position of the array.
I'd welcome improvements to the binning string algorithm, I just didn't come up with anything better...
I think the main problem with the histograms for non-indexed attributes is that we use default bounds for the histogram, which doesn't work very well. I'm going to add some options for the user to specify the bounds.

docs/user/commandline_tools.rst
elahrvivaz: fixed.

docs/user/commandline_tools.rst
elahrvivaz: fixed

docs/user/commandline_tools.rst
elahrvivaz: added some bounds options - you can now either use the default bounds, specify them on the command line, or look them up via another stats query.

Afterglow375:
@elahrvivaz I noticed using the removeschema command line tool command does not clear the corresponding stats for the removed feature type. Shouldn't the removeschema command also delete the stats metadata?

elahrvivaz:
@Afterglow375 yeah, good catch

elahrvivaz:
@Afterglow375 fixed the deletion - table will still be there but not the data

docs/user/installation_and_configuration.rst
```diff
@@ -184,6 +184,10 @@ Test the command that invokes the GeoMesa Tools:
         list             List GeoMesa features for a given catalog
         queryrasterstats Export queries and statistics about the last X number of queries to a CSV file.
         removeschema     Remove a schema and associated features from a GeoMesa catalog
+        stats-analyze    Analyze statistics on a GeoMesa feature type
+        stats-bounds     View bounds on attributes in a GeoMesa schema
+        stats-count      View feature counts in a GeoMesa schema
+        stats-histogram  View statistics on a GeoMesa feature type
```
Afterglow375: Add stats-enumerate here.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
```diff
@@ -8,254 +8,280 @@
...
-import scala.collection.mutable
 
 /**
  * GeoMesa Metadata/Catalog abstraction using key/value String pairs storing
- * them on a per-featurename basis
+ * them on a per-typeName basis
  */
-trait GeoMesaMetadata {
-  def delete(featureName: String, numThreads: Int)
+trait GeoMesaMetadata[T] {
```
Afterglow375: Just curious, are there plans for the other data stores like the Kafka data store to support stats as well?

docs/user/installation_and_configuration.rst
elahrvivaz: fixed.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
elahrvivaz: that's an interesting idea, and i can see the utility of it. but having accumulo lets us do a lot of nice things, like scan over ranges and use combiners. and i don't think that hbase, cassandra, etc use zookeeper, so we might not gain that much.

Afterglow375: True. Also just assumed all the other stores like hbase and cassandra used zookeeper.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/stats/StatsHistogramCommand.scala
```diff
@@ -0,0 +1,304 @@
...
+    sb.append(s"[${Threshold4._1}${Threshold4._2}${Console.RESET}  ${denormalize(0.8f)} to $max]")
+
+    sb.toString
+  }
+
+  // ANSI colors and unicode values for various density counts
+  private val Threshold1 = (Console.BOLD + Console.CYAN,   '\u26AB') // 26AB - black circle
+  private val Threshold2 = (Console.BOLD + Console.GREEN,  '\u25CF') // 25CF - medium black circle
+  private val Threshold3 = (Console.BOLD + Console.YELLOW, '\u25CF') // 25CF - medium black circle
+  private val Threshold4 = (Console.BOLD + Console.RED,    '\u2B24') // 2B24 - black circle large
```
jnh5y: What?!  No CartoCSS support for the command line map?  Tsk tsk.;)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/stats/StatsHistogramCommand.scala
```diff
@@ -0,0 +1,304 @@
...
+              }
+            }
+            if (lower == upper) {
+              bounds.put(attribute, Histogram.buffer(lower))
+            } else {
+              bounds.put(attribute, (lower, upper))
+            }
+          }
+        } else {
+          return // cancel operation
```
jnh5y: Would it be worth providing feedback that we are stopping?

geomesa-tools/README.md
```diff
@@ -229,8 +234,259 @@ To remove a feature type and it's associated data from a catalog table, use the
 ####Example:
     geomesa removeschema -u username -p password -i instance -z zoo1,zoo2,zoo3 -c test_catalog -f testfeature1
     geomesa removeschema -u username -p password -i instance -z zoo1,zoo2,zoo3 -c test_catalog --pattern 'testfeatures\d+'
-    
-    
+
+
+### stats-analyze
```
jnh5y: @mdzimmerman, to document the new GeoMesa tools for stats, is this README.md the only place that we need words?

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaOutputFormat.scala
```diff
@@ -134,6 +136,7 @@ class GeoMesaRecordWriter(params: Map[String, String],
         case (table, writer) => (new Text(table), writer)
       }
     })
+    val stats = statsCache.getOrElseUpdate(sftName, ds.stats.statUpdater(sft))
```
jnh5y: The outputformat handles stats?!?!?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/BoundsFilterVisitor.scala
```diff
@@ -0,0 +1,83 @@
...
+      case f: BegunBy                        => visitor.visit(f, envelope)
+      case f: During                         => visitor.visit(f, envelope)
+      case f: EndedBy                        => visitor.visit(f, envelope)
+      case f: Ends                           => visitor.visit(f, envelope)
+      case f: Meets                          => visitor.visit(f, envelope)
+      case f: MetBy                          => visitor.visit(f, envelope)
+      case f: OverlappedBy                   => visitor.visit(f, envelope)
+      case f: TContains                      => visitor.visit(f, envelope)
+      case f: TEquals                        => visitor.visit(f, envelope)
+      case f: TOverlaps                      => visitor.visit(f, envelope)
```
jnh5y: Ok, I'll bite, what's going on here?!

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/BoundsFilterVisitor.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.geotools.filter.visitor.ExtractBoundsFilterVisitor
+import org.geotools.geometry.jts.ReferencedEnvelope
+import org.locationtech.geomesa.utils.geotools.wholeWorldEnvelope
+import org.opengis.filter._
+import org.opengis.filter.spatial._
+import org.opengis.filter.temporal._
+
+/**
+  * Helper for extracting bounds from a filter. For some reason, ExtractBoundsFilterVisitor does
+  * not accept a generic Filter...
```
jnh5y: Oh, that's what's going on?!

jnh5y: Ok, we might be using this filter visitor backwards.  See https://github.com/geotools/geotools/blob/master/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java#L323

geomesa-blobstore/geomesa-blobstore-core/src/test/resources/log4j.xml
```diff
@@ -1,12 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-  ~ Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
-  ~ All rights reserved. This program and the accompanying materials
-  ~ are made available under the terms of the Apache License, Version 2.0 which
-  ~ accompanies this distribution and is available at
-  ~ http://www.opensource.org/licenses/apache2.0.php.
-  -->
-
```
jnh5y: Shouldn't we leave this?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/MultiIteratorTest.scala
```diff
@@ -33,6 +35,10 @@ class MultiIteratorTest extends Specification with TestWithMultipleSfts with Laz
 
   val spec = SimpleFeatureTypes.encodeType(TestData.featureType, includeUserData = true)
 
+  // noinspection LanguageFeature
+  // note: size returns an estimated amount, instead we need to actually count the features
+  implicit def collectionToIter(c: SimpleFeatureCollection): Iterator[SimpleFeature] = SelfClosingIterator(c)
```
jnh5y: Could/should this be typed as a SelfClosingIterator?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
```diff
@@ -205,18 +206,29 @@ object Z3IdxStrategy extends StrategyProvider {
...
-  // TODO also if very small bbox, z2 has ~10 more bits of lat/lon info
-    if (filter.primary.exists(isSpatialFilter)) 200 else 401
+    * More than id lookups (at 1), high-cardinality attributes (at 1).
+    * Less than unknown cardinality attributes (at 999).
+    * With a spatial component, less than z2, otherwise more than z2 (at 400)
+    */
+  override protected def indexBasedCost(sft: SimpleFeatureType,
+                                        filter: QueryFilter,
+                                        transform: Option[SimpleFeatureType]): Long = {
+    if (filter.primary.exists(isSpatialFilter)) 200L else 401L
```
jnh5y: So far these numbers are still completely heuristic, right?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z2IdxStrategy.scala
```diff
@@ -175,16 +176,24 @@ object Z2IdxStrategy extends StrategyProvider {
...
+
   /**
-    * TODO update description
-    * Gets the estimated cost of running the query. Currently, cost is hard-coded to sort between
-    * strategies the way we want. Z2 should be more than id lookups (at 1), high-cardinality attributes (at 1)
-    * and less than STidx (at 400) and unknown cardinality attributes (at 999).
-    *
-    * Eventually cost will be computed based on dynamic metadata and the query.
+    * More than id lookups (at 1), high-cardinality attributes (at 1), z3 (at 200).
+    * Less than spatial-only z3 (at 401), unknown cardinality attributes (at 999).
```
jnh5y: Do we have a preference for using Z2 if available for spatial-only queries?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Strategy.scala
```diff
@@ -201,7 +209,29 @@ object Strategy extends LazyLogging {
...
+              filter: QueryFilter,
+              transform: Option[SimpleFeatureType],
+              stats: GeoMesaStats,
+              eval: CostEvaluation): Long = {
+    if (eval == CostEvaluation.Stats) {
+      statsBasedCost(sft, filter, transform, stats).getOrElse(indexBasedCost(sft, filter, transform))
+    } else if (eval == CostEvaluation.Index) {
+      indexBasedCost(sft, filter, transform)
+    } else {
+      throw new NotImplementedError(s"Unknown cost evaluation type $eval")
```
jnh5y: Hmm...  I wonder if this should be a) completely pluggable (SPI), or b) a sealed trait.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryStrategyDecider.scala
```diff
@@ -49,14 +52,24 @@ object QueryStrategyDecider extends QueryStrategyDecider with MethodProfiling {
...
                                 requested: Option[StrategyType],
                                 output: ExplainerOutputType): Seq[Strategy] = {
 
     implicit val timings = new Timing()
 
     // get the various options that we could potentially use
-    val options = new QueryFilterSplitter(sft).getQueryOptions(query.getFilter, output)
+    val options = {
+      val all = new QueryFilterSplitter(sft).getQueryOptions(query.getFilter, output)
+      // don't evaluate z2 index if there is a z3 option, it will always be picked
```
jnh5y: Wait... Z2 should be preferred for spatial queries...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategyV5.scala
```diff
@@ -208,18 +210,23 @@ class AttributeIdxStrategyV5(val filter: QueryFilter) extends Strategy with Lazy
...
-    if (cost == 0) Int.MaxValue else cost // cost == 0 if somehow the filters don't match anything
-  }
+  override def getCost(sft: SimpleFeatureType,
+                       filter: QueryFilter,
+                       transform: Option[SimpleFeatureType],
+                       stats: GeoMesaStats,
+                       evaluation: CostEvaluation) =
+    AttributeIdxStrategy.getCost(sft, filter, transform, stats, evaluation)
+
+  // note: won't get called as we override getCost method, above
```
jnh5y: And that's sensible since the V5 Attribute table is somewhat deprecated?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
```diff
@@ -203,7 +204,45 @@ object AttributeIdxStrategy extends StrategyProvider {
...
+  }
+
+  /**
+    * full index:
+    *   high cardinality - 1
+    *   unknown cardinality - 101
+    * join index:
+    *   high cardinality - 10
+    *   unknown cardinality - 1010
+    * low cardinality - Long.MaxValue
```
jnh5y: Is the 10 to 1 ratio between join and full based on anything empirical?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
```diff
@@ -0,0 +1,465 @@
...
+            val lt = Z3Table.getWeekAndSeconds(bounds.min.getTime)._1
+            val ut = Z3Table.getWeekAndSeconds(bounds.max.getTime)._1
+            Range.inclusive(lt, ut).map(_.toShort)
+          }.getOrElse(Seq.empty)
+        }
+        val histograms = weeks.map(histogramKey(geom, dtg, _)).flatMap(readStat[Z3Histogram](sft, _))
+        // combine the week splits into a single stat
+        Z3Histogram.combine(histograms)
+      }
+      z.flatten.toSeq
```
jnh5y: This is a tad dense.  Seems like there's only going to be one (geom, dtg) pair considered...  yielding over that seems clunky.  Not a huge blocker, but if there's a way to clean this up, it's be nice.

jnh5y:
I have finished a read-through.  As soon as @Afterglow375 is happy from a testing side and my idle comments are addressed/dismissed, I think we'll be good to merge this PR!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/stats/StatsHistogramCommand.scala
elahrvivaz: fixed.

geomesa-tools/README.md
elahrvivaz: It's also updated in the docs

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaOutputFormat.scala
elahrvivaz: well yeah :)

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/BoundsFilterVisitor.scala
elahrvivaz: There's a comment at the top - basically the geotools bounds visitor for some reason doesn't accept a generic filter :/

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/BoundsFilterVisitor.scala
elahrvivaz: ah, haha. I'll fix.

geomesa-blobstore/geomesa-blobstore-core/src/test/resources/log4j.xml
elahrvivaz: we don't have copyrights in our other logging files - we discussed at one point and decided no need to copyright things that don't show any 'originality'

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/MultiIteratorTest.scala
elahrvivaz: Not sure it really matters... do you have a strong opinion?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategy.scala
elahrvivaz: yeah, these are the old way we calculated costs - left for data stores that don't have stats, or if the stats query planning proves disastrous, you can fall back to this with a system property.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Z2IdxStrategy.scala
elahrvivaz: yeah, it will use z2 for spatial only queries. because it's not binned by week, it will be a lot faster.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Strategy.scala
elahrvivaz: It's an enumeration - not good enough? I think the SPI is a bit overboard at this point :)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryStrategyDecider.scala
elahrvivaz: yeah, there will only be a z3 and z2 option if there is a spatial and a temporal component - slightly confusing in that those checks are done in the query splitter, and not here. The thinking is that eventually we probably want to evaluate both z2 and z3 options when there is a temporal component, as if you search for e.g. 5 years z3 is going to struggle with all the ranges per week. However, since we don't consider that now, we just filter out the z2's up front so we don't have to calculate estimated costs.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategyV5.scala
elahrvivaz: well it uses the same logic as the newer table for the cost, so it just delegates to that. b/c I made the delegate cost methods protected, this class can't access them on the new attribute strategy class, so I had to override the main method - but still had to implement the delegate methods to satisfy the trait. Kind of smelly, but as you said it's deprecated so I don't mind it so much.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIdxStrategy.scala
elahrvivaz: no, it's been that way for a while.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
elahrvivaz: did my best... see if it's any clearer. I like the for/yield when needing two options, otherwise you have to do `opt1.flatMap(o1 => opt2.map((_, o1)))` or something...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/BoundsFilterVisitor.scala
elahrvivaz: fixed.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/MultiIteratorTest.scala
jnh5y: Let's try SelfClosingIterator.  I've had some requests to be able to close resources and having access to the proper 'close' method seems beneficial.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/Strategy.scala
jnh5y: Yeah, since it is an enumeration, I was thinking that a sealed trait might be another option.

We need an SPI to decide which components are SPI-loadable.... then the ice cream will consume itself...

Meh, this one ain't worth thinking about more.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryStrategyDecider.scala
jnh5y: Ok, shout if I have this wrong:  spatial-only goes to Z2.  ST queries go to Z3.  

At some point, we could calculate costs and route long-temporal ST queries to Z2, but not now, right?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/MultiIteratorTest.scala
elahrvivaz: this is just a local method in a unit test... i'll change it though

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryStrategyDecider.scala
elahrvivaz: yep, you got it.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/MultiIteratorTest.scala
elahrvivaz: updated

jnh5y: Doh; missed that it was a unit test.

geomesa-tools/README.md
mdzimmerman: There's a shorter summary of the command line arguments in ``docs/user/commandline_tools.rst``, but as @elahrvivaz says, it's already updated in this PR.

jnh5y:
Merged as 5e1dc4ea6c8b0c26514cfe21626c8c66b2556815.

PR #902, opened 2016-05-27
GEOMESA-978 Encode SFTs in TypeSafe Config

jahhulbert-ccri:

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
@@ -22,11 +22,11 @@ class GetSftCommand(parent: JCommander) extends CommandWithCatalog(parent) with 
   override def execute() = {
     logger.info(s"Getting SFT for feature ${params.featureName} from catalog ${catalog}")
     try {
-      params.format match {
+      params.format.toLowerCase match {
         case "typesafe" =>
-          println(SimpleFeatureTypes.toConfigString(ds.getSchema(params.featureName), params.userData, params.concise))
+          println(SimpleFeatureTypes.toConfigString(ds.getSchema(params.featureName), !params.excludeUserData, params.concise))
         case "spec" =>
-          println(SimpleFeatureTypes.encodeType(ds.getSchema(params.featureName), params.userData))
+          println(SimpleFeatureTypes.encodeType(ds.getSchema(params.featureName), !params.excludeUserData))
         case _ =>
           logger.error(s"Unknown config format: ${params.format}")
       }
@@ -43,14 +43,14 @@ class GetSftCommand(parent: JCommander) extends CommandWithCatalog(parent) with 
 object GetSftCommand {
   @Parameters(commandDescription = "Get the SimpleFeatureType of a feature")
   class GetSftParameters extends GeoMesaConnectionParams with FeatureTypeNameParam {
-    @Parameter(names = Array("--concise"), description = "Render in concise format (true/false)", required = false, arity =1)
+    @Parameter(names = Array("--concise"), description = "Render in concise format", required = false)
     var concise: Boolean = false
 
     @Parameter(names = Array("--format"), description = "Formats for sft (comma separated string, allowed values are typesafe, spec)", required = false)
     var format: String = "typesafe"
 
-    @Parameter(names = Array("--with-user-data"), description = "Include user data (true/false)", required = false, arity = 1)
-    var userData: Boolean = true
+    @Parameter(names = Array("--exclude-user-data"), description = "Exclude user data", required = false)
+    var excludeUserData: Boolean = false
   }
 }
 
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
@@ -8,7 +8,7 @@
 
 package org.locationtech.geomesa.tools.accumulo.commands
 
-import com.beust.jcommander.{JCommander, Parameters}
+import com.beust.jcommander.{JCommander, Parameter, Parameters}
 import com.typesafe.scalalogging.LazyLogging
 import org.locationtech.geomesa.tools.accumulo.GeoMesaConnectionParams
 import org.locationtech.geomesa.tools.accumulo.commands.GetSftCommand._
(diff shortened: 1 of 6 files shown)


jahhulbert-ccri:
the printing looks like this: note the user data but it worked when i reingested and only created the records table :smile: 

```
geomesa {
    sfts {
        csv2 {
            fields=[
                {
                    index=join
                    name=name
                    type=String
                },
                {
                    index=join
                    name=lastseen
                    type=Date
                },
                {
                    name=age
                    type=Integer
                },
                {
                    name=id
                    type=Integer
                },
                {
                    name=talents
                    type="Map[String,Integer]"
                },
                {
                    index=join
                    name=friends
                    type="List[String]"
                },
                {
                    default="true"
                    name=geom
                    srid="4326"
                    type=Point
                }
            ]
            type-name=csv2
            user-data {
                geomesa {
                    index {
                        dtg=lastseen
                    }
                    table {
                        sharing="true"
                    }
                    version="8"
                    indexes {
                       enabled="records"
                    }
                }
            }
        }
    }
}
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
```diff
@@ -22,7 +22,14 @@ class GetSftCommand(parent: JCommander) extends CommandWithCatalog(parent) with
   override def execute() = {
     logger.info(s"Getting SFT for feature ${params.featureName} from catalog ${catalog}")
     try {
-      println(SimpleFeatureTypes.encodeType(ds.getSchema(params.featureName)))
+      params.format match {
```
elahrvivaz: maybe add a toLowerCase here just to be more lenient

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
```diff
@@ -35,7 +42,15 @@ class GetSftCommand(parent: JCommander) extends CommandWithCatalog(parent) with
 
 object GetSftCommand {
   @Parameters(commandDescription = "Get the SimpleFeatureType of a feature")
-  class GetSftParameters extends GeoMesaConnectionParams
-    with FeatureTypeNameParam {}
+  class GetSftParameters extends GeoMesaConnectionParams with FeatureTypeNameParam {
+    @Parameter(names = Array("--concise"), description = "Render in concise format (true/false)", required = false, arity =1)
```
elahrvivaz: how about making this arity 0 so you specify '--concise' instead of '--concise true' ? (i think that's the way arity works...)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
```diff
@@ -35,7 +42,15 @@ class GetSftCommand(parent: JCommander) extends CommandWithCatalog(parent) with
...
+    var concise: Boolean = false
+
+    @Parameter(names = Array("--format"), description = "Formats for sft (comma separated string, allowed values are typesafe, spec)", required = false)
+    var format: String = "typesafe"
+
+    @Parameter(names = Array("--with-user-data"), description = "Include user data (true/false)", required = false, arity = 1)
```
elahrvivaz: same thing here - how about changing to '--exclude-user-data' and make arity 0?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -161,6 +174,41 @@ object SimpleFeatureTypes {
...
 
+  def toConfig(sft: SimpleFeatureType, includeUserData: Boolean = true): Config = {
+    import scala.collection.JavaConverters._
+    var fieldOptMap: Map[String, java.util.Map[String, String]] = sft.getAttributeDescriptors.map { ad =>
+      ad.getLocalName -> AttributeSpecFactory.fromAttributeDescriptor(sft, ad).toMap.asJava
+    }.toMap[String, java.util.Map[String, String]]
+
+    Option(sft.getGeometryDescriptor).foreach { gd =>
+      val g = gd.getLocalName
+      fieldOptMap = fieldOptMap.updated(g, fieldOptMap(g).updated("default", "true").asJava)
```
elahrvivaz: do we need to do this for the date too, or will that be handled already?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -161,6 +174,41 @@ object SimpleFeatureTypes {
...
+        })
+      base.withValue(UserDataPath, userData.root())
+    } else {
+      base
+    }
+    updated.atPath(s"${ConfigSftParsing.path}.${sft.getTypeName}")
+  }
+
+  def toConfigString(sft: SimpleFeatureType, includeUserData: Boolean = true, concise: Boolean = false): String = {
+    val opts = if (concise) ConfigRenderOptions.concise else ConfigRenderOptions.defaults().setFormatted(true).setComments(false).setOriginComments(false).setJson(false)
```
elahrvivaz: [picky] long line

elahrvivaz:
nice!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
jahhulbert-ccri: will do

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
jahhulbert-ccri: yeah that's how it works...will do!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/GetSftCommand.scala
jahhulbert-ccri: +1

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
jahhulbert-ccri: hmmmm good question i'll look into it

jnh5y:
@jahhulbert-ccri @elahrvivaz would it be worth suppressing certain userData when writing out the SFTs?  The version='8' seems silly/bad.  On the other hand, index.enabled ='records' seems good...

jnh5y:
Is there anyway to make the update more terse/concise?  I expect to see lines like index.enabled='records' rather than lots of {}'s.

jahhulbert-ccri:
@jnh5y what do you mean by the "update" ? the config? if so then no there's no default way that i know of ... you can pass the -concise flag if you desire...i'll look into seeing if we can get it keyed like "geomesa.index.enabled" = "records" but by default the formatter doesn't do that

jahhulbert-ccri:
we could possibly add a --all flag if you wanted to suppress certain data, i'd suggest a followon PR. I believe that currently it only gives things that start with "geomesa" anyways.

elahrvivaz:
I checked briefly when you mentioned the formatting of the user data keys, but I didn't see an option to flatten them.

jahhulbert-ccri:
yeah i couldn't find anything either

PR #903, opened 2016-05-27
For Comment: Load Converters and Config via annotations

jahhulbert-ccri:
Example of how we can load things via annotations and not have to worry about META-INF/resources SPI files...Users can just annotate things. We could also do a SimpleFeatureType loading this way as well.

Lemme know what you think.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1876,6 +1876,11 @@
                 <artifactId>jcommander</artifactId>
                 <version>1.48</version>
             </dependency>
+            <dependency>
+                <groupId>org.reflections</groupId>
+                <artifactId>reflections</artifactId>
+                <version>0.9.10</version>
+            </dependency>
 
             <!-- test dependencies -->
             <dependency>
--- a/geomesa-convert/geomesa-convert-common/pom.xml
+++ b/geomesa-convert/geomesa-convert-common/pom.xml
@@ -44,6 +44,10 @@
             <artifactId>jai_core</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.reflections</groupId>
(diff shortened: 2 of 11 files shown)


jnh5y:
@jahhulbert-ccri missing headers ahoy!
[WARNING] Missing header in: /home/travis/build/locationtech/geomesa/geomesa-convert/geomesa-convert-common/src/main/java/org/locationtech/geomesa/convert/annotations/ConverterConfig.java

[WARNING] Missing header in: /home/travis/build/locationtech/geomesa/geomesa-convert/geomesa-convert-common/src/main/java/org/locationtech/geomesa/convert/annotations/ConverterFactory.java

[WARNING] Missing header in: /home/travis/build/locationtech/geomesa/geomesa-convert/geomesa-convert-common/src/main/java/org/locationtech/geomesa/convert/annotations/TransformFactory.java

pom.xml
```diff
@@ -1876,6 +1876,11 @@
                 <artifactId>jcommander</artifactId>
                 <version>1.48</version>
             </dependency>
+            <dependency>
+                <groupId>org.reflections</groupId>
+                <artifactId>reflections</artifactId>
+                <version>0.9.10</version>
```
jnh5y: What kinda license we looking at?

jnh5y:
The nice thing about SPI is that editing a text file can control what is available.  Is there a good way to setup annotations after the fact?  I.e., if we miss an annotation, we'd have to recompile, right?

Other than preference of advertisement, are there other reasons to consider annotations?  

I think I'd need a little convincing...

jahhulbert-ccri:
closing. not doing it looks like.

PR #904, opened 2016-05-30
Parsing keywords from user data into feature sources

atallahhezbor:


jnh5y:
Hi @atallahhezbor, I created a ticket for this work here: https://geomesa.atlassian.net/browse/GEOMESA-1226

As you enter a new PR, can you make the first line "GEOMESA-1226 Map userData to keywords in FeatureSources".  (Feel free to edit the title of the JIRA ticket and use that in the new PR.)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -246,5 +251,8 @@ object RichSimpleFeatureType {
       Seq(GEOMESA_PREFIX) ++ userData[String](USER_DATA_PREFIX).map(_.split(",")).getOrElse(Array.empty)
 
     def userData[T](key: AnyRef): Option[T] = Option(sft.getUserData.get(key).asInstanceOf[T])
+
+    def getKeywords: java.util.Set[String] = userData[String](KEYWORDS_KEY).map(_.split(KEYWORDS_DELIMITTER).toSet.asJava).getOrElse(java.util.Collections.emptySet[String])
```
jnh5y: Did you want to add 'add|removeKeyword' functions in this PR or another one?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -56,6 +56,8 @@ class AccumuloDataStoreTest extends Specification with TestWithMultipleSfts {
 
   val defaultGeom = WKTUtils.read("POINT(45.0 49.0)")
 
+  val keywordsDelimitter = "|"
```
jnh5y: I'd suggest moving this to geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -191,6 +193,9 @@ object RichSimpleFeatureType {
   val DEFAULT_DATE_KEY    = "geomesa.index.dtg"
   val ST_INDEX_SCHEMA_KEY = "geomesa.index.st.schema"
   val USER_DATA_PREFIX    = "geomesa.user-data.prefix"
+  val KEYWORDS_KEY        = "geomesa.keywords"
+
+  val KEYWORDS_DELIMITTER = """\|"""
```
jnh5y: This is actually the regex to split a (Java) String.

As an interesting note...  String.split() takes a regex pattern.  https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#split-java.lang.String-

Scala Stringlike adds a method which takes a Char:
http://www.scala-lang.org/api/2.11.5/index.html#scala.collection.immutable.StringLike@split(separator:Char):Array[String]

We could do...
val s = "foo|bar"
s.split('|')

But we'd still have to call mkString with a string...
Seq("bar", "baz").mkString("|").

Anyhow, might be worth having a separate regex and joiner defined.  Either that or @elahrvivaz  and @jahhulbert-ccri  will have a better idea;).

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -431,6 +431,8 @@ object SimpleFeatureTypes {
     }
   }
 
+  // Do something smart here for keyword-userData??!?
```
jnh5y: Eh, my comment can go.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraFeatureStore.scala
```diff
@@ -65,6 +66,12 @@ class CassandraFeatureStore(entry: ContentEntry) extends ContentFeatureStore(ent
   }
 
 
+  override def getInfo: ResourceInfo = {
```
jnh5y: Sine we have this four places, I wonder if it makes sense to extract a tiny function to do this keyword enrichment?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
atallahhezbor: I was thinking another PR, but it wouldn't be too bad to throw that in here

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraFeatureStore.scala
atallahhezbor: Sounds good to me. Is there a place for helper functions like that?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureSource.scala
```diff
@@ -228,7 +232,7 @@ class DelegatingResourceInfo(source: SimpleFeatureSource) extends ResourceInfo {
 
   import scala.collection.JavaConversions._
 
-  private val keywords = Collections.unmodifiableSet(Set("features", getName))
+  private val keywords: java.util.Set[String] = new java.util.HashSet[String](List("features", getName))
```
elahrvivaz: as a defensive programming strategy, it might be better to keep this as an unmodifiable set, and just add source.getSchema.getKeywords when it's created (instead of line 103 above)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -72,6 +74,21 @@ class AccumuloDataStoreTest extends Specification with TestWithMultipleSfts {
...
+      val keywords: java.util.Set[String] = Set("keywordA", "keywordB", "keywordC")
+      sftWithKeywords.getUserData.put("geomesa.keywords", keywords.mkString(keywordsDelimitter))// Put keywords in userData
+
+      val spec = SimpleFeatureTypes.encodeType(sftWithKeywords, true)
+      val newType = SimpleFeatureTypes.createType("keywordsTest", spec)
+      ds.createSchema(newType)
+
+      val fs = ds.getFeatureSource(newType.getTypeName)
+
+      (fs.getInfo.getKeywords containsAll keywords) mustEqual true
```
elahrvivaz: [minor] this would work better as `fs.getInfo.getKeywords must containAllOf(keywords)` (syntax may be slightly off).
The advantage is that if that fails, you will get a more descriptive error message (e.g. list(foo, bar) contains foo and bar but does not contain baz), vs as currently written you would just get a 'value is false' failure.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: [formatting] long line

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: Any reason not to use a comma? that is the usual default delimiter for things like this, and it's not a special regex character so you don't have to escape it for splitting.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: In general, we try to avoid comments like this, as they tend to get lost in the code. If it's something that you want to follow up on, create an issue in JIRA and reference it here as a TODO (if you don't have jira permissions someone can create it for you).
That said, it's probably easiest just to leave it as a string so that it will get persisted correctly, and split it in the RichSimpleFeatureType (like you're doing). Otherwise we will need to start adding converters on read/write for different user-data keys.

jnh5y: ( @elahrvivaz, the comment is mine from when I started the branch:) )

elahrvivaz:
looks good @atallahhezbor!
Did you open a CLA with eclipse? You'll need to do that - instructions are in the contributing doc here: https://github.com/locationtech/geomesa/blob/master/CONTRIBUTING.md
You should also update your commit with a 'sign-off', using the -s flag (or someone can do it as they merge).

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: lol, ok - well see comment above :)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
jnh5y: Could a keyword indicate a field splitter/delimiter?  

Due to that, I had suggested '|'.

atallahhezbor:
Thank you both for looking over this!
@elahrvivaz Yup! I signed that after realizing I needed to yesterday.

I'll work on these fixes and hopefully push a passing commit sometime today

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: Not sure - is there a specific use case in mind for this? Presumably these are going to be added to e.g. an SFT conf file? Seems like for e.g. stealth, there aren't any commas

elahrvivaz: Maybe we shouldn't jam them together - use any user-data that has a key starting with 'geomesa.keyword'? Kind of would create a lot of weakly identified conventions to follow though.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureSource.scala
atallahhezbor: @elahrvivaz Running into issues with having it as unmodifiable. I get an UnsupportedOperationException when I try to combine source.getSchema.getKeywords with Set("features", getName). Even making it explicit like

    val l = source.getSchema.getKeywords  

    l.addAll(List("features", getName))      

    Collections.unmodifiableSet(l)

still throws the error.


Would it be reasonable to have "features" and getName be returned from source.getSchema.getKeywords?

elahrvivaz: does it work like:
`Collections.unmodifiableSet(Set("features", getName) ++ source.getSchema.getKeywords)`
?
++ will take two unmodifiable collections and create a new one from them.

atallahhezbor: Ah! That's the operator I was looking for! Thanks!

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraFeatureStore.scala
Afterglow375: I would say it should go in a new trait here: https://github.com/locationtech/geomesa/tree/master/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools

but @elahrvivaz would have a better idea.

elahrvivaz: seems reasonable to me. definitely in geomesa-utils, I think that's the only module imported into all projects.

atallahhezbor:
Had to squash these commits into a new signed-off one. 
Continue the trail [here](https://github.com/locationtech/geomesa/pull/907/)

elahrvivaz:
replaced with #907

PR #907, opened 2016-05-31
GEOMESA-1226 Map userData to keywords in FeatureSources

atallahhezbor:


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -191,6 +193,11 @@ object RichSimpleFeatureType {
   val DEFAULT_DATE_KEY    = "geomesa.index.dtg"
   val ST_INDEX_SCHEMA_KEY = "geomesa.index.st.schema"
   val USER_DATA_PREFIX    = "geomesa.user-data.prefix"
+  val KEYWORDS_KEY        = "geomesa.keywords"
+
+  val KEYWORDS_SPLITTER = "\\|"
```
atallahhezbor: Looking ahead to having a command-line tool add keywords, maybe a comma would actually be better here? I've seen that used in other command line arguments (ex: tzoo1,tzoo2,tzoo3)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -72,6 +73,21 @@ class AccumuloDataStoreTest extends Specification with TestWithMultipleSfts {
       ds.getSchema(defaultSft.getTypeName) mustEqual defaultSft
     }
 
+    "create a schema with keywords" in {
```
cne1x: Given that this is in the ```AccumuloDataStoreTest```, is there any merit to making this a fuller round-trip through a MockAccumulo instance in a real data store?

jnh5y: I believe ```ds``` is backed by a MockAccumulo datastore.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/ContentFeatureSourceInfo.scala
```diff
@@ -0,0 +1,23 @@
...
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.utils.geotools
+
+import org.geotools.data.store.ContentFeatureSource
+import org.geotools.data.ResourceInfo
+
+
+trait ContentFeatureSourceInfo extends ContentFeatureSource {
```
jnh5y: Definitely not for this ticket... 

... but, are there other shared goodies we could pull up into this trait?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -202,9 +203,43 @@ class AccumuloDataStore(val connector: Connector,
...
+        "Schema version" -> SCHEMA_VERSION_KEY,
+        "Table sharing status" -> TABLE_SHARING_KEY,
+        "Sharing prefix" -> SHARING_PREFIX_KEY,
+        "Default date" -> DEFAULT_DATE_KEY,
+        "St index schema" -> ST_INDEX_SCHEMA_KEY,
+        "User data prefix" -> USER_DATA_PREFIX)
+
+    unmodifiableUserdataKeys.foreach({ case (key, value) =>
+      if (sft.getUserData.contains(value) && sft.userData[String](value) != previousSft.userData[String](value)) {
+        throw new UnsupportedOperationException("Updating " + key + " is not allowed")
```
atallahhezbor: @elahrvivaz I was not sure if this was exhaustive. 
Or should it be more permissive if there is other user data that is okay to change?

elahrvivaz: I think what we need to maintain is the following:

`RichSimpleFeatureType.SCHEMA_VERSION_KEY`
`RichSimpleFeatureType.TABLE_SHARING_KEY`
`RichSimpleFeatureType.SHARING_PREFIX_KEY`
`RichSimpleFeatureType.DEFAULT_DATE_KEY`
`RichSimpleFeatureType.ST_INDEX_SCHEMA_KEY`
`SimpleFeatureTypes.ENABLED_INDEXES`

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -202,9 +203,43 @@ class AccumuloDataStore(val connector: Connector,
...
+    val unmodifiableUserdataKeys = HashMap(
+        "Geomesa prefix" -> GEOMESA_PREFIX,
+        "Schema version" -> SCHEMA_VERSION_KEY,
+        "Table sharing status" -> TABLE_SHARING_KEY,
+        "Sharing prefix" -> SHARING_PREFIX_KEY,
+        "Default date" -> DEFAULT_DATE_KEY,
+        "St index schema" -> ST_INDEX_SCHEMA_KEY,
+        "User data prefix" -> USER_DATA_PREFIX)
+
+    unmodifiableUserdataKeys.foreach({ case (key, value) =>
```
jahhulbert-ccri: don't think you need the ```(``` and ```)``` here just the curlies

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -202,9 +203,43 @@ class AccumuloDataStore(val connector: Connector,
...
+    // Currently this method only allows for updating of keywords in user data. All other attempted changes will fail
+
+    // Get previous schema and user data
+    val previousSft = getSchema(typeName)
+    val schemaTypeName = sft.getTypeName()
+    val existingUserData = metadata.read(schemaTypeName, ATTRIBUTES_KEY)
+
+    // Check that unmodifiable user data has not changed
+    // Currently only keywords at KEYWORDS_KEY are allowed
+    val unmodifiableUserdataKeys = HashMap(
```
jahhulbert-ccri: any reason to use hashmap here instead of the default Map ?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaConsumerFeatureSource.scala
```diff
@@ -38,7 +38,10 @@ abstract class KafkaConsumerFeatureSource(entry: ContentEntry,
                                           query: Query)
   extends ContentFeatureSource(entry, query)
   with ContentFeatureSourceSecuritySupport
-  with ContentFeatureSourceReTypingSupport {
+  with ContentFeatureSourceReTypingSupport
+  with ContentFeatureSourceInfo{
```
jahhulbert-ccri: [picky] space before ```{```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
jahhulbert-ccri: i would at minimum add a check to make sure that incoming keywords don't contain the delimiting symbol

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: I would actually make this a Set, and in the exception put the actual key instead of trying to make it human readable. If someone is mucking with this stuff, then they should know what the prefixes are or at least how to find them.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
jahhulbert-ccri: I'd probably also vote for a comma as the default list delimiter for anything

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -202,9 +203,43 @@ class AccumuloDataStore(val connector: Connector,
...
-  override def updateSchema(typeName: Name, sft: SimpleFeatureType): Unit =
-    throw new UnsupportedOperationException("AccumuloDataStore does not support updateSchema")
+  override def updateSchema(typeName: Name, sft: SimpleFeatureType): Unit = {
+    import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType._
+
+    // Currently this method only allows for updating of keywords in user data. All other attempted changes will fail
+
+    // Get previous schema and user data
+    val previousSft = getSchema(typeName)
+    val schemaTypeName = sft.getTypeName()
```
jahhulbert-ccri: whats the behavior if the type names don't match? do we intend it to change it or keep it the same

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
atallahhezbor: I'm actually reworking this per a discussion with @jnh5y to use a null byte instead. This way incoming keywords can contain any symbol

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: off the top of my head, i think we should be able to support changing type names for the latest schema version (8?). that might mess up geoserver layers and such though, so maybe we should be careful with it.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: how are you going to allow null bytes in e.g. an sft string passed in from the command line?

atallahhezbor: @elahrvivaz sorry I was referring to the delimiter for encoding the keywords array being a null byte. Command line args will likely be a comma.

elahrvivaz: I would just keep it as a comma everywhere unless we have a compelling use case not to... otherwise we are going to have to do a lot of special handling to check for commas and replace them, etc - plus if we're using commas in the input strings, doesn't that imply that restrict us from having commas in the list values?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
atallahhezbor: Current behavior is if the type names don't match (A and B) then the metadata from B will overwrite the metadata for A. Good catch. I can add a check and throw an exception for now.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
atallahhezbor: Jim usually links to [this](https://ccristudio.jira.com/wiki/pages/viewpage.action?spaceKey=STEALTH&title=Keywords) as the compelling use case not to. 

It does restrict us in the command line. We also discussed maybe only allowing adding one keyword at a time

atallahhezbor: As an update on this, there is a workaround for the command line where JCommander will construct a list of arguments if a flag is repeated multiple times.
Something like 

    geomesa keyword -a keywordA -a keyword=1,2,3

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -24,7 +24,7 @@ import org.opengis.feature.simple.SimpleFeatureType
 import scala.collection.JavaConversions._
 import scala.util.parsing.combinator.JavaTokenParsers
 import scala.util.{Failure, Success, Try}
-
+import RichSimpleFeatureType._
```
jnh5y: I'd use the fully qualified import.  Also, is this import necessary?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -246,5 +252,27 @@ object RichSimpleFeatureType {
       Seq(GEOMESA_PREFIX) ++ userData[String](USER_DATA_PREFIX).map(_.split(",")).getOrElse(Array.empty)
 
     def userData[T](key: AnyRef): Option[T] = Option(sft.getUserData.get(key).asInstanceOf[T])
+
+    def getKeywords: java.util.Set[String] = {
+      userData[String](KEYWORDS_KEY).map(_.split(KEYWORDS_DELIMITER).toSet.asJava).getOrElse(java.util.Collections.emptySet[String])
```
jnh5y: Long line. Breaking it out with the map and the getOrElse on their own line might make this more readable as well.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -202,9 +203,42 @@ class AccumuloDataStore(val connector: Connector,
...
+
+    // Check that the rest of the schema has not changed (columns, types, etc)
+    val previousColumns = previousSft.getAttributeDescriptors
+    val currentColumns = sft.getAttributeDescriptors
+    if (previousColumns != currentColumns) {
+      throw new UnsupportedOperationException("Updating schema columns is not allowed")
+    }
+
+    // If all is well, update the metadata
+    writeMetadata(sft)
```
elahrvivaz: instead of writing the metadata with this function (which will likely cause some issues...), just write the attributes with
```
val attributesValue   = SimpleFeatureTypes.encodeType(sft, includeUserData = true)
metadata.write(schemaTypeName, ATTRIBUTES_KEY, attributesValue)
```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
atallahhezbor: RichSimpleFeatureType holds the constants for the key and the delimiter, but I can fully qualify

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
jnh5y: Alternatively, this line
```
 userData[String](KEYWORDS_KEY).toSet.flatMap(_.split(KEYWORDS_DELIMITER))
```
would do the same and allow for removing the ```import scala.collection.JavaConverters._```

I'd have a small preference for that since I prefer JavaConversions over JavaConverters;)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -72,6 +74,97 @@ class AccumuloDataStoreTest extends Specification with TestWithMultipleSfts {
...
+
+      val sftWithKeywords: SimpleFeatureType = createNewSchema("name:String", dtgField = None)
```
elahrvivaz: this test could be simplified a lot if you just put the keywords at the end of the spec string. e.g.
`createNewSchema(s"name:String;$KEYWORDS_KEY='keywordA,keywordB,keywordC'")
Then you don't have to create a second schema

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -246,5 +252,27 @@ object RichSimpleFeatureType {
       Seq(GEOMESA_PREFIX) ++ userData[String](USER_DATA_PREFIX).map(_.split(",")).getOrElse(Array.empty)
 
     def userData[T](key: AnyRef): Option[T] = Option(sft.getUserData.get(key).asInstanceOf[T])
+
+    def getKeywords: java.util.Set[String] = {
```
elahrvivaz: make this a scala set instead of a java set - we only use java classes for externally facing things.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -246,5 +252,27 @@ object RichSimpleFeatureType {
...
+      userData[String](KEYWORDS_KEY).map(_.split(KEYWORDS_DELIMITER).toSet.asJava).getOrElse(java.util.Collections.emptySet[String])
+    }
+
+    def addKeywords(keywordsString: String): Unit = {
+      val currentKeywords = userData[String](KEYWORDS_KEY).map(_.split(KEYWORDS_DELIMITER).toSet).get
```
elahrvivaz: isn't this .get unsafe? like there may not be any keywords set?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
jnh5y: @elahrvivaz getKeywords seems to only be used in places which are also immediately needing a java.util.Set.  In some sense, we'd just push that conversion elsewhere.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -76,7 +76,11 @@ object SimpleFeatureTypes {
     } else {
       Map.empty[String, AnyRef]
     }
-    val opts = userData.map { case (k, v) => new GenericOption(k, v) }.toSeq
+    val opts = userData.map {
+      // Special case to handle adding keywords
+      case (KEYWORDS_KEY, v) => new GenericOption(KEYWORDS_KEY, v.asInstanceOf[java.util.ArrayList[String]].mkString(KEYWORDS_DELIMITER))
+      case (k, v) => new GenericOption(k, v)
+    }.toSeq
```
elahrvivaz: if we're going to do this, then we need to also do the reverse when writing it back out... still not sold on having special handling for the user data though...
question: can we delimit things using the spec string? e.g. normally you can encode user data like https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/Z3IdxStrategyTest.scala#L40

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
atallahhezbor: Yup. Changed to getOrElse

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: ah, possibly... although aren't we joining it with a scala Set in the data store trait?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -76,7 +76,11 @@ object SimpleFeatureTypes {
     } else {
       Map.empty[String, AnyRef]
     }
-    val opts = userData.map { case (k, v) => new GenericOption(k, v) }.toSeq
+    val opts = userData.map {
+      // Special case to handle adding keywords
+      case (KEYWORDS_KEY, v) => new GenericOption(KEYWORDS_KEY, v.asInstanceOf[java.util.ArrayList[String]].mkString(KEYWORDS_DELIMITER))
```
elahrvivaz: it'd be good to handle either a string or an array here.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -8,10 +8,11 @@
 
 package org.locationtech.geomesa.accumulo.data
 
-import java.io.IOException
+import java.io.{File, IOException}
```
jnh5y: Is File actually imported?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -38,6 +39,7 @@ import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import org.opengis.filter.Filter
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
+import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType._
```
jnh5y: Actually, run 'optimize imports'...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -29,6 +30,7 @@ import org.opengis.temporal.Instant
 import scala.reflect.ClassTag
 import scala.util.Try
 
+import scala.collection.JavaConverters._
```
jnh5y: Optimize imports.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -253,5 +259,33 @@ object RichSimpleFeatureType {
 
     /** For additional rendering options see {@link SimpleFeautureTypes#toConfigString} **/
     def toConfigString: String = SimpleFeatureTypes.toConfigString(sft)
+
+    def getKeywords: java.util.Set[String] = {
+      userData[String](KEYWORDS_KEY).map(_.split(KEYWORDS_DELIMITER).toSet.asJava)
+        .getOrElse(java.util.Collections.emptySet[String])
+    }
+
+
```
jnh5y: Extra space.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -253,5 +259,33 @@ object RichSimpleFeatureType {
...
+    def addKeywords(keywordsString: String): Unit = {
+      val currentKeywords = userData[String](KEYWORDS_KEY).map(_.split(KEYWORDS_DELIMITER).toSet).getOrElse(Set())
```
jnh5y: Call to getKeywords.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
atallahhezbor: Oops. Yeah that hung around from when I tried reading example conf files for tests

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/ContentFeatureSourceInfo.scala
```diff
@@ -0,0 +1,23 @@
...
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.utils.geotools
+
+import org.geotools.data.store.ContentFeatureSource
+import org.geotools.data.ResourceInfo
+
+
```
jnh5y: Extra newline.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraFeatureStore.scala
```diff
@@ -23,10 +23,12 @@ import org.geotools.referencing.crs.DefaultGeographicCRS
 import org.joda.time.{DateTime, Interval}
 import org.locationtech.geomesa.filter.FilterHelper
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+import org.locationtech.geomesa.utils.geotools.ContentFeatureSourceInfo
```
jnh5y: Optimize imports.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -24,7 +24,7 @@ import org.opengis.feature.simple.SimpleFeatureType
 import scala.collection.JavaConversions._
 import scala.util.parsing.combinator.JavaTokenParsers
 import scala.util.{Failure, Success, Try}
-
```
jnh5y: Optimize imports.

PR #908, opened 2016-06-06
GEOMESA-1241 Add command line tools for adding/removing keywords

atallahhezbor:

### DIFF
--- a/docs/user/commandline_tools.rst
+++ b/docs/user/commandline_tools.rst
@@ -117,7 +117,7 @@ keywords
 
 Add or remove keywords to a specified schema::
 Repeat the -a or -r flags to add or remove multiple keywords
-The ``-ra``` option removes all keywords
+The ``--removeAll``` option removes all keywords
 The ```-l``` option lists the schema's keywords following all operations
 If there is whitespace within a keyword, enclose it in quotes for proper functionality
 
--- a/docs/user/commandline_tools.rst
+++ b/docs/user/commandline_tools.rst
@@ -30,6 +30,7 @@ Run ``geomesa`` without any arguments to produce the following usage text::
         help                Show help
         ingest              Ingest/convert various file formats into GeoMesa
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
+        keyword             Add/remove keywords on an existing schema
         list                List GeoMesa features for a given catalog
         queryrasterstats    Export queries and statistics about the last X number of queries to a CSV file.
         removeschema        Remove a schema and associated features from a GeoMesa catalog
@@ -110,6 +111,20 @@ Get the specified feature type as a typesafe config::
 Get the specified feature type as an encoded feature schema string::
 
     $ geomesa getsft -u username -p password -c test_catalog -f test_feature --format spec
+
+keyword
+~~~~~~~
+
+Add or remove keywords to a specified schema::
+Repeat the -a or -r flags to add or remove multiple keywords
+The ``-ra``` option removes all keywords
+The ```-l``` option lists the schema's keywords following all operations
+If there is whitespace within a keyword, enclose it in quotes for proper functionality
+
+    $ geomesa keyword -u username -p password \
+        -a keywordB -a keywordC -r keywordA -l\
+        -i instance -z zoo1,zoo2,zoo3 \
+        -c catalog -f featureTypeName
 
 list
 ~~~~
--- a/docs/user/commandline_tools.rst
+++ b/docs/user/commandline_tools.rst
@@ -30,7 +30,7 @@ Run ``geomesa`` without any arguments to produce the following usage text::
         help                Show help
         ingest              Ingest/convert various file formats into GeoMesa
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
-        keyword             Add/remove keywords on an existing schema
+        keywords            Add/remove keywords on an existing schema
         list                List GeoMesa features for a given catalog
         queryrasterstats    Export queries and statistics about the last X number of queries to a CSV file.
         removeschema        Remove a schema and associated features from a GeoMesa catalog
@@ -112,8 +112,8 @@ Get the specified feature type as an encoded feature schema string::
 
     $ geomesa getsft -u username -p password -c test_catalog -f test_feature --format spec
 
-keyword
-~~~~~~~
+keywords
+~~~~~~~~
 
 Add or remove keywords to a specified schema::
 Repeat the -a or -r flags to add or remove multiple keywords
@@ -121,7 +121,7 @@ The ``-ra``` option removes all keywords
 The ```-l``` option lists the schema's keywords following all operations
 If there is whitespace within a keyword, enclose it in quotes for proper functionality
 
-    $ geomesa keyword -u username -p password \
+    $ geomesa keywords -u username -p password \
         -a keywordB -a keywordC -r keywordA -l\
         -i instance -z zoo1,zoo2,zoo3 \
         -c catalog -f featureTypeName
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -181,6 +181,7 @@ Test the command that invokes the GeoMesa Tools:
         help             Show help
         ingest           Ingest a file of various formats into GeoMesa
         ingestraster     Ingest a raster file or raster files in a directory into GeoMesa
+        keyword          Add/remove keywords on an existing schema
         list             List GeoMesa features for a given catalog
         queryrasterstats Export queries and statistics about the last X number of queries to a CSV file.
         removeschema     Remove a schema and associated features from a GeoMesa catalog
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -181,7 +181,7 @@ Test the command that invokes the GeoMesa Tools:
         help             Show help
         ingest           Ingest a file of various formats into GeoMesa
         ingestraster     Ingest a raster file or raster files in a directory into GeoMesa
-        keywords          Add/remove keywords on an existing schema
+        keywords         Add/remove keywords on an existing schema
         list             List GeoMesa features for a given catalog
         queryrasterstats Export queries and statistics about the last X number of queries to a CSV file.
         removeschema     Remove a schema and associated features from a GeoMesa catalog
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -181,7 +181,7 @@ Test the command that invokes the GeoMesa Tools:
         help             Show help
         ingest           Ingest a file of various formats into GeoMesa
         ingestraster     Ingest a raster file or raster files in a directory into GeoMesa
-        keyword          Add/remove keywords on an existing schema
+        keywords          Add/remove keywords on an existing schema
         list             List GeoMesa features for a given catalog
         queryrasterstats Export queries and statistics about the last X number of queries to a CSV file.
         removeschema     Remove a schema and associated features from a GeoMesa catalog
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -223,9 +223,9 @@ class AccumuloDataStore(val connector: Connector,
     val unmodifiableUserdataKeys = Set(SCHEMA_VERSION_KEY, TABLE_SHARING_KEY, SHARING_PREFIX_KEY,
                                    DEFAULT_DATE_KEY, ST_INDEX_SCHEMA_KEY, SimpleFeatureTypes.ENABLED_INDEXES)
 
-    unmodifiableUserdataKeys.foreach { case (key) =>
+    unmodifiableUserdataKeys.foreach { key =>
       if (sft.getUserData.contains(key) && sft.userData[String](key) != previousSft.userData[String](key)) {
-        throw new UnsupportedOperationException("Updating " + key + " is not allowed")
+        throw new UnsupportedOperationException(s"Updating $key is not allowed")
       }
     }
 
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -214,7 +214,11 @@ class AccumuloDataStore(val connector: Connector,
 
     // Prevent modifying wrong type if type names don't match
     if (!schemaTypeName.equals(typeName.toString)) {
-      throw new UnsupportedOperationException("Updating the type name of a schema is not allowed " + schemaTypeName + " " + typeName)
+      throw new IllegalArgumentException(s"Provided type name $typeName does not match schema type name $schemaTypeName")
+    }
+
+    if (previousSft == null) {
+      throw new IllegalArgumentException(s"No schema found for given type name $typeName")
     }
 
     val existingUserData = metadata.read(schemaTypeName, ATTRIBUTES_KEY)
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
@@ -18,6 +18,7 @@ import org.geotools.data.DataAccessFactory.Param
 import org.geotools.data.store.{ContentDataStore, ContentEntry, ContentFeatureSource}
 import org.geotools.data.{DataStore, DataStoreFactorySpi, Query}
 import org.geotools.feature.NameImpl
+import org.opengis.feature.simple.SimpleFeatureType
 import org.locationtech.geomesa.kafka.KafkaDataStore.FeatureSourceFactory
 import org.opengis.feature.`type`.Name
 import org.opengis.filter.Filter
@@ -69,6 +70,14 @@ class KafkaDataStore(override val zookeepers: String,
     super[ContentDataStore].dispose()
     super[KafkaDataStoreSchemaManager].dispose()
   }
+
+  def getKafkaSchema(typeName: String): SimpleFeatureType = getSchema(new NameImpl(typeName))
+
+  override def getSchema(name: Name): SimpleFeatureType = {
+    val typeName = name.getLocalPart
+    getFeatureConfig(typeName).sft
+  }
+
 }
 
 object KafkaDataStoreFactoryParams {
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
@@ -71,13 +71,6 @@ class KafkaDataStore(override val zookeepers: String,
     super[KafkaDataStoreSchemaManager].dispose()
   }
 
-  def getKafkaSchema(typeName: String): SimpleFeatureType = getSchema(new NameImpl(typeName))
-
-  override def getSchema(name: Name): SimpleFeatureType = {
-    val typeName = name.getLocalPart
-    getFeatureConfig(typeName).sft
-  }
-
 }
 
 object KafkaDataStoreFactoryParams {
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
@@ -153,9 +153,9 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
     val unmodifiableUserdataKeys = Set(SCHEMA_VERSION_KEY, TABLE_SHARING_KEY, SHARING_PREFIX_KEY,
       DEFAULT_DATE_KEY, ST_INDEX_SCHEMA_KEY, SimpleFeatureTypes.ENABLED_INDEXES)
 
-    unmodifiableUserdataKeys.foreach { case (key) =>
+    unmodifiableUserdataKeys.foreach { key =>
       if (sft.getUserData.keySet().contains(key) && sft.userData[String](key) != previousSft.userData[String](key)) {
-        throw new UnsupportedOperationException("Updating " + key + " is not allowed")
+        throw new UnsupportedOperationException(s"Updating $key is not allowed")
       }
     }
 
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
@@ -20,6 +20,7 @@ import org.locationtech.geomesa.kafka.common.ZkUtils
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.opengis.feature.`type`.Name
 import org.opengis.feature.simple.SimpleFeatureType
+import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType._
 
 import scala.collection.JavaConverters._
 import scala.util.Try
@@ -135,6 +136,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
         zkUtils.deleteTopic(fc.topic)
       }
     }
+  }
+
+  def updateKafkaSchema(typeName: String, sft: SimpleFeatureType) = {
+
+    // Get previous schema and user data
+    val previousSft = getSchema(typeName)
+    val schemaTypeName = sft.getTypeName()
+
+    // Prevent modifying wrong type if type names don't match
+    if (!schemaTypeName.equals(typeName.toString)) {
+      throw new UnsupportedOperationException("Updating the type name of a schema is not allowed " + schemaTypeName + " " + typeName)
+    }
+
+    // Check that unmodifiable user data has not changed
+    val unmodifiableUserdataKeys = Set(SCHEMA_VERSION_KEY, TABLE_SHARING_KEY, SHARING_PREFIX_KEY,
+      DEFAULT_DATE_KEY, ST_INDEX_SCHEMA_KEY, SimpleFeatureTypes.ENABLED_INDEXES)
+
+    unmodifiableUserdataKeys.foreach { case (key) =>
+      if (sft.getUserData.keySet().contains(key) && sft.userData[String](key) != previousSft.userData[String](key)) {
+        throw new UnsupportedOperationException("Updating " + key + " is not allowed")
+      }
+    }
+
+    // Check that the rest of the schema has not changed (columns, types, etc)
+    val previousColumns = previousSft.getAttributeDescriptors
+    val currentColumns = sft.getAttributeDescriptors
+    if (previousColumns != currentColumns) {
+      throw new UnsupportedOperationException("Updating schema columns is not allowed")
+    }
+
+    // If all is well, write to zookeeper metadata
+    zkUtils.zkClient.writeData(getSchemaPath(typeName), SimpleFeatureTypes.encodeType(sft, includeUserData = true))
   }
 
   override def dispose(): Unit = {
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
@@ -146,7 +146,11 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
 
     // Prevent modifying wrong type if type names don't match
     if (!schemaTypeName.equals(typeName.toString)) {
-      throw new UnsupportedOperationException("Updating the type name of a schema is not allowed " + schemaTypeName + " " + typeName)
+      throw new IllegalArgumentException(s"Provided type name $typeName does not match schema type name $schemaTypeName")
+    }
+
+    if (previousSft == null) {
+      throw new IllegalArgumentException(s"No schema found for given type name $typeName")
     }
 
     // Check that unmodifiable user data has not changed
--- a/geomesa-tools/README.md
+++ b/geomesa-tools/README.md
@@ -48,7 +48,7 @@ This should print out the following usage text:
         help                Show help
         ingest              Ingest/convert various file formats into GeoMesa
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
-        keyword             Add/remove/list keywords on a given schema
+        keywords             Add/remove/list keywords on a given schema
         list                List GeoMesa features for a given catalog
         queryrasterstats    Export queries and statistics about the last X number of queries to a CSV file.
         removeschema        Remove a schema and associated features from a GeoMesa catalog
@@ -801,13 +801,13 @@ To list the features on a specified catalog table, use the `list` command.
 #### Example command:
     geomesa list -u username -p password -c test_catalog
     
-### keyword
-To add, remove, or list all the keywords on a specified catalog table, use the `keyword` command
+### keywords
+To add, remove, or list all the keywords on a specified catalog table, use the `keywords` command
 
 #### Usage (required options denoted with star):
-    $geomesa help keyword
+    $geomesa help keywords
     Add/Remove/List keywords on an existing schema
-    Usage: keyword [options]
+    Usage: keywords [options]
       Options:
         -a, --add
            A keyword to add. Can be specified multiple times
@@ -841,7 +841,7 @@ To add, remove, or list all the keywords on a specified catalog table, use the `
 
     
 #### Example command:
-    geomesa keyword -u username -p password -c test_catalog -f feature_name -i instance \
+    geomesa keywords -u username -p password -c test_catalog -f feature_name -i instance \
     -a keywordA -a keywordB=foo,bar -r keywordC -l
 
 
--- a/geomesa-tools/README.md
+++ b/geomesa-tools/README.md
@@ -48,7 +48,7 @@ This should print out the following usage text:
         help                Show help
         ingest              Ingest/convert various file formats into GeoMesa
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
-        keywords             Add/remove/list keywords on a given schema
+        keywords            Add/remove/list keywords on a given schema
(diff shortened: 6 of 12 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,93 @@
...
+  override val command: String = "keyword"
+  override val params = new KeywordParameters()
+
+  override def execute(): Unit = {
+    val dsParams = new DataStoreHelper(params).paramMap
+    val tryDs = DataStoreFinder.getDataStore(dsParams)
+    if (tryDs == null) {
+      throw new ParameterException("Could not load a data store with the provided parameters")
+    }
+    tryDs.dispose()
```
elahrvivaz: you don't want to call dispose here if you're still using the data store (e.g. on the next line)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,93 @@
...
+      sft.addKeywords(params.keywordsToAdd.mkString(KEYWORDS_DELIMITER))
+    }
+
+    if (params.keywordsToRemove != null) {
+      sft.removeKeywords(params.keywordsToRemove.mkString(KEYWORDS_DELIMITER))
+    }
+
+    if (params.removeAll) {
+      val confirm = StdIn.readLine("Remove all keywords? (y/n): ")
+      if (confirm.equals("y")) {
```
elahrvivaz: for usability, i would allow 'y' or 'yes' in any case (e.g. do a .toLowerCase on the readLine)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+  }
+
+
+}
+
+// Overrides JCommander's split on comma
+class KeywordParameterSplitter extends IParameterSplitter {
+  override def split(s : String): java.util.List[String] = s.split("\u0000").toList
```
elahrvivaz: is it possible to input this character on the command line? jcommander allows you to just specify the argument more than once (like in your example), so i would say don't do any splitting on single strings at all.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,93 @@
...
+    @Parameter(names = Array("-a", "--add"), description = "A keyword to add. Can be specified multiple times", splitter = classOf[KeywordParameterSplitter])
+    var keywordsToAdd: java.util.List[String] = null
+
+    @Parameter(names = Array("-r", "--remove"), description = "A keyword to remove. Can be specified multiple times", splitter = classOf[KeywordParameterSplitter])
+    var keywordsToRemove: java.util.List[String] = null
+
+    @Parameter(names = Array("-l", "--list"), description = "List all keywords on the schema")
+    var list: Boolean = false
+
+    @Parameter(names = Array("-ra", "--removeAll"), description = "Remove all keywords on the schema")
```
elahrvivaz: we try to keep the single dash arguments to a single character to conform with the gnu(?) spec. maybe make it -R or just don't have a single dash argument.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Overrides JCommander's split on comma
+class KeywordParameterSplitter extends IParameterSplitter {
+  override def split(s : String): java.util.List[String] = s.split("\u0000").toList
+}
+
+object KeywordCommand {
+
+  @Parameters(commandDescription = "Add/Remove keywords on an existing schema")
```
elahrvivaz: I would add 'List' to the description - e.g. 'add/remove/list'

elahrvivaz:
@atallahhezbor looks good. Could we move it into the 'common' package and allow it to be called for kafka as well?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
jahhulbert-ccri: +1

elahrvivaz:
One other thing, it would be good to update the tools/README.md (in addition to the docs which you've already done). You might also need to update the docs/developer/tools to match the readme.

docs/user/commandline_tools.rst
```diff
@@ -30,6 +30,7 @@ Run ``geomesa`` without any arguments to produce the following usage text::
         help                Show help
         ingest              Ingest/convert various file formats into GeoMesa
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
+        keyword             Add/remove keywords on an existing schema
```
jahhulbert-ccri: just wondering if ```keywords``` with an ```s``` would be better. doesn't matter much.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    if (params.removeAll) {
+      val confirm = StdIn.readLine("Remove all keywords? (y/n): ")
+      if (confirm.equals("y")) {
+        sft.removeAllKeywords()
+      }
+    }
+
+    tryDs.updateSchema(params.featureName, sft)
```
jahhulbert-ccri: if the user does remove all and then cancels from the prompt do we still want to execute this?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
atallahhezbor: I've looked into this and it may be possible depending on your character set and shell, but I haven't been able to get it working in bash. [See here](http://stackoverflow.com/questions/7572801/type-null-character-in-terminal)
 In either case, we would need this class to override split such that the keywords can contain commas without being split automatically by jcommander. I could just return [s]

elahrvivaz: ok, just returning `s` would be good then.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
atallahhezbor: My thinking here was that multiple actions could be executed in one command. For example, adding a few keywords and removing all, but then realizing that they didn't want to remove all? It's kind of an odd case so I can see aborting everything being reasonable functionality.

elahrvivaz: the way it's ordered, it would end up removing the keywords you just added...

atallahhezbor:
@elahrvivaz I don't see why not. I'll move it to common and make sure all is still well.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
atallahhezbor: Yes, but if they answer "no" then it won't. Again, it's an odd case, so it's hard to tell what the desired functionality would be.
Thinking about it more, I can see answering no as a change in whatever the user was trying to do, so in that sense I agree, the whole operation should stop

atallahhezbor:
@elahrvivaz An issue with moving it to common is that Kafka requires zookeeper parameters, whereas Accumulo has them as optional This leads to conflicting parameters when I extend both. I'm looking for a clean way to override it but something like 

        override var zookeepers: String = super[GeoMesaConnectionParams].zookeepers

is not possible unless zookeepers is a method.

elahrvivaz:
hmm, I guess so far we have just duplicated common commands... if you figure out something to let us re-use code, that would be nice, but otherwise just copy into both places.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
```diff
@@ -69,6 +70,14 @@ class KafkaDataStore(override val zookeepers: String,
     super[ContentDataStore].dispose()
     super[KafkaDataStoreSchemaManager].dispose()
   }
+
+  def getKafkaSchema(typeName: String): SimpleFeatureType = getSchema(new NameImpl(typeName))
```
elahrvivaz: Looking in ContentDataStore, it looks like you don't want this method or to override getSchema. Instead, you want to modify the schema coming back from the feature source in `createFeatureSource`, which in this case would mean changing the cache-loader on line 48.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,105 @@
...
+        return
+      }
+    }
+
+    // Attempt to replace the old schema with the updated one
+    if (keywordsModified) {
+      Try {
+        ds.removeSchema(sft.getTypeName)
+      } match {
+        case Failure(ex) => println(ex.toString)
```
elahrvivaz: if this fails we probably don't want to continue - maybe just remove the try/catch and let the exception bubble up

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,105 @@
...
+      }
+      ds.asInstanceOf[KafkaDataStoreSchemaManager].createSchema(sft)
+    }
+
+    ds.dispose()
```
elahrvivaz: need to move this down to after printing the keywords

docs/user/commandline_tools.rst
elahrvivaz: +1

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
atallahhezbor: I decided to continue because I would run into a "NodeExists" exception even though the schema is deleted correctly. I noticed this exception also happens with the RemoveSchema command [here](https://github.com/locationtech/geomesa/blob/77a676462b99f060b178aa27e055709a455c3c25/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/RemoveSchemaCommand.scala#L54) 
Though I should at least match that specific exception

jahhulbert-ccri: can you send the stacktrace?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
atallahhezbor: My thinking was if the user adds keywords and also lists them in the same command, then they should see the resulting set as opposed to the previous set

elahrvivaz: that's fine - just you're not supposed to use the datastore after calling dispose on it. And you use it again to reload the sft. currently our dispose method doesn't do anything, but just in case it changes in the future...

atallahhezbor: Oh I thought you were referring to the whole block. Gotcha

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
atallahhezbor: org.I0Itec.zkclient.exception.ZkNodeExistsException: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /admin/delete_topics/geomesa-ds-kafka-example-json
	at org.I0Itec.zkclient.exception.ZkException.create(ZkException.java:55)
	at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:685)
	at org.I0Itec.zkclient.ZkClient.create(ZkClient.java:304)
	at org.I0Itec.zkclient.ZkClient.createPersistent(ZkClient.java:243)
	at kafka.utils.ZkUtils$.createPersistentPath(ZkUtils.scala:306)
	at kafka.admin.AdminUtils$.deleteTopic(AdminUtils.scala:159)
	at org.locationtech.geomesa.kafka08.ZkUtils08.deleteTopic(ZkUtils08.scala:23)
	at org.locationtech.geomesa.kafka.KafkaDataStoreSchemaManager$$anonfun$removeSchema$1.apply(KafkaDataStoreSchemaManager.scala:135)
	at org.locationtech.geomesa.kafka.KafkaDataStoreSchemaManager$$anonfun$removeSchema$1.apply(KafkaDataStoreSchemaManager.scala:133)
	at org.locationtech.geomesa.kafka.KafkaDataStoreSchemaManager$class.removeSchema(KafkaDataStoreSchemaManager.scala:133)
	at org.locationtech.geomesa.kafka.KafkaDataStore.removeSchema(KafkaDataStore.scala:28)
	at org.locationtech.geomesa.tools.kafka.commands.KeywordCommand.execute(KeywordCommand.scala:60)
	at org.locationtech.geomesa.tools.accumulo.commands.KeywordCommandTest$$anonfun$1$$anonfun$apply$3.apply(KeywordCommandTest.scala:84)
	at org.locationtech.geomesa.tools.accumulo.commands.KeywordCommandTest$$anonfun$1$$anonfun$apply$3.apply(KeywordCommandTest.scala:75)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:283)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:173)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:153)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:128)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:203)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:155)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
Caused by: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /admin/delete_topics/geomesa-ds-kafka-example-json
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:119)
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:51)
	at org.apache.zookeeper.ZooKeeper.create(ZooKeeper.java:783)
	at org.I0Itec.zkclient.ZkConnection.create(ZkConnection.java:87)
	at org.I0Itec.zkclient.ZkClient$1.call(ZkClient.java:308)
	at org.I0Itec.zkclient.ZkClient$1.call(ZkClient.java:304)
	at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:675)
	at org.I0Itec.zkclient.ZkClient.create(ZkClient.java:304)
	at org.I0Itec.zkclient.ZkClient.createPersistent(ZkClient.java:243)
	at kafka.utils.ZkUtils$.createPersistentPath(ZkUtils.scala:306)
	at kafka.admin.AdminUtils$.deleteTopic(AdminUtils.scala:159)
	at org.locationtech.geomesa.kafka08.ZkUtils08.deleteTopic(ZkUtils08.scala:23)
	at org.locationtech.geomesa.kafka.KafkaDataStoreSchemaManager$$anonfun$removeSchema$1.apply(KafkaDataStoreSchemaManager.scala:135)
	at org.locationtech.geomesa.kafka.KafkaDataStoreSchemaManager$$anonfun$removeSchema$1.apply(KafkaDataStoreSchemaManager.scala:133)
	at org.locationtech.geomesa.kafka.KafkaDataStoreSchemaManager$class.removeSchema(KafkaDataStoreSchemaManager.scala:133)
	at org.locationtech.geomesa.kafka.KafkaDataStore.removeSchema(KafkaDataStore.scala:28)
	at org.locationtech.geomesa.tools.kafka.commands.KeywordCommand.execute(KeywordCommand.scala:60)
	at org.locationtech.geomesa.tools.accumulo.commands.KeywordCommandTest$$anonfun$1$$anonfun$apply$3.apply(KeywordCommandTest.scala:84)
	at org.locationtech.geomesa.tools.accumulo.commands.KeywordCommandTest$$anonfun$1$$anonfun$apply$3.apply(KeywordCommandTest.scala:75)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.geotools.data.DataStoreFinder
+import org.locationtech.geomesa.tools.accumulo.commands.KeywordCommand.KeywordParameters
+import org.locationtech.geomesa.tools.accumulo.{DataStoreHelper, GeoMesaConnectionParams}
+import org.locationtech.geomesa.tools.common.FeatureTypeNameParam
+import org.locationtech.geomesa.tools.common.commands.Command
+
+import scala.collection.JavaConversions._
+import scala.io.StdIn
+
+class KeywordCommand(parent: JCommander) extends Command(parent) {
```
Afterglow375: Make this command extend CommandWithCatalog for consistency with the other commands which use an AccumuloDataStore.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,92 @@
...
+        sft.removeAllKeywords()
+      } else {
+        println("Aborting operation")
+        tryDs.dispose()
+        return
+      }
+    }
+
+    tryDs.updateSchema(params.featureName, sft)
+    tryDs.dispose()
```
Afterglow375: I believe this call is unnecessary, but doesn't hurt.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
atallahhezbor: Ah I see. Well would it be alright to use `getFeatureSource.getSchema` if I'm going that route?

elahrvivaz: you shouldn't need this method at all - there is already a getSchema(String) method in ContentDataStore that you can call - it delegates to getFeatureSource.getSchema, but no point in re-implementing it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    val sft = tryDs.getSchema(params.featureName)
+
+    if (params.keywordsToAdd != null) {
+      sft.addKeywords(params.keywordsToAdd.mkString(KEYWORDS_DELIMITER))
+    }
+
+    if (params.keywordsToRemove != null) {
+      sft.removeKeywords(params.keywordsToRemove.mkString(KEYWORDS_DELIMITER))
```
Afterglow375: I looked at the implementation of removeKeywords and it looks like it's possible to remove all the keywords but still set the key/value in the userData (where the value will be an empty string). Don't know if that may have unforeseen consequences or not.

atallahhezbor: That actually is an issue that I accounted for with removeAllKeywords but forgot about the case when all are manually deleted. 
When the retrieved keywords is an empty string, (which i'm guessing is somehow encoded with a null byte) an additional empty keyword is retrieved.
We could handle this by deleting the key value pair when removing the last keyword. Or maybe switching to another un-used delimiter.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
Afterglow375: Calling removeSchema on a KDS is risky business, as its behavior can be different depending on how Kafka is configured. In this case, what you're seeing is that the Kafka topic deletion setting for kafka isn't enabled (which is the default). If that setting is disabled, all that will happen when calling removeSchema is the Kafka topic will be "marked" for deletion (i.e. there'll be a new zk node at /admin/delete_topics/<topic_name> created indicating to Kafka that the topic is marked for deletion). On Kafka restart, the topic will be deleted, but not during actual runtime. This is something that hasn't really been addressed with the KDS... 

Either way, I don't think removeSchema is what you want to use. removeSchema on a KDS will delete the schema AND all the data associated with it. In your corresponding keywords command for Accumulo, you only call updateSchema, which doesn't remove any data. It'd probably be best for you to implement an updateSchema method for the KDS similiar to the one for the Accumulo data store (if you do, should go in KafkaDataStoreSchemaManager).

atallahhezbor: Ah okay. I was under the impression that the schema was separate and safe to delete and recreate.

My new approach is to use `zkUtils.zkClient.writeData` to update the schema

docs/user/commandline_tools.rst
```diff
@@ -111,6 +112,20 @@ Get the specified feature type as an encoded feature schema string::
 
     $ geomesa getsft -u username -p password -c test_catalog -f test_feature --format spec
 
+keywords
+~~~~~~~~
+
+Add or remove keywords to a specified schema::
+Repeat the -a or -r flags to add or remove multiple keywords
+The ``-ra``` option removes all keywords
```
elahrvivaz: we removed the short version of the flag - need to update the docs here to --removeAll

docs/user/installation_and_configuration.rst
```diff
@@ -181,6 +181,7 @@ Test the command that invokes the GeoMesa Tools:
         help             Show help
         ingest           Ingest a file of various formats into GeoMesa
         ingestraster     Ingest a raster file or raster files in a directory into GeoMesa
+        keywords          Add/remove keywords on an existing schema
```
elahrvivaz: [formatting] extra space

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -137,6 +138,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
...
+    // Prevent modifying wrong type if type names don't match
+    if (!schemaTypeName.equals(typeName.toString)) {
+      throw new UnsupportedOperationException("Updating the type name of a schema is not allowed " + schemaTypeName + " " + typeName)
+    }
+
+    // Check that unmodifiable user data has not changed
+    val unmodifiableUserdataKeys = Set(SCHEMA_VERSION_KEY, TABLE_SHARING_KEY, SHARING_PREFIX_KEY,
+      DEFAULT_DATE_KEY, ST_INDEX_SCHEMA_KEY, SimpleFeatureTypes.ENABLED_INDEXES)
+
+    unmodifiableUserdataKeys.foreach { case (key) =>
```
elahrvivaz: [scala] for single values, you can just put `.foreach { key => }`, you only need the case if you're unapplying something (e.g. `case (tuple1, tuple2)`, `case CaseClass(valOne, valTwo)` etc)

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -137,6 +138,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
...
+      if (sft.getUserData.keySet().contains(key) && sft.userData[String](key) != previousSft.userData[String](key)) {
+        throw new UnsupportedOperationException("Updating " + key + " is not allowed")
```
elahrvivaz: [scala] in general prefer using string interpolation: `s"Updating $key is not allowed"`

geomesa-tools/README.md
```diff
@@ -48,6 +48,7 @@ This should print out the following usage text:
         help                Show help
         ingest              Ingest/convert various file formats into GeoMesa
         ingestraster        Ingest a raster file or raster files in a directory into GeoMesa
+        keywords             Add/remove/list keywords on a given schema
```
elahrvivaz: [formatting] extra space

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -137,6 +138,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
...
 
+  def updateKafkaSchema(typeName: String, sft: SimpleFeatureType) = {
+
+    // Get previous schema and user data
+    val previousSft = getSchema(typeName)
+    val schemaTypeName = sft.getTypeName()
+
+    // Prevent modifying wrong type if type names don't match
+    if (!schemaTypeName.equals(typeName.toString)) {
+      throw new UnsupportedOperationException("Updating the type name of a schema is not allowed " + schemaTypeName + " " + typeName)
```
Afterglow375: I'm not sure this is the proper Exception to throw. This gets thrown when the user passes in incorrect args, so maybe an IllegalArgumentException? If the typeName and sft.getTypeName that the user passes in are different, it's more likely the user made an error rather than them trying to update the type name of a schema.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -137,6 +138,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
...
+    }
+
+    // Check that unmodifiable user data has not changed
+    val unmodifiableUserdataKeys = Set(SCHEMA_VERSION_KEY, TABLE_SHARING_KEY, SHARING_PREFIX_KEY,
+      DEFAULT_DATE_KEY, ST_INDEX_SCHEMA_KEY, SimpleFeatureTypes.ENABLED_INDEXES)
+
+    unmodifiableUserdataKeys.foreach { key =>
+      if (sft.getUserData.keySet().contains(key) && sft.userData[String](key) != previousSft.userData[String](key)) {
```
Afterglow375: I may be wrong, but isn't it possible to hit a NullPointerException where previousSft may be null? If the schema doesn't exist in zookeeper, it'll be null.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -137,6 +138,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
...
+
+    // Check that the rest of the schema has not changed (columns, types, etc)
+    val previousColumns = previousSft.getAttributeDescriptors
+    val currentColumns = sft.getAttributeDescriptors
+    if (previousColumns != currentColumns) {
+      throw new UnsupportedOperationException("Updating schema columns is not allowed")
+    }
+
+    // If all is well, write to zookeeper metadata
+    zkUtils.zkClient.writeData(getSchemaPath(typeName), SimpleFeatureTypes.encodeType(sft, includeUserData = true))
```
Afterglow375: Have you tested this? Does this overwrite the existing data properly?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
```diff
@@ -137,6 +138,38 @@ trait KafkaDataStoreSchemaManager extends DataStore with LazyLogging {
     }
   }
 
+  def updateKafkaSchema(typeName: String, sft: SimpleFeatureType) = {
```
Afterglow375: Maybe create some unit tests? I do know that the unit tests for Kafka are buggy and are thus turned off by default, so maybe not. @jnh5y @elahrvivaz Are unit tests worth doing?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
atallahhezbor: Good catch. I'll add a check for that.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
atallahhezbor: Yup and it seemed to work fine. I have a test file but I didn't push it since it connects to a real instance

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/KeywordCommand.scala
Afterglow375: Yeah, the if statement approach makes sense to me.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,99 @@
...
+class KeywordCommand(parent: JCommander) extends CommandWithKDS(parent) {
+
+  import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType._
+  override val command: String = "keywords"
+  override val params = new KeywordParameters()
+
+  override def execute(): Unit = {
+
+    if (ds == null) {
+      throw new ParameterException("Could not load a data store with the provided parameters")
```
Afterglow375: Unnecessary if statement. The `new DataStoreHelper(params).getDataStore` call checks the parameters.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/kafka/commands/KeywordCommand.scala
```diff
@@ -0,0 +1,99 @@
...
+      val reloadedSft = ds.getSchema(params.featureName)
+      println("Keywords: " + reloadedSft.getKeywords.toString)
+    }
+
+    ds.dispose()
+  }
+}
+
+// Overrides JCommander's split on comma to allow single keywords containing commas
+class KeywordParameterSplitter extends IParameterSplitter {
```
Afterglow375: This should go somewhere into the `common` module (which the accumulo and kafka modules can use).

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStoreSchemaManager.scala
atallahhezbor: IIRC the issue we ran into with unit tests for this functionality is that we couldn't figure out how to get the keyword command to connect to the same mock data store as the unit test. Using the parameters, the keyword command would just spin up a new mock datastore without the keywords it needed to modify.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -214,7 +214,11 @@ class AccumuloDataStore(val connector: Connector,
 
     // Prevent modifying wrong type if type names don't match
     if (!schemaTypeName.equals(typeName.toString)) {
-      throw new UnsupportedOperationException("Updating the type name of a schema is not allowed " + schemaTypeName + " " + typeName)
+      throw new IllegalArgumentException(s"Provided type name $typeName does not match schema type name $schemaTypeName")
```
elahrvivaz: not a big deal, and I know @Afterglow375 just said to change it, but i think UnsupportedOperationException is actually correct here - theoretically the geotools API we implement would allow you to do this, but we don't have it implemented.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaDataStore.scala
```diff
@@ -18,6 +18,7 @@ import org.geotools.data.DataAccessFactory.Param
 import org.geotools.data.store.{ContentDataStore, ContentEntry, ContentFeatureSource}
 import org.geotools.data.{DataStore, DataStoreFactorySpi, Query}
 import org.geotools.feature.NameImpl
+import org.opengis.feature.simple.SimpleFeatureType
```
elahrvivaz: [minor] presumably left over imports that should be removed? Also can remove the extra space added below just to keep the history of the file cleaner.

elahrvivaz:
i think everything has been addressed - merging.

docs/user/commandline_tools.rst
```diff
@@ -111,6 +112,20 @@ Get the specified feature type as an encoded feature schema string::
...
     $ geomesa getsft -u username -p password -c test_catalog -f test_feature --format spec
 
+keywords
+~~~~~~~~
+
+Add or remove keywords to a specified schema::
+Repeat the -a or -r flags to add or remove multiple keywords
+The ``--removeAll``` option removes all keywords
+The ```-l``` option lists the schema's keywords following all operations
+If there is whitespace within a keyword, enclose it in quotes for proper functionality
```
mdzimmerman: The double colon needs to be at the end of this line for Sphinx to properly render the code block below.

elahrvivaz: i'll fix in the merge

elahrvivaz:
c72abe861a831c5d7eb9264a03f2733bb834ea24

PR #918, opened 2016-06-13
GEOMESA-1235 Fixing key and counts for histogram ascii map

elahrvivaz:


jnh5y:
Is it possible to write a quick unit test for this issue?  If not, I'm good to merge it.

elahrvivaz:
I could, but it would be comparing ascii values in a string so it would probably be very brittle... what do you think?

PR #928, opened 2016-06-16
GEOMESA-1245 Using BindingFilterVisitor to improve filter evaluation times

elahrvivaz:
We might want to hold off on merging this for 1.2.3 - it changes the CQL, so the ECQL.toFilter might break in some cases. It didn't happen in any of our unit tests, but might be good to have it in a snapshot for a while before being released.


jahhulbert-ccri:
+1 lets hold off. @elahrvivaz remind me and I can try to do some testing later this week

jnh5y:
@jahhulbert-ccri, have you had a chance to test this?

jahhulbert-ccri:
looks good i'm going to test it out...looks like just query correctness testing

elahrvivaz:
originally I put the binding method later on in the pipeline, after we have manipulated the filter, and that caused some queries to fail - I believe it was related to our IDL splitting/handling. So there are at least some scenarios where it seems to change the filter.

cne1x:
@elahrvivaz I believe you are correct about the filter transformation via anti-meridian handling.  If the geometry spans that 0-degrees margin, we translate it into an OR of the two decomposed sub-geometries.

jahhulbert-ccri:
this looks good to me. similar usage in geoserver... @cne1x you cool with it?

PR #920, opened 2016-06-14
GEOMESA-1237 Compacting stats table after deletes to ensure combiners…

elahrvivaz:
… run correctly

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #929, opened 2016-06-17
GEOMESA-1251,GEOMESA-1250 Adding top-k stat, stats for non-indexed attributes

elahrvivaz:
~~note: stacks on top of #920, please only consider last commit(s)~~

Stats currently tracked:
1. indexed attributes - bounds, frequency, histograms, top-k
2. default date, geometry - bounds, histogram
3. attributes marked as 'keep-stats:true' - bounds, top-k

Note that there may be some overlap between the categories


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
```diff
@@ -45,6 +46,16 @@ class GeoMesaMetadataStats(val ds: AccumuloDataStore, statsTable: String)
 
   private val metadata = new AccumuloBackedMetadata(ds.connector, statsTable, new StatsMetadataSerializer(ds))
 
+  private val compactionScheduled = new AtomicBoolean(false)
+  private val lastCompaction = new AtomicLong(0L)
+
+  executor.scheduleWithFixedDelay(new Runnable() {
+    override def run(): Unit = if (compactionScheduled.compareAndSet(true, false) &&
```
cne1x: Do we want to reverse the comparisons in this test?  Don't we want to check if the last compaction is at least an hour old before risking updating ```compactionScheduled```?  Or am I just misunderstanding this test?

elahrvivaz: good point, I'll fix in the other PR.

PR #934, opened 2016-06-19
WIP GEOMESA-1253 Per-attribute level visibility

elahrvivaz:
Known things that aren't supported:
- Join/partial attribute indices
- Pre-calculated BIN values
- Density/heatmap weighting

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/AttributeTable.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/AttributeTable.scala
@@ -26,6 +26,7 @@ import org.locationtech.geomesa.accumulo.data._
 import org.locationtech.geomesa.utils.geotools.RichAttributeDescriptors.RichAttributeDescriptor
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.locationtech.geomesa.utils.index.VisibilityLevel
 import org.locationtech.geomesa.utils.stats.IndexCoverage
 import org.opengis.feature.`type`.AttributeDescriptor
 import org.opengis.feature.simple.SimpleFeatureType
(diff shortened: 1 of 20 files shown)


elahrvivaz:
Didn't see a good spot to put docs, so putting them here:

Per-attribute visibility can be enabled when creating your simple feature type by setting the appropriate user data value:
```
sft.getUserData().put("geomesa.visibility.level", "attribute");
dataStore.createSchema(sft);
```
When writing each feature, the per-attribute visibilities must be set in a comma-delimited string in the user data. Each attribute must have a corresponding value in the delimited string, otherwise an error will be thrown.

Example for a simple feature type with four attributes:
```
SecurityUtils.setFeatureVisibility(feature, "admin,user,admin,user")
```
or
```
feature.getUserData().put("geomesa.feature.visibility", "admin,user,admin,user");
```

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/AttributeTable.scala
```diff
@@ -48,36 +49,86 @@ object AttributeTable extends GeoMesaTable with LazyLogging {
...
   private type TryEncoder = Try[(TypeEncoder[Any, String], TypeEncoder[_, String])]
 
-  override def supports(sft: SimpleFeatureType) =
-    sft.getSchemaVersion > 5 && sft.getAttributeDescriptors.exists(_.isIndexed)
+  override def supports(sft: SimpleFeatureType) = {
+    val ok = sft.getSchemaVersion > 5 && sft.getAttributeDescriptors.exists(_.isIndexed)
+    if (ok && sft.getVisibilityLevel == VisibilityLevel.Attribute &&
+        !sft.getAttributeDescriptors.filter(_.isIndexed).forall(_.getIndexCoverage() == IndexCoverage.FULL)) {
+      // TODO GEOMESA-1254 support index values
+      throw new IllegalArgumentException("Attribute level visibility is currently only supported for fully" +
```
jahhulbert-ccri: should we return false instead or is this more of an illegal state/request of the usre

elahrvivaz: I figure it's better to fail fast, rather than silently suppress indexing of attributes (which returning false would do)

jahhulbert-ccri: seems reasonable

elahrvivaz:
moved to #937

PR #935, opened 2016-06-20
GEOMESA-1238 Docs reorganization

mdzimmerman:
* Move text from many READMEs into main docs; leave stubs that point to
  RST that should properly redirect on GitHub.
* Create separate chapters for each data store
* Add general data stores page (for docs common to all or most GM stores)
* Trim down developer manual, remove modules
* Add modules to user manual
  * geomesa-blobstore
  * geomesa-convert
  * geomesa-jobs
  * geomesa-native-api
  * geomesa-stream
  * geomesa-web-data
* Added installation docs for all data stores to
  installation_and_configuration.rst
* Added info on using all datastores to geoserver.rst

I have not moved all of the READMEs into the manuals; specifically the
ones that appear to be old/unsupported, and ones that are bundled in the
binary distribution (tools, gs-plugins).


mdzimmerman:
Test build here: http://geomesa.org/documentation/test/

PR #936, opened 2016-06-21
Fixes describe command keyword output

atallahhezbor:


geomesa-tools/bin/download-data.sh
```diff
@@ -0,0 +1,64 @@
+#! /usr/bin/env bash
+
+type=$1
```
jahhulbert-ccri: what happens if ```$1``` is empty

geomesa-tools/bin/download-data.sh
```diff
@@ -0,0 +1,64 @@
...
+  wget "http://research.microsoft.com/pubs/152883/User_guide_T-drive.pdf" -P $GEOMESA_HOME/data/tdrive
+
+else
+
+  if [ -n "$type" ]; then
+    PREFIX="Unknown data type '$type'."
+  else
+    PREFIX="Please enter a data type."
+  fi
+  echo "$PREFIX Available types: gdelt, geolife, osm-gpx, tdrive"
```
jahhulbert-ccri: prefer this syntax in bash for var substitution when possible ```"${PREFIX}"```

geomesa-tools/bin/download-data.sh
elahrvivaz: it's handled in the else on line 60

geomesa-tools/bin/geomesa
```diff
@@ -146,10 +146,15 @@ function findJars() {
...
 
+function findSftDirs() {
+    DIRS=$(find $GEOMESA_CONF_DIR/sfts -mindepth 1 -type d -exec printf {}: \;)
+    echo ${DIRS%:}
+}
 # Start constructing GEOMESA_CP (classpath)
 # include geomesa first so that the correct log4j.properties is picked up
 GEOMESA_CP="${GEOMESA_CONF_DIR}:$(findJars $GEOMESA_LIB/common)"
 GEOMESA_CP="${GEOMESA_CP}:${ACCUMULO_CONF_DIR}:${HADOOP_CONF_DIR}"
+GEOMESA_CP="${GEOMESA_CP}:${GEOMESA_CONF_DIR}:$(findSftDirs)"
```
jahhulbert-ccri: does this work OK with no dirs?

geomesa-tools/conf/sfts/geonames/reference.conf
```diff
@@ -0,0 +1,64 @@
+geomesa =
```
jahhulbert-ccri: picky - move { up a line multiple times in this file to match the others

jahhulbert-ccri:
i'd say the only thing to think about is if we ever wanted to reserve ```$GEOMESA_HOME/data``` for something else. I can't think of what it would be so that's probl moot.

geomesa-tools/bin/geomesa
atallahhezbor: Yup. They simple don't get added to the classpath and don't show up in ```geomesa env``` or ```geomesa classpath```.

PR #937, opened 2016-06-21
GEOMESA-1253 Initial support for attribute level visibility

elahrvivaz:
* Enable in the sft with the user data: geomesa.visibility.level='attribute'
* Put comma-delimited visibilities in the simple feature user data corresponding to each attribute
* WARNING: Experimental, does not support all query types and compatibility is not gauranteed going forward


elahrvivaz:
replaces #934

PR #946, opened 2016-06-22
GEOMESA-1266 Fix libjars, prioritize guava 17 for mapred

jahhulbert-ccri:

### DIFF
--- a/geomesa-tools/examples/ingest/csv/example.conf
+++ b/geomesa-tools/examples/ingest/csv/example.conf
@@ -2,7 +2,7 @@ geomesa {
   sfts {
     "example-csv" = {
       attributes = [
-        { name = "id",       type = "Integer",         index = false                             }
+        { name = "fid",      type = "Integer",         index = false                             }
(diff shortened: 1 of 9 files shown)


geomesa-tools/examples/ingest/csv/example.conf
```diff
@@ -2,7 +2,7 @@ geomesa {
   sfts {
     "example-csv" = {
       attributes = [
-        { name = "id",       type = "Integer",         index = false                             }
+        { name = "fid",      type = "Integer",         index = false                             }
```
jnh5y: I'm curious about the change between id and fid...  Can you comment on that?

elahrvivaz:
thanks for fixing that @jahhulbert-ccri. Odd that guava would be required, that's only a test-scope dependency for stream-lib.

geomesa-tools/examples/ingest/csv/example.conf
elahrvivaz: having an attribute named ID causes some geotools issues.

PR #947, opened 2016-06-22
GEOMESA-1268 Removing guava cache putAll call for guava 11 compatibility

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
@@ -183,7 +183,7 @@ class AccumuloBackedMetadata[T](connector: Connector, catalogTable: String, seri
     writer.addMutation(insert)
     writer.close()
     // also pre-fetch into the cache
-    // note: don't use putAll, it breaks guava compatibility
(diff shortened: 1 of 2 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
```diff
@@ -183,8 +183,8 @@ class AccumuloBackedMetadata[T](connector: Connector, catalogTable: String, seri
     writer.addMutation(insert)
     writer.close()
     // also pre-fetch into the cache
-    val toCache = kvPairs.map(kv => (typeName, kv._1) -> Option(kv._2))
-    metaDataCache.putAll(toCache)
+    // note: don't use putAll, it breaks guava compatibility
```
cne1x: I like this comment, but I'm wondering whether we shouldn't also add a note in the main POM to explain how we have a core dependency (because of Accumulo) on Guava 17.0, but that most of our client-side deployments use GeoTools/GeoServer that have a dependency on Guava 11.0.2, so we are trying to use the calls in Guava 17 that also existed in Guava 11.  Just a thought.

elahrvivaz: I'll add some more comments

PR #948, opened 2016-06-22
GEOMESA-1269 Docs: Accumulo runtime required for ingest

mdzimmerman:
Moves runtime installation before command-line tools install.

### DIFF
--- a/docs/user/jobs.rst
+++ b/docs/user/jobs.rst
@@ -1,13 +1,13 @@
 GeoMesa Jobs
 ============
 
-Building Instructions
----------------------
-
 This project (``geomesa-jobs`` in the source distribution) contains Map-Reduce
 and Scalding jobs for maintaining GeoMesa.
 
-If you wish to build this project separately, you can with maven:
+Building Instructions
+---------------------
+
+If you wish to build ``geomesa-jobs`` separately, you can with Maven:
 
 .. code-block:: shell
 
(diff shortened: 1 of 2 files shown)


jahhulbert-ccri:
@mdzimmerman can we add something about versioning somewhere? the client must match the distributed runtime and they need to be kept in sync.

mdzimmerman:
The "Upgrading" section near the bottom of this chapter does mention matching versions: https://github.com/ccri/geomesa/blob/0eb95ce655b7d1d165192c6ac6f2f02e28ad2855/docs/user/installation_and_configuration.rst#upgrading

That having been said we could add a note box in this section about that.

I am not so sure we explicitly mention that data ingested from previous versions can be read with newer versions. Has that been true for all point releases in the 1.2.x series? I was under the impression (perhaps mistakenly) that this was not entirely the case for 1.2.2.

jahhulbert-ccri:
@mdzimmerman yeah you can use data ingested from rc7 as well.

docs/user/jobs.rst
```diff
@@ -130,7 +132,7 @@ slightly):
 
 .. code-block:: shell
 
-    yarn jar geomesa-jobs/target/geomesa-jobs-1.0.0-shaded.jar \
+    geomesa$ yarn jar geomesa-jobs/target/geomesa-jobs-1.0.0-shaded.jar \
```
jahhulbert-ccri: should this also have the $VERSION plugged in?

PR #961, opened 2016-06-27
GEOMESA-1256 WPS join process for querying two different schemas

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
@@ -1,0 +1,121 @@
+/***********************************************************************
+* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.accumulo.process
+
+import com.typesafe.scalalogging.LazyLogging
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.geotools.data.store.ReTypingFeatureCollection
+import org.geotools.feature.DefaultFeatureCollection
+import org.geotools.process.ProcessException
+import org.geotools.process.factory.{DescribeParameter, DescribeProcess, DescribeResult}
+import org.geotools.process.vector.VectorProcess
+import org.geotools.util.NullProgressListener
+import org.locationtech.geomesa.accumulo.process.query.{QueryResult, QueryVisitor}
+import org.locationtech.geomesa.accumulo.process.unique.UniqueProcess
+import org.locationtech.geomesa.accumulo.util.SelfClosingIterator
+import org.locationtech.geomesa.filter.function.BinaryOutputEncoder
+import org.locationtech.geomesa.filter.function.BinaryOutputEncoder.EncodingOptions
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.opengis.util.ProgressListener
+
+/**
+  * Returns features from a feature type based on a join against a second feature type.
+  */
+@DescribeProcess(
+  title = "Join Process",
+  description = "Queries a feature type based on attributes from a second feature type"
+)
+class JoinProcess extends VectorProcess with LazyLogging {
+
+  // TODO this is currently aimed at a 1.2.2 backport - instead of the UniqueProcess we could run a stats job
+  /**
+    * @param unique result of running UniqueProcess
+    * @param join joined data set from which return features will be taken
+    * @param attribute attribute to join on
+    * @param bins flag to return results in BIN format
+    * @param binDtg date field for bin records - will use default date if not specified
+    * @param binTrackId track ID field for bin records - will not include trackId if not specified
+    * @param binLabel label field for bin record (optional)
+    * @param monitor listener to monitor progress
+    * @throws org.geotools.process.ProcessException if something goes wrong
+    * @return
+    */
+  @throws(classOf[ProcessException])
+  @DescribeResult(name = "result", description = "Output features")
+  def execute(@DescribeParameter(name = "unique", description = "Result of unique process for joined attribute")
+              unique: SimpleFeatureCollection,
+              @DescribeParameter(name = "join", description = "Joined features")
+              join: SimpleFeatureCollection,
+              @DescribeParameter(name = "attribute", description = "Attribute field to join on", min = 1)
+              attribute: String,
+              @DescribeParameter(name = "bins", description = "Return BIN records or regular records", min = 0)
+              bins: java.lang.Boolean,
+              @DescribeParameter(name = "binDtg", description = "Date field to use for BIN records", min = 0)
+              binDtg: String,
+              @DescribeParameter(name = "binTrackId", description = "Track field to use for BIN records", min = 0)
+              binTrackId: String,
+              @DescribeParameter(name = "binLabel", description = "Label field to use for BIN records", min = 0)
+              binLabel: String,
+              monitor: ProgressListener): SimpleFeatureCollection = {
+
+    import org.locationtech.geomesa.filter.ff
+
+    import scala.collection.JavaConversions._
+
+    logger.trace(s"Attempting join query on ${join.getClass.getName}")
+
+    if (join.isInstanceOf[ReTypingFeatureCollection]) {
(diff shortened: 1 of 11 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+      None
+    }
+
+    val result = if (joinFilters.isEmpty) {
+      new DefaultFeatureCollection(null, join.getSchema)
+    } else {
+      val filter = ff.or(joinFilters.toList)
+      val visitor = new QueryVisitor(join, filter)
+      join.accepts(visitor, new NullProgressListener)
```
cne1x: Just to make sure I understand...  If there are a million unique join-field values, then the filter will consist of a million ```equals``` filters all joined into a single ```OR```, and that massive filter will be applied to every feature in the ```join``` collection?  Is that right?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/resources/process/join-bin.xml
```diff
@@ -0,0 +1,102 @@
...
+        <wps:LiteralData>pk</wps:LiteralData>
+      </wps:Data>
+    </wps:Input>
+    <wps:Input>
+      <ows:Identifier>bins</ows:Identifier>
+      <wps:Data>
+        <wps:LiteralData>true</wps:LiteralData>
+      </wps:Data>
+    </wps:Input>
+    <!--<wps:Input>
```
cne1x: Do we need to preserve the comments?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
elahrvivaz: the query visitor gets intercepted by our data store, so it results in a new geomesa query and doesn't run directly against each feature. But yeah, in general you would end up with a massive OR query. In general we handle those pretty well, but the onus is on the user to make suitable queries.
The impetus for this functionality is something like a summary feature combined with multiple detail features. We look up the summary by ID (so only a single value comes back), then join against the detailed features. Instead of making it a 2-stage query handled by the client, it's all handled server side in this WPS.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/resources/process/join-bin.xml
elahrvivaz: I left them in there as examples of the possible input configuration available. If they're not included, it will try to use sensible defaults.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,121 @@
...
+      visitor.getResult.asInstanceOf[QueryResult].results
+    }
+
+    // pass bin parameters off to the output format
+    binOptions match {
+      case None    => BinaryOutputEncoder.CollectionEncodingOptions.remove(result.getID)
+      case Some(o) => BinaryOutputEncoder.CollectionEncodingOptions.put(result.getID, o)
+    }
+
+    result
```
cne1x: Apologies for being pedantic, but is this really a JOIN process, or is it a sub-select process?  The difference being that we are not creating a new synthetic feature-type from two separate feature types that are defined by a set of criteria; instead, we are selecting the records from a second feature collection whose foreign-key values come from an external set.  The actual join would have to be done externally to this process, yes?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/JoinProcessTest.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class JoinProcessTest extends Specification with TestWithMultipleSfts {
+
+  sequential
+
+  val sft1 = createNewSchema("pk:String,age:Int,weight:Int,dtg:Date,*geom:Point:srid=4326")
+  val sft2 = createNewSchema("pk:String:index=true,dtg:Date,*geom:Point:srid=4326")
```
cne1x: [picky] Wouldn't the field in ```sft2``` be called ```fk``` (as it's a foreign-key reference) instead of ```pk```?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
elahrvivaz: true, there is no joining of attributes. Cross-schema select? What do you think it should be called?

elahrvivaz: actually I should verify with @BrittEvans that he doesn't want a true join

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/JoinProcessTest.scala
elahrvivaz: pk is probably a bad name for it - I'll change to 'track'. the names have to match between the two sfts though. It's not actually meant to be a foreign key, just a common key stored in both sfts.

geomesa-process/src/main/scala/org/locationtech/geomesa/process/ProcessFactory.scala
```diff
@@ -33,5 +33,6 @@ class ProcessFactory
     classOf[UniqueProcess],
     classOf[HashAttributeProcess],
     classOf[HashAttributeColorProcess],
-    classOf[SamplingProcess]
+    classOf[SamplingProcess],
+    classOf[JoinProcess]
```
jnh5y: Can you add a one-line description to geomesa-process/README.md?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,121 @@
...
+              @DescribeParameter(name = "attribute", description = "Attribute field to join on", min = 1)
+              attribute: String,
+              @DescribeParameter(name = "bins", description = "Return BIN records or regular records", min = 0)
+              bins: java.lang.Boolean,
+              @DescribeParameter(name = "binDtg", description = "Date field to use for BIN records", min = 0)
+              binDtg: String,
+              @DescribeParameter(name = "binTrackId", description = "Track field to use for BIN records", min = 0)
+              binTrackId: String,
+              @DescribeParameter(name = "binLabel", description = "Label field to use for BIN records", min = 0)
+              binLabel: String,
```
jnh5y: Hmm...  two things:
1.  Is there any sensible way to separate out the bin file format options so that they aren't required?  (I could imagine using this in other settings.)

2.  Would be it be sensible to provide additional filters for the second feature collections?  For instance, I could see someone wanting only the data in a particular view port, etc.

elahrvivaz: 1. the bin arguments are optional (that's the min = 0), but I couldn't find a good way to pass then in other than this. The WPS output formats don't have any mechanism for passing information in, so I had to hack it with a static object.
2. I could see how that would be useful. Not sure it's a use-case at the moment though... I'll try to add it anyway.

elahrvivaz: I'm going to update this with actual joining of attributes.

elahrvivaz:
temporarily closing

elahrvivaz:
updated with joining of attributes

geomesa-process/src/main/scala/org/locationtech/geomesa/process/ProcessFactory.scala
elahrvivaz: added

elahrvivaz:
won't let me re-open, moved to #969

PR #963, opened 2016-06-28
GEOMESA-1267 Checking for iterator versions on getSchema

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -11,10 +11,12 @@ package org.locationtech.geomesa.accumulo.data
 
 import java.io.IOException
 import java.nio.charset.StandardCharsets
+import java.util.concurrent.atomic.AtomicLong
 import java.util.{List => jList}
 
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.accumulo.core.client._
+import org.apache.accumulo.core.security.Authorizations
 import org.geotools.data._
 import org.geotools.factory.Hints
 import org.geotools.feature.{FeatureTypes, NameImpl}
@@ -26,6 +28,7 @@ import org.locationtech.geomesa.accumulo.data.stats._
 import org.locationtech.geomesa.accumulo.data.tables._
 import org.locationtech.geomesa.accumulo.index.Strategy.StrategyType.StrategyType
 import org.locationtech.geomesa.accumulo.index._
+import org.locationtech.geomesa.accumulo.iterators.ProjectVersionIterator
 import org.locationtech.geomesa.accumulo.util.{DistributedLocking, GeoMesaBatchWriterConfig, Releasable}
 import org.locationtech.geomesa.accumulo.{AccumuloVersion, GeomesaSystemProperties}
 import org.locationtech.geomesa.features.SerializationOption.SerializationOptions
@@ -33,6 +36,7 @@ import org.locationtech.geomesa.features.SerializationType.SerializationType
 import org.locationtech.geomesa.features.kryo.KryoFeatureSerializer
 import org.locationtech.geomesa.features.{SerializationType, SimpleFeatureSerializers}
 import org.locationtech.geomesa.security.{AuditProvider, AuthorizationsProvider}
+import org.locationtech.geomesa.utils.conf.GeoMesaProperties
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.{FeatureSpec, NonGeomAttributeSpec}
 import org.locationtech.geomesa.utils.geotools.{GeoToolsDateFormat, SimpleFeatureTypes}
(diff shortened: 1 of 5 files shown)


cne1x:
Looks good to me.

geomesa-utils/pom.xml
```diff
@@ -136,4 +136,24 @@
...
+                <excludes>
+                    <exclude>org/locationtech/geomesa/geomesa.properties</exclude>
+                </excludes>
+            </resource>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+                <includes>
+                    <include>org/locationtech/geomesa/geomesa.properties</include>
+                </includes>
```
jnh5y: Ok, I'll bite.  These two whacks seem to be at odds....

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -684,6 +684,31 @@ class AccumuloDataStore(val connector: Connector,
...
+      val clientVersion = GeoMesaProperties.GeoMesaProjectVersion
+      val scanner = connector.createScanner(catalogTable, new Authorizations())
+      val iteratorVersion = try {
+        ProjectVersionIterator.scanProjectVersion(scanner)
+      } catch {
+        case NonFatal(e) => "unavailable"
+      } finally {
+        scanner.close()
+      }
+      if (iteratorVersion != clientVersion) {
```
jnh5y: Hrmm... this check seems to be rather strict.  

Are there any obvious options which would let us have some room for mixing and matching, but still let folks know when their versions are too different?

elahrvivaz: the version is the geomesa pom version, e.g. 1.2.3. Currently, I'm pretty sure Bad Things will happen if it doesn't match. Maybe later we can relax this requirement when we get serious about min/maj.

geomesa-utils/pom.xml
elahrvivaz: one excludes the props file (and implicitly includes everything else) and _doesn't_ do filtering, the second includes the props file (and implicitly excludes everything else) and _does_ do filtering.

jnh5y: Derp.  Can't read.  My XML eyes glossed over the difference between includes and excludes:(.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: @jnh5y maybe add a flag in the data store config to ignore differing versions?

jahhulbert-ccri: I vote to keep it strict for now. Add a note perhaps that one day we can not be strict...as you mention we should enforce the gs plugin matching the iter version until we start really caring about verisons. The only exception though I think is older ingest clients should be able to keep writing? maybe that doesn't matter

elahrvivaz: Yeah, it would prevent older clients from writing - that may be a good thing, as if they tried to do an update it might fail. If they really want to keep using the old clients, they can always use the ignore flag.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaProperties.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package org.locationtech.geomesa.utils.conf
+
+import java.util.Properties
+
+import com.typesafe.scalalogging.LazyLogging
+
+object GeoMesaProperties extends LazyLogging {
+
+  private val EmbeddedFile = "/org/locationtech/geomesa/geomesa.properties"
```
jnh5y: Is this configuration file documented?

jahhulbert-ccri:
@jnh5y i think the main issue right now is that our code actually doesn't support mixing versions so it's probably best to not proceed if the system figures out its in an unsupported situation?

jahhulbert-ccri:
ok...verified these test cases:
1. ingest with 1.2.3, upgrade to 1.2.4 worked, export with either worked
2. ingest with 1.2.4, export with 1.2.4 worked
3. ingest with 1.2.3, no upgrade, export with 1.2.4 failed (expected)

jnh5y:
Maybe we could split this up into a smaller PR which would satisfy  https://geomesa.atlassian.net/browse/GEOMESA-1348 and then sort out what to do for mismatches?

elahrvivaz:
note: going to split out the versioning piece with the checking piece so that we have it available

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaProperties.scala
elahrvivaz: no, but it currently only has the version from the pom - not something anyone should be modifying.

elahrvivaz:
Removed throwing of exception - note this will likely just cause other failures further down the line, but ¯\_(ツ)_/¯

geomesa-utils/pom.xml
jahhulbert-ccri: I had a friend who named their cat derp...
image

jahhulbert-ccri:
gonna play with it and see how it works and then merge if I like it.

PR #965, opened 2016-06-29
GEOMESA-1287,GEOMESA-1286 Using guava executors for stats to not hang…

elahrvivaz:
… up JVM shutdown, refactoring query stats writing to tie to datastore lifecycle


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatReader.scala
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      val scanner = connector.createScanner(table, auths)
+      val rangeStart = s"$typeName~${UsageStatTransform.dateFormat.print(dates.getStart)}"
+      val rangeEnd = s"$typeName~${UsageStatTransform.dateFormat.print(dates.getEnd)}"
+      scanner.setRange(new Range(rangeStart, rangeEnd))
+      transform.iterator(scanner)
+    }
+  }
+
+  private def checkTable: Boolean = synchronized {
```
cne1x: Isn't this functionally equivalent to declaring the following?

```scala
private lazy var tableExists = synchronized {
  tableExists = connector.tableOperations().exists(table)
}
```

cne1x:
On a quick read-through, this looks reasonable.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatReader.scala
elahrvivaz: hmm, does that work? you're setting the var to a Unit statement and updating the var inside the statement? I think I need the method so that if the table doesn't exist initially, but gets created later it will pick up the change.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -76,10 +75,15 @@ class AccumuloDataStore(val connector: Connector,
 
   private val tableOps = connector.tableOperations()
   private val statsTable = GeoMesaTable.concatenateNameParts(catalogTable, "stats")
+  private val usageStatsTable = GeoMesaTable.concatenateNameParts(catalogTable, "queries")
 
   override val metadata: GeoMesaMetadata[String] =
     new AccumuloBackedMetadata(connector, catalogTable, MetadataStringSerializer)
+
   override val stats: GeoMesaStats = new GeoMesaMetadataStats(this, statsTable, config.generateStats)
+  override val usageStats: GeoMesaUsageStats =
```
jnh5y: Should this be instantiated if collectUsageStats is false?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -407,7 +411,10 @@ class AccumuloDataStore(val connector: Connector,
    *
    * @see org.geotools.data.DataAccess#dispose()
    */
-  override def dispose(): Unit = {}
+  override def dispose(): Unit = {
+    stats.close()
+    usageStats.close()
```
jnh5y: If the above is refactored, we should only close a stats class which is used;).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -79,24 +80,21 @@ class AccumuloDataStoreFactory extends DataStoreFactorySpi {
...
 
-    // stats defaults to true if not specified
-    val collectQueryStats = !useMock && collectQueryStatsParam.lookupWithDefault[Boolean](params)
-
-    if (collectQueryStats) {
-      new AccumuloDataStore(connector, tableName, authProvider, auditProvider, visibility, config) with StatWriter
-    } else {
-      new AccumuloDataStore(connector, tableName, authProvider, auditProvider, visibility, config)
-    }
+    new AccumuloDataStore(connector, tableName, authProvider, auditProvider, visibility, config)
```
jnh5y: nice refactoring!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: you're still able to query stats through this even if collect is false

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
```diff
@@ -49,13 +50,24 @@ class GeoMesaMetadataStats(val ds: AccumuloDataStore, statsTable: String, genera
...
+    override def run(): Unit = {
       if (lastCompaction.get < DateTime.now(DateTimeZone.UTC).minusHours(1).getMillis &&
           compactionScheduled.compareAndSet(true, false) ) {
         compact()
       }
-  }, 1, 1, TimeUnit.HOURS)
+      if (running.get) {
+        scheduledCompaction = executor.schedule(this, 1, TimeUnit.HOURS)
+      } else {
+        scheduledCompaction = null
```
jnh5y: Does setting the scheduledCompaction to null stop/cancel the previously scheduled compactions?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
```diff
@@ -212,6 +224,13 @@ class GeoMesaMetadataStats(val ds: AccumuloDataStore, statsTable: String, genera
 
   override def clearStats(sft: SimpleFeatureType): Unit = metadata.delete(sft.getTypeName)
 
+  override def close(): Unit = {
+    running.set(false)
+    if (scheduledCompaction != null) {
+      scheduledCompaction.cancel(false)
+    }
```
jnh5y: Hmm.... if I'm understanding correctly, running.set(false) will start a race between 1) scheduledCompaction to set to null and 2) getting into the if block....

Seems like one could see either A) an NPE on 230 (after the check passes, the other thread nulls out scheduledCompaction) or B) the check fails and the compactions aren't canceled properly.

This seems like a problem in a web container situation where dispose a datastore could be common.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
```diff
@@ -0,0 +1,83 @@
...
+class UsageStatWriter(connector: Connector, table: String) extends Runnable with Closeable with LazyLogging {
+
+  // initial schedule
+  UsageStatWriter.executor.schedule(this, writeDelayMillis, TimeUnit.MILLISECONDS)
+
+  private val writeDelayMillis = 5000
+
+  private val batchWriterConfig = GeoMesaBatchWriterConfig().setMaxMemory(10000L).setMaxWriteThreads(5)
+
+  private var maybeWriter: BatchWriter = null
```
jnh5y: Great variable name.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
```diff
@@ -0,0 +1,83 @@
...
+   */
+  def queueStat[T <: UsageStat](stat: T)(implicit transform: UsageStatTransform[T]): Unit =
+    queue.offer(() => transform.statToMutation(stat))
+
+  override def run() = {
+    var toMutation = queue.poll()
+    if (toMutation != null) {
+      val writer = getWriter
+      do {
+        writer.addMutation(toMutation())
```
jnh5y: Ok, I'm confused by the () in the toMutation()...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
elahrvivaz: yeah, I decided to not worry about synchronizing it because the compactions are only run once an hour, so the odds of an NPE are low (and the data store would have to be held for at least an hour before disposing)... but true it could throw during dispose, which would be unfortunate. I'll update

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
elahrvivaz: no, but the previously scheduled one is **this** one, if that makes sense...

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureReaderTest.scala
```diff
@@ -38,10 +39,20 @@ class AccumuloFeatureReaderTest extends Specification with TestWithDataStore {
...
-  def dataStoreWithStats(statArray: ArrayBuffer[Stat]) =
+  def dataStoreWithStats(statArray: ArrayBuffer[UsageStat]) =
     new AccumuloDataStore(ds.connector, ds.catalogTable, ds.authProvider,
-      ds.auditProvider, ds.defaultVisibilities, ds.config) with StatWriter {
-    override def writeStat(stat: Stat): Unit = statArray.append(stat)
+      ds.auditProvider, ds.defaultVisibilities, ds.config.copy(collectUsageStats = true)) {
+    override val usageStats = new StatWriter(statArray)
+  }
+
+  class StatWriter(statArray: ArrayBuffer[UsageStat]) extends GeoMesaUsageStats {
```
jnh5y: Is it worth adding 'Mock' or 'Test' to this class name?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
elahrvivaz: it's invoking the function to get the actual mutation - lazy evaluation essentially. see the queue.offer on line 44

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureReaderTest.scala
elahrvivaz: i'll update

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
```diff
@@ -146,5 +147,6 @@ class GeoMesaRecordWriter(params: Map[String, String], delegate: RecordWriter[Te
   override def close(reporter: Reporter) = {
     delegate.close(reporter)
     statsCache.values.foreach(_.close())
+    ds.dispose()
```
jnh5y: Nice job on adding ds.dispose's everywhere.  Is it *everywhere*?;)

jnh5y:
After a read through, this looks good.  I'm going to test out with the GM-Tutorials Accumulo QuickStart before merging..

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapred/GeoMesaOutputFormat.scala
elahrvivaz: everywhere I could find! :)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
jnh5y: *nice*  a queue of computation...  *mind blown*.

elahrvivaz: haha, I was trying to figure out a way to pass the implicit transform around without losing the stat type in the queue, and this is what I came up with :)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/GeoMesaMetadataStats.scala
elahrvivaz: fixed

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureReaderTest.scala
elahrvivaz: fixed

jnh5y:
Worked for me!

PR #966, opened 2016-06-29
GEOMESA-1289 Download script for tdrive doesn't work

atallahhezbor:
Updates URLs for TDrive and handles scraping complaints

### DIFF
--- a/geomesa-tools/bin/download-data.sh
+++ b/geomesa-tools/bin/download-data.sh
@@ -53,14 +53,16 @@ case "$type" in
   tdrive)
 
     echo "Note: each zip file contains approximately one million points"
-    read -p "Download how many zip files? " NUM
+    read -p "Download how many zip files? (9 total) " NUM
 
     for i in `seq 1 $NUM`; do
-	  echo "Downloading zip $i of $NUM"
-	  wget "http://research.microsoft.com/pubs/152883/0$i.zip" -P $GEOMESA_HOME/data/tdrive
+      echo "Downloading zip $i of $NUM"
+      ZIP_ID=$((i+5))
+      wget "https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/0$ZIP_ID.zip" -P $GEOMESA_HOME/data/tdrive
+      sleep 5
     done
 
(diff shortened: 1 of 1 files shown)


geomesa-tools/bin/download-data.sh
```diff
@@ -53,14 +53,16 @@ case "$type" in
   tdrive)
 
     echo "Note: each zip file contains approximately one million points"
-    read -p "Download how many zip files? " NUM
+    read -p "Download how many zip files? (9 total) " NUM
```
jnh5y: Are there only 9 files now?  I previously downloaded like ~4800....

atallahhezbor: They may have grouped some them together, but i don't feel like their claim of 15 million points still holds.
[See here](https://www.microsoft.com/en-us/research/publication/t-drive-trajectory-data-sample/)

elahrvivaz: When I downloaded it a few months ago, there were zips 01-014. But is seems that 01-05 are no longer hosted...
That page doesn't render fully for me in chrome, either.

elahrvivaz:
actually the links to the first 5 still work: https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/02.zip

atallahhezbor:
Should we count on those remaining hosted and always download them? Worst case the user gets what will look like a bad zip

atallahhezbor:
Or we could check the length of the file first

elahrvivaz:
@atallahhezbor I pushed a change that works for me - can you try it out?

elahrvivaz:
actually doing a length or md5 check on the files is a pretty good idea - there are only 14 of them so we could compare them in the script

elahrvivaz:
actually not sure if you can access it but wget shows the response type - we could just check that it's application/zip.

elahrvivaz:
oops, i pushed to the wrong branch - here's the change I made: https://github.com/ccri/geomesa/commit/c201bbb5a46381e50198ce463a011b3afc4b2b1d

atallahhezbor:
That worked for me! Not sure why I was getting 403s with my user agent.
For checking type, we could grep the output of wget --spider?

jnh5y:
Merged as 034fcad6230d3db834832316b64d616f7847cd56.  Thanks @atallahhezbor @elahrvivaz!

PR #968, opened 2016-06-30
GEOMESA-1292 Fixing assignment of geometry as an indexed attribute

elahrvivaz:


cne1x:
This looks good to me.

One (small, probably irrelevant) question, though...  It looks like we have multiple places in the ```GeomAttributeSpec``` where we are working around the ```OPT_INDEX``` map entry.  What if we simply removed that up front from the ```options``` that are passed in?  If that entry is always Bad for this class, why not remove it?  Just a thought.

elahrvivaz:
Yeah, I can see it both ways - there are multiple places that we construct the attribute spec from what i can see - figured it was safer to remove it at the end then try to catch it up front. Should probably do both for extra security...

cne1x:
I hadn't realized there were multiple ways to get those options in.  In that case, forget I mentioned it.  :)

jnh5y:
Does this PR handle the situation where an old table has geom:index=true already set?

elahrvivaz:
@jnh5y yep

PR #969, opened 2016-06-30
GEOMESA-1256 WPS join process for querying two different schemas

elahrvivaz:
I rebased on master and it wouldn't let me re-open the old PR, so this replaces #961 

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
@@ -1,0 +1,250 @@
+/***********************************************************************
+* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.accumulo.process
+
+import com.typesafe.scalalogging.LazyLogging
+import org.geotools.data.simple.{SimpleFeatureCollection, SimpleFeatureIterator}
+import org.geotools.data.store.ReTypingFeatureCollection
+import org.geotools.feature.collection.DecoratingSimpleFeatureCollection
+import org.geotools.feature.simple.SimpleFeatureTypeBuilder
+import org.geotools.feature.{AttributeTypeBuilder, DefaultFeatureCollection}
+import org.geotools.process.ProcessException
+import org.geotools.process.factory.{DescribeParameter, DescribeProcess, DescribeResult}
+import org.geotools.process.vector.VectorProcess
+import org.geotools.util.NullProgressListener
+import org.locationtech.geomesa.accumulo.process.query.{QueryResult, QueryVisitor}
+import org.locationtech.geomesa.accumulo.util.SelfClosingIterator
+import org.locationtech.geomesa.features.ScalaSimpleFeature
+import org.locationtech.geomesa.filter.function.BinaryOutputEncoder
+import org.locationtech.geomesa.filter.function.BinaryOutputEncoder.EncodingOptions
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+import org.opengis.filter.Filter
+import org.opengis.util.ProgressListener
+
+/**
+  * Returns features from a feature type based on a join against a second feature type.
+  */
+@DescribeProcess(
+  title = "Join Process",
+  description = "Queries a feature type based on attributes from a second feature type"
+)
+class JoinProcess extends VectorProcess with LazyLogging {
+
+  /**
+    *
+    * @param primary main feature collection to query
+    * @param secondary secondary feature collection to query with results from first feature collection
+    * @param joinAttribute attribute to join on
+    * @param joinFilter additional filter to apply to joined features
+    * @param attributes attributes to return, from both collections, qualified by schema name
+    * @param bins flag to return results in BIN format
+    * @param binDtg date field for bin records - will use default date if not specified
+    * @param binTrackId track ID field for bin records - will not include trackId if not specified
+    * @param binLabel label field for bin record (optional)
+    * @param monitor listener to monitor progress
+    * @throws org.geotools.process.ProcessException if something goes wrong
+    * @return
+    */
+  @throws(classOf[ProcessException])
+  @DescribeResult(name = "result", description = "Output features")
+  def execute(@DescribeParameter(name = "primary", description = "Primary feature collection being queried", min = 1)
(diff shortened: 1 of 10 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,250 @@
...
+              binLabel: String,
+              monitor: ProgressListener): SimpleFeatureCollection = {
+
+    import org.locationtech.geomesa.filter.ff
+
+    import scala.collection.JavaConversions._
+
+    logger.trace(s"Attempting join query on ${joinAttribute.getClass.getName}")
+
+    if (primary.isInstanceOf[ReTypingFeatureCollection] || secondary.isInstanceOf[ReTypingFeatureCollection]) {
```
jnh5y: Is it a hard requirement that both datasets be a GeoMesa datastore?

elahrvivaz: no, but if it's not a geomesa feature collection, then it will manually iterate over all the features. this warning is in case you have a geomesa feature collection, but it's been wrapped by geoserver, which prevents us from processing the query visitor, so you end up manually iterating over all the features.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,250 @@
...
+    */
+  @throws(classOf[ProcessException])
+  @DescribeResult(name = "result", description = "Output features")
+  def execute(@DescribeParameter(name = "primary", description = "Primary feature collection being queried", min = 1)
+              primary: SimpleFeatureCollection,
+              @DescribeParameter(name = "secondary", description = "Secondary feature collection to be joined", min = 1)
+              secondary: SimpleFeatureCollection,
+              @DescribeParameter(name = "joinAttribute", description = "Attribute field to join on", min = 1)
+              joinAttribute: String,
+              @DescribeParameter(name = "joinFilter", description = "Additional filter to apply to joined features", min = 0)
```
jnh5y: Thanks for the joinFilter!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,250 @@
...
+    // check for too many features coming back - limit is somewhat arbitrary, but this
+    // class is mainly intended for a single feature lookup
+    val primaryFeatures = SelfClosingIterator(primary).toList
+    require(primaryFeatures.length < 129,
+      s"Too many features returned from primary query - got ${primaryFeatures.length}, max 128")
+
+    val joinProperty = ff.property(joinAttribute)
+    val joinFilters = {
+      val values = primaryFeatures.map(_.getAttribute(joinAttribute)).distinct
+      values.map(p => ff.equals(joinProperty, ff.literal(p)))
```
jnh5y: Does it make sense to do any deduplication here?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
```diff
@@ -0,0 +1,250 @@
...
+
+      new DecoratingSimpleFeatureCollection(results) {
+        override def getSchema: SimpleFeatureType = returnSft
+        override def features(): SimpleFeatureIterator = new SimpleFeatureIterator {
+          val delegate = results.features
+          override def next(): SimpleFeature = {
+            val secondarySf = delegate.next()
+            val toJoin = secondarySf.getAttribute(joinAttribute)
+            val primarySf = primaryFeatures.find(_.getAttribute(joinAttribute) == toJoin).getOrElse {
+              throw new RuntimeException("No feature joined from attribute query")
```
jnh5y: Ok, so if there's not a match, exceptions are thrown?  

It seems like there could be misses...

jnh5y:
This looks reasonable to me.  Since it is a WPS, I'm fine without reviewer testing.  @jahhulbert-ccri thoughts?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
elahrvivaz: there's a .distinct right above that does the dedupe

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
elahrvivaz: shouldn't be possible for there to be any misses. The join filter is created based on the attribute values from the primary features, so each join feature should match at least one primary feature.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
jahhulbert-ccri: yeah historically this is for debugging other cases like problems solved by the layer name in geoserver not matching the type name in accumulo i think it would make a retyping feature collection and our visitors then wouldn't work so we looked for log statements like this :sob:

jnh5y: We should be bettter GT/GS citizens and go fix the ReTypingFeatureCollection:).

jahhulbert-ccri:
looks good to me. i'll run a simple test and merge it for 1.2.4

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/JoinProcess.scala
elahrvivaz: +1

jnh5y: I've added it to my 'nice-to-do' list.  

There's gonna be a GT/GS remote code sprint on July 22nd...  Might could pull together something before then...

jahhulbert-ccri:
deployed...tested...worked in wps demo in geoserver

PR #970, opened 2016-06-30
GEOMESA-1249 Fixing kafka data store appending feature writer

elahrvivaz:
Note: not all unit tests pass in master currently - this fixes some tests but not all.

This should adhere better to the geotools API, but that might cause issues for current users who expect the previous behavior.

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
@@ -9,6 +9,7 @@
 package org.locationtech.geomesa.kafka
 
 import java.io.Closeable
+import java.util.concurrent.atomic.AtomicLong
 import java.{util => ju}
 
 import com.typesafe.scalalogging.LazyLogging
@@ -17,7 +18,6 @@ import kafka.producer.{Producer, ProducerConfig}
 import org.geotools.data.store.{ContentEntry, ContentFeatureStore}
 import org.geotools.data.{FeatureReader, FeatureWriter, Query}
 import org.geotools.feature.FeatureCollection
-import org.geotools.feature.collection.BridgeIterator
 import org.geotools.geometry.jts.ReferencedEnvelope
 import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.locationtech.geomesa.kafka.KafkaDataStore.FeatureSourceFactory
@@ -34,10 +34,10 @@ class KafkaProducerFeatureStore(entry: ContentEntry,
                                 topic: String,
                                 broker: String,
                                 producer: Producer[Array[Byte], Array[Byte]],
-                                q: Query = null)
+                                q: Query)
   extends ContentFeatureStore(entry, q) with Closeable with LazyLogging {
 
-  private val writerPool = ObjectPoolFactory(new ModifyingFeatureWriter(query), 5)
+  private val writerPool = ObjectPoolFactory(getWriterInternal(null, 0).asInstanceOf[KafkaFeatureWriter], 5)
 
   override def getBoundsInternal(query: Query) =
     ReferencedEnvelope.create(new Envelope(-180, 180, -90, 90), CRS_EPSG_4326)
@@ -47,13 +47,13 @@ class KafkaProducerFeatureStore(entry: ContentEntry,
   override def addFeatures(featureCollection: FeatureCollection[SimpleFeatureType, SimpleFeature]): ju.List[FeatureId] = {
     writerPool.withResource { fw =>
       val ret = Array.ofDim[FeatureId](featureCollection.size())
-      fw.setIter(new BridgeIterator[SimpleFeature](featureCollection.features()))
+      val iter = featureCollection.features()
       var i = 0
-      while(fw.hasNext) {
+      while (iter.hasNext) {
         val sf = fw.next()
         ret(i) = sf.getIdentifier
-        i+=1
-        fw.write()
+        fw.write(sf)
+        i += 1
       }
       ret.toList
     }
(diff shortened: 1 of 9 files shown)


elahrvivaz:
@jahhulbert-ccri do you have time to test this out? I'm not super confident that something won't break, given the state of kafka unit tests.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
```diff
@@ -64,86 +64,83 @@ class KafkaProducerFeatureStore(entry: ContentEntry,
...
   }
 
-  def clearFeatures(): Unit = {
-    val msg = GeoMessage.clear()
-    logger.debug("sending message: {}", msg)
+  def clearFeatures(): Unit = writerPool.withResource { fw => fw.send(GeoMessage.clear()) }
 
-    val encoder = new KafkaGeoMessageEncoder(sft)
-    producer.send(encoder.encodeClearMessage(topic, msg))
+  override def getWriterInternal(query: Query, flags: Int) = {
```
Afterglow375: What's the `flags` parameter for? I don't see it being used.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
```diff
@@ -64,86 +64,83 @@ class KafkaProducerFeatureStore(entry: ContentEntry,
...
+  override def getReaderInternal(query: Query): FeatureReader[SimpleFeatureType, SimpleFeature] = null
 
-  class ModifyingFeatureWriter(query: Query) extends FeatureWriter[SimpleFeatureType, SimpleFeature] with LazyLogging {
+  override def close(): Unit = producer.close()
+}
 
-    val msgEncoder = new KafkaGeoMessageEncoder(sft)
-    val reuse = new ScalaSimpleFeature("", sft)
+abstract class KafkaFeatureWriter(sft: SimpleFeatureType, producer: Producer[Array[Byte], Array[Byte]], topic: String)
+    extends FeatureWriter[SimpleFeatureType, SimpleFeature]with LazyLogging {
```
Afterglow375: [picky] add a space between the `]w` in `SimpleFeature]with`

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
```diff
@@ -64,86 +64,83 @@ class KafkaProducerFeatureStore(entry: ContentEntry,
...
-    def setIter(iter: Iterator[SimpleFeature]): Unit = {
-      toModify = iter
-    }
+  override def getFeatureType: SimpleFeatureType = sft
+  override def close(): Unit = {}
+}
 
-    var curFeature: SimpleFeature = null
-    override def getFeatureType: SimpleFeatureType = sft
+class KafkaFeatureWriterAppend(sft: SimpleFeatureType, producer: Producer[Array[Byte], Array[Byte]], topic: String)
```
Afterglow375: General question... when using feature writer append, is it okay to modify features? Or as the name suggests is it just for appending?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
```diff
@@ -64,86 +64,83 @@ class KafkaProducerFeatureStore(entry: ContentEntry,
...
+class KafkaFeatureWriterModify(sft: SimpleFeatureType, producer: Producer[Array[Byte], Array[Byte]], topic: String, query: Query)
+    extends KafkaFeatureWriterAppend(sft, producer, topic) {
 
-    private def send(msg: GeoMessage): Unit = {
-      logger.debug("sending message: {}", msg)
-      producer.send(msgEncoder.encodeMessage(topic, msg))
-    }
+  private val ids = query.getFilter match {
+    case ids: Id => ids.getIDs.iterator
+    case _ => throw new NotImplementedError("Only modify by ID is supported")
```
Afterglow375: Why is only modify by ID supported?

Afterglow375:
Do these changes necessitate doc updates?

jnh5y:
@Afterglow375 nice review; go ahead and test as you see fit!

elahrvivaz:
I don't believe these need doc updates - we are basically conforming to the geotools API now, where before we weren't. But since we don't document the geotools API we shouldn't need to update anything.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
elahrvivaz: limitation of the design - when we 'modify' a feature, we're actually just sending it off on a kafka queue. With a normal feature writer, you can do a scan to return 'live' features that can then be modified - but the kafka producer feature store doesn't actually have access to any features. And the kafka consumer feature store, which does have access to the features, can't modify them.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
elahrvivaz: With append it's implied that you are creating new features. You still modify the feature coming back from the writer.next(), but it's a blank new feature vs an existing one. I guess the geotools API is ambiguous about what happens if you create a new feature with the same ID as an existing feature - in our case it would end up being a modify.

elahrvivaz: I think we consolidate createOrUpdate because we don't actually know if a message is an update or not, since the producer doesn't have access to any of the features after they're written.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
elahrvivaz: fixed

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
elahrvivaz: it's part of the content data store API - I believe it indicates whether this is a modify or append operation. But since we only support modify by ID, we ignore it. Technically if the query was 'INCLUDE' and the flag was modify, we should return all the existing features for modification. Since that's not possible due to the producer not having any features cached, we treat it the same as append.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/KafkaProducerFeatureStore.scala
jahhulbert-ccri: Yeah I'm not really sure...but I assumed it that ```getFeaturerWriterAppend``` was mostly an optimization because it wouldn't need to update anything features or read features it could simply write them which for some datastores would be a different access pattern or access different parts of the underlying store's code

jahhulbert-ccri:
@elahrvivaz i'll give this a test unless @Afterglow375 does it

jahhulbert-ccri:
so far seems to be working with kafka/geomesa nifi stuff

PR #974, opened 2016-07-06
GEOMESA-1297 Fix LaTeX build

mdzimmerman:
(Mostly) fixes LaTeX build of docs, which was not changed when we
reorganized how the different manuals were built with HTML.


cne1x:
There are some places where the LaTeX output is mangled -- overly large tables whose cell contents overlap; images whose dimensions in inches exceed their native resolution; etc. -- but it does render.

PR #976, opened 2016-07-07
GEOMESA-1301 Allow sft initialization in Spark without a context restart

atallahhezbor:

### DIFF
--- a/docs/user/spark.rst
+++ b/docs/user/spark.rst
@@ -48,8 +48,8 @@ types in each executor's registrator.
     // Broadcast sft encodings to executors
     val broadcastedSfts = sc.broadcast(sfts.map{sft => (name, SimpleFeatureTypes.encodeType(sft)})
     // Populate the type cache on each partition
-    someRdd.foreachPartition{ iter =>
-        broadcastedSfts.foreach{ case (name, spec) =>
+    someRdd.foreachPartition { iter =>
+        broadcastedSfts.foreach { case (name, spec) =>
             val sft = SimpleFeatureTypes.createType(name, spec)
             GeoMesaSparkKryoRegistrator.putType(sft)
         }
--- a/docs/user/spark.rst
+++ b/docs/user/spark.rst
@@ -1,0 +1,86 @@
+GeoMesa Spark
+=============
+
+GeoMesa Spark allows for execution of jobs on Apache Spark using data stored in GeoMesa. Through an API provided in the
+geomesa-compute module, we allow creation of Spark RDDs, writing of Spark RDDs to GeoMesa data stores, and serialization
+of simple features using Kryo.
+
+Usage
+-----
+
+To use GeoMesa with Spark, the executors must know how to serialize and deserialize Simple Features. There are two ways
+to accomplish this.
+
+Restart the Spark Context
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+One option is initialize a new Spark Context with the desired data store or SimpleFeatureType.
+This involves calling ``GeoMesaSpark.init``, which will take an existing Spark Configuration, and return a new one
+that is set to use our GeoMesaSparkKryoRegistrator to serialize Simple Features of the provided types. This return
+value must be used to initialize a new Spark Context as ``init`` will also set system properties for all executors
+such that they can serialize features of those types.
+Multiple data stores are able to be initialized by continually passing the resulting configuration into multiple calls
+to ``init`` and restarting the context using the final result.
+
+.. code-block:: scala
+
+    // Call initialize, retrieving a new spark config
+    val newConfig = GeoMesaSpark.init(new SparkConf(), dataStore)
+    // Stop all other running Spark Contexts
+    Option(sc).foreach(_.stop())
+    // Initialize a new one with the desired config
+    val sc = new SparkContext(newConfig)
+
+Broadcast Simple Feature Types
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Alternatively, if a restart of the Spark Context is undesirable, one is able to register classes directly into the Kryo
+Registrator. To do this, take advantage of the fact that basic Spark configuration values can be set in ``spark-defaults.conf``
+In this pattern, a call to ``GeoMesaSpark.register`` will register the Simple Feature Types of a provided data store,
+skipping the need use shared system properties that require a restart. The caveat, however, is that before serialization,
+the SimpleFeatureType encodings must be sent to the executors via a Spark Broadcast and then used to create the corresponding
+types in each executor's registrator.
+
+.. code-block:: scala
+
+    // Register the sfts of a given data store
+    GeoMesaSpark.register(dataStore)
+    // Broadcast sft encodings to executors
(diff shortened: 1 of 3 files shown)


geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -216,3 +215,15 @@ class GeoMesaSparkKryoRegistrator extends KryoRegistrator {
     SimpleFeatureSerializers.simpleFeatureImpls.foreach(kryo.register(_, serializer, kryo.getNextRegistrationId))
   }
 }
+
+object GeoMesaSparkKryoRegistrator {
+  val typeCache: ConcurrentHashMap[String, SimpleFeatureType] = new ConcurrentHashMap[String, SimpleFeatureType]()
+
+  def putType(sft: SimpleFeatureType): Unit = {
+    GeoMesaSparkKryoRegistrator.typeCache.putIfAbsent(sft.getTypeName, sft)
```
atallahhezbor: Wondering if there's a case where a type would need to be updated. I've noticed in Jupyter that static variables tend to hang around even after a kernel restart

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -186,27 +192,20 @@ class GeoMesaSparkKryoRegistrator extends KryoRegistrator {
 
   override def registerClasses(kryo: Kryo): Unit = {
     val serializer = new com.esotericsoftware.kryo.Serializer[SimpleFeature]() {
-      val typeCache = CacheBuilder.newBuilder().build(
-        new CacheLoader[String, SimpleFeatureType] {
-          override def load(key: String): SimpleFeatureType = {
-            val spec = System.getProperty(GeoMesaSpark.typeProp(key))
```
elahrvivaz: i think we still need this to pass sfts between the jvms in a spark cluster. the static hashmap below will only work inside a single jvm. the key piece is that the system props get passed around the cluster through the 'extraJavaOptions' config.

atallahhezbor: Right. The new pattern is to broadcast the sfts to the partitions as needed.

elahrvivaz: how do they get broadcast? I feel like I'm missing something...

atallahhezbor: Quite literally with `sc.broadcast(sfts)` where `sfts` is a map of type name to string encoding. Each partition then grabs this, creates the type, and puts it in their cache

atallahhezbor:
Is there a place I can document this new pattern? All I could find was `spark.rst` which is more of a tutorial

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -224,6 +224,12 @@ object GeoMesaSparkKryoRegistrator {
   }
 
   def getType(typeName: String): SimpleFeatureType = {
-    GeoMesaSparkKryoRegistrator.typeCache.get(typeName)
+    val spec = System.getProperty(GeoMesaSpark.typeProp(typeName))
+    if (spec == null) {
+      GeoMesaSparkKryoRegistrator.typeCache.get(typeName)
+    }
+    else {
```
atallahhezbor: Oops extra new line. I'll fix

elahrvivaz:
@atallahhezbor if you're up for it, maybe you could add a section to the docs for spark. you're right, I think right now there is just a tutorial.

docs/user/spark.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. code-block:: scala
+
+    // Register the sfts of a given data store
+    GeoMesaSpark.register(dataStore)
+    // Broadcast sft encodings to executors
+    val broadcastedSfts = sc.broadcast(sfts.map{sft => (name, SimpleFeatureTypes.encodeType(sft)})
+    // Populate the type cache on each partition
+    someRdd.foreachPartition{ iter =>
+        broadcastedSfts.foreach{ case (name, spec) =>
```
elahrvivaz: [picky] space before `{` (and on line above)

docs/user/spark.rst
```diff
@@ -0,0 +1,86 @@
+GeoMesa Spark
```
elahrvivaz: this looks really good, thanks for adding it.

mdzimmerman: Yeah, this is great! Make sure you add an entry to ``docs/user/index.rst`` so that this new chapter is included in the manual build.

elahrvivaz:
looks good - can be merged IMO

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
jahhulbert-ccri: is it using the same jvm?

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -66,6 +68,10 @@ object GeoMesaSpark extends LazyLogging {
 
   def jOpt(typeName: String, spec: String) = s"-D${typeProp(typeName)}=$spec"
 
+  def register(ds: DataStore): Unit = register(ds.getTypeNames.map(ds.getSchema))
+
+  def register(sfts: Seq[SimpleFeatureType]): Unit = sfts.foreach { GeoMesaSparkKryoRegistrator.putType }
```
jahhulbert-ccri: picky - i think we use ```foreach( )``` with parens instead of curlies for oneliners

jahhulbert-ccri:
@elahrvivaz i agree looks good. i guess i could try to test it on a cloud local with spark i'll see how hard that is

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
atallahhezbor: ```putType``` can be called on the jvm on the executors too (where the typecache is likely empty), but I've only noticed it happen on the driver's jvm. Though I may have just forgotten to restart the kernel. I'll see if I can reproduce it tomorrow. But either way, I feel like sfts are final enough to just be put once

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
jnh5y: NB: We don't seem to have a fixed style; there are plenty of curly oneliners.

PR #977, opened 2016-07-07
GEOMESA-1299 Better back-compatibility tests

elahrvivaz:


elahrvivaz:
question: do we care about things before 1.1.0-rc7?

jahhulbert-ccri:
good question...i think we may have schemas from rc7 and tables from rc6 that we need to support or else have a well tested transition to the new schemas using the copy job.

elahrvivaz:
@jahhulbert-ccri you mean 1.0.0-rc6 and 7? I'll add those.

jahhulbert-ccri:
i think 1.1.0-rc6 is as far back as we have supported or might support recently

elahrvivaz:
ok, cool - I'll just add rc6 then

elahrvivaz:
@jahhulbert-ccri added 1.1.0-rc.6

PR #986, opened 2016-07-11
GEOMESA-1306 Splitting stats table

elahrvivaz:
* Metadata is split into multiple rows instead of a single row
* First time a schema is read in the single-row format, it will be converted to multi-row


elahrvivaz:
Note: this will prevent rolling back a geomesa version update without some manual steps to restore the old metadata table.

jahhulbert-ccri:
@elahrvivaz we can always blow away the stats table and the command line tools can add them or adding it to the datastore config?

elahrvivaz:
@jahhulbert-ccri yeah - I wanted to make the update seamless from a user perspective, so that there aren't any manual steps. But that means that it re-writes the stats and the catalog metadata into the new format. So once you call `getSchema` with the new data store, older versions will no longer be able to read the metadata...
If you think that's a bad idea I can make it a manual step. Or else add some clear steps on rolling back (which would involve re-writing the table in the other direction).

elahrvivaz:
@jahhulbert-ccri In order for the data store to regenerate all the stats, you'd have to delete the 'stats-generated' key from the metadata table.

PR #993, opened 2016-07-14
GEOMESA-1315 - Attach the web-install-assembly to the profile's build

jnh5y:


jahhulbert-ccri:
just wondering what the motivation was for this (it's good i assume)

jnh5y:
Yeah, @ddseapy asked for the artifact to created internally, and without this change, I could get the maven deploy bits to pick up the assembly out as something with a GAV to be deployed.:)

I don't know if that qualifies as 'good' or not; just Maven config mojo.

PR #994, opened 2016-07-15
Complex avro

xandris:
Rather than try to integrate this into all the existing code myself, I created a separate module that can encode complex features. ComplexEncoderTest isn't a unit test, just a main method. I've just completed all the main functionality so I haven't had a chance to make actual unit tests. Forgive the bad jokes in the example data.

To use it, create a new `FeatureCodecs`; this is required to be mixed into a `CodecRegistry` and requires a FieldNameEncoder to be initialized early:

```
    val reg = new { val fne = FieldNameEncoderV1 } with StatefulCodecs(new WKBWriter(), new WKBReader()) with FeatureCodecs
```

You can request an Avro `Schema` from it given a `ComplexType`:

```
    val schema = reg.schemaFor(carType)
```

To actually use this on an Avro `Encoder`, request a codec from the registry:

```
    val codec = reg.find(car.getDescriptor)
```

And then call `.enc`:

```
      codec.enc(car, EncoderFactory.get().binaryEncoder(baos, avroenc))
```

There is a corresponding `.dec` method, of course.

The `FeatureCodecs` mixin attempts to bind an Avro `Schema` to a `ComplexType` once before any encoding or decoding is done; this way, the schemas need only be interrogated when they change and the binding can be reused. There isn't any cache management of these bindings yet; memory is leaked when the schemas change, which could be an issue if they change a lot.

The `*Codecs` classes contain mappings between Avro primitives and Java classes. They're used to determine how to generate the `Schema` and how to do the actual encoding and decoding. Multiple Java classes can be bound to the same Avro type, and an Avro type can be bound to many Java classes. When generating a schema, the first Avro type bound to a particular Java type is chosen for the binding. This means that if you define a codec for `Short` that binds to `Schema.Type.INT` but then later define a mapping between `Short` and `Schema.Type.LONG`, the schema generator will prefer `Schema.Type.INT`.

The codecs include support for strings, booleans, numeric types, byte arrays, byte buffers, dates, UUIDs, URIs, and geometries.

`Map` and `List` valued properties are supported as well. I'm not sure if I got the in-memory format right; the Key/Value or List element classes are stored on the `AttributeDescriptor` and retrieved using `RichAttributeDescriptor`. The codecs expect `java.util.List` and `java.util.Map`. Let me know if that's not right. They're encoded directly into Avro and so show up in the schema. The same codecs are used to encode list element and key/value pairs. Maps are encoded as an array of records with `k` and `v` fields.

Properties inherited from GML AbstractFeatureType are also supported, as well as extensions of anySimpleType with attributes (e.g. gml:description, which has a codespace and an href if you want to store the description elsewhere).

I tried to profile and improve performance where I could. On my laptop, 500K 'encode' operations followed by 500K 'decode' operations takes 15s, which isn't great but still acceptable. There are a number of un-Scala like bits of code in order to speed things up.

How can we integrate this? What's the next step?

Lastly, I am open to changing _anything_. I've been working on this pretty hard this week and my eyes are starting to go a little blurry, so chances are there are mistakes or I've misunderstood something.

### DIFF
--- a/geomesa-features/geomesa-feature-avro-complex/pom.xml
+++ b/geomesa-features/geomesa-feature-avro-complex/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>geomesa-features</artifactId>
         <groupId>org.locationtech.geomesa</groupId>
-        <version>1.2.2-SNAPSHOT</version>
+        <version>1.2.5-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/geomesa-features/geomesa-feature-avro-complex/pom.xml
+++ b/geomesa-features/geomesa-feature-avro-complex/pom.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>geomesa-features</artifactId>
+        <groupId>org.locationtech.geomesa</groupId>
+        <version>1.2.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>geomesa-feature-avro-complex</artifactId>
+
+
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-feature-avro</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-utils</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-main</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.specs2</groupId>
+            <artifactId>specs2_2.11</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/geomesa-features/geomesa-feature-avro-complex/pom.xml
+++ b/geomesa-features/geomesa-feature-avro-complex/pom.xml
@@ -31,6 +31,11 @@
             <artifactId>gt-main</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.geotools.xsd</groupId>
+            <artifactId>gt-xsd-gml3</artifactId>
+            <version>${gt.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.avro</groupId>
             <artifactId>avro</artifactId>
         </dependency>
--- a/geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
+++ b/geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
@@ -1,0 +1,697 @@
+package org.locationtech.geomesa.features.avro.complex
+
+import java.nio.ByteBuffer
+import java.sql.Timestamp
+import java.{net, sql, util}
+import java.util.function.Function
+import java.util.{Calendar, Date, GregorianCalendar, UUID, ArrayList => JArrayList, Collection => JCollection, Collections => JCollections, HashMap => JHashMap}
+
+import com.vividsolutions.jts.geom.Geometry
+import com.vividsolutions.jts.io.{WKBReader, WKBWriter}
+import org.apache.avro.Schema
+import org.apache.avro.io.{Decoder, Encoder}
+import org.geotools.feature.`type`.AttributeDescriptorImpl
+import org.geotools.feature.{AttributeImpl, ComplexAttributeImpl, FeatureImpl}
+import org.geotools.filter.identity.FeatureIdImpl
+import org.geotools.gml2.GMLSchema
+import org.locationtech.geomesa.features.avro.complex.AvroSchemas.{FieldNameEncoder, _}
+import org.locationtech.geomesa.utils.geotools.RichAttributeDescriptors.RichAttributeDescriptor
+import org.opengis.feature.`type`.{PropertyDescriptor, _}
+import org.opengis.feature.{Attribute, ComplexAttribute, Property}
+import org.opengis.filter.identity.FeatureId
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+object AvroCodecs {
+
+  val primitiveTypes =
+    List(
+      classOf[String],
+      classOf[java.lang.Integer],
+      classOf[Int],
+      classOf[java.lang.Long],
+      classOf[Long],
+      classOf[java.lang.Double],
+      classOf[Double],
+      classOf[java.lang.Float],
+      classOf[Float],
+      classOf[java.lang.Boolean],
+      classOf[Boolean]
+    )
+
+  abstract case class AvroCodec[T](schema:Schema) {
+    val clazz: Class[T]
+    def enc(t: T, enc: Encoder): Unit
+    def dec(dec: Decoder): T
+  }
+
+  trait CodecRegistry {
+    protected val registry = mutable.Map.empty[Class[_],mutable.LinkedHashMap[Schema.Type,AvroCodec[_]]]
+    protected val cache = mutable.Map.empty[Class[_], Option[mutable.LinkedHashMap[Schema.Type,AvroCodec[_]]]]
+
+    def bind[T](encMethod: (Encoder,T)=>Unit, decMethod: (Decoder)=>T, schema: Schema)(implicit tag: ClassTag[T]) = {
+      val codec = new AvroCodec[T](schema) {
+        val clazz = tag.runtimeClass.asInstanceOf[Class[T]]
+        override def enc(t: T, out: Encoder): Unit = encMethod(out, t)
+        override def dec(in: Decoder): T = decMethod(in)
+      }
+      registry.getOrElseUpdate(tag.runtimeClass,mutable.LinkedHashMap.empty)(schema.getType) = codec
+      codec
+    }
+
+    val JBOOLEAN = classOf[java.lang.Boolean]
+    val JBYTE = classOf[java.lang.Byte]
+    val JCHAR = classOf[java.lang.Character]
+    val JSHORT = classOf[java.lang.Short]
+    val JINT = classOf[java.lang.Integer]
+    val JLONG = classOf[java.lang.Long]
+    val JFLOAT = classOf[java.lang.Float]
+    val JDOUBLE = classOf[java.lang.Double]
+
+    def normalizePrimitive(clazz:Class[_]) = clazz match {
+      case JBOOLEAN | java.lang.Boolean.TYPE => classOf[Boolean]
+      case JBYTE | java.lang.Byte.TYPE => classOf[Byte]
+      case JCHAR | java.lang.Character.TYPE => classOf[Char]
+      case JSHORT | java.lang.Short.TYPE => classOf[Short]
+      case JINT | java.lang.Integer.TYPE => classOf[Int]
+      case JLONG | java.lang.Long.TYPE => classOf[Long]
+      case JFLOAT | java.lang.Float.TYPE => classOf[Float]
+      case JDOUBLE | java.lang.Double.TYPE => classOf[Double]
+      case x => x
+    }
+
+    def findAll[T](clazz:Class[T]): Option[mutable.LinkedHashMap[Schema.Type, AvroCodec[_]]] =
+      cache.getOrElseUpdate(clazz, registry.get(normalizePrimitive(clazz))
+        .orElse(Option(clazz.getSuperclass).flatMap(findAll(_)))
+        .orElse {
+          for (ifc <- clazz.getInterfaces; map <- findAll(ifc))
+            return Some(map)
+          None
+        }
+      )
+
+    def find[T](clazz:Class[T], s:Schema.Type): Option[AvroCodec[_>:T]] =
+      findAll(clazz).flatMap(m=>m.get(s)).asInstanceOf[Option[AvroCodec[_>:T]]]
+
+    def find[T](clazz:Class[T]): Option[AvroCodec[_>:T]] =
+      findAll(clazz).flatMap(m=>m.headOption).map{case(t,c)=>c}.asInstanceOf[Option[AvroCodec[_>:T]]]
+  }
+
+  trait PrimitiveCodecs extends CodecRegistry {
+    implicit val STRING = bind(_.writeString(_:String), _.readString(), Schema.create(Schema.Type.STRING))
+    implicit val BYTES = bind(_.writeBytes(_:scala.Array[Byte]), d=>{
+      val buf = d.readBytes(null)
+      val pos = buf.position()
+      val len = buf.limit() - pos
+      if(len == 0)
+        scala.Array.emptyByteArray
+      else if(pos == 0 && buf.hasArray)
+        buf.array()
+      else {
+        val a = scala.Array.ofDim[Byte](len)
+        buf.get(a, 0, len)
+        a
+      }
+    }, Schema.create(Schema.Type.BYTES))
+    implicit val BYTEBUFFER = bind(_.writeBytes(_:ByteBuffer), _.readBytes(null), Schema.create(Schema.Type.BYTES))
+    implicit val BOOLEAN = bind(_.writeBoolean(_:Boolean), _.readBoolean(), Schema.create(Schema.Type.BOOLEAN))
+
+    val BYTE_AS_INT = bind[Byte](_.writeInt(_), _.readInt().toByte, Schema.create(Schema.Type.INT))
+    val SHORT_AS_INT = bind[Short](_.writeInt(_), _.readInt().toShort, Schema.create(Schema.Type.INT))
+    implicit val INT = bind[Int](_.writeInt(_), _.readInt(),  Schema.create(Schema.Type.INT))
+    val LONG_AS_INT = bind[Long]((e,l)=>e.writeInt(l.toInt), _.readInt(),  Schema.create(Schema.Type.INT))
+    val FLOAT_AS_INT = bind[Float]((e,l)=>e.writeInt(l.toInt), _.readInt(),  Schema.create(Schema.Type.INT))
+    val DOUBLE_AS_INT = bind[Double]((e,l)=>e.writeInt(l.toInt), _.readInt(),  Schema.create(Schema.Type.INT))
+
+    val BYTE_AS_LONG = bind[Short](_.writeLong(_), _.readLong().toByte,  Schema.create(Schema.Type.LONG))
+    val SHORT_AS_LONG = bind[Short](_.writeLong(_), _.readLong().toShort,  Schema.create(Schema.Type.LONG))
+    val INT_AS_LONG = bind[Int](_.writeLong(_), _.readLong().toInt,  Schema.create(Schema.Type.LONG))
+    implicit val LONG = bind[Long](_.writeLong(_), _.readLong(),  Schema.create(Schema.Type.LONG))
+    val FLOAT_AS_LONG = bind[Float]((e,l)=>e.writeLong(l.toLong), _.readLong(),  Schema.create(Schema.Type.LONG))
+    val DOUBLE_AS_LONG = bind[Double]((e,l)=>e.writeLong(l.toLong), _.readLong(),  Schema.create(Schema.Type.LONG))
+
+    val BYTE_AS_FLOAT = bind[Byte](_.writeFloat(_), _.readFloat().toByte,  Schema.create(Schema.Type.FLOAT))
+    val SHORT_AS_FLOAT = bind[Short](_.writeFloat(_), _.readFloat().toShort,  Schema.create(Schema.Type.FLOAT))
+    val INT_AS_FLOAT = bind[Int](_.writeFloat(_), _.readFloat().toInt,  Schema.create(Schema.Type.FLOAT))
+    val LONG_AS_FLOAT = bind[Long](_.writeFloat(_), _.readFloat().toLong,  Schema.create(Schema.Type.FLOAT))
+    implicit val FLOAT = bind[Float](_.writeFloat(_), _.readFloat(),  Schema.create(Schema.Type.FLOAT))
+    val DOUBLE_AS_FLOAT = bind[Double]((e,f)=>e.writeFloat(f.toFloat), _.readFloat(),  Schema.create(Schema.Type.FLOAT))
+
+    val BYTE_AS_DOUBLE = bind[Byte](_.writeDouble(_), _.readDouble().toByte,  Schema.create(Schema.Type.DOUBLE))
+    val SHORT_AS_DOUBLE = bind[Short](_.writeDouble(_), _.readDouble().toShort,  Schema.create(Schema.Type.DOUBLE))
+    val INT_AS_DOUBLE = bind[Int](_.writeDouble(_), _.readDouble().toInt,  Schema.create(Schema.Type.DOUBLE))
+    val LONG_AS_DOUBLE = bind[Long](_.writeDouble(_), _.readDouble().toLong,  Schema.create(Schema.Type.DOUBLE))
+    val FLOAT_AS_DOUBLE = bind[Float](_.writeDouble(_), _.readDouble().toFloat,  Schema.create(Schema.Type.DOUBLE))
+    implicit val DOUBLE = bind[Double](_.writeDouble(_), _.readDouble(),  Schema.create(Schema.Type.DOUBLE))
+  }
+
+  trait DerivedCodecs extends PrimitiveCodecs {
+    def derive[T,U](encMethod: T=>U, decMethod: U=>T)(implicit primitiveCodec: AvroCodec[U], tag: ClassTag[T]) = {
+      val codec = new AvroCodec[T](primitiveCodec.schema) {
+        override val clazz = tag.runtimeClass.asInstanceOf[Class[T]]
+        override def enc(t: T, out: Encoder) = primitiveCodec.enc(encMethod(t), out)
+        override def dec(in: Decoder) = decMethod(primitiveCodec.dec(in))
+      }
+      registry.getOrElseUpdate(tag.runtimeClass, mutable.LinkedHashMap.empty)(primitiveCodec.schema.getType)=codec
+      codec
+    }
+
+    implicit val DATE = derive[Date,Long](_.getTime, new Date(_))
+    implicit val CALENDAR = derive[Calendar,Long](_.getTimeInMillis, d=>{val gc=new GregorianCalendar(); gc.setTimeInMillis(d); gc})
+    implicit val SQL_DATE = derive[sql.Date,Long](_.getTime, new sql.Date(_))
+    implicit val SQL_TIMESTAMP = derive[Timestamp,Long](_.getTime, new Timestamp(_))
+    implicit val _UUID = derive(encodeUUID, decodeUUID)
+    implicit val URI = derive[net.URI,String](_.toString, net.URI.create)
+  }
+
+  class StatefulCodecs(wkbWriter: WKBWriter, wkbReader: WKBReader) extends DerivedCodecs {
+    val GEOMETRY = derive[Geometry,scala.Array[Byte]](wkbWriter.write, wkbReader.read)
+  }
+
+  def encodeUUID(uuid: UUID) =
+    ByteBuffer.allocate(16)
+      .putLong(uuid.getMostSignificantBits)
+      .putLong(uuid.getLeastSignificantBits)
+      .flip.asInstanceOf[ByteBuffer]
+
+  def decodeUUID(bb: ByteBuffer): UUID = new UUID(bb.getLong, bb.getLong)
+
+
+  /*
+   * These classes are specializations of Function to improve performance
+   */
+
+  private[this] abstract class EncodeComplexAttribute extends ((ComplexAttribute,JHashMap[Name,JArrayList[Property]], Encoder)=>Unit) {
+    def apply(ca:ComplexAttribute, props: JHashMap[Name,JArrayList[Property]], e:Encoder)
+  }
+
+  private[this] abstract class EncodeProperties extends ((JArrayList[Property],Encoder)=>Unit) {
+    def apply(props:JArrayList[Property], e:Encoder)
+  }
+
+  private[this] abstract class EncodeProperty extends ((Encoder,Property)=>Unit) {
+    def apply(e: Encoder, p: Property)
+  }
+
+  private[this] abstract class DecodeToAttribute extends (Decoder=>Attribute) {
+    def apply(d:Decoder): Attribute
+  }
+
+  private[this] abstract class DecodeToAttributes extends (Decoder=>JCollection[Attribute]) {
+    def apply(d:Decoder): JCollection[Attribute]
+  }
+
+  private[this] val arrayFactory = new Function[Name,JArrayList[Property]] { def apply(t: Name) = new JArrayList[Property]() }
+
+  trait FeatureCodecs { self: CodecRegistry =>
+    protected[this] val fne: FieldNameEncoder
+    private[this] val schemaFac = new SchemaFactory(fne, this)
+    private[this] val codecs = mutable.Map.empty[AttributeDescriptor,AvroCodec[Property]]
+    private[this] val constructing = mutable.Set.empty[Record]
+    private[this] val encoders = mutable.Map.empty[Record,EncodeProperty]
+    private[this] val decoders = mutable.Map.empty[Record, DecodeToAttribute]
+
+    def schemaFor(ct:ComplexType) = schemaFac(ct)
+
+    def find(ct:ComplexType): AvroCodec[Property] = {
+      find(new AttributeDescriptorImpl(ct, ct.getName, 1, 1, false, null))
+    }
+
+    def find(ad:AttributeDescriptor): AvroCodec[Property] = {
+      codecs.getOrElseUpdate(ad, {
+        val schema = schemaFac(ad.getType.asInstanceOf[ComplexType])
+        val schemaType = getType(schema, fne)
+        val enc = bindEnc(schemaType, ad)
+        val dec = bindDec(schemaType, ad)
+        bind[Property](enc,dec,schema)
+      })
+    }
+
+    private[this] def bindEnc(s:SchemaType, ad:PropertyDescriptor):EncodeProperty = s match {
+      case r@Record(fields) if ad.getType.isInstanceOf[ComplexType] =>
+        val ct = ad.getType.asInstanceOf[ComplexType]
+      encoders.get(r) match {
+        case Some(enc) => return enc
+        case _ =>
+      }
+
+      if(constructing(r)) {
+        return new EncodeProperty {
+          private[this] lazy val binding=encoders(r)
+          override def apply(e: Encoder, p: Property): Unit = binding(e,p)
+        }
+      }
+
+      constructing(r) = true
+
+      val descByName = AvroSchemas.allProps(ct).map(d => d.getName -> d).toMap
+
+      // Each field MUST be mapped
+      val bindings = r.fields.map { case (name, st) =>
+        if (AvroSchemas.AVRO_NAMESPACE == name.getNamespaceURI) {
+          name.getLocalPart match {
+            case "__version__" =>
+              new EncodeComplexAttribute {
+                def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) = e.writeInt(1)
+              }
+
+            case "__fid__" =>
+              new EncodeComplexAttribute {
+                def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) =
+                  e.writeString(Option(ca.getIdentifier) flatMap (i => Option(i.getID)) map (_.toString) getOrElse "")
+              }
+
+            case "simpleContent" => st match {
+              case n@Nullable(Primitive(t)) =>
+                val binding = bindEncPrimitive(AvroSchemas.nearestSimpleType(ct), t)
+                new EncodeComplexAttribute {
+                  def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) =
+                    Option(ca.getProperty(GEOT_SIMPLE_CONTENT)) match {
+                      case Some(value) if value.getValue != null =>
+                        n.writeNonNull(e)
+                        binding(e, value)
+                      case _ =>
+                        n.writeNull(e)
+                    }
+                }
+
+              case _ =>
+                throw new IllegalStateException(s"The property $name must be a nullable primitive type")
+            }
+
+            case other => st match {
+              case n: Nullable =>
+                new EncodeComplexAttribute {
+                  def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) = n.writeNull(e)
+                }
+              case _ => ???
+            }
+          }
+        } else {
+          descByName.get(name) match {
+            case Some(desc) =>
+              val binding = bindEncMany(st, desc.asInstanceOf[AttributeDescriptor])
+              new EncodeComplexAttribute {
+                def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) =
+                  binding(p.computeIfAbsent(name, arrayFactory), e)
+              }
+            case None =>
+              st match {
+                case n: Nullable =>
+                  new EncodeComplexAttribute {
+                    def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) = n.writeNull(e)
+                  }
+                case _ =>
+                  throw new IllegalStateException()
+              }
+          }
+        }
+      }.toArray
+
+
+      val ret = new EncodeProperty {
+        def apply(e: Encoder, p: Property) = {
+          val ca = p.asInstanceOf[ComplexAttribute]
+          val props = new JHashMap[Name, JArrayList[Property]]()
+          val pIt = ca.getProperties.iterator()
+          while (pIt.hasNext) {
+            val p = pIt.next()
+            props.computeIfAbsent(p.getName, arrayFactory).add(p)
+          }
+          val sz = bindings.length
+          var i = 0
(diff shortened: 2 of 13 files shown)


geomesa-features/geomesa-feature-avro-complex/pom.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>geomesa-features</artifactId>
+        <groupId>org.locationtech.geomesa</groupId>
+        <version>1.2.2-SNAPSHOT</version>
```
cne1x: GeoMesa 1.2.4 was released last night.  Do we need this to be on the 1.2.2-SNAPSHOT, or can we upgrade?

geomesa-features/geomesa-feature-avro-complex/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>geomesa-feature-avro-complex</artifactId>
+
+
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-feature-avro</artifactId>
+            <version>${project.version}</version>
```
cne1x: I believe the GeoMesa convention has been to use the ```${geomesa.version}``` POM property in lieu of ```${project.version}```, not that it makes much difference.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,697 @@
...
+      case JCHAR | java.lang.Character.TYPE => classOf[Char]
+      case JSHORT | java.lang.Short.TYPE => classOf[Short]
+      case JINT | java.lang.Integer.TYPE => classOf[Int]
+      case JLONG | java.lang.Long.TYPE => classOf[Long]
+      case JFLOAT | java.lang.Float.TYPE => classOf[Float]
+      case JDOUBLE | java.lang.Double.TYPE => classOf[Double]
+      case x => x
+    }
+
+    def findAll[T](clazz:Class[T]): Option[mutable.LinkedHashMap[Schema.Type, AvroCodec[_]]] =
```
jnh5y: Admittedly this is my first read-through.  Could this just be a HashMap and the None case just have the map being empty?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,697 @@
...
+        .orElse(Option(clazz.getSuperclass).flatMap(findAll(_)))
+        .orElse {
+          for (ifc <- clazz.getInterfaces; map <- findAll(ifc))
+            return Some(map)
+          None
+        }
+      )
+
+    def find[T](clazz:Class[T], s:Schema.Type): Option[AvroCodec[_>:T]] =
+      findAll(clazz).flatMap(m=>m.get(s)).asInstanceOf[Option[AvroCodec[_>:T]]]
```
jnh5y: Ah, ok, that might be why... you are trying to expose AvroCodec if you find them...

geomesa-features/geomesa-feature-avro-complex/pom.xml
xandris: Oh! No, I just missed that when I was doing the rebase. I'll change it to match the parent

geomesa-features/geomesa-feature-avro-complex/pom.xml
xandris: I tried to change that just now and there didn't seem to be a property by that name? There `${geomesa.release.version}`, but that's at `1.2.4` and I don't have 1.2.4 installed :P

xandris:
I made those changes and pushed them to the PR

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+
+    implicit val DATE = derive[Date,Long](_.getTime, new Date(_))
+    implicit val CALENDAR = derive[Calendar,Long](_.getTimeInMillis, d=>{val gc=new GregorianCalendar(); gc.setTimeInMillis(d); gc})
+    implicit val SQL_DATE = derive[sql.Date,Long](_.getTime, new sql.Date(_))
+    implicit val SQL_TIMESTAMP = derive[Timestamp,Long](_.getTime, new Timestamp(_))
+    implicit val _UUID = derive(encodeUUID, decodeUUID)
+    implicit val URI = derive[net.URI,String](_.toString, net.URI.create)
+  }
+
+  class StatefulCodecs(wkbWriter: WKBWriter, wkbReader: WKBReader) extends DerivedCodecs {
```
jnh5y: Other places in the GeoMesa codebase, we use object pools or ThreadLocals to prevent issues with these classes not being thread-safe. 

If we try to address concurrency, that might be useful.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      codecs.getOrElseUpdate(ad, {
+        val schema = schemaFac(ad.getType.asInstanceOf[ComplexType])
+        val schemaType = getType(schema, fne)
+        val enc = bindEnc(schemaType, ad)
+        val dec = bindDec(schemaType, ad)
+        bind[Property](enc,dec,schema)
+      })
+    }
+
+    private[this] def bindEnc(s:SchemaType, ad:PropertyDescriptor):EncodeProperty = s match {
```
jnh5y: For readability, I'm tempted to suggest expanding this to bindEncoder

jnh5y: This method is way too long...  let's see what we can come up...

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      case r@Record(fields) if ad.getType.isInstanceOf[ComplexType] =>
+        val ct = ad.getType.asInstanceOf[ComplexType]
+      encoders.get(r) match {
+        case Some(enc) => return enc
+        case _ =>
```
jnh5y: If I'm reading correctly, the '_' case is the rest of the method.  Perhaps extract it as a buildEncoder/computeEncoder method?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+        val schema = schemaFac(ad.getType.asInstanceOf[ComplexType])
+        val schemaType = getType(schema, fne)
+        val enc = bindEnc(schemaType, ad)
+        val dec = bindDec(schemaType, ad)
+        bind[Property](enc,dec,schema)
+      })
+    }
+
+    private[this] def bindEnc(s:SchemaType, ad:PropertyDescriptor):EncodeProperty = s match {
+      case r@Record(fields) if ad.getType.isInstanceOf[ComplexType] =>
```
jnh5y: I think some of the whitespace here is tripping me up. 

What happens if ad.getType is not ComplexType?

jnh5y: Maybe avoid the ```isInstanceOf``` with something like
```
(s, ad.getType) match {
   case (r@Records(fields), _: ComplexType) => 
}
```

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+              case _ =>
+                throw new IllegalStateException(s"The property $name must be a nullable primitive type")
+            }
+
+            case other => st match {
+              case n: Nullable =>
+                new EncodeComplexAttribute {
+                  def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) = n.writeNull(e)
+                }
+              case _ => ???
```
jnh5y: Ok, how can one get here?  

Can throw an exception which might be helpful to the user?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+                  binding(p.computeIfAbsent(name, arrayFactory), e)
+              }
+            case None =>
+              st match {
+                case n: Nullable =>
+                  new EncodeComplexAttribute {
+                    def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) = n.writeNull(e)
+                  }
+                case _ =>
+                  throw new IllegalStateException()
```
jnh5y: What's the Illegal State?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+          val sz = bindings.length
+          var i = 0
+          while (i < sz) {
+            bindings(i)(ca, props, e)
+            i = i + 1
+          }
+        }
+      }
+      encoders(r) = ret
+      constructing(r) = false
```
jnh5y: Are you using constructing as manage concurrency? 

If so, it seems like two calls to this method could overlap so that this flag has just been set to false again as the second is starting to construct/build the encoder.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      }
+
+    private[this] def bindEncCodec[T](codec:AvroCodec[T]): EncodeProperty =
+      new EncodeProperty { override def apply(e: Encoder, p: Property) = codec.enc(p.getValue.asInstanceOf[T],e) }
+
+
+    private[this] def bindDec(s:SchemaType, ad:PropertyDescriptor): DecodeToAttribute = s match {
+      case r@Record(fields) =>
+        val ct=ad.getType.asInstanceOf[ComplexType]
+        decoders.get(r) match { case Some(dec) => return dec case _ => }
```
jnh5y: Same deal here obviously:)  This method should be broken up.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+              i = i + 1
+            }
+            b.build()
+          }
+        }
+        decoders(r)=ret
+        constructing(r)=false
+        ret
+      case _ =>
+        ???
```
jnh5y: Under what conditions can we get here?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+    }
+
+    private[this] def bindDecMany(s: SchemaType, ad:AttributeDescriptor): DecodeToAttributes = s match {
+      case n@Nullable(st) =>
+        new DecodeToAttributes {
```
jnh5y: Consider extracting functions to create each of these.  I think it'd improve readability.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      case r:Record =>
+        bindDec(s,ad)
+
+      case Primitive(t) =>
+        self.find(ad.getType.getBinding, t) match {
+          case Some(codec) =>
+            new DecodeToAttribute {
+              def apply(d: Decoder) = new AttributeImpl(codec.dec(d), ad, null)
+            }
+          case None => ???
```
jnh5y: Ditto.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      // Each field MUST be mapped
+      val bindings = r.fields.map { case (name, st) =>
+        if (AvroSchemas.AVRO_NAMESPACE == name.getNamespaceURI) {
+          name.getLocalPart match {
+            case "__version__" =>
+              new EncodeComplexAttribute {
+                def apply(ca: ComplexAttribute, p: JHashMap[Name, JArrayList[Property]], e: Encoder) = e.writeInt(1)
+              }
+
+            case "__fid__" =>
```
jnh5y: Is it worth using FEATURE_ID_AVRO_FIELD_NAME since it's defined?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+          lazy val fieldSkipper = new Applier {
+            def apply(d:Decoder,props:JArrayList[Property], b:Builder)=st.skip(d)
+          }
+
+          if(AVRO_NAMESPACE == name.getNamespaceURI) name.getLocalPart match {
+            case "__version__" =>
+              new Applier {
+                def apply(d:Decoder,props:JArrayList[Property],b:Builder)=d.readInt()
+              }
+            case "__fid__" =>
```
jnh5y: Same with FEATURE_ID_AVRO_FIELD_NAME.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+          def apply(d:Decoder, props:JArrayList[Property], b:Builder): Unit
+        }
+
+        val bindings:scala.Array[Applier] = (for((name,st)<-r.fields) yield {
+          lazy val fieldSkipper = new Applier {
+            def apply(d:Decoder,props:JArrayList[Property], b:Builder)=st.skip(d)
+          }
+
+          if(AVRO_NAMESPACE == name.getNamespaceURI) name.getLocalPart match {
+            case "__version__" =>
```
jnh5y: Ditto for __version__ and its static val friend.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
```diff
@@ -0,0 +1,253 @@
...
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+import scala.collection.immutable.{BitSet, Queue}
+import scala.collection.mutable
+
+object AvroSchemas {
+  val FEATURE_ID_AVRO_FIELD_NAME = "__fid__"
+  val AVRO_SIMPLE_FEATURE_VERSION = "__version__"
+  val AVRO_SIMPLE_FEATURE_USERDATA = "__userdata__"
+  val VERSION = 4
```
jnh5y: Ok, just for my own education and clarity.  Who's version is version 4 here?

Avro?  GeoMesa's Avro schema?  OGC's ComplexFeature?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
```diff
@@ -0,0 +1,253 @@
...
+  val GEOT_SIMPLE_CONTENT = new NameImpl("simpleContent")
+  val USERDATAITEM_SCHEMA = SchemaBuilder.record("userDataItem")
+    .namespace(AVRO_NAMESPACE)
+    .fields()
+    .name("class").`type`.stringType().noDefault()
+    .name("key").`type`.stringType().noDefault()
+    .name("value").`type`.stringType().noDefault().endRecord()
+
+  trait FieldNameEncoder {
+    def encode(name:String): String
```
jnh5y: (style)  Unfortunately we haven't documented the GeoMesa code style documented (we do vaguely point at http://docs.scala-lang.org/style).  Here are a few things which would be nice to address:

1.  There's usually one space after colons and commas (i.e., ",", ":").
2.  There's usually a space on both sides of a production ("=>").
3.  For control structures, we almost never one-line them.  Despite the notes here (http://docs.scala-lang.org/style/control-structures), we tend to use {} for each branch.  It is verbose, but you don't get silly situations like this:

```
if (control);
foo
```
What does the above do:)?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
```diff
@@ -0,0 +1,253 @@
...
+    }
+    def decode(f:Field, parent:Schema): Name = {
+      val ns = Option(f.getProp("namespace")).getOrElse(parent.getNamespace)
+      new NameImpl(decode(ns), decode(f.name))
+    }
+  }
+
+  object FieldNameEncoderV1 extends FieldNameEncoder{
+    private val radix = "0123456789ABCDEF"
+    private val allowedCharsInit:BitSet = ('A'.toInt.to('Z')++'a'.toInt.to('z')).map(identity)(scala.collection.breakOut)
```
jnh5y: Also, since we review code on GitHub, we tend to wrap before 100 (or maybe 120) characters so that lines are completely visible on the screen.

This is despite everyone having 24+ inch monitors;).

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
```diff
@@ -0,0 +1,253 @@
...
+    private val allowedChars:BitSet = allowedCharsInit ++ '0'.toInt.to('9')
+
+    private def encodeCharInit(b:Byte) = if(allowedCharsInit(b)) b.toChar.toString else "_"+radix((b>>4)&0xf)+radix(b&0xf)
+    private def encodeChar(b:Byte) = if(allowedChars(b)) b.toChar.toString else "_"+radix((b>>4)&0xf)+radix(b&0xf)
+    private def fromHex(c:Char):Int = if(c>='0' && c<='9') c-'0' else if(c>='A' && c<='F') c-'A'+10 else if(c>='a' && c<='f') c-'a'+10 else throw new IllegalArgumentException()
```
jnh5y: (style) Wrap the long lines.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
```diff
@@ -0,0 +1,253 @@
...
+      if(nillable) Schema.createUnion(List(Schema.create(Schema.Type.NULL), maybeMulti)) else maybeMulti
+    }
+
+    private[this] def computeFields[R](parent: ComplexType): Iterable[Field] = {
+      allProps(parent).map(_.asInstanceOf[AttributeDescriptor]).flatMap { ad=>
+        ad.getType match {
+          case ct: ComplexType =>
+            Some(computeField(parent, ad.getName, computeType(this(ct), ad)))
+          case _ if ad.isList =>
+            for(clazz<-ad.getListType(); codec<-registry.find(clazz)) yield
```
jnh5y: Ah!  We also leave a space on either side of a '<-'.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/FeatureTypeBuilder.scala
```diff
@@ -0,0 +1,119 @@
...
+  def makeDescriptor(t:ComplexType): AttributeDescriptor = new AttributeDescriptorImpl(t, t.getName, 1, 1, false, null)
+
+  def isDescendedFrom(child: PropertyType, ancestor: PropertyType): Boolean =
+    Iterator.iterate(child)(_.getSuper).takeWhile(_ != null).contains(ancestor)
+
+  def subclassByRestriction(child: PropertyType) =
+    isDescendedFrom(child, GMLSchema.FEATUREPROPERTYTYPE_TYPE)
+
+  def allProps(ct:ComplexType):Iterable[PropertyDescriptor] = ct.getSuper match {
+    case st:ComplexType if subclassByRestriction(st) => allProps(st).drop(ct.getDescriptors.size())++ct.getDescriptors
```
jnh5y: It seems like the first several descriptors are replaced here.

Is it clear why the ones to replace are up first?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/FeatureTypeBuilder.scala
```diff
@@ -0,0 +1,119 @@
...
+      this
+    }
+
+    /*def set(name:String): SubFeatureBuilder[self.type] = {
+      val desc=byName(name)
+      if(!desc.getType.isInstanceOf[ComplexType]) {
+        throw new IllegalArgumentException(s"Not a complex type: ${desc.getType.getName}")
+      }
+      new SubFeatureBuilder[FeatureBuilder.this.type](desc, this)
+    }*/
```
jnh5y: Removed commented code?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/FeatureTypeBuilder.scala
```diff
@@ -0,0 +1,119 @@
...
+    protected[builder] def internalSet(desc:AttributeDescriptor, prop:Property): self.type = {
+      if(desc.getMaxOccurs != 1) {
+        props.getOrElseUpdate(desc, mutable.Buffer.empty).append(prop)
+      } else {
+        props(desc)=mutable.Buffer(prop)
+      }
+      this
+    }
+
+    def end: ComplexAttribute =
```
jnh5y: Is it more typical to call the builder method 'build'?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/SubFeatureTypeBuilder.scala
```diff
@@ -0,0 +1,15 @@
+package org.locationtech.geomesa.utils.geotools.builder
+import org.opengis.feature.`type`.{ComplexType,Name}
+
+class SubFeatureTypeBuilder[ReturnType](val target: DescriptorBuilder[ReturnType]) extends AbstractFeatureTypeBuilder[DescriptorBuilder[ReturnType]] {
+  override def namespace(ns: String) = super.namespace(ns)
+
+  def end = {
+    target.finish(makeType)
```
jnh5y: For all these builders, I'd standardize on 'build'.  Unless there's a reason not to...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/TopLevelDescriptorBuilder.scala
```diff
@@ -0,0 +1,24 @@
...
+  def name(n: String): DescriptorBuilder[_<:AttributeDescriptor] = {
+    _name = new NameImpl(targetNs, n)
+    `type`(n + "Type")
+  }
+
+  def end = new AttributeDescriptorImpl(makeType, _name, 1, 1, false, null)
+
+  val typeMap: scala.collection.mutable.Map[Name,ComplexType] = mutable.Map.empty
+
+  override def saveType[T <: ComplexType](t: T) = {typeMap(t.getName)=t;t}
```
jnh5y: (style) My preference would be to call this 'putType'.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/TopLevelDescriptorBuilder.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  override def getType(name:Name) = typeMap.get(name)
+
+  override def getType(name:Name, computeFn: =>ComplexType) = typeMap.getOrElseUpdate(name, computeFn)
```
jnh5y: I'd suggest renaming this getTypeOrUpdate.

As a sidebar, we went through a similar naming debate regarding a map of feature types elsewhere.  The suggestion from that was to reflect the underlying methods/operations as best as possible in the function names.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/AbstractFeatureTypeBuilder.scala
```diff
@@ -0,0 +1,56 @@
...
+  final val descriptors: mutable.Buffer[PropertyDescriptor] = mutable.Buffer.empty
+
+  def namespace(ns: String): this.type = {
+    targetNs = ns
+    this
+  }
+
+  def `type`(name: String): DescriptorBuilder[_<:ReturnType] = `type`(name, GMLSchema.ABSTRACTFEATURETYPE_TYPE)
+
+  def `type`(name: String, superType: ComplexType): DescriptorBuilder[ReturnType] = {
```
jnh5y: Hmm...  are there any suggestions/possibilities for getting around naming the function `type`.  

'typename' seems to fit for the first use, but it seems sloppy for the second.

jnh5y:
How are you running this code?  

I was to get it going with
```
mvn exec:java -Dexec.mainClass=org.locationtech.geomesa.features.avro.complex.ComplexEncoderTest -Dexec.classpathScope=test
```
(I'm asking and mentioning so that other reviewers know how to run the sample test.)

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
+package org.locationtech.geomesa.features.avro.complex
+
+import java.nio.ByteBuffer
+import java.sql.Timestamp
+import java.{net, sql, util}
+import java.util.function.Function
+import java.util.{Calendar, Date, GregorianCalendar, UUID, ArrayList => JArrayList, Collection => JCollection, Collections => JCollections, HashMap => JHashMap}
```
jnh5y: Since you mentioned performance, I loaded up the test code in YourKit.  About 10% of the time was spent in ArrayList.addAll.

I'm still trying to understand things.  As an optimization suggestion, I think it'd make sense for the interface to specify the java.util.List interface rather than the implementation.  

Internally, one could use (Scala) collections which would not require array creation/copies.  At the interface with the client, that collection could be wrapped by a JavaConverter or array-copied one time.

Thoughts?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
```diff
@@ -0,0 +1,253 @@
...
+  final case class Nullable(schema: SchemaType) extends SchemaType {
+    def this(schema: SchemaType, nullidx:Int) = { this(schema); this.nullidx = nullidx; }
+    private[this] var nullidx: Int = 0
+    def writeNull(e:Encoder) = { e.writeIndex(nullidx); e.writeNull(); }
+    def writeNonNull(e:Encoder) = e.writeIndex(1 - nullidx)
+    def readNull(d:Decoder) = if(d.readIndex() == nullidx) { d.readNull(); true; } else false
+    def skip(d:Decoder) = if(!readNull(d)) schema.skip(d)
+  }
+
+  case class Array(schema: SchemaType) extends SchemaType {
```
jnh5y: For the implementations of SchemaType, I'd ask for longer class names.

In particular, 'Array' is a no-go as a class name;).

geomesa-features/geomesa-feature-avro-complex/pom.xml
jahhulbert-ccri: actually, ideally we'd move this up to the root parent pom to have it managed there like other submodules and go with project.version to track the specific maven version https://github.com/locationtech/geomesa/blob/master/pom.xml#L1463 then drop the version from the entry here

geomesa-features/geomesa-feature-avro-complex/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+            <artifactId>gt-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-main</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools.xsd</groupId>
+            <artifactId>gt-xsd-gml3</artifactId>
+            <version>${gt.version}</version>
```
jahhulbert-ccri: ideally move version to be managed by parent pom as well like entries above

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroSchemas.scala
jahhulbert-ccri: +1 yeah throw another word on there for the class name

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/FeatureTypeBuilder.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.geotools.feature.{AttributeImpl, ComplexAttributeImpl, FeatureImpl, NameImpl}
+import org.geotools.filter.identity.FeatureIdImpl
+import org.geotools.gml3.GMLSchema
+import org.opengis.feature.`type`._
+import org.opengis.feature.{ComplexAttribute, Feature, Property}
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable
+
+object FeatureTypeBuilder {
```
jahhulbert-ccri: is this only for complex types? we have the SftBuilder for simple types so i assume it is...if so lets name it ComplexTypeBuilder or something like that

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/builder/SubFeatureTypeBuilder.scala
jahhulbert-ccri: +1

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+import org.opengis.feature.`type`.{PropertyDescriptor, _}
+import org.opengis.feature.{Attribute, ComplexAttribute, Property}
+import org.opengis.filter.identity.FeatureId
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+object AvroCodecs {
+
+  val primitiveTypes =
```
jahhulbert-ccri: This bit of code closely mimics https://github.com/locationtech/geomesa/blob/master/geomesa-features/geomesa-feature-avro/src/main/scala/org/locationtech/geomesa/features/avro/AvroSimpleFeatureUtils.scala#L90 and I think ideally within geomesa a lot of this code should be shared with the simple feature serialization?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      classOf[java.lang.Float],
+      classOf[Float],
+      classOf[java.lang.Boolean],
+      classOf[Boolean]
+    )
+
+  abstract case class AvroCodec[T](schema:Schema) {
+    val clazz: Class[T]
+    def enc(t: T, enc: Encoder): Unit
+    def dec(dec: Decoder): T
```
jahhulbert-ccri: prefer full name of encode and decode since I get mixed up with "decrement"

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+        }
+      }.toArray
+
+
+      val ret = new EncodeProperty {
+        def apply(e: Encoder, p: Property) = {
+          val ca = p.asInstanceOf[ComplexAttribute]
+          val props = new JHashMap[Name, JArrayList[Property]]()
+          val pIt = ca.getProperties.iterator()
+          while (pIt.hasNext) {
```
jahhulbert-ccri: I assume that this isn't called often? like not in an inner loop when serializing...if so prefer rich iterator and use a foreach instead of while loop

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+            val p = pIt.next()
+            props.computeIfAbsent(p.getName, arrayFactory).add(p)
+          }
+          val sz = bindings.length
+          var i = 0
```
jahhulbert-ccri: similar to above...perhaps a zipWithIndex if its not called often for scala-style iteratation

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
jahhulbert-ccri: yeah you'd need a reentrant lock here but i'd rather do something else like a synchronized block

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+    )
+
+  abstract case class AvroCodec[T](schema:Schema) {
+    val clazz: Class[T]
+    def enc(t: T, enc: Encoder): Unit
+    def dec(dec: Decoder): T
+  }
+
+  trait CodecRegistry {
+    protected val registry = mutable.Map.empty[Class[_],mutable.LinkedHashMap[Schema.Type,AvroCodec[_]]]
```
jahhulbert-ccri: question on thread safety here on the mutable map...i may see it later but can this be accessed from multiple threads? if so we need a concurrent hash map or similar mixin

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      JDOUBLE -> classOf[Double],
+      java.lang.Boolean.TYPE -> classOf[Boolean],
+      java.lang.Byte.TYPE -> classOf[Byte],
+      java.lang.Character.TYPE -> classOf[Char],
+      java.lang.Short.TYPE -> classOf[Short],
+      java.lang.Integer.TYPE -> classOf[Int],
+      java.lang.Long.TYPE -> classOf[Long],
+      java.lang.Float.TYPE -> classOf[Float],
+      java.lang.Double.TYPE -> classOf[Double]
+    ).withDefault(k=>k)
```
jahhulbert-ccri: what types would use the withDefault?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+  trait PrimitiveCodecs extends CodecRegistry {
+    implicit val STRING = bind(_.writeString(_:String), _.readString(), Schema.create(Schema.Type.STRING))
+    implicit val BYTES = bind(_.writeBytes(_:scala.Array[Byte]), d=>{
+      val buf = d.readBytes(null)
+      val pos = buf.position()
+      val len = buf.limit() - pos
+      if(len == 0)
+        scala.Array.emptyByteArray
+      else if(pos == 0 && buf.hasArray)
+        buf.array()
```
jahhulbert-ccri: I need to think if this is a safe ByteBuffer thing to do before merge.

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+    implicit val SQL_TIMESTAMP = derive[Timestamp,Long](_.getTime, new Timestamp(_))
+    implicit val _UUID = derive(encodeUUID, decodeUUID)
+    implicit val URI = derive[net.URI,String](_.toString, net.URI.create)
+  }
+
+  class StatefulCodecs(wkbWriter: WKBWriter, wkbReader: WKBReader) extends DerivedCodecs {
+    val GEOMETRY = derive[Geometry,scala.Array[Byte]](wkbWriter.write, wkbReader.read)
+  }
+
+  def encodeUUID(uuid: UUID) =
```
jahhulbert-ccri: this should be shared with the simple feature code

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
```diff
@@ -0,0 +1,704 @@
...
+      constructing(r) = false
+      ret
+    }
+
+    private[this] def bindEncMany(s:SchemaType, ad:AttributeDescriptor):EncodeProperties = s match {
+      case n@Nullable(st) =>
+        val binding = bindEncMany(st,ad)
+        new EncodeProperties { def apply(props: JArrayList[Property], e: Encoder) = {
+          // We want to be as gentle on the GC as possible, so don't allocate a list unless absolutely necessary
+          var out: JArrayList[Property] = props
```
jahhulbert-ccri: simplify using foreach hopefully make this method smaller

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
jahhulbert-ccri: yeah...possibly share with the simple feature class too so its common

jahhulbert-ccri:
Hi @xandris thanks for the PR! I started looking through things...it might take me a little while since there's a lot of code here and I haven't worked all that much with complex features. But I'll try to look over it as soon as possible to get some feedback to you.

A few top level notes/questions...

1. Did you intended to separate it completely from the simple feature avro code? Seems there may be some overlap there with encoding and schema making. I think ideally low level things like serializing/writing things would be nice to keep together but it might be more intricate to merge.

2. There seem to be some while loops and I want to just make sure they are inner loop type code instead of less intense blocks. If they are less rarely called then I'd prefer the foreach and iteration in the functional scala style to cut down on the length of methods.

3. Concurrency - we may need to do a little thought to make sure the registry, etc is threadsafe and if so mark it with an ```@ThreadSafe``` annotation

4. Jim mentioned some style differences from the geomesa codebase at large and a few longer method blocks that perhaps should be broken up and the other match statement exceptional cases

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
jnh5y: And 'enc' could be 'enchant':)?

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
jnh5y: zipWithIndex shouldn't be in an inner loop; only use infrequently;)

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
xandris: I was imagining pushing the concurrency requirements to the code using this API, but I can think on it some more

geomesa-features/geomesa-feature-avro-complex/src/main/scala/org/locationtech/geomesa/features/avro/complex/AvroCodecs.scala
xandris: Only to manage binding of cyclic types, not concurrent calls to these methods on the same instance

PR #995, opened 2016-07-15
GEOMESA-1163 Fixing decomposition of polygons into z-index

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/Z2Table.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/Z2Table.scala
@@ -22,7 +22,6 @@ import org.locationtech.geomesa.accumulo.data.AccumuloFeatureWriter.FeatureToMut
 import org.locationtech.geomesa.accumulo.data._
 import org.locationtech.geomesa.curve.Z2SFC
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
-import org.locationtech.sfcurve.zorder.{Z2, ZPrefix}
 import org.opengis.feature.simple.SimpleFeatureType
 
 object Z2Table extends GeoMesaTable {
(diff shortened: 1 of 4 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/Z2Table.scala
```diff
@@ -141,43 +141,28 @@ object Z2Table extends GeoMesaTable {
...
-      if (zbits < GEOM_Z_NUM_BYTES * 8) {
-        // divide the range into two smaller ones using tropf litmax/bigmin
-        val (litmax, bigmin) = Z2.zdivide(Z2((min + max) / 2), Z2(min), Z2(max))
-        in.enqueue((min, litmax.z), (bigmin.z, max))
+  // gets z values that cover the bounding box
+  private def getZPrefixes(xmin: Double, ymin: Double, xmax: Double, ymax: Double): Set[Long] = {
+    Z2SFC.ranges((xmin, xmax), (ymin, ymax), 8 * GEOM_Z_NUM_BYTES).flatMap { range =>
+      val lower = range.lower & GEOM_Z_MASK
+      val upper = range.upper & GEOM_Z_MASK
+      if (lower == upper) {
```
cne1x: Is it ever possible for ```(upper - lower) < GEOM_Z_STEP```?

elahrvivaz: hmm, I need to think about that a little... I was thinking that they would map to the same value and be covered by the `:+ upper` on line 163, but maybe it's possible they wouldn't. let me look into it.

elahrvivaz: ok, it shouldn't be possible due to the `GEOM_Z_MASK`. They will either be equal, or be at least `GEOM_Z_STEP` apart.

elahrvivaz: `
scala> val GEOM_Z_NUM_BYTES = 3
GEOM_Z_NUM_BYTES: Int = 3

scala>  val GEOM_Z_MASK: Long = Long.MaxValue << (64 - 8 * GEOM_Z_NUM_BYTES)
GEOM_Z_MASK: Long = -1099511627776

scala> val GEOM_Z_STEP: Long = 1L << (64 - 8 * GEOM_Z_NUM_BYTES)
GEOM_Z_STEP: Long = 1099511627776

scala> val lower = 1099511627775L & GEOM_Z_MASK
lower: Long = 0

scala> val upper = 1099511627777L & GEOM_Z_MASK
upper: Long = 1099511627776

scala> upper - lower
res0: Long = 1099511627776

scala> (upper - lower) / GEOM_Z_STEP
res1: Long = 1
`

cne1x:
Looks good to me.

elahrvivaz:
@jnh5y once we merge the xz index, this might not kill the travis build, since the z2 poly index won't be hit in most cases - if so, it would be good to get it in 1.2.5.

PR #1000, opened 2016-07-18
GEOMESA-1323 Example attr level vis ingest

jahhulbert-ccri:

### DIFF
--- a/geomesa-tools/conf/application.conf
+++ b/geomesa-tools/conf/application.conf
@@ -11,7 +11,20 @@ geomesa {
         { name = "geom",     type = "Point",           index = true, srid = 4326, default = true }
       ]
     }
-    "example-xml" = {
+    "example-csv-multivis" = {
+      attributes = [
+        { name = "name",     type = "String",          index = true  }
+        { name = "age",      type = "Integer",         index = false }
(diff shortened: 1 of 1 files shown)


geomesa-tools/conf/application.conf
```diff
@@ -78,7 +91,29 @@ geomesa {
...
-    "example-xml" = {
+    "example-csv-per-attr-visibilities" = {
+      type   = "delimited-text"
+      format = "CSV"
+      options {
+        skip-lines = 1
+      }
+      id-field = "toString($fid)",
+      user-data = {
+        geomesa.feature.visibility = "'owl,owl,owl,newt,newt,owl'"
```
elahrvivaz: are the single quotes required and/or problematic? we should probably update the docs examples if you actually need them.

jahhulbert-ccri: required...unfortunately...to do a string constant in this case it looks like

elahrvivaz: btw, I already added a (non-attribute-level) vis example here: https://github.com/locationtech/geomesa/blob/master/geomesa-tools/conf/application.conf#L58
I guess we have multiple places for stuff...

PR #1007, opened 2016-07-25
GEOMESA-1318 Added geomesa distributed runtime install script

Afterglow375:

### DIFF
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -150,6 +150,20 @@ Optional parameters:
      -g (Path of GeoMesa distributed runtime JAR)
      -h (HDFS URI e.g. hdfs://localhost:54310)
 
+Alternatively you could manually install the GeoMesa distributed runtime JAR with these commands:
+
+.. code::
+    $ accumulo shell -u root
+    > createnamespace myNamespace
+    > grant NameSpace.CREATE_TABLE -ns myNamespace -u myUser
+    > config -s general.vfs.context.classpath.myNamespace=hdfs://NAME_NODE_FDQN:54310/accumulo/classpath/myNamespace/[^.].*.jar
+    > config -ns myNamespace -s table.classpath.context=myNamespace
+
+Then copy the distributed runtime jar into HDFS under the path you specified.
+The path above is just an example; you can included nested folders with project
+names, version numbers, and other information in order to have different versions of GeoMesa on
+the same Accumulo instance.
+
 .. note::
 
     When connecting to a data store using Accumulo namespaces, you must prefix
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -138,7 +138,7 @@ Execute the ``install-geomesa-namespace.sh`` script in the ``geomesa-$VERSION/di
 
 .. code::
 
-    $ ./install-geomesa-namespace.sh -u root -n myNamespace
+    $ ./install-geomesa-namespace.sh -u myUser -n myNamespace
 
 The command line arguments the script accepts are:
 
@@ -162,7 +162,8 @@ Alternatively you could manually install the GeoMesa distributed runtime JAR wit
 Then copy the distributed runtime jar into HDFS under the path you specified.
 The path above is just an example; you can included nested folders with project
 names, version numbers, and other information in order to have different versions of GeoMesa on
-the same Accumulo instance.
+the same Accumulo instance. You should remove any GeoMesa JARs under
+``$ACCUMULO_HOME/lib/ext`` to prevent any classpath conflicts.
 
 .. note::
 
--- a/docs/user/installation_and_configuration.rst
+++ b/docs/user/installation_and_configuration.rst
@@ -132,26 +132,23 @@ For Accumulo 1.6+
 
 Copying the runtime JAR to each tablet server as for Accumulo 1.5 above will
 still work, but in Accumulo 1.6, we can leverage namespaces to isolate the
-GeoMesa classpath from the rest of Accumulo. First, you have to create the
-namespace in the Accumulo shell:
+GeoMesa classpath from the rest of Accumulo.
+
+Execute the ``install-geomesa-namespace.sh`` script in the ``geomesa-$VERSION/dist/accumulo`` directory.
 
 .. code::
 
-    $ accumulo shell -u root
-    > createnamespace myNamespace
-    > grant NameSpace.CREATE_TABLE -ns myNamespace -u myUser
-    > config -s general.vfs.context.classpath.myNamespace=hdfs://NAME_NODE_FDQN:54310/accumulo/classpath/myNamespace/[^.].*.jar
-    > config -ns myNamespace -s table.classpath.context=myNamespace
+    $ ./install-geomesa-namespace.sh -u root -n myNamespace
 
-.. note::
+The command line arguments the script accepts are:
 
-    Depending on Hadoop version, you may need to use ``hdfs://NAME_NODE_FDQN:8020``.
-
-Then copy the distributed runtime jar into HDFS under the path you specified.
-The path above is just an example; you can included nested folders with project
-names, version numbers, and other information in order to have different versions of GeoMesa on
-the same Accumulo instance. You should remove any GeoMesa JARs under
-``$ACCUMULO_HOME/lib/ext`` to prevent any classpath conflicts.
+Required parameters:
+     -u (Accumulo username)
+     -n (Accumulo namespace)
+Optional parameters:
+     -p (Accumulo password)
+     -g (Path of GeoMesa distributed runtime JAR)
+     -h (HDFS URI e.g. hdfs://localhost:54310)
 
 .. note::
 
--- a/geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
+++ b/geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
@@ -1,0 +1,96 @@
+#! /usr/bin/env bash
+#
+# Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License, Version 2.0 which
+# accompanies this distribution and is available at
+# http://www.opensource.org/licenses/apache2.0.php.
+#
+
+# Installs a GeoMesa distributed runtime JAR into an Accumulo namespace
+
+while getopts ":u:p:n:g:h:" opt; do
+  case $opt in
+    u)
+      ACCUMULO_USER=$OPTARG
+      ;;
+    p)
+      ACCUMULO_PASSWORD=$OPTARG
+      ;;
+    n)
+      ACCUMULO_NAMESPACE=$OPTARG
+      ;;
+    g)
+      GEOMESA_JAR=$OPTARG
+      ;;
+    h)
+      HDFS_URI=$OPTARG
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      exit 1
+      ;;
+    :)
+      echo "Option -$OPTARG requires an argument." >&2
+      exit 1
+      ;;
+  esac
+done
+
+if [ -z "$ACCUMULO_USER" ]; then
+    echo "Accumulo username parameter is required: -u" >&2
+    ERROR=1
+fi
+
+if [ -z "$ACCUMULO_NAMESPACE" ]; then
+    echo "Accumulo namespace parameter is required: -n" >&2
+    ERROR=1
+fi
+
+if [ -z "$GEOMESA_JAR" ]; then
+    # cd to directory of script and look for GeoMesa JAR
+    cd "$( dirname "${BASH_SOURCE[0]}" )"
+    GEOMESA_JAR=$(ls | grep geomesa-accumulo-distributed-runtime)
+    if [ -z "$GEOMESA_JAR" ]; then
+        echo "Null GeoMesa JAR parameter encountered and could not find a GeoMesa distributed runtime JAR"
+        ERROR=1
+    else
+        echo "Null GeoMesa JAR parameter encountered, using $GEOMESA_JAR"
+    fi
+fi
+
+if [[ -z "$HDFS_URI" ]]; then
+    HDFS_URI=`hdfs getconf -confKey fs.defaultFS`
+    echo "Null HDFS URI parameter encountered, using $HDFS_URI"
+fi
+
+if [[ -z $ERROR && -z "$ACCUMULO_PASSWORD" ]]; then
+    read -s -p "Enter Accumulo Password: " ACCUMULO_PASSWORD
+fi
+
+if [ -n "$ERROR" ]; then
+    echo -e "\nRequired parameters:\n\t" \
+      "-u (Accumulo username)\n\t" \
+      "-n (Accumulo namespace)"
+    echo -e "Optional parameters:\n\t" \
+      "-p (Accumulo password)\n\t" \
+      "-g (Path of GeoMesa distributed runtime JAR)\n\t" \
+      "-h (HDFS URI e.g. hdfs://localhost:54310)"
+    exit 1
+fi
+
+echo "Copying GeoMesa JAR for Accumulo namespace $ACCUMULO_NAMESPACE ..."
+hadoop fs -mkdir -p /accumulo/classpath/${ACCUMULO_NAMESPACE}
+hadoop fs -copyFromLocal -f $GEOMESA_JAR /accumulo/classpath/${ACCUMULO_NAMESPACE}/
+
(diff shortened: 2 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,96 @@
...
+    GEOMESA_JAR=$(ls | grep geomesa-accumulo-distributed-runtime)
+    if [ -z "$GEOMESA_JAR" ]; then
+        echo "Null GeoMesa JAR parameter encountered and could not find a GeoMesa distributed runtime JAR"
+        ERROR=1
+    else
+        echo "Null GeoMesa JAR parameter encountered, using $GEOMESA_JAR"
+    fi
+fi
+
+if [[ -z "$HDFS_URI" ]]; then
```
cne1x: [picky] This is the second syntax for a conditional statement; any reason for the inconsistency?

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,96 @@
...
+hadoop fs -copyFromLocal -f $GEOMESA_JAR /accumulo/classpath/${ACCUMULO_NAMESPACE}/
+
+if hadoop fs -ls /accumulo/classpath/${ACCUMULO_NAMESPACE}/geomesa*.jar > /dev/null 2>&1
+then
+    echo -e "createnamespace ${ACCUMULO_NAMESPACE}\n" \
+      "grant NameSpace.CREATE_TABLE -ns ${ACCUMULO_NAMESPACE} -u root\n" \
+      "config -s general.vfs.context.classpath.${ACCUMULO_NAMESPACE}=${HDFS_URI}/accumulo/classpath/${ACCUMULO_NAMESPACE}/.*.jar\n" \
+      "config -ns ${ACCUMULO_NAMESPACE} -s table.classpath.context=${ACCUMULO_NAMESPACE}\n" \
+      | accumulo shell -u $ACCUMULO_USER -p $ACCUMULO_PASSWORD
+    echo -e "Finished installing GeoMesa distributed runtime jar."
```
cne1x: [picky] This is one of the very few instances of "jar" rather than "JAR".

cne1x:
This looks reasonable to me.  Nicely done.

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,96 @@
...
+      exit 1
+      ;;
+    :)
+      echo "Option -$OPTARG requires an argument." >&2
+      exit 1
+      ;;
+  esac
+done
+
+if [ -z "$ACCUMULO_USER" ]; then
```
jahhulbert-ccri: I've always heard go with the double brackets for any conditional testing in bash

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,96 @@
...
+      "-n (Accumulo namespace)"
+    echo -e "Optional parameters:\n\t" \
+      "-p (Accumulo password)\n\t" \
+      "-g (Path of GeoMesa distributed runtime JAR)\n\t" \
+      "-h (HDFS URI e.g. hdfs://localhost:54310)"
+    exit 1
+fi
+
+echo "Copying GeoMesa JAR for Accumulo namespace $ACCUMULO_NAMESPACE ..."
+hadoop fs -mkdir -p /accumulo/classpath/${ACCUMULO_NAMESPACE}
```
jahhulbert-ccri: this should be configurable. not everyone is in ```/accumulo/classpath```

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,96 @@
...
+hadoop fs -copyFromLocal -f $GEOMESA_JAR /accumulo/classpath/${ACCUMULO_NAMESPACE}/
+
+if hadoop fs -ls /accumulo/classpath/${ACCUMULO_NAMESPACE}/geomesa*.jar > /dev/null 2>&1
+then
+    echo -e "createnamespace ${ACCUMULO_NAMESPACE}\n" \
+      "grant NameSpace.CREATE_TABLE -ns ${ACCUMULO_NAMESPACE} -u root\n" \
```
jahhulbert-ccri: does this only use the root user? if so lets make that configurable at least at the top of the file

jahhulbert-ccri: actually do we really need this grant? i think the root user can always do this by default and if its another user then that user will need to have the admin grant the permissions to them anyways?

docs/user/installation_and_configuration.rst
```diff
@@ -132,26 +132,23 @@ For Accumulo 1.6+
...
 
-.. code::
+Execute the ``install-geomesa-namespace.sh`` script in the ``geomesa-$VERSION/dist/accumulo`` directory.
 
-    $ accumulo shell -u root
-    > createnamespace myNamespace
-    > grant NameSpace.CREATE_TABLE -ns myNamespace -u myUser
-    > config -s general.vfs.context.classpath.myNamespace=hdfs://NAME_NODE_FDQN:54310/accumulo/classpath/myNamespace/[^.].*.jar
-    > config -ns myNamespace -s table.classpath.context=myNamespace
+.. code::
```
mdzimmerman: I would not remove the instructions for configuring the JAR in the namespace manipulation, even if the script takes care of it. Maybe we could put all of it into an admonition box or move it to a section in the Accumulo section?

jnh5y: +1  I'd like to see both approaches described.  It'll definitely help folks understand that it is an either/or choice.

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
jnh5y: We are doing the same as here: http://www.geomesa.org/documentation/user/installation_and_configuration.html#for-accumulo-1-6

jnh5y: And yeah, I think root should be ${ACCUMULO_USER}.

Thinking more about @jahhulbert-ccri's point, the user doing this means require privileges to do this...  For many installations, this should be useful.  In more secure cases, the admins might just use the documentation to figure out what we are trying to do.

jahhulbert-ccri: i'm a little confused as why myUser is in there at all?

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
Afterglow375: Why expose this to the user? Why does it matter exactly where the namespaces are set up.

docs/user/installation_and_configuration.rst
```diff
@@ -132,26 +132,37 @@ For Accumulo 1.6+
...
 still work, but in Accumulo 1.6, we can leverage namespaces to isolate the
-GeoMesa classpath from the rest of Accumulo. First, you have to create the
-namespace in the Accumulo shell:
+GeoMesa classpath from the rest of Accumulo.
+
+Execute the ``install-geomesa-namespace.sh`` script in the ``geomesa-$VERSION/dist/accumulo`` directory.
 
 .. code::
 
+    $ ./install-geomesa-namespace.sh -u root -n myNamespace
```
elahrvivaz: does the user have to be root? have we tested with anything else? since it grants permissions, a regular user maybe wouldn't work?
if it has to be root, we probably need 2 sets of users - the root user/pwd and then the runtime user that will be granted the permissions.

docs/user/installation_and_configuration.rst
```diff
@@ -132,26 +132,37 @@ For Accumulo 1.6+
...
+
+The command line arguments the script accepts are:
+
+Required parameters:
+     -u (Accumulo username)
+     -n (Accumulo namespace)
+Optional parameters:
+     -p (Accumulo password)
+     -g (Path of GeoMesa distributed runtime JAR)
+     -h (HDFS URI e.g. hdfs://localhost:54310)
```
elahrvivaz: could you indicate when these would be needed, or if it will just prompt you if you don't include them?

docs/user/installation_and_configuration.rst
```diff
@@ -132,26 +132,37 @@ For Accumulo 1.6+
...
 
-.. note::
-
-    Depending on Hadoop version, you may need to use ``hdfs://NAME_NODE_FDQN:8020``.
-
 Then copy the distributed runtime jar into HDFS under the path you specified.
 The path above is just an example; you can included nested folders with project
 names, version numbers, and other information in order to have different versions of GeoMesa on
-the same Accumulo instance. You should remove any GeoMesa JARs under
-``$ACCUMULO_HOME/lib/ext`` to prevent any classpath conflicts.
```
elahrvivaz: this still seems like a good caveat to leave in, any reason we removed it?

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+    echo "Accumulo namespace parameter is required: -n" >&2
+    ERROR=1
+fi
+
+if [[ -z "$GEOMESA_JAR" ]]; then
+    # cd to directory of script and look for GeoMesa JAR
+    cd "$( dirname "${BASH_SOURCE[0]}" )"
+    GEOMESA_JAR=$(ls | grep geomesa-accumulo-distributed-runtime)
+    if [ -z "$GEOMESA_JAR" ]; then
+        echo "Null GeoMesa JAR parameter encountered and could not find a GeoMesa distributed runtime JAR"
```
elahrvivaz: I would re-word this to something like:
> Could not find GeoMesa distributed runtime JAR - please specify the jar using the 'foo' flag

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+        ERROR=1
+    else
+        echo "Null GeoMesa JAR parameter encountered, using $GEOMESA_JAR"
```
elahrvivaz: Same thing here - change to `Using $GEOMESA_JAR` - we don't want to refer to nulls in a user-facing script

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+    fi
+fi
+
+if [[ -z "$NAMESPACE_DIR" ]]; then
+    NAMESPACE_DIR="/accumulo/classpath"
+    echo "Null namespace directory parameter encountered, using $NAMESPACE_DIR"
```
elahrvivaz: ditto the rest of these statements about Null

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+fi
+
+if [[ -z "$HDFS_URI" ]]; then
+    HDFS_URI=`hdfs getconf -confKey fs.defaultFS`
+    echo "Null HDFS URI parameter encountered, using $HDFS_URI"
+fi
+
+if [[ -z $ERROR && -z "$ACCUMULO_PASSWORD" ]]; then
```
elahrvivaz: do we need quotes around the `$ERROR`? not sure, but other examples are quoted

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+    read -s -p "Enter Accumulo Password: " ACCUMULO_PASSWORD
```
elahrvivaz: would be nice to include the user here - "Please enter Accumulo password for user $ACCUMULO_USER: "

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+      "-p (Accumulo password)\n\t" \
+      "-g (Path of GeoMesa distributed runtime JAR)\n\t" \
+      "-d (Directory to create namespace in, defaults to /accumulo/classpath)\n\t" \
+      "-h (HDFS URI e.g. hdfs://localhost:54310)"
+    exit 1
+fi
+
+echo "Copying GeoMesa JAR for Accumulo namespace $ACCUMULO_NAMESPACE ..."
+hadoop fs -mkdir -p ${NAMESPACE_DIR}/${ACCUMULO_NAMESPACE}
+hadoop fs -copyFromLocal -f $GEOMESA_JAR ${NAMESPACE_DIR}/${ACCUMULO_NAMESPACE}/
```
elahrvivaz: probably good to quote these two paths and the jar path just in case there is a space or something

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+
+if hadoop fs -ls ${NAMESPACE_DIR}/${ACCUMULO_NAMESPACE}/geomesa*.jar > /dev/null 2>&1
```
elahrvivaz: huh, haven't seen that syntax before :)

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+then
+    echo -e "createnamespace ${ACCUMULO_NAMESPACE}\n" \
+      "grant NameSpace.CREATE_TABLE -ns ${ACCUMULO_NAMESPACE} -u $ACCUMULO_USER\n" \
+      "config -s general.vfs.context.classpath.${ACCUMULO_NAMESPACE}=${HDFS_URI}${NAMESPACE_DIR}/${ACCUMULO_NAMESPACE}/.*.jar\n" \
+      "config -ns ${ACCUMULO_NAMESPACE} -s table.classpath.context=${ACCUMULO_NAMESPACE}\n" \
+      | accumulo shell -u $ACCUMULO_USER -p $ACCUMULO_PASSWORD
+    echo -e "Finished installing GeoMesa distributed runtime JAR."
```
elahrvivaz: might be good to check the return value from the shell command for errors

geomesa-accumulo/geomesa-accumulo-distributed-runtime/src/main/resources/install-geomesa-namespace.sh
```diff
@@ -0,0 +1,105 @@
...
+fi
+
+if [[ -z "$NAMESPACE_DIR" ]]; then
+    NAMESPACE_DIR="/accumulo/classpath"
+    echo "Null namespace directory parameter encountered, using $NAMESPACE_DIR"
+fi
+
+if [[ -z "$HDFS_URI" ]]; then
+    HDFS_URI=`hdfs getconf -confKey fs.defaultFS`
+    echo "Null HDFS URI parameter encountered, using $HDFS_URI"
```
elahrvivaz: probably worth another check that we actually got something usable

docs/user/installation_and_configuration.rst
Afterglow375: Yeah, the script will try to find defaults for omitted optional params. Password will be prompted for.

elahrvivaz:
looks good, i'm going to merge

elahrvivaz:
538fb2d7c7105799bc9efad6a32ac6b4240a3252

PR #978, opened 2016-07-07
GEOMESA-1302 Added Swagger documentation to the stats REST API

Afterglow375:
Signed-off-by: Alex Tatusko <atatusko@ccri.com>

PR #1009, opened 2016-07-25
GEOMESA-1331 Fixing histogram out-of-bounds exception in stats estimation

elahrvivaz:
Note: stacks on top of #978, please only consider last commit(s)


jnh5y:
The final commit looks reasonable to me.

PR #1011, opened 2016-07-26
GEOMESA-1337 - Add tutorial and example code for GMSpark Shallow Join

atallahhezbor:


docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+      "tableName"  -> "countries_catalog")
+
+    val gdeltDs = DataStoreFinder.getDataStore(gdeltParams)
+    val countriesDs = DataStoreFinder.getDataStore(countriesParams)
+
+
+Next, initialize a Spark context from the Simple Feature Types of the data stores.
+
+.. code-block:: scala
+    val types = gdeltDs.getTypeNames.map(ds.getSchema) ++ countriesDs.getTypeNames.map(ds.getSchema)
```
mdzimmerman: Needs to be an empty line between the code-block directive and the first line of the block.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+
+Our new RDD is now of type ``RDD[(String, SimpleFeature)]`` and can be used for a Spark ``reduceByKey`` operation, but
+first, we need to create a simple feature type to represent the aggregated data.
+
+Creating a New Simple Feature Type
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We first loop through the attributes of a sample feature from the GDELT RDD to decide what fields can be aggregated.
+
+.. code-block: scala
```
mdzimmerman: Need two colons.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+-------------
+
+.. warning::
+
+    You will need access to a Hadoop |hadoop_version| installation with Yarn as well as an Accumulo |accumulo_version| database.
+
+    You will need to have ingested GDELT data using GeoMesa. Instructions are available in :doc:`geomesa-examples-gdelt`.
+
+    You will need to have ingested a shapefile of polygons outlining your regions of choice. In this tutorial we use
+    `this <http://thematicmapping.org/downloads/world_borders.php>` shapefile of countries.
```
mdzimmerman: Need two underscores after the backtic.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+
+You will also need:
+
+-  a `Spark <http://spark.apache.org/>`__ 1.5.0 or later distribution
+-  an Accumulo user that has appropriate permissions to query your data
+-  `Java JDK 8 <http://www.oracle.com/technetwork/java/javase/downloads/index.html>`__,
+-  a `Jupyter Notebook <https://github.com/jupyter/notebook>`__ server with the
+    `Apache Toree https://toree.incubator.apache.org/documentation/user/installation.html`__ Scala kernel installed
```
mdzimmerman: Need angle brackets around the URL.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+            None
+        }
+    }).toArray
+    val countable = sc.broadcast(countableIndices)
+
+With these fields, we can create a Simple Feature Type to store their averages and totals, prefixing each one with
+"total_" and "avg_". Of course, it may not make sense to aggregate features like "ID" should it appear, but this
+approach makes it easy if the fields are not known ahead of time.
+
+.. code-block: scala
```
mdzimmerman: Two colons.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+    val coverSft = SimpleFeatureTypes.createType("aggregate",sftBuilder.getSpec)
+
+
+Aggregating by Key
+^^^^^^^^^^^^^^^^^^
+
+To begin aggregating we first send our new Simple Feature Type to each of the executors so that they are
+able to create and serialize Simple Features of that type.
+
+.. code-block: scala
```
mdzimmerman: Two colons.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+Now we can apply a ``reduceByKey`` operation to the keyed RDD. This Spark operation will take pairs of RDD elements of
+the same key, apply the given function, and replace them with the result. Here, we have three cases for reduction.
+
+1. The two Simple Features have not been aggregated into one of a new type.
+2. The two Simple Features have both been aggregated into one of a new type.
+3. One of the Simple Features has been aggregated (but not both)
+
+For the sake of brevity, we will only show the first case, with the other three following similar patterns.
+
+.. code-block: scala
```
mdzimmerman: Two colons.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+    val geoJsonString = geoJsonWriters.collect.mkString("[",",","]")
+
+    // Write to file
+
+In order to modify the DOM from within a Jupyter cell, we must set up a Mutation Observer to correctly respond to asynchronous
+changes. We attach the observer to ``element``, which refers to the cell from which the javascript code is run. Within
+this observer, we instantiate a new Leaflet map, and add a base layer from OSM. Further, inside the leaflet, we create a
+new layer either from GeoServer as WMS, or a tile layer from the GeoJSON file we created.
+
+(new MutationObserver(function() {
```
mdzimmerman: Can you put this in a ``.. code-block:: javascript``?

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
+Visualizing Analytics in Jupyter
```
mdzimmerman: You need to add an entry to ``docs/tutorials/index.rst`` so that this tutorial is navigable from the main TOC.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+for large-scale data processing". Spark presents an abstraction called a
+Resilient Distributed Dataset (RDD) that facilitates expressing
+transformations, filters, and aggregations, and efficiently executes the
+computation across a distributed set of resources. Spark manages the
+lineage of a block of transformed data so that if a node goes down,
+Spark can restart the computation for just the missing blocks.
+
+`Jupyter Notebook <https://github.com/jupyter/notebook>`__ is an interactive web interface
+for a kernel. Jupyter allows you quickly prototype by writing code in runnable cells, computing a final result
+as you go. Visualization of data is also easily done through integration Jupyter cell 'magics' to create JavaScript and
```
mdzimmerman: (picky) I would put magics in double quotes.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+            GeoMesaSparkKryoRegistrator.putType(newSft)
+        }
+    }
+
+Now we can apply a ``reduceByKey`` operation to the keyed RDD. This Spark operation will take pairs of RDD elements of
+the same key, apply the given function, and replace them with the result. Here, we have three cases for reduction.
+
+1. The two Simple Features have not been aggregated into one of a new type.
+2. The two Simple Features have both been aggregated into one of a new type.
+3. One of the Simple Features has been aggregated (but not both)
```
mdzimmerman: (picky) missing period

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+
+Visualization
+^^^^^^^^^^^^^
+
+While there are many ways to visualize data from an RDD, here we choose to demonstrate the use of leaflet for easy integration
+with Jupyter Notebook. To use, either install it through Jupyter's ``nbextensions`` tool, or place the following HTML
+magic in your notebook to import it properly.
+
+.. code-block: HTML
+    %%HTML
```
mdzimmerman: Blank line at start of block.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+    <link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.css" />
+    <script src="http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.js"></script>
+
+The problem of getting data from an RDD in the Scala Kernel to client-side JavaScript can also be solved in many ways.
+One option is to save the RDD to a GeoMesa schema and use the GeoServer Manager API to publish a WMS layer. Leaflet
+is capable of then reading a WMS layer into it's map via HTTP. A more direct route, however, is to export the RDD as GeoJSON.
+To do this, use Toree's ``AddDeps`` magic to add the GeoTool GeoJSON dependency on the fly.
+
+.. code-block: bash
+    %AddDeps org.geotools gt-geojson 14.1 --transitive --repository http://download.osgeo.org/webdav/geotools
```
mdzimmerman: Blank line at start of block.

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+magic in your notebook to import it properly.
+
+.. code-block: HTML
+    %%HTML
+    <link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.css" />
+    <script src="http://cdn.leafletjs.com/leaflet/v0.7.7/leaflet.js"></script>
+
+The problem of getting data from an RDD in the Scala Kernel to client-side JavaScript can also be solved in many ways.
+One option is to save the RDD to a GeoMesa schema and use the GeoServer Manager API to publish a WMS layer. Leaflet
+is capable of then reading a WMS layer into it's map via HTTP. A more direct route, however, is to export the RDD as GeoJSON.
```
mdzimmerman: ``s/it's/its/``

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+    var map = L.map('map').setView([35.4746,-44.7022],3);
+    // Add the base layer
+    L.tileLayer("http://{s}.tile.osm.org/{z}/{x}/{y}.png").addTo(map);
+
+
+    this.disconnect()
+})).observe(element[0], {childList: true})
+
+There are many opportunities here to style these layers such as coloring polygons by attributes. Here we color each
+country's polygon by it's average goldstein scale, indicating how events are contributing to the stability of a country
```
mdzimmerman: ``s/it's/its/``

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,359 @@
...
+                case _ => None
+            }
+        } else {
+            None
+        }
+    }).toArray
+    val countable = sc.broadcast(countableIndices)
+
+With these fields, we can create a Simple Feature Type to store their averages and totals, prefixing each one with
+"total_" and "avg_". Of course, it may not make sense to aggregate features like "ID" should it appear, but this
```
mdzimmerman: GeoMesa doesn't let you have attributes named "ID", does it?

atallahhezbor: Not sure, but I meant it as a general case of something like GDELT's globalEventId. I'll word it a bit better

docs/tutorials/shallow-join.rst
```diff
@@ -0,0 +1,381 @@
...
+
+-  a `Spark <http://spark.apache.org/>`__ 1.5.0 or later distribution
+-  an Accumulo user that has appropriate permissions to query your data
+-  `Java JDK 8 <http://www.oracle.com/technetwork/java/javase/downloads/index.html>`__,
+-  a `Jupyter Notebook <https://github.com/jupyter/notebook>`__ server with the
+    `Apache Toree <https://toree.incubator.apache.org/documentation/user/installation.html>`__ Scala kernel installed
+
+
+Create RDDs
+^^^^^^^^^^^
```
mdzimmerman: All of the subsequent headings should have dashes rather than carets under them; otherwise it makes it seem like each of the following sections are subsections of "Prerequisites".

mdzimmerman:
Alright, looks good to me. Sorry to be so picky, but it wasn't rendering properly when I built the document with Sphinx.

docs/tutorials/shallow-join.rst
jnh5y: We have warnings for that; check out https://github.com/locationtech/geomesa/blob/master/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala.;)

jahhulbert-ccri: Ideally we wouldn't use ID because it screws stuff up.

jahhulbert-ccri:
cool. @jnh5y  or @elahrvivaz got anything else?

atallahhezbor:
@jahhulbert-ccri after some discussion with Bob, I plan on expanding the tutorial to talk more about visualization

docs/tutorials/shallow-join.rst
```diff
@@ -379,3 +386,6 @@ country's polygon by its average goldstein scale, indicating how events are cont
 during that time range.
 
 .. figure:: _static/img/tutorials/2016-07-26-shallow-join/aggregate-GDELT.png
+
+The final result of the analysis described in this tutorial is found in the Jupyter notebook: :download:`_static/shallow-join-gdelt.ipynb`.
+A static render of this notebook can be found on Github (https://github.com/locationtech/geomesa/docs/tutorials/_static/shallow-join-gdelt.ipynb).
```
mdzimmerman: Paths here have to match where you put the notebook (e.g. ``_static/geomesa-examples-jupyter/shallow-join-gdelt.ipynb``).

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
     groupedByDay.map { case (date, iter) => (date, iter.size) }
   }
 
+  def shallowJoin(sc: SparkContext, coveringSet: RDD[SimpleFeature], data: RDD[SimpleFeature], key: String): RDD[SimpleFeature] = {
+    // Broadcast sfts to executors
+    val sfts = sc.broadcast(GeoMesaSparkKryoRegistrator.typeCache.values.map{ sft =>
```
elahrvivaz: [picky] space before an opening `{` - noticed it a few other places too

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+    val broadcastedCover = sc.broadcast(coveringSet.collect)
+
+    // Key data by cover name
+    val keyedData = data.mapPartitions { iter =>
+      import org.locationtech.geomesa.utils.geotools.Conversions._
+
+      iter.flatMap { sf =>
+        // Iterate over covers until a match is found
+        val it = broadcastedCover.value.iterator
+        var container = ""
```
elahrvivaz: [minor] would be more standard to use `var container: String = null` and check for nulls, or alternatively just make it an option to begin with.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+          Some(container, sf)
+        } else {
+          None
+        }
+      }
+    }
+
+    // Get the indices and types of the attributes that can be aggregated and send them to the partitions
+    val featureAttributes = data.first.getAttributes.toSeq
+    val countableIndices = featureAttributes.toIndexedSeq.indices.flatMap( { index =>
```
elahrvivaz: [picky] don't need the `()` since you have `{}`

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+      // Skip the id
+      if (index != 0) {
+        featureAttributes(index) match {
+          case attr if attr != null && attr.getClass == classOf[Integer] =>
+            Some(index, "Integer")
+          case attr if attr != null && attr.getClass == classOf[java.lang.Long] =>
+            Some(index, "Long")
+          case attr if attr != null && attr.getClass == classOf[java.lang.Double] =>
```
elahrvivaz: could we just use pattern matching here? e.g.
```
featureAttributes(index) match {
  case attr: Integer if attr != null =>
```

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+    sftBuilder.intType("count")
+    val featureProperties = data.first.getProperties.toSeq
+    countableIndices.foreach{ case (index, clazz) =>
+        val featureName = featureProperties.apply(index).getName
+        clazz match {
+          case "Integer" => sftBuilder.intType("total_" + featureName)
+          case "Long" => sftBuilder.longType("total_" + featureName)
+          case "Double" => sftBuilder.doubleType("total_" + featureName)
+        }
+        sftBuilder.doubleType("avg_"+featureProperties.apply(index).getName)
```
elahrvivaz: [picky] space around `+`

elahrvivaz: don't have to change, but in general prefer string interpolation - `s"avg_${featureProperties.apply(index).getName}"`

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+          sf.setDefaultGeometry(broadcastedCoverMap.value.getOrElse(coverName, null))
+          Some(sf)
+        } else {
+          None
+        }
+      }
+    }
+  }
+
+  def reduceAndAggregate(keyedData: RDD[(String, SimpleFeature)], countable: Broadcast[Array[(Int, String)]]): RDD[(String, SimpleFeature)] = {
```
elahrvivaz: [minor] long line - put `countable` down on the next line

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+      iter.flatMap{ case (coverName, sf) =>
+        if (sf.getType.getTypeName == "aggregate") {
+          sf.getProperties.foreach{ prop =>
+            val name = prop.getName.toString
+            if (name.startsWith("total_")) {
+              val count = sf.get[Integer]("count")
+              val avg = prop.getValue match {
+                case (a: Integer) => a / count
+                case (a: java.lang.Long) => a / count
+                case (a: java.lang.Double) => a / count
```
elahrvivaz: [minor] don't need `()`

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+      import org.locationtech.geomesa.utils.geotools.Conversions.RichSimpleFeature
+
+      iter.flatMap{ case (coverName, sf) =>
+        if (sf.getType.getTypeName == "aggregate") {
+          sf.getProperties.foreach{ prop =>
+            val name = prop.getName.toString
+            if (name.startsWith("total_")) {
+              val count = sf.get[Integer]("count")
+              val avg = prop.getValue match {
+                case (a: Integer) => a / count
```
elahrvivaz: could `count` ever be 0?

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+        countable.value.foreach { case (index, clazz) =>
+          val propA = featurePropertiesA(index)
+          val propB = featurePropertiesB(index)
+          val valA = if (propA == null) 0 else propA.getValue
+          val valB = if (propB == null) 0 else propB.getValue
+
+          val sum = (valA, valB) match {
+            case (a: Integer, b: Integer) => a + b
+            case (a: java.lang.Long, b: java.lang.Long) => a + b
+            case (a: java.lang.Double, b: java.lang.Double) => a + b
```
elahrvivaz: could this cause a problem if propA is null, making it an Int, and propB is e.g. a Double?

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+            case _ => throw new Exception("Couldn't match countable type.")
+          }
+          // Set the total
+          if (propA != null) {
+            aggregateFeature.setAttribute("total_" + propA.getName.toString, sum)
```
elahrvivaz: since you're only using the sum if porpA is not null, you could move the sum calc inside this if block

elahrvivaz:
looks good, my comments were mainly style and picky things

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+
+    // Get the indices and types of the attributes that can be aggregated and send them to the partitions
+    val featureAttributes = data.first.getAttributes.toSeq
+    val countableIndices = featureAttributes.toIndexedSeq.indices.flatMap( { index =>
+      // Skip the id
+      if (index != 0) {
+        featureAttributes(index) match {
+          case attr if attr != null && attr.getClass == classOf[Integer] =>
+            Some(index, "Integer")
+          case attr if attr != null && attr.getClass == classOf[java.lang.Long] =>
```
vpipkt: Doesn't this mean that if the `.first` feature has a null, then the field will be omitted from the analysis for all features?

atallahhezbor: It was that or an NPE. Though now that I think about it more I may be able to get the type from the property instead of the attribute.

vpipkt: In general one would use `dataStore.getSchema` and then call `getBinding` on types, but I _think_ this would require a refactor since this function only takes the RDD's.

A minimal refactor would be to have this function take the  `SimpleFeatureType` of the `data` RDD as well.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,224 @@ object GeoMesaSpark extends LazyLogging {
...
+      import org.locationtech.geomesa.utils.geotools.Conversions._
+
+      iter.flatMap { sf =>
+        // Iterate over covers until a match is found
+        val it = broadcastedCover.value.iterator
+        var container = ""
+        while (it.hasNext && container == "") {
+          val cover = it.next()
+          // If the cover's polygon contains the feature, set the container
+          if (cover.geometry.contains(sf.geometry)) {
```
vpipkt: I think `intersects` may be a bit more generally applicable and would be _almost_ equivalent in the case of a polygon covering set and point data. The difference would be points on the boundary of a polygon would count for the polygon.

The intersect is really interesting and useful for `data` with LineString or Polygon geometries. In this case one expects an individual data to possibly `intersect  more than one covering set. And plausibly `contain`ed by none.

docs/tutorials/shallow-join.rst
cne1x: We may want to reconsider our terminology, @jnh5y.  "shallow join" appears to be a new term for the previously well established practice of aggregating over an inner join (albeit, a join using a geospatial predicate, which again is not novel).  Perhaps, "jupyter-aggregation" is a better name for this tutorial?

docs/tutorials/_static/geomesa-examples-jupyter/shallow-join-gdelt.ipynb
```diff
@@ -0,0 +1,814 @@
...
+    "    import org.locationtech.geomesa.utils.geotools.Conversions.RichSimpleFeature\n",
+    "\n",
+    "    iter.flatMap{ case (countryName, sf) =>\n",
+    "        if (sf.getType.getTypeName == \"countryInformation\") {\n",
+    "            sf.getProperties.foreach{ prop =>\n",
+    "                val name = prop.getName.toString\n",
+    "                if (name.startsWith(\"total_\")) {\n",
+    "                    val count = sf.get[Integer](\"count\")\n",
+    "                    val avg = (prop.getValue) match {\n",
+    "                            case (a: Integer) => a / count\n",
```
cne1x: I'm not as familiar with Spark and Jupyter as I ought to be, but aren't there quite a few extraneous parentheses here?  Another way to shorten this, if appropriate, might be:
```scala
val avg = Try { a.toDouble / count } match {
  case Success(a) => a
  case _ => throw new Exception(s\"couldn't match $name\")\n"
}
```
It also has the advantage of catching the divide-by-zero (although that seems exceedingly unlikely in this case, as there almost certainly wouldn't exist a "count" record if there hadn't been at least one contributing data point).

docs/tutorials/_static/geomesa-examples-jupyter/shallow-join-gdelt.ipynb
```diff
@@ -0,0 +1,814 @@
...
+    "                case (a: java.lang.Long, b: java.lang.Long) => a + b\n",
+    "                case (a: java.lang.Double, b: java.lang.Double) => a + b\n",
+    "                case _ => throw new Exception(\"Couldn't match countable type.\")\n",
+    "            }\n",
+    "            // Set the total\n",
+    "            if( propA != null)\n",
+    "                aggregateFeature.setAttribute(\"total_\"+ propA.getName.toString, sum)                      \n",
+    "                  \n",
+    "        }\n",
+    "        aggregateFeature.setAttribute(\"count\", new Integer(2))              \n",
```
cne1x: [picky] If this code is going to be called many times, it may be more efficient to pre-compute the index of these constant string attribute names, and set them by index.

docs/tutorials/_static/geomesa-examples-jupyter/shallow-join-gdelt.ipynb
elahrvivaz: is avg always a double? would make sense, but the current code makes it the same binding as the values being averaged. i.e. `1/2 == 0` vs `1.0/2 == 0.5`. @cne1x's fix would make it a double. Need to make sure that the simple feature type attribute binding for avg is the right type, regardless.

docs/tutorials/shallow-join.rst
jnh5y: Yeah, if there's an existing name, let's use it.  Explaining that we are forming an inner join with a covering geo set is useful.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
atallahhezbor: Since I'm skipping over features that aren't of type "aggregate", no. 
When I create features of that type, `count` is initialized to two, and when features of that type are merged, `count` only increases. The only case I could see is some sort of overflow... I could add a check for it being 0 or negative, but then what would I use instead?

docs/tutorials/_static/geomesa-examples-jupyter/shallow-join-gdelt.ipynb
atallahhezbor: `avg` is always a double https://github.com/ccri/geomesa/blob/e99e068265967becf8da6a4ee2d0864f3ab6fed3/geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala#L264

So yes, I would I would need to make `a` a double before dividing.

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
atallahhezbor: I could switch to intersects, but currently the loop stops when the first cover is found. I did that to be slightly more efficient, but I could change that too to handle other geometries.

Though in that case, is it okay to have features get "double counted"? I can see this, for example, if the LineStrings were roads, and a single road would be counted for each region it was in, but the total count would be off. I'm also not sure if there are examples of those geometries that would only need to be counted once across all covers.

I could just let the function be a parameter and default to this.

jnh5y: Just to wade in, how about 'intersects' and *not* worrying with double-counts?

For instance, if an interstate is in multiple states, counting it for each seems reasonable.  (Am I making sense?  I'm jumping in in the middle:))

atallahhezbor: Yeah that makes sense, but I was just worried of examples where that wouldn't seem reasonable. We wouldn't lose much of anything by switching to intersects for our current examples (points and roads)

docs/tutorials/_static/geomesa-examples-jupyter/shallow-join-gdelt.ipynb
atallahhezbor: Though the reason I can't make is as simple as your example, @cne1x is that `getValue` returns an Object, so I need to pattern match before calling `toDouble`. Also I can't just do `asInstanceOf[Double]` because types like `java.lang.Long` cannot be cast to `Double`

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
atallahhezbor: @elahrvivaz if the attribute was null, would it still pattern match to a known type? The attributes are of type AnyRef, and I'm not that familiar with how Scala deals with those in this context.

elahrvivaz: null objects wouldn't pattern match to a type - it would be caught by your `case _ =>`
```
scala> val i: Integer = null
i: Integer = null

scala> i match { case m: Integer => println("matched integer"); case _ => println("matched null") }
matched null
```

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
atallahhezbor: By moving the sum calculation into block of the null check like you suggested below, I get around this issue. I also better handle the vice versa case (when propB is null)

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
atallahhezbor: Turns out I can derive the SimpleFeatureType from the first element of the RDD, which isn't affected by null attributes. I think that'd be the cleanest route

geomesa-compute/src/main/scala/org/locationtech/geomesa/compute/spark/GeoMesaSpark.scala
```diff
@@ -187,6 +188,229 @@ object GeoMesaSpark extends LazyLogging {
...
+          Some(container.get, sf)
+        } else {
+          None
+        }
+      }
+    }
+
+    // Get the indices and types of the attributes that can be aggregated and send them to the partitions
+    val countableTypes = Seq("Integer", "Long", "Double")
+    val typeNames = data.first.getType.getTypes.toIndexedSeq.map{t => t.getBinding.getSimpleName.toString}
```
vpipkt: Nice!

PR #1012, opened 2016-07-29
GEOMESA-1343 XZ Spatial Index

elahrvivaz:
* Replaces Z2 index for polygons


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/XZ2Table.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.hadoop.io.Text
+import org.locationtech.geomesa.accumulo.data.AccumuloFeatureWriter.FeatureToMutations
+import org.locationtech.geomesa.accumulo.data._
+import org.locationtech.geomesa.curve.XZ2SFC
+import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
+import org.opengis.feature.simple.SimpleFeatureType
+
+object XZ2Table extends GeoMesaTable {
+
+  val SFC = new XZ2SFC(12)
```
Afterglow375: Will this be configurable?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/XZ2Table.scala
```diff
@@ -0,0 +1,94 @@
...
+      Seq(mutation)
+    }
+  }
+
+  override def getIdFromRow(sft: SimpleFeatureType): (Text) => String = {
+    val offset = if (sft.isTableSharing) 10 else 9 // table sharing + shard + 8 byte long
+    (row: Text) => new String(row.getBytes, offset, row.getLength - offset, StandardCharsets.UTF_8)
+  }
+
+  // table sharing (0-1 byte), split(1 byte), z value (8 bytes), id (n bytes)
```
Afterglow375: [picky] comment says z value but it should be xz value

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/XZ2IdxStrategy.scala
```diff
@@ -0,0 +1,142 @@
...
+    val numThreads = ds.getSuggestedThreads(sft.getTypeName, XZ2Table)
+
+    val ranges = if (filter.primary.isEmpty) {
+      if (sft.isTableSharing) {
+        Seq(aRange.prefix(new Text(sft.getTableSharingBytes)))
+      } else {
+        Seq(new aRange())
+      }
+    } else {
+      // setup Z2 iterator
```
Afterglow375: [picky] another comment that should say xz2

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,392 @@
...
+    // normalize inputs to [0,1]
+    val (nxmin, nymin, nxmax, nymax) = XZ2SFC.normalize(xmin, ymin, xmax, ymax)
+
+    // calculate the length of the sequence code (section 4.1 of XZ-Ordering paper)
+
+    val w = nxmax - nxmin
+    val h = nymax - nymin
+
+    // el-one is a bit confusing to read, but corresponds with the paper's definitions
+    val l1 = math.min(g, math.floor(math.log(math.max(w, h)) / math.log(0.5)).toInt)
```
Afterglow375: As a minor optimization, hardcode in the value of `math.log(0.5)`?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,392 @@
...
+
+    // predicate for checking how many axis the polygon intersects
+    def predicate(value: Double, wh: Double): Boolean = math.floor((value / l1) + 2) * l1 <= value + wh
```
Afterglow375: This predicate doesn't make sense to me. It's only true if l1 is 0 (which happens when `math.max(w, h)` is > .5). As such, there are several cases where the length is calculated at a finer resolution than it should be.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,392 @@
...
+
+    // note: this part of the paper is slightly confusing - do both x AND y have to satisfy the predicate?
+    // from experimental results seems that way
+    val length = if (predicate(nxmin, w) && predicate(nymin, h)) l1 else l1 + 1
```
Afterglow375: You should only need to calculate the predicate for the max of w and h.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,392 @@
...
+  */
+class XZ2SFC(g: Short) {
+
+  // TODO see what the max value of g can be where we can use Ints instead of Longs and possibly refactor to use Ints
+
+  /**
+    * Index a polygon by it's bounding box
+    *
+    * @param bounds (xmin, ymin, xmax, ymax)
+    * @return
```
Afterglow375: Document what the return types are for the methods in this class?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,392 @@
...
+
+    // values remaining to process - initial size of 100 in general saves us some re-allocation
+    val remaining = new java.util.ArrayDeque[XElement](100)
+
+    // checks if a quad is contained in the search space
+    def isContained(quad: XElement): Boolean = {
+      var i = 0
+      while (i < query.length) {
+        if (quad.isContained(query(i))) {
+          return true
```
Afterglow375: omit the `return`

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
cne1x: Generally, the compiler should be smart enough to recognize that part of the expression as constant, and perform precisely the optimization you are recommending.  It's not a bad idea, anyway.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/XZ2Table.scala
elahrvivaz: added a ticket for it here: https://geomesa.atlassian.net/browse/GEOMESA-1358

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/tables/XZ2Table.scala
elahrvivaz: still technically a z value ;)
fixed

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/XZ2IdxStrategy.scala
elahrvivaz: fixed

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: fixed

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: fixed

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: Need the return to short-circuit the evaluation of the other polygons - it's not very scala-y but it should be faster than a query.exists() I believe... @jahhulbert-ccri @jnh5y any thoughts?

jnh5y: How about 
```
def isContained(quad: XElement): Boolean = (0 to query.length).find { i => quad.isContained(query(i)) }
```

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: thanks, fixed

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: hmm, do you think the paper is wrong? or did i just implement it wrong?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: hmm, i thought Range was bad to iterate over?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,392 @@
...
+  def index(xmin: Double, ymin: Double, xmax: Double, ymax: Double): Long = {
+    // normalize inputs to [0,1]
+    val (nxmin, nymin, nxmax, nymax) = XZ2SFC.normalize(xmin, ymin, xmax, ymax)
+
+    // calculate the length of the sequence code (section 4.1 of XZ-Ordering paper)
+
+    val w = nxmax - nxmin
+    val h = nymax - nymin
+
+    val (wh, lower, upper) = if (w > h) (w, nxmin, nxmax) else (h, nymin, nymax)
```
elahrvivaz: @Afterglow375 I was thinking about this, and I believe we need to check each dimension, not just the longest. Even if the longest dimension doesn't escape the non-extended element, the shorter one might.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,394 @@
...
+    * @param length length of the sequence code that will be generated
+    * @return
+    */
+  private def sequenceCode(x: Double, y: Double, length: Int): Long = {
+    var xmin = 0.0
+    var ymin = 0.0
+    var xmax = 1.0
+    var ymax = 1.0
+
+    var cs = 1L
```
jnh5y: The expression in the paper has the +1 inside the loop.  This should 0...

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
```diff
@@ -0,0 +1,394 @@
...
+
+    var i = 0
+    while (i < length) {
+      val xCenter = (xmin + xmax) / 2.0
+      val yCenter = (ymin + ymax) / 2.0
+      (x < xCenter, y < yCenter) match {
+        case (true,  true)  =>                                                   xmax = xCenter; ymax = yCenter
+        case (false, true)  => cs += 1L * (math.pow(4, g - i).toLong - 1L) / 3L; xmin = xCenter; ymax = yCenter
+        case (true,  false) => cs += 2L * (math.pow(4, g - i).toLong - 1L) / 3L; xmax = xCenter; ymin = yCenter
+        case (false, false) => cs += 3L * (math.pow(4, g - i).toLong - 1L) / 3L; xmin = xCenter; ymin = yCenter
```
jnh5y: And and each of these should have +1.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: i originally tried it that way, but it didn't work - summation is higher priority than addition, so the + actually goes outside the sum, from what I can tell

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: same as above

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
elahrvivaz: ok, whatever changes have been made since I tried it seem to have made the difference - it works now inside the loop, so I'll update.

elahrvivaz:
I believe everything has been fixed and this is ready to merge.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala
jnh5y: Yeah, it does allocate more objects:(.  I think this should be ok as is.  Otherwise, one could use a var with false as an initial value and bail out of the loop as it is set to true and then return the variable?

elahrvivaz: there's no `break` in scala though right? so we'd have to check the value each time through the loop. admittedly, these are very small issues...

geomesa-z3/src/test/scala/org/locationtech/geomesa/curve/XZ2SFCTest.scala
```diff
@@ -0,0 +1,129 @@
...
+        (-180.0, -90.0, 8.0, 8.0),
+        (0.0, 0.0, 8.0, 8.0),
+        (9.0, 9.0, 9.5, 9.5),
+        (20.0, 20.0, 180.0, 90.0)
+      )
+      forall(containing ++ overlapping) { bbox =>
+        val ranges = sfc.ranges(Seq(bbox)).map(r => (r.lower, r.upper))
+        val matches = ranges.exists(r => r._1 <= poly && r._2 >= poly)
+        if (!matches) {
+          println(s"$bbox - no match")
```
jnh5y: printlns yo!

geomesa-z3/src/test/scala/org/locationtech/geomesa/curve/XZ2SFCTest.scala
```diff
@@ -0,0 +1,129 @@
...
+        (0.0, 0.0, 8.0, 8.0),
+        (9.0, 9.0, 9.5, 9.5),
+        (12.5, 12.5, 13.5, 13.5),
+        (20.0, 20.0, 180.0, 90.0)
+      )
+      forall(containing ++ overlapping) { bbox =>
+        val ranges = sfc.ranges(Seq(bbox)).map(r => (r.lower, r.upper))
+        val matches = ranges.exists(r => r._1 <= poly && r._2 >= poly)
+        if (!matches) {
+          println(s"$bbox - no match")
```
jnh5y: println

geomesa-z3/src/test/scala/org/locationtech/geomesa/curve/XZ2SFCTest.scala
elahrvivaz: fixed

geomesa-z3/src/test/scala/org/locationtech/geomesa/curve/XZ2SFCTest.scala
elahrvivaz: fixed

jnh5y:
Committed as 64eab5e303789ab36387238d5cf27c0ff76a41e9.

PR #1014, opened 2016-08-01
Added dependency (htrace) to install-hadoop-accumulo.sh

acerioni:
htrace is required in order for the GeoServer plugin to work with Accumulo v. 1.7

### DIFF
--- a/geomesa-tools/bin/install-hadoop-accumulo.sh
+++ b/geomesa-tools/bin/install-hadoop-accumulo.sh
@@ -21,6 +21,9 @@ guava_version="11.0.2"
 com_log_version="1.1.3"
 commons_vfs2_version="2.0"
 
+# for the GeoServer plugin to work with Accumulo 1.7 we need htrace
+htrace_version="3.1.0-incubating"
+
 base_url="https://search.maven.org/remotecontent?filepath="
 
 if [[ (-z "$1") ]]; then
@@ -49,6 +52,7 @@ else
             "${base_url}commons-logging/commons-logging/${com_log_version}/commons-logging-${com_log_version}.jar"
             "${base_url}com/google/guava/guava/${guava_version}/guava-${guava_version}.jar"
             "${base_url}org/apache/commons/commons-vfs2/${commons_vfs2_version}/commons-vfs2-${commons_vfs2_version}.jar"
+            "${base_url}org/apache/htrace/htrace-core/${htrace_version}/htrace-core-${htrace_version}.jar"
             )
 
         for x in "${urls[@]}"; do


geomesa-tools/bin/install-hadoop-accumulo.sh
```diff
@@ -49,6 +52,7 @@ else
             "${base_url}commons-logging/commons-logging/${com_log_version}/commons-logging-${com_log_version}.jar"
             "${base_url}com/google/guava/guava/${guava_version}/guava-${guava_version}.jar"
             "${base_url}org/apache/commons/commons-vfs2/${commons_vfs2_version}/commons-vfs2-${commons_vfs2_version}.jar"
+            "${base_url}org/apache/htrace/htrace-core/${htrace_version}/htrace-core-${htrace_version}.jar"
```
elahrvivaz: does this affect accumulo 1.6 at all?

acerioni: I actually don't know, as I'm using Accumulo 1.7.0. I suppose that the "extra" dependency would be ignored in case Accumulo 1.6 is used, but I agree that it would be better to test it before accepting the pull-request!

jahhulbert-ccri: 1.6 doesn't need it only 1.7 as far as I've noticed...also commons-vfs2 changes versions with 1.7

jahhulbert-ccri: @acerioni can you add an argument  for 1.6 vs 1.7 in the script perhaps and then select the list of users from that?

jahhulbert-ccri:
@jnh5y I guess that @acerioni needs to sign that Eclipse paperwork?

jnh5y:
Hi @acerioni regarding Eclipse paperwork, we have a few things detailed in the Contributing.md here: https://github.com/locationtech/geomesa/blob/master/CONTRIBUTING.md

Let me know when you sign a CLA (and for which email address).  Thanks for the contribution!

acerioni:
Hello everybody,

I've just signed a CLA, with the following address: alessandro [dot] cerioni [at] gmail [dot] com.

@jahhulbert-ccri: upon installing GeoMesa (v. 1.2.4), I realized that no other .sh script requires arguments. Instead, the user is ask to modify the script with convenient version numbers before its execution. If we wanted to be coherent with the rest of the distribution, I guess we'd better update the script so to trigger the installation of htrace only if the (major) Accumulo version is >= 1.7. What do you think?

Cheers,
Alessandro

jahhulbert-ccri:
Hi @acerioni that should work for now! thanks!

jnh5y:
Hi Alessandro, 

Thanks for signing the CLA!  

As a really, really quick option, maybe we could duplicate the script and have a 'install-hadoop-accumulo17.sh'.  Currently, this script is actually filtered by the Maven properties in the build...  because of that, I don't think it'd be easy to do filtering on the Accumulo version.  (Maybe there are options to use conditionals in the filtering...)

Thoughts?

PR #1017, opened 2016-08-02
GEOMESA-1349 Fixing writing of query stats

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
@@ -25,9 +25,7 @@ import org.locationtech.geomesa.accumulo.util.GeoMesaBatchWriterConfig
 class UsageStatWriter(connector: Connector, table: String) extends Runnable with Closeable with LazyLogging {
 
   // initial schedule
-  UsageStatWriter.executor.schedule(this, writeDelayMillis, TimeUnit.MILLISECONDS)
-
-  private val writeDelayMillis = 5000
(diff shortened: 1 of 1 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/usage/UsageStatWriter.scala
```diff
@@ -25,9 +25,7 @@ import org.locationtech.geomesa.accumulo.util.GeoMesaBatchWriterConfig
 class UsageStatWriter(connector: Connector, table: String) extends Runnable with Closeable with LazyLogging {
 
   // initial schedule
-  UsageStatWriter.executor.schedule(this, writeDelayMillis, TimeUnit.MILLISECONDS)
-
-  private val writeDelayMillis = 5000
+  UsageStatWriter.executor.schedule(this, UsageStatWriter.writeDelayMillis, TimeUnit.MILLISECONDS)
```
jnh5y: Ok, I'll bite... how does this fix things?

elahrvivaz: i think the problem was using the val writeDelayMillis before it is declared. although in that case you would think it would default to 0 and execute immediately...

jnh5y: Well, it'd execute immediately, but maybe not get scheduled?

elahrvivaz: i'm still not sure, but the run method was never getting called, not even the first time.

PR #1020, opened 2016-08-02
GEOMESA-1356 The Native API's query-builder should allow temporal bou…

cne1x:
…nds to be (in|ex)clusive

Added a new builder function and corresponding unit tests.

### DIFF
--- a/geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaQuery.java
+++ b/geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaQuery.java
@@ -12,7 +12,7 @@ import org.apache.hadoop.classification.InterfaceStability;
 import org.geotools.factory.CommonFactoryFinder;
 import org.opengis.filter.Filter;
 import org.opengis.filter.FilterFactory2;
-import org.opengis.filter.PropertyIsBetween;
+import org.opengis.filter.expression.PropertyName;
 import org.opengis.filter.spatial.BBOX;
 
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


geomesa-native-api/src/test/java/org/locationtech/geomesa/api/GeoMesaQueryTest.java
```diff
@@ -0,0 +1,86 @@
+package org.locationtech.geomesa.api;
```
jnh5y: Headers yo!  (This'll cause the mvn license:check in Travis CI to barf.)

elahrvivaz:
looks good to me

geomesa-native-api/src/main/java/org/locationtech/geomesa/api/GeoMesaQuery.java
```diff
@@ -45,11 +46,62 @@ public GeoMesaQueryBuilder within(double lx, double ly, double ux, double uy) {
...
+
             return this;
         }
 
+        /**
+         * Converts the current start- and end-point into a valid temporal filter.
+         *
+         * @return the net temporal filter that results from the current temporal bounds
+         */
+        public Filter getTemporalFilter() {
```
jnh5y: I like this implementation.

Given it, it looks like we could offer 'after' and 'before', no?

Also, those could be combined.  Builder.after(start).before(end).build would return a between.

jnh5y:
This is good to go once the headers are added.  I'd really like to see 'before' and 'after' added.  

Repeated calls to before/after can just reset the value.

jnh5y:
Looks good to me.  As soon as Travis says 'Green', let's merge it.

PR #1023, opened 2016-08-05
GEOMESA-1363 - The Native API should allow an AccumuloGeoMesaIndex to…

cne1x:
… be instantiated with an Accumulo Connector

GEOMESA-1364 The Native API should allow the user to compute the list of Accumulo tables for a feature type before the tables exist in Accumulo

Added unit tests for both of these functions.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
@@ -181,22 +181,6 @@ object AccumuloDataStoreFactory {
     security.getAuthorizationsProvider(params, auths)
   }
 
-  def createDataStore(connector: Connector, params: JMap[String, Serializable] = EmptyParams): AccumuloDataStore = {
-    val visibility = visibilityParam.lookupOpt[String](params).getOrElse("")
-
-    val tableName = tableNameParam.lookUp(params).asInstanceOf[String]
-    val connector = connParam.lookupOpt[Connector](params).getOrElse {
-      buildAccumuloConnector(params, java.lang.Boolean.valueOf(mockParam.lookUp(params).asInstanceOf[String]))
-    }
-
-    val authProvider = buildAuthsProvider(connector, params)
-    val auditProvider = buildAuditProvider(params)
-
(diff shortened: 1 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -135,6 +87,8 @@ object AccumuloDataStoreFactory {
   val DISPLAY_NAME = "Accumulo (GeoMesa)"
   val DESCRIPTION = "Apache Accumulo\u2122 distributed key/value store"
 
+  val EmptyParams: JMap[String, Serializable] = new java.util.HashMap[String, Serializable]()
```
elahrvivaz: might be good to wrap this in a `Collections.immutableMap` just to be safe

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -166,6 +120,82 @@ object AccumuloDataStoreFactory {
...
+      throw new IllegalArgumentException("Forcing empty auths is checked, but explicit auths are provided")
+    }
+    val auths: List[String] =
+      if (forceEmptyAuths || configuredAuths.length > 0) configuredAuths.toList
+      else masterAuthsStrings.toList
+
+    security.getAuthorizationsProvider(params, auths)
+  }
+
+  def createDataStore(connector: Connector, params: JMap[String, Serializable] = EmptyParams): AccumuloDataStore = {
```
elahrvivaz: since we can pass the connector in through the param map in the other createDataStore method, do we need this at all? Not sure if it's doing something slightly different.

cne1x: That's an excellent point.  I had missed that we had a parameter defined for the connection.  Looking at the ```createDataStore``` method, though, it doesn't appear as though we have explicit support for re-using a provided ```Connection```, which I will add in lieu of the explicitly retyped routine I created.  Thanks!

PR #1024, opened 2016-08-05
GEOMESA-1359 Fix KafkaDataStore addFeatures

jnh5y:


elahrvivaz:
guess we never figured out why the test was passing?

jnh5y:
Nope... indeed not.:(

PR #1029, opened 2016-08-08
GEOMESA-1369 Fixing scalatra path handling

elahrvivaz:

### DIFF
--- a/geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/PathHandlingServletRequest.scala
+++ b/geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/PathHandlingServletRequest.scala
@@ -1,0 +1,39 @@
+/***********************************************************************
+* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.web.core
+
+import javax.servlet.http.{HttpServletRequest, HttpServletRequestWrapper}
+
+import scala.annotation.tailrec
+
+/**
+  * Reverts geoserver's path handling in order to fix scalatra's servlet mapping.
+  *
+  * @see @org.geoserver.platform.AdvancedDispatchFilter
+  * @param request request to wrap
+  */
+class PathHandlingServletRequest(request: HttpServletRequest) extends HttpServletRequestWrapper(request) {
+
+  private lazy val unwrapped = PathHandlingServletRequest.unwrap(request)
+
+  override def getPathInfo: String = unwrapped.getPathInfo
+
+  override def getServletPath: String = unwrapped.getServletPath
+}
+
+object PathHandlingServletRequest {
+
+  @tailrec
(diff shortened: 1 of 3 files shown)


geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/SpringScalatraBootstrap.scala
```diff
@@ -30,9 +30,12 @@ trait GeoMesaScalatraServlet extends ScalatraServlet with LazyLogging {
...
   // https://geomesa.atlassian.net/browse/GEOMESA-1062
-  override def handle(req: HttpServletRequest, res: HttpServletResponse): Unit = req match {
-    case r: HttpServletRequestWrapper => super.handle(r.getRequest.asInstanceOf[HttpServletRequest], res)
-    case _ => super.handle(req, res)
+  override def handle(req: HttpServletRequest, res: HttpServletResponse): Unit = {
+    val wrapped = req match {
+      case r: HttpServletRequestWrapper => new PathHandlingServletRequest(r)
+      case _ => req
+    }
+    super.handle(wrapped, res)
```
jnh5y: Can you make this same change in GeoMesaStatsEndpoint.scala?

This code is copied around since the override was stomped by the other Scalatra mixins.

elahrvivaz: will do, didn't realize that

geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/SpringScalatraBootstrap.scala
```diff
@@ -30,9 +30,12 @@ trait GeoMesaScalatraServlet extends ScalatraServlet with LazyLogging {
 
   // This may be causing issues within scalatra, to paraphrase a comment: "Wrapped requests are probably wrapped for a reason."
   // https://geomesa.atlassian.net/browse/GEOMESA-1062
```
elahrvivaz: Not sure if we should close that old ticket - I haven't tested file upload support but since we're not unwrapping the request it might work now.

geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/SpringScalatraBootstrap.scala
elahrvivaz: fixed

geomesa-web/geomesa-web-core/src/main/scala/org/locationtech/geomesa/web/core/PathHandlingServletRequest.scala
```diff
@@ -0,0 +1,39 @@
...
+  private lazy val unwrapped = PathHandlingServletRequest.unwrap(request)
+
+  override def getPathInfo: String = unwrapped.getPathInfo
+
+  override def getServletPath: String = unwrapped.getServletPath
+}
+
+object PathHandlingServletRequest {
+
+  @tailrec
```
elahrvivaz: interestingly i didn't think this annotation would work, but i guess the match is considered the tail position, and the order of the case statements doesn't matter.

jnh5y: Maybe the compiler gave up and isn't actually doing anything?;)

PR #1030, opened 2016-08-08
GEOMESA-1322 Handle unsupported tilde in feature name

cfkelly:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
@@ -384,5 +384,5 @@ object MultiRowAccumuloMetadata extends AccumuloMetadataRow {
 
   override def getRange(typeName: String): Range = Range.prefix(s"$typeName~")
 
-  override def getTypeName(row: Text): String = Text.decode(row.getBytes, 0, row.find("~"))
+  override def getTypeName(row: Text): String = Text.decode(row.getBytes, 0, row.toString.lastIndexOf("~"))
 }
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/IngestCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/IngestCommand.scala
@@ -36,7 +36,6 @@ class IngestCommand(parent: JCommander) extends Command(parent) with LazyLogging
     val fmtParam = Option(params.format).flatMap(f => Try(Formats.withName(f.toLowerCase(Locale.US))).toOption)
     lazy val fmtFile = params.files.flatMap(f => Try(Formats.withName(getFileExtension(f))).toOption).headOption
     val fmt = fmtParam.orElse(fmtFile).getOrElse(Other)
-
     if (fmt == SHP) {
       val ds = new DataStoreHelper(params).getDataStore()
       params.files.foreach(GeneralShapefileIngest.shpToDataStore(_, ds, params.featureName))


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
```diff
@@ -384,5 +384,5 @@ object MultiRowAccumuloMetadata extends AccumuloMetadataRow {
 
   override def getRange(typeName: String): Range = Range.prefix(s"$typeName~")
 
-  override def getTypeName(row: Text): String = Text.decode(row.getBytes, 0, row.find("~"))
+  override def getTypeName(row: Text): String = Text.decode(row.getBytes, 0, row.toString.lastIndexOf("~"))
```
elahrvivaz: won't this just push the problem down so that `~`s in attribute names cause issues?

jnh5y: Do you have an example?  Are you saying that field names like "first~name" would break?

jnh5y:
This one definitely needs a unit test to show what would break without this change.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/IngestCommand.scala
```diff
@@ -36,7 +36,6 @@ class IngestCommand(parent: JCommander) extends Command(parent) with LazyLogging
     val fmtParam = Option(params.format).flatMap(f => Try(Formats.withName(f.toLowerCase(Locale.US))).toOption)
     lazy val fmtFile = params.files.flatMap(f => Try(Formats.withName(getFileExtension(f))).toOption).headOption
     val fmt = fmtParam.orElse(fmtFile).getOrElse(Other)
-
```
jahhulbert-ccri: undo removal of line...

jahhulbert-ccri:
Can we have a unit test that encapsulates the fix?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/GeoMesaMetadata.scala
elahrvivaz: @jnh5y exactly

jnh5y: Since the field names are separate from the typename, I'm tempted to go ahead and merge this as is.

The typename which could go wrong is tower:type from OSM.

elahrvivaz: mmm, knowingly introducing another bug doesn't seem like such a good idea (`~` in attribute names would work, currently).

elahrvivaz: @jnh5y what would be wrong with tower:type?

jnh5y: Oh, tower:type as a field name is a completely separate issue.  Our SFT parsing does something similar with ':' as a delimiter...

I don't see how we are introducing any additional bugs.  This is about splitting the rowId only, right?  Am I missing something;)?

elahrvivaz: for stats, the rowId is encoded as `s"${typeName}~{attribute}"` - so attribute names with `~` would also break things if we're not splitting on the first `~`.

elahrvivaz: maybe stats doesn't actually use this method though - let me double check.

elahrvivaz: yeah, it does use it: https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/StatsCombiner.scala#L49

jnh5y: Doh.  Looks like our life would be better if we could use \x00 as the separator...

Ok.... let's reconsider this...  typename is more configurable/change-able than a datasets field names...  What are our options:
1.  Introduce really, really complex code to sort out all the possibilities?  
2.  Not support typenames with ~'s in the name.  Throw an exception.
3.  Pull some kind of escaping?
4.  Others?

elahrvivaz: i think 2 is good for now, and we can re-consider for 1.3.0

elahrvivaz:
I'm going to close this for now as we need a more robust solution

PR #1031, opened 2016-08-09
GEOMESA-1370 XZ3 Spatio-temporal Index

elahrvivaz:
Note: stacks on top of #1012 

### DIFF
--- a/geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ3SFC.scala
+++ b/geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ3SFC.scala
@@ -8,7 +8,8 @@
 
 package org.locationtech.geomesa.curve
 
-import org.locationtech.geomesa.curve.XZ2SFC.{QueryWindow, XElement}
+import org.locationtech.geomesa.curve.TimePeriod.TimePeriod
+import org.locationtech.geomesa.curve.XZ3SFC.{QueryWindow, XElement}
 import org.locationtech.sfcurve.IndexRange
 
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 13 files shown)


geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ3SFC.scala
```diff
@@ -0,0 +1,447 @@
...
+    *
+    * Based on Definition 2 from the XZ-Ordering paper
+    *
+    * @param x normalized x value [0,1]
+    * @param y normalized y value [0,1]
+    * @param z normalized z value [0,1]
+    * @param length length of the sequence code that will be generated
+    * @return
+    */
+  private def sequenceCode(x: Double, y: Double, z: Double, length: Int): Long = {
```
Afterglow375: Should there maybe unit tests for this? and other functions in this class like `index`. I assume the math holds up here, but it's hard to be sure just by looking at it. There are unit tests which test range generation, but I'm not sure that covers this.

elahrvivaz: the unit tests for the range generation also cover this, as we index a geometry and then verify that the ranges match. See e.g. https://github.com/locationtech/geomesa/pull/1031/files/#diff-7179fffb1f698d44792fbd90ee197475R24
I'm not sure of a better way to test this since we don't really know what the output is supposed to be, other than what we coded :P

Afterglow375: Yeah hopefully it's fine. I'm only saying this because sequenceCode for XZ2 was improper but tests still passed.

PR #1032, opened 2016-08-09
FilterHelper.extractGeometries() handles null attribute parameter.

mteldridge:


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -209,9 +209,10 @@ object FilterHelper extends LazyLogging {
 
       // Note: although not technically required, all known spatial predicates are also binary spatial operators
       case f: BinarySpatialOperator if isSpatialFilter(f) =>
+        if (attribute == null) logger.warn("Geometry attribute is not set.  Will include all geometries from filter")
```
elahrvivaz: i don't think we need a logger.warn here since this is an intended use case, no?

elahrvivaz:
looks good to me

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
mteldridge: That was at Anthony's suggestion. If there are two differently named geometry attributes in the filter, then both geometries get returned which might not be what the caller intends. He figured it might be prudent to log a warning.

elahrvivaz:
@mteldridge did you sign the CLA described in CONTRIBUTING.md? Could you also amend your commit with -s to add a sign-off footer, e.g. `git commit --amend -s`

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
elahrvivaz: I guess another thing you could do would be:
`FilterHelper.propertyNames(filter, null).flatMap(FilterHelper.extractGeometries(filter, _, false))`
that would actually avoid having to change this method...

elahrvivaz: actually you just want the envelope right? another thing you could do is to call `BoundsFilterVisitor(filter)` - that might be the best solution.

mteldridge: Does `BoundsFilterVisiter.visit(filter)` recurse through the filter?

elahrvivaz: yeah, it should do exactly what you want.

mteldridge: Using the property names filter (first example).

elahrvivaz: nice. should we close this unmerged then?

elahrvivaz:
i think we have a work-around for the underlying use case - i'm going to close this. @mteldridge shout if you want it re-opened

PR #1033, opened 2016-08-10
Gt15.1

jnh5y:


jnh5y:
The CQs for GT 15.1 and JAI-Ext 1.0.9 have been approved.

cne1x:
Looks good to me; we have tested this out, and it appears to work.

PR #1035, opened 2016-08-10
GEOMESA-1108 Pull version for docs from pom.xml properly

mdzimmerman:
Setting versions via Sphinx command line was not doing all of the
other substitions properly. This solution sets the versions via
the maven-resources-plugin prior to Sphinx invocation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,8 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
         <!-- bump version right before performing releases for the README and the documentation -->
-        <geomesa.release.version>1.2.4</geomesa.release.version>
+        <geomesa.release.version>1.2.5</geomesa.release.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -48,7 +48,8 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
         <!-- bump version right before performing releases for the README and the documentation -->
-        <geomesa.release.version>1.2.4</geomesa.release.version>
+        <geomesa.release.version>1.2.5</geomesa.release.version>
```
elahrvivaz: are we still using this or can we get rid of it?

elahrvivaz:
looks good to me

pom.xml
mdzimmerman: It's still used in ``build/README.md``.

PR #1041, opened 2016-08-12
GEOMESA-1378 Fixing serialization of feature IDs in m/r output

elahrvivaz:


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaOutputFormat.scala
```diff
@@ -140,10 +144,15 @@ class GeoMesaRecordWriter(params: Map[String, String],
     val stats = statsCache.getOrElseUpdate(sftName, ds.stats.statUpdater(sft))
 
     val withFid = AccumuloFeatureWriter.featureWithFid(sft, value)
-    val encoder = encoderCache.getOrElseUpdate(sftName, SimpleFeatureSerializers(sft, ds.getFeatureEncoding(sft)))
+    val serializer = serializerCache.getOrElseUpdate(sftName,
```
jnh5y: Do we want to extract this as utility function?  (And when?)

I'd be fine with a (yes, later) answer.  If so, then maybe we should add a ticket?

elahrvivaz: fixed

PR #1043, opened 2016-08-12
GEOMESA-1384 Verifying calls to geometry.getCentroid

elahrvivaz:
Note: this bug is likely to creep back in unless we are super-vigilant about not calling getCentroid. Probably it should be fixed in geotools.

### DIFF
--- a/geomesa-tools/conf/sfts/twitter/reference.conf
+++ b/geomesa-tools/conf/sfts/twitter/reference.conf
@@ -177,7 +177,7 @@ geomesa  {
 
         { name = polyWkt, transform = "concat( 'polygon((', mkstring(',', $c0, $c1, $c2, $c3, $c4), '))')"}
 
-        { name = geom, transform = "cql:centroid(polygon($polyWkt))" }
+        { name = geom, transform = "cql:centroid(cql:envelope(polygon($polyWkt)))" }
       ]
     }
   }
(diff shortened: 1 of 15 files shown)


geomesa-tools/conf/sfts/twitter/reference.conf
```diff
@@ -177,7 +177,7 @@ geomesa  {
 
         { name = polyWkt, transform = "concat( 'polygon((', mkstring(',', $c0, $c1, $c2, $c3, $c4), '))')"}
 
-        { name = geom, transform = "cql:centroid(polygon($polyWkt))" }
+        { name = geom, transform = "cql:centroid(cql:envelope(polygon($polyWkt)))" }
```
jnh5y: @jahhulbert-ccri this is the fix I was suggesting to you.

jnh5y:
+1

jahhulbert-ccri:
looks good to me. yeah I think we should really contact geotools and put in a ticket to remove this when its fixed since its not intuitive to call envelope before centroid...

jnh5y:
@jahhulbert-ccri I agree.  The problem is actually in getCentroid in JTS.  I've created https://geomesa.atlassian.net/browse/GEOMESA-1387 and assigned it to myself.  This workaround will be in place for awhile...

jahhulbert-ccri:
cool, i'll merge it...anything else we need that you know of?

jnh5y:
Thanks, any additional testing before the release would be great; I don't think we've changed anything 'dangerous' since Friday...

jahhulbert-ccri:
yup i will rerun...i'm doing backwards compat upgrade testing...so it kinda implicitly tests some stuff

PR #1044, opened 2016-08-12
GEOMESA-1385 Allow for override of reserved words in schema

elahrvivaz:


jnh5y:
+1

PR #1049, opened 2016-08-22
GEOMESA-1391 Use Saxon library for faster xml processing

anthonyccri:
- Uses the Saxon library for better xml and xpath
  processing
- Enhanced tests to show namespace wildcards

Signed-off-by: Anthony Fox anthonyfox@ccri.com


geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XMLConverterTest.scala
```diff
@@ -246,14 +246,14 @@ class XMLConverterTest extends Specification {
...
-          |   feature-path = "Feature" // can be any xpath - relative to the root, or absolute
+          |   feature-path = "*:Feature" // can be any xpath - relative to the root, or absolute
           |   xsd          = "xml-feature.xsd" // looked up by class.getResource
           |   fields = [
           |     // paths can be any xpath - relative to the feature-path, or absolute
-          |     { name = "number", path = "number",           transform = "$0::integer" }
-          |     { name = "color",  path = "color",            transform = "trim($0)" }
-          |     { name = "weight", path = "physical/@weight", transform = "$0::double" }
-          |     { name = "source", path = "/doc/DataSource/name/text()" }
+          |     { name = "number", path = "*:number",           transform = "$0::integer" }
```
jahhulbert-ccri: are these backwards compat with the old fields?

anthonyccri: Yes.

elahrvivaz:
+1

geomesa-convert/geomesa-convert-xml/pom.xml
```diff
@@ -20,6 +20,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>net.sf.saxon</groupId>
+            <artifactId>Saxon-HE</artifactId>
+            <version>9.7.0-7</version>
```
elahrvivaz: this could be a problem:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=5368
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=1903

@jnh5y think it's still worth opening a CQ?

jnh5y:
@elahrvivaz, 7 years is a long time.  Maybe they fixed the IP issue?

We either have to push Eclipse legal to review or we have to identify another library.

elahrvivaz:
doesn't sound like they have identified everyone: http://www.saxonica.com/documentation/#!conditions/contributors

geomesa-convert/geomesa-convert-xml/pom.xml
elahrvivaz: opened CQ here: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11876

jahhulbert-ccri:
@anthonyccri here's an idea...make the implementation configurable so you can set in the geomesa config the XML driver you want to use? I could mock that up that way we can have documentation on how to do fast XML processing.

jnh5y:
@jahhulbert-ccri, if this is doable, we may be able to provide a script which downloads Saxon.  Potentially, the solution should check for the presence of an alternative and *just* use it.  Otherwise, the script could set some config which indicates that the downloaded Saxon should be used.

Basically, I'd like folks to be able to turn on the turbo mode easily:).

elahrvivaz:
I've made saxon optionally load from the classpath - alternate implementations can be specified by the system/geomesa prop `geomesa.convert.xpath.provider`. Also added an install-saxon script to tools

elahrvivaz:
had to roll back test changes as the enhanced xpath isn't supported by the default impl

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -31,6 +31,13 @@
     </property>
 
     <property>
+        <name>geomesa.convert.xpath.provider</name>
+        <value>net.sf.saxon.xpath.XPathFactoryImpl</value>
```
jnh5y: Uhmm... does this deserve a comment here in the template?

jnh5y:
@elahrvivaz should this be documented?

Second, what is everyone thinking in terms of getting this into 1.3.0?

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
elahrvivaz: I'll add more info and add it to the docs too

elahrvivaz:
moved the xpath configuration into the converter definition instead of a system property. added section on xml in converter docs.

elahrvivaz:
tested using the geomesa tools and the example-xml with and without saxon in the lib folder

PR #1050, opened 2016-08-22
Quick fix - not for merge - Spark hack

jahhulbert-ccri:

### DIFF
--- a/geomesa-compute/pom.xml
+++ b/geomesa-compute/pom.xml
@@ -98,7 +98,7 @@
             <groupId>com.clearspring.analytics</groupId>
             <artifactId>stream</artifactId>
             <version>2.9.2</version>
-            <scope>provided</scope>
+            <scope>compile</scope>
             <exclusions>
                 <exclusion>
                     <groupId>it.unimi.dsi</groupId>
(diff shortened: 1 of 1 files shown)


jahhulbert-ccri:
@elahrvivaz @jnh5y we have some options:
1. Shade like this (works as long as the compute jar is used) (doesn't work for custom build spark analytics!)
2. copy/paste clearspring classes and relocate them to org.locationtech.geomesa.xyz in our codebase

Also,  spark uses 2.7.0 and we use 2.9.2

Thoughts? Would like to get something out soon.

Andrew

jnh5y:
@jahhulbert-ccri does this work as an option:
3.  Rollback our dependency to 2.7.0 to match Spark?

Option #2 might require special permission from Eclipse.  We usually are not asking to distribute code; there might be some way to do that...

jahhulbert-ccri:
@jnh5y this issue is not that its not matching.

jahhulbert-ccri:
we should prob also roll back though...

elahrvivaz:
is it worth trying to contribute the changes back to clearspring? (if we're downgrading versions maybe not)

jahhulbert-ccri:
@elahrvivaz @jnh5y what if we go the same way as we did with geotools which is copy in the class, modify it, attribute it, and then put in a bug to upgrade?

jnh5y:
@jahhulbert-ccri copying code and patching in that manner was ok for a bug-fix release since we weren't going through a full Eclipse review.  (I.e., that is a compromise kinda thing.)

I don't think it'll work out for GeoMesa 1.3.0 (which would be fully reviewed).

I think that 1) making a request to Clearspring (with a reason) would be fair.  

At the moment, how bad is using reflection?  It ain't a 'good' option, but I have a feeling it might be the best among weak options...

jahhulbert-ccri:
I put up a PR for that: https://github.com/locationtech/geomesa/pull/1051 i need to fix the header i guess. I think it works well enough

PR #1051, opened 2016-08-22
GEOMESA-1396 RichCountMinSketch doesn't work with Spark classloaders

jahhulbert-ccri:


jahhulbert-ccri:
need to verify the logic...

@jnh5y take a look

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RichCountMinSketch.scala
```diff
@@ -0,0 +1,108 @@
...
+  private val tableField = classOf[CountMinSketch].getDeclaredField("table")
+  tableField.setAccessible(true)
+
+  private val sizeField = classOf[CountMinSketch].getDeclaredField("size")
+  sizeField.setAccessible(true)
+
+  private val hashAField = classOf[CountMinSketch].getDeclaredField("hashA")
+  hashAField.setAccessible(true)
+
+  def refWidth(cms: CountMinSketch) = widthField.getInt(cms)
```
elahrvivaz: maybe rename these to 'reflectFoo' for clarity. @cne1x amiright?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RichCountMinSketch.scala
```diff
@@ -0,0 +1,108 @@
...
+      return false
+    }
+    var i = 0
+    while (i < depth) {
+      if (refHashA(sketch)(i) != refHashA(other)(i)) {
+        return false
+      }
+      val t = table
+      val row = table(i)
+      val otherRow = refTable(other)(i)
```
elahrvivaz: since these are all declaredFields, I'm guessing repeated calling doesn't really have a performance penalty? or would it be better to call once and store to a local val?

elahrvivaz:
looks good

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RichCountMinSketch.scala
jahhulbert-ccri: i'll change em...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/RichCountMinSketch.scala
jahhulbert-ccri: good question...it should be faster to store them than dereference every time...

jahhulbert-ccri: actually i guess the issue i was thinking about is if the underlying reference changed...

jahhulbert-ccri:
I dropped this to 2.7.0 to match spark...note that it appears to be compatible but I wouldn't swear by it...we should probably test upgrade as @elahrvivaz mentioned...If we don't drop to 2.7.0 we need to shade it. thoughts?

elahrvivaz:
we already shade guava and commons-codec in geomesa-compute - adding one more might not be an issue...

jahhulbert-ccri:
closing in favor of the shaded version https://github.com/locationtech/geomesa/pull/1055

PR #1063, opened 2016-08-31
GEOMESA-1320 Command line blasts terminal with errors on ingest

aheyne:
I'm not sure if this is the correct way to address this issue but it should redirect any errors thrown by BinnedArrays to the geomesa log.

### DIFF
--- a/geomesa-tools/conf/log4j.properties
+++ b/geomesa-tools/conf/log4j.properties
@@ -4,7 +4,9 @@ log4j.rootLogger=info,file
 geomesa.log.dir=.
 
 # set logging levels to appropriate values
-log4j.logger.org.locationtech.geomesa.tools=info
+log4j.logger.org.locationtech.geomesa.tools.accumulo.ingest=info,file
+log4j.additivity.org.locationtech.geomesa.tools.accumulo.ingest=false
+log4j.logger.org.locationtech.geomesa.tools=info,stderr
 log4j.logger.org.locationtech.geomesa=info
 log4j.logger.org.apache.zookeeper=warn
 log4j.logger.hsqldb.db=warn
--- a/geomesa-tools/conf/log4j.properties
+++ b/geomesa-tools/conf/log4j.properties
@@ -4,7 +4,7 @@ log4j.rootLogger=info,file
 geomesa.log.dir=.
 
 # set logging levels to appropriate values
-log4j.logger.org.locationtech.geomesa.tools=info,stderr
+log4j.logger.org.locationtech.geomesa.tools=info
 log4j.logger.org.locationtech.geomesa=info
 log4j.logger.org.apache.zookeeper=warn
 log4j.logger.hsqldb.db=warn


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -25,7 +26,7 @@ import scala.reflect.ClassTag
   * @param bounds upper and lower bounds for the input values
   * @tparam T type of input value
   */
-abstract class BinnedArray[T](val length: Int, val bounds: (T, T)) {
+abstract class BinnedArray[T](val length: Int, val bounds: (T, T)) extends LazyLogging {
```
elahrvivaz: @aheyne have you verified this as a fix? i don't think adding the logging trait will do anything, as that only exposes a logger object you can use afaik.

jnh5y: Ah, maybe the logger is designed to have errors go to the log file instead of stdout?

jahhulbert-ccri: perhaps require() send to std out before

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -131,13 +132,17 @@ abstract class WholeNumberBinnedArray[T](length: Int, bounds: (T, T)) extends Bi
   private val min = convertToLong(bounds._1)
   private val max = convertToLong(bounds._2)
 
-  require(min < max,
-    s"Upper bound must be greater than lower bound: lower='${bounds._1}'($min) upper='${bounds._2}'($max)")
+  if(min > max) {
```
jahhulbert-ccri: [style] space after ```if``` and before ```(```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -131,13 +132,17 @@ abstract class WholeNumberBinnedArray[T](length: Int, bounds: (T, T)) extends Bi
...
+    val e = new IllegalArgumentException
```
jahhulbert-ccri: doesn't look idiomatic...maybe just throw the error instead of logging/throwing:
```
if (min > max) {
  throw new IllegalArgumentException(msg...)
}
```

and make the call handle it...else do:
```
if (min > max) {
  val msg = "msg here"
  logger.error(msg)
  throw new IAE(msg)
}
```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -229,8 +234,11 @@ class BinnedGeometryArray(length: Int, bounds: (Geometry, Geometry))
 
 class BinnedFloatArray(length: Int, bounds: (jFloat, jFloat)) extends BinnedArray[jFloat](length, bounds) {
 
-  require(bounds._1 < bounds._2,
-    s"Upper bound must be greater than lower bound: lower='${bounds._1}' upper='${bounds._2}'")
+  if(bounds._1 > bounds._2){
```
jahhulbert-ccri: space again

jahhulbert-ccri:
actaully...require does the same thing...we shouldn't need to do this change it's not the root cause

jahhulbert-ccri:
looking at the scala source:
```
  /** Tests an expression, throwing an `IllegalArgumentException` if false.
   *  This method is similar to `assert`, but blames the caller of the method
   *  for violating the condition.
   *
   *  @param requirement   the expression to test
   *  @param message       a String to include in the failure message
   */
  @inline final def require(requirement: Boolean, message: => Any) {
    if (!requirement)
      throw new IllegalArgumentException("requirement failed: "+ message)
  }
```

aheyne:
I think your right. This seemed very spit and glue as require is basically what I was replacing it with. What I don't get is why stderr isn't going to the log file in this situation. I'm going to reset this branch and go in a new direction.

jahhulbert-ccri:
it may be useful to duplicate the error with a 1 line twitter file and then you can use the debugger to catch it and see why

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/BinnedArray.scala
```diff
@@ -131,13 +132,17 @@ abstract class WholeNumberBinnedArray[T](length: Int, bounds: (T, T)) extends Bi
   private val min = convertToLong(bounds._1)
   private val max = convertToLong(bounds._2)
 
-  require(min < max,
-    s"Upper bound must be greater than lower bound: lower='${bounds._1}'($min) upper='${bounds._2}'($max)")
+  if (min > max) {
+    val msg = s"Upper bound must be greater than lower bound: lower='${bounds._1}'($min) upper='${bounds._2}'($max)"
+    logger.error(msg)
+    throw new IllegalArgumentException(msg)
```
elahrvivaz: aren't these are still going to spam the console? now it will just log as well. (`require` throws an IllegalArgumentException)

jahhulbert-ccri:
any updates on this?

jahhulbert-ccri:
we can hold it off for a while

aheyne:
I'm working on it now, no updates though.

aheyne:
This working properly now from my tests.

geomesa-tools/conf/log4j.properties
```diff
@@ -4,7 +4,7 @@ log4j.rootLogger=info,file
 geomesa.log.dir=.
 
 # set logging levels to appropriate values
-log4j.logger.org.locationtech.geomesa.tools=info,stderr
+log4j.logger.org.locationtech.geomesa.tools=info
```
jahhulbert-ccri: hahaha i bet this will gitblame to either me or @elahrvivaz that's funny that its the fix

elahrvivaz: ah, i think we should keep stderr actually - i think we use it for messages to the user

elahrvivaz: this is only for the 'tools' package - maybe we can add a more specific one for the ingestion package?

jahhulbert-ccri: good catch

aheyne:
This now sends errors from geomesa.tools.accumulo.ingest to only the log file and keeps geomesa.tools error in the console.

PR #1066, opened 2016-09-01
GEOMESA-1407 Fix Accumulo quick start instructions about time dimension.

aheyne:
Accumulo quickstart says to configure the time dimension. Doing this only allows one point to be visible, diverging experience from the example results.


aheyne:
Actually, I need to change some other references in the file. I'll re-pr this later.

PR #1071, opened 2016-09-05
GEOMESA-1417 Allow custom quote characters in DelimitedTextConverter

anthonyccri:
* Added an optional field, 'options.quote', that controls what character
  is used for quoting


geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
```diff
@@ -46,20 +46,24 @@ class DelimitedTextConverterFactory extends AbstractSimpleFeatureConverterFactor
...
+    conf.getStringOpt("options.quote").foreach { q =>
+      baseFmt = baseFmt.withQuote(q.toCharArray()(0))
+    }
+
     new DelimitedTextConverter(baseFmt, sft, idBuilder, fields, userDataBuilder, opts, validating)
   }
 }
 
-class DelimitedOptions(var skipLines: Int = 0, var pipeSize: Int = 16 * 1024)
+class DelimitedOptions(var skipLines: Int = 0, var pipeSize: Int = 16 * 1024, var quote: String = "\"")
```
elahrvivaz: is this actually used, or could it be removed?

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
```diff
@@ -46,20 +46,24 @@ class DelimitedTextConverterFactory extends AbstractSimpleFeatureConverterFactor
...
     val opts = {
-      import org.locationtech.geomesa.utils.conf.ConfConversions._
       val o = "options"
       val dOpts = new DelimitedOptions()
       conf.getIntOpt(s"$o.skip-lines").foreach(s => dOpts.skipLines = s)
       conf.getIntOpt(s"$o.pipe-size").foreach(p => dOpts.pipeSize = p)
       dOpts
     }
 
+    conf.getStringOpt("options.quote").foreach { q =>
```
jahhulbert-ccri: [maybe not important] if we wanted to disable quotes what would we do?

elahrvivaz: i believe most of the formats don't consider quotes - just the 'QUOTE*' one

jahhulbert-ccri:
maybe in the future we should also allow a "escape" char so we can enable backslash escaping.

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
jahhulbert-ccri: cool

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
jahhulbert-ccri: doesn't seem to be used

elahrvivaz:
looks like MYSQL is the only format that sets an escape char (`\`), so yeah exposing that would be good

jahhulbert-ccri:
@anthonyccri can you add an argument for escaping? maybe ```options.escape```

jahhulbert-ccri:
@elahrvivaz if you are ok with my additions we're good to go

elahrvivaz:
looks good - merging

PR #1076, opened 2016-09-06
GEOMESA-1282 Docs: generalize build_and_mirror.sh script

mdzimmerman:

### DIFF
--- a/docs/bin/build-and-mirror.sh
+++ b/docs/bin/build-and-mirror.sh
@@ -2,17 +2,44 @@
 #
 # Builds the HTML documentation tree and mirrors it to the geomesa.github.io repository.
 
-# IMPORTANT: Update these paths for your system!
+# default paths for docs, can be overridden by command line switches
 GEOMESA=/opt/devel/src/geomesa
 GEOMESA_GITHUB_IO=/opt/devel/src/geomesa.github.io
 DOCS=$GEOMESA_GITHUB_IO/documentation
 
+function check_repo {
+    # check that repo in $dir is a Git repo and has $url as a remote
+    dir="$1"
+    url="$2"
+    echo "check_repo $dir $url"
+    if [[ ! -d "$dir" ]] ; then
+        echo "ERROR: $dir not a directory"
+        exit 3
+    fi
+    if [[ ! -d "$dir/.git" ]] ; then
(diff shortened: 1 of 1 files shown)


docs/bin/build-and-mirror.sh
```diff
@@ -27,9 +54,58 @@ function build_and_mirror {
...
+    echo "        --gmweb <dir>    set \$GEOMESA_GITHUB_IO [$GEOMESA_GITHUB_IO]"
+    echo
+}
+
+### MAIN ###
+
+# parse options
+while :
+do
+   case "$1" in
```
jahhulbert-ccri: maybe use ```getopts``` instead ?

docs/bin/build-and-mirror.sh
```diff
@@ -2,17 +2,44 @@
 #
 # Builds the HTML documentation tree and mirrors it to the geomesa.github.io repository.
 
-# IMPORTANT: Update these paths for your system!
+# default paths for docs, can be overridden by command line switches
 GEOMESA=/opt/devel/src/geomesa
```
jahhulbert-ccri: Also, these should respect existing ENV entries like this:
```GEOMESA=/path/to GEOMESA_GITHUB_IO=/path/to/this build_and_mirror.sh``` so that devs can set it on the command line or in their bashrc env

I actually prefer not to have /opt/devel in here at all :angel:

jahhulbert-ccri:
a few improvements but overall ok

jahhulbert-ccri:
@mdzimmerman lemme know if you can address these by tomorrow?

mdzimmerman:
Yeah, I'll do it today.

docs/bin/build-and-mirror.sh
mdzimmerman: ``getopts`` doesn't support long command line options.

docs/bin/build-and-mirror.sh
mdzimmerman: Alright, done.

docs/bin/build-and-mirror.sh
jahhulbert-ccri: ah that sucks ok

PR #1058, opened 2016-08-24
WIP Added addindex command to geomesa command line tools

Afterglow375:

PR #1078, opened 2016-09-06
GEOMESA-1421 Better handling of optional arguments for the Accumulo GeoMesa command line tools

Afterglow375:


elahrvivaz:
+1

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/Utils.scala
```diff
@@ -8,7 +8,7 @@
 
 package org.locationtech.geomesa.tools.accumulo
 
-import java.io.{BufferedReader, File, InputStreamReader}
+import java.io.{FileNotFoundException, BufferedReader, File, InputStreamReader}
```
jahhulbert-ccri: can you organize imports here?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/AccumuloParams.scala
```diff
@@ -43,9 +45,45 @@ trait AccumuloConnectionParams extends OptionalZookeepersParam {
...
+
+  /**
+    * Get a handle to a datastore for a pre-existing catalog table
+    *
+    * @throws Exception if the catalog table does not exist in accumulo
+    */
+  def createDataStore(): AccumuloDataStore = {
+    import scala.collection.JavaConversions._
+    Option(DataStoreFinder.getDataStore(dataStoreParams).asInstanceOf[AccumuloDataStore]).getOrElse {
+      throw new Exception("Could not load a data store with the provided parameters: " +
```
jahhulbert-ccri: maybe an illegal argument exception instead of vanilla exception? Not sure if it really matters but I guess generally its nice to have it be IAE for the java world

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/AccumuloParams.scala
```diff
@@ -25,7 +27,7 @@ trait AccumuloRasterTableParam {
   var table: String = null
 }
 
-trait AccumuloConnectionParams extends OptionalZookeepersParam {
```
jahhulbert-ccri: The only downside here is that now we are mixing the params classes with some logic for the environment...may not matter...I thought we separated it out earlier so we could use these standalone but not sure...let me think

elahrvivaz: we could move this logic outside the trait - the main takeaway we wanted from this PR is having the parameters set in the trait before we start processing. that's being accomplished by the invocation in `Runner.createCommand` - as long as something equivalent happens there i'm happy.

jahhulbert-ccri: cool

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/TableConfCommand.scala
```diff
@@ -114,7 +114,7 @@ object TableConfCommand {
     @Parameter(names = Array("-t", "--table-suffix"), description = "Table suffix to operate on (attr_idx, st_idx, or records)", required = true)
     var tableSuffix: String = null
 
-    lazy val ds = new DataStoreHelper(this).getDataStore()
+    lazy val ds = DataStoreParamsHelper.createDataStore(this)
     lazy val tableName = getTableName(ds, this)
```
elahrvivaz: while we're changing this, is ds accessed anywhere else? it should be disposed. i would change to:
```
lazy val tableName = {
  val ds = DataStoreParamsHelper.createDataStore(this)
  val name = getTableName(ds, this)
  ds.dispose()
  name
}
```
you'll need to fix `getTableName` to use the `ds` object that's passed in directly instead of the one from `params`

jahhulbert-ccri:
is this 1.2.7?

elahrvivaz:
@jahhulbert-ccri don't think so - it's some precursor work to #1058

elahrvivaz:
@Afterglow375 this ended up getting merged in as part of #1082 - sorry to not get it as a separate commit :/

PR #1079, opened 2016-09-07
GEOMESA-1283 Multiple Distributions

jahhulbert-ccri:
* Combine the tools into the distribution so there is a single tar to unzip (a la accumulo, kafka, spark, hadoop, everyone else)
* Create separate distributions/tools tarballs for accumulo, kafka, hbase, etc.
* Create separate tarballs for kafka 08/09 distros
* Rename kafka artifacts to 08

TODO:
* Remove tarballs from maven build profile and have them upload to static website space instead of plopping them into nexus
* Provide a make distribution script or something to allow user to bundle accumulo/hadoop/kafka jars (for specific versions) with the tools/distribution similar to spark (also allows provided)

SEE TICKET for more info: https://geomesa.atlassian.net/browse/GEOMESA-1283


geomesa-dist/geomesa-dist-accumulo/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+
+    <artifactId>geomesa-dist-accumulo</artifactId>
+    <name>GeoMesa Accumulo Distribution</name>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <dependencies>
+        <!-- Tools -->
```
cne1x: +1 for commenting why the various dependencies are here

geomesa-dist/geomesa-dist-kafka/src/main/assemblies/component.xml
```diff
@@ -0,0 +1,100 @@
...
+            <includes>
+                <include>org.locationtech.geomesa:geomesa-kafka-gs-plugin*</include>
+            </includes>
+        </dependencySet>
+        <dependencySet>
+            <outputDirectory>lib</outputDirectory>
+            <directoryMode>755</directoryMode>
+            <fileMode>0644</fileMode>
+            <useProjectArtifact>false</useProjectArtifact>
+            <useTransitiveDependencies>true</useTransitiveDependencies>
```
cne1x: The only two places where we use transitive dependencies are here and in the corresponding spot for the Accumulo distribution.  We should test this carefully, because we've seen this cause problems before, at least when packaging up GeoServer's WEB-INF lib, where it dragged in conflicting versions of things.

geomesa-dist/geomesa-dist-bigtable/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+
+    <artifactId>geomesa-dist-bigtable</artifactId>
+    <name>GeoMesa BigTable Distribution</name>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+
+    <profiles>
```
cne1x: Do we need this profile here if we have it defined in the parent POM?  (Assuming the parent POM only includes this module if the profile is active, that is.)

geomesa-gs-plugin/geomesa-bigtable-gs-plugin/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>geomesa-gs-plugin</artifactId>
         <groupId>org.locationtech.geomesa</groupId>
-        <version>1.2.0-SNAPSHOT</version>
+        <version>1.2.7-SNAPSHOT</version>
```
cne1x: Arent' we using ```${project.version}``` elsewhere, rather than hard-coding this, or is this a special case?

geomesa-hbase/geomesa-bigtable-datastore/pom.xml
```diff
@@ -13,7 +13,7 @@
     <parent>
         <artifactId>geomesa-hbase</artifactId>
         <groupId>org.locationtech.geomesa</groupId>
-        <version>1.2.0-SNAPSHOT</version>
+        <version>1.2.7-SNAPSHOT</version>
```
cne1x: ... similar question about hard-coded version.

geomesa-tools/geomesa-tools-accumulo/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-compiler</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-csv</artifactId>
+        </dependency>
+
+        <!-- Kafka CLI dependencies -->
```
cne1x: Do we still need these Kafka dependencies for the Accumulo tools, or should those be specific to the Kafka tools?

geomesa-tools/geomesa-tools-common/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+        </dependency>
+
+        <!-- needed in the lib dir -->
+        <dependency>
+            <groupId>org.scala-lang.modules</groupId>
+            <artifactId>scala-xml_2.11</artifactId>
+        </dependency>
+
+        <!-- provided dependencies -->
+        <dependency>
```
cne1x: Wouldn't Accumulo and Hadoop dependencies not be needed by the common tools components?

cne1x:
I have a few questions/comments posted here, but this looks good to me.  Thanks very much, @jahhulbert-ccri !

geomesa-dist/geomesa-dist-bigtable/pom.xml
jahhulbert-ccri: yeah probably not...i'll investigate

geomesa-gs-plugin/geomesa-bigtable-gs-plugin/pom.xml
jahhulbert-ccri: special only bc its in a parent tag...all the parent tag references are updated by maven release:prepare the reason it was off before is bc our build for releases doesn't use the bigtable profile

geomesa-hbase/geomesa-bigtable-datastore/pom.xml
jahhulbert-ccri: see above

geomesa-tools/geomesa-tools-accumulo/pom.xml
jahhulbert-ccri: nope...left over from the copypaste

geomesa-tools/geomesa-tools-common/pom.xml
jahhulbert-ccri: yeah shouldn't be i think that's a copy/paste err

geomesa-dist/geomesa-dist-kafka/src/main/assemblies/component.xml
jahhulbert-ccri: Yeah, actually commons-logging conflicts alot too :) minor versions but it does

jahhulbert-ccri:
I have some kafka 08/09 stuff coming

elahrvivaz:
do you think it makes any sense to re-organize by distro? so have a layout like:
```
geomesa-accumulo/
geomesa-accumulo/geomesa-accumulo-datastore
geomesa-accumulo/geomesa-accumulo-distributed-runtime
geomesa-accumulo/geomesa-accumulo-gs-plugin
geomesa-accumulo/geomesa-accumulo-tools
```
I'm conflicted between keeping things under functional groups or distro groups...

jahhulbert-ccri:
@elahrvivaz this is what it would look like moving things around...what do you think?

elahrvivaz:
@jahhulbert-ccri i like it

jahhulbert-ccri:
@elahrvivaz sounds good the rest are coming. hopefully  jhugs won't mind :)

jahhulbert-ccri:
@jnh5y i moved the gs plugins back...i need to verify that the dists are happy and have all the correct jars for tools and stuff i'm going to compare them to the old dists from 1.2.6

PR #1083, opened 2016-09-09
GEOMESA-1425 Accept [Enter] in read prompts

aheyne:
All command line tools scripts now accept [Enter] for the indicated default value when prompting the user for input.


jahhulbert-ccri:
looks good

PR #1091, opened 2016-09-14
GEOMESA-1429 Replace Scala StdIn

jahhulbert-ccri:


jahhulbert-ccri:
needed for scala 2.10/2.11 compat

elahrvivaz:
looks good. as a further improvement, we have some common methods in the `Prompt` class that might replace some of these.

PR #1099, opened 2016-09-15
GEOMESA-1435 Add terminating newlines to pom.xml if missing

tkunicki:
Results of command:
```
for x in `find . -name pom.xml`; do if [ -n "$(tail -c 1 <"$x")" ]; then echo >>"$x"; fi; done
```

This is to make sure last line in file is actually a line per [POSIX standards](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).


elahrvivaz:
do we need a ticket or anything for this? what's the rationale - better sed output?

tkunicki:
POSIX compliant tooling expects a trailing newline and will typically insert one if missing.  Having a trailing newline will allow for consistency in operations with these tools.  The use of `sed` in our scala version change script will add a newline, if missing.  Making sure our `pom.xml` have a newline will reduce listing of spurious (non functional) changes with `git diff` after files are touched by POSIX compliant tooling.

PR #1101, opened 2016-09-19
GEOMESA-1437 Cleaning up test logging

elahrvivaz:


geomesa-utils/src/test/resources/log4j.xml
```diff
@@ -2,8 +2,8 @@
 <!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
 <log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/" debug="false">
     <appender name="CONSOLE" class="org.apache.log4j.ConsoleAppender">
-        <layout class="org.apache.log4j.PatternLayout">
-            <param name="ConversionPattern" value="[%d] %5p %c: %m%n"/>
+        <layout class="org.apache.log4j.EnhancedPatternLayout">
+            <param name="ConversionPattern" value="[%d] %5p %c{1}: %m%n"/>
```
jnh5y: Do we risk any classname collisions/confusion with only printing the classname without any additional info?

elahrvivaz: possibly - it's just for unit tests though so in general I think it would be ok. It makes the logs a lot shorter. We could also make it use just a single char per package like `o.l.g.a.d.AccumuloDataStore` if you'd prefer that.

jnh5y: @elahrvivaz, ah! I missed that it was for unit tests only.

Does this need to be applied to 1.2 and 1.3?

elahrvivaz: probably not for 1.2 unless someone really wants it

PR #1111, opened 2016-09-26
GEOMESA-1224 Geomesa gs plugin install/uninstall script

aheyne:
Added script to allow easy installation and uninstallation of geomesa's geoserver plugins and provides information on versions.


geomesa-tools-common/bin/validate-jars.sh
```diff
@@ -0,0 +1,403 @@
...
+function removeJars() {
+  install_dir="${1}"
+  jars=("${@}")
+  echo "Jars to be removed:"
+  for jar in "${jars}"; do
+    echo "${jar}"
+  done
+  read -p "Are you sure you want to Continue? [Y/n]" confirm
+  confirm=${confirm,,} #lowercasing
+  if [[ $confirm =~ ^(yes|y) || $confirm == "" ]]; then
```
elahrvivaz: maybe we should count empty string as a cancel out of an abundance of caution

aheyne:
This should be for 1.3.x only. 

In testing I discovered that stripping the version from the old kafka jars doesn't work quite right due to how they were named. 
e.g. Both geomesa-kafka-09-datastore-1.2.6.jar and geomesa-kafka-09-gs-plugin-1.2.6.jar are stripped to geomesa-kafka. 
This shouldn't cause an issue for uninstalling or upgrading old versions of the plugins since they're just being removed but this causes ambiguity when repairing or installing.

aheyne:
Renamed the Accumulo dist/geoserver folder to dist/gs-plugins so that it matches the kafka distribution and to allow agnostic scripts to predict the location of the geoserver plugins.

PR #1119, opened 2016-09-29
GEOMESA-1453 extract and use profiles and system properties when determining scala version

tkunicki:


elahrvivaz:
@tkunicki ready for merge?

tkunicki:
Ends up we can remove all scala version checks with regard to our use of the zinc server.  Further investigation by examining the results `zinc -status` after zinc-facilitated compilation showed that the values that were passed to the zinc server at startup are ignored. The required `scala-compiler` and `scala-library` versions are specified by `scala-maven-plugin` invocations of the zinc server [as verified by an examination of the plugin source](https://github.com/davidB/scala-maven-plugin/blob/3.2.1/src/main/java/sbt_inc/SbtIncrementalCompiler.java#L96)

PR #1120, opened 2016-09-29
GEOMESA-1454 Add SFT name getter to GenerateFeatureWrappers

vpipkt:
Method in generated wrapper is called `getSimpleFeatureTypeName`, which is unlikely to be the name of an attribute on a simple feature.

Signed-off-by: Jason T. Brown <jason.brown@ccri.com>

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/GenerateFeatureWrappers.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/GenerateFeatureWrappers.scala
@@ -74,6 +74,7 @@ object GenerateFeatureWrappers {
     import scala.collection.JavaConversions._
 
     val attrs = sft.getAttributeDescriptors.map(AttributeDetails(_, sft))
+    val sftName = sft.getTypeName()
 
     val sb = new StringBuilder()
     sb.append(s"${tab}implicit class ${sft.getTypeName}")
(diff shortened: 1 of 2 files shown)


elahrvivaz:
looks good - you know you can do feature.getSimpleFeatureType.getTypeName, right? ;)

vpipkt:
@elahrvivaz of course. But there are cases where we are working with the simple feature type and have no simple feature objects. E.g. we are doing a simple feature builder, and need to the sft name to intialize it...

elahrvivaz:
doesn't the implicit value class require a SimpleFeature instance though?

vpipkt:
@elahrvivaz great point. This Maybe what I am after in the output is more like this:

```
object SimpleFeatureWrappers {

  val test1 = "test1"

  implicit class test1(val sf: org.opengis.feature.simple.SimpleFeature) extends AnyVal {
  //...
  }
}
```

And then could use SimpleFeatureWrappers.test1 instead of "test1"

elahrvivaz:
yeah, that would make sense

jnh5y:
@elahrvivaz @vpipkt Are you guys good with this PR?

elahrvivaz:
I don't think it does what @vpipkt actually wanted - might want to hold off

elahrvivaz:
@vpipkt going to close this as I think you decided you didn't want it - let me know if otherwise.

PR #1124, opened 2016-09-29
GEOMESA-702 Support for 'exists' predicate

elahrvivaz:
* Note: will run a full table scan

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
@@ -214,6 +214,22 @@ class AccumuloDataStoreQueryTest extends Specification with TestWithMultipleSfts
       (urNum + llNum) mustEqual (orNum + andNum)
     }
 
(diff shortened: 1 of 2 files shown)


jnh5y:
In terms of the full-table scan, is there a way to use an attribute index for this?

Is exists the same as "not null"?  (And vice versa for does-not-exist?)

elahrvivaz:
exists seems to be implemented that the simple feature type contains the descriptor with the appropriate name. As such, I think it would return true for every feature in a schema. does-not-exist is the opposite... they don't seem very useful for actually querying, so I figure as long as we have nominal support we're ok.
```
scala> val filter = ECQL.toFilter("name EXISTS")
filter: org.opengis.filter.Filter = [ PropertyExists('name') = true ]
scala> val sft = SimpleFeatureTypes.createType("foo", "name:String")
sft: org.opengis.feature.simple.SimpleFeatureType = SimpleFeatureTypeImpl foo identified extends Feature(name:name)
scala> val sf = new ScalaSimpleFeature("0", sft, Array("name"))
sf: org.locationtech.geomesa.features.ScalaSimpleFeature = ScalaSimpleFeature:0
scala> filter.evaluate(new ScalaSimpleFeature("0", sft, Array("not null")))
res0: Boolean = true
scala> filter.evaluate(new ScalaSimpleFeature("0", sft, Array(null)))
res1: Boolean = true
```

PR #1126, opened 2016-09-30
GEOMESA-1065 Fail travis build if CQ deps change

aheyne:
* Added CQManifest.tsv, an exact copy of cqs.tsv. It acts as an authoratative source that must be changed for new cqs to be added.
* Added travis now runs the cq check and compares results to CQManifest.


aheyne:
As a note, when you want to update the CQs you can just replace the manifest with the result of calculate-cqs.

elahrvivaz:
does travis have a git client? maybe we could just run `git diff build/cqs.tsv` vs having a separate copy to compare

aheyne:
Now uses git diff.

PR #1129, opened 2016-09-30
GEOMESA-1220 Don't freeze on bad formats

cfkelly:
* Removes PipedWriter and PipedReader
* Talked with Chris about how this is more of a band-aid and that a more elegant solution will be needed 


geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
```diff
@@ -80,44 +76,16 @@ class DelimitedTextConverter(format: CSVFormat,
...
-    }
-  })
-
   override def processInput(is: Iterator[String], ec: EvaluationContext): Iterator[SimpleFeature] = {
     ec.counter.incLineCount(options.skipLines)
     super.processInput(is.drop(options.skipLines), ec)
   }
 
   override def fromInputType(string: String): Seq[Array[Any]] = {
-    // empty strings cause deadlock
```
mdzimmerman: Does this mean that empty strings will no longer cause deadlocks? Are empty string/lines now tested for in the test suite changes above?

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
```diff
@@ -80,44 +76,16 @@ class DelimitedTextConverter(format: CSVFormat,
...
   extends LinesToSimpleFeatureConverter {
 
-  var curString: String = null
-  val q = Queues.newArrayBlockingQueue[String](32)
-  // if the record to write is bigger than the buffer size of the PipedReader
-  // then the writer will block until the reader reads data off of the pipe.
-  // For this reason, we have to separate the reading and writing into two
-  // threads
-  val writer = new PipedWriter()
-  val reader = new PipedReader(writer, options.pipeSize)  // record size
```
elahrvivaz: @anthonyccri was there a specific reason for the piped reader/writer, or is it ok to remove it?

jahhulbert-ccri: the purpose I think was to allow blocks of text to be read off kafka or something en mass instead of line at a time

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
jahhulbert-ccri: i believe i previously fixed empty strings and made a test case...i think maybe this is it: https://github.com/locationtech/geomesa/blob/master/geomesa-convert/geomesa-convert-text/src/test/scala/org/locationtech/geomesa/convert/text/NewLinesTest.scala

jahhulbert-ccri:
@cfkelly if you get a chance can you take a look at the review changes?

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverter.scala
cfkelly: I had talked with @cne1x about the best approach to fixing the issue, and we agreed that it would be best to remove the PipedWriter/Reader and go with the proposed line-by-line parsing for the short term. The catch is that this will be slower*, but it will not hang on bad formats. The plan is to work towards a more elegant solution that uses threading.

* most likely

elahrvivaz: @anthonyccri can you confirm this approach is ok?

jahhulbert-ccri:
I'd probably rename the city1 and city2 to city-tsv.conf and city-mysql.conf or whatnot

jahhulbert-ccri:
@cfkelly this looks good now, just waiting to see if some customers are ok with the change

PR #1132, opened 2016-10-03
GEOMESA-1341 Fixing creation of query plans for spark, m/r

elahrvivaz:
* Attribute row-value queries do transformation in the iterator stack, instead of on the client
  (m/r input formats do not currently have a way to modify client transform)
* Spark serialization changed to support different schemas with the same name (e.g. transformations)
* Adds a new iterator - clients with this patch will not work against older distributed runtimes for attribute row-value queries


elahrvivaz:
@tkunicki would you have time to test out this branch real quick? I want to make sure serialization across a cluster still works ok, but my local setup isn't really working.

build/cqs.tsv
```diff
@@ -88,7 +88,7 @@ org.apache.commons:commons-csv	1.0	compile
 org.apache.commons:commons-lang3	3.3.2	compile
 org.apache.commons:commons-math	2.1	compile
 org.apache.commons:commons-pool2	2.2	compile
-org.apache.commons:commons-vfs2	2.0	compile
+org.apache.commons:commons-vfs2	2.1	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12053

pom.xml
```diff
@@ -1443,25 +1443,6 @@
             </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>
-                <artifactId>commons-vfs2</artifactId>
-                <version>2.0</version>
```
elahrvivaz: this was meant to exclude certain transitive deps from accumulo, but we didn't update the version with accumulo 1.7 which now imports 2.1. this caused some mock tests to fail.

jahhulbert-ccri: we should update, right?

elahrvivaz: it pulls 2.1 in transitively from accumulo now

PR #1139, opened 2016-10-05
GEOMESA-1467 Multithreading join scans in BatchMultiScanner

elahrvivaz:


geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/Utils.scala
```diff
@@ -49,6 +49,7 @@ object Utils {
     val BIN     = Value("bin")
     val AVRO    = Value("avro")
     val XML     = Value("xml")
+    val NULL    = Value("null")
```
jnh5y: Is the Null export to help us benchmark more precisely?

elahrvivaz: i added it for testing and figured it would be useful to have

PR #1149, opened 2016-10-06
1.2 - GEOMESA-1358 Configurable precision for XZ indices

elahrvivaz:

### DIFF
--- a/docs/user/data_management.rst
+++ b/docs/user/data_management.rst
@@ -186,6 +186,29 @@ user data using the hint ``geomesa.z3.interval``:
 
 See below for alternate ways to set the user data.
 
+.. _customizing_xz_index:
+
+Customizing the XZ-Index
+------------------------
+
+GeoMesa uses an extended z-curve index for storing geometries with extents. The index can be customized
+by specifying the resolution level used to store geometries. By default, the resolution level is 12. If
+you have very small geometries, or very large geometries, you may want to tweak this value.
+
+The resolution level for an index is set when calling ``createSchema``. It may be specified through
+the simple feature type user data using the hint ``geomesa.xz.precision``:
+
+.. code-block:: java
+
+    // set the hint directly
+    SimpleFeatureType sft = ...
+    sft.getUserData().put("geomesa.xz.precision", 12);
+
+See below for alternate ways to set the user data.
+
+For more information on resolution level (g), see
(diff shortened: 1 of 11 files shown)


docs/user/data_management.rst
```diff
@@ -186,6 +186,29 @@ user data using the hint ``geomesa.z3.interval``:
...
 See below for alternate ways to set the user data.
 
+.. _customizing_xz_index:
+
+Customizing the XZ-Index
+------------------------
+
+GeoMesa uses an extended z-curve index for storing geometries with extents. The index can be customized
+by specifying the resolution level used to store geometries. By default, the resolution level is 12. If
+you have very small geometries, or very large geometries, you may want to tweak this value.
```
cne1x: ... in which direction?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z2/XZ2WritableIndex.scala
```diff
@@ -20,29 +20,31 @@ import org.locationtech.geomesa.accumulo.AccumuloVersion
...
 import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 import org.opengis.feature.simple.SimpleFeatureType
 
 trait XZ2WritableIndex extends AccumuloWritableIndex {
 
   import AccumuloWritableIndex.{DefaultNumSplits, DefaultSplitArrays}
-  import XZ2Index.SFC
 
   override def writer(sft: SimpleFeatureType, ops: AccumuloDataStore): (WritableFeature) => Seq[Mutation] = {
+    val sfc = XZ2SFC(sft.getXZPrecision)
```
cne1x: Could we still have a lazy variable at the trait level that gets shared among these routines?

cne1x: Never mind.  :(

elahrvivaz: the sfcs are cached in the apply method, so it should mostly be re-using the same instance object.

docs/user/data_management.rst
elahrvivaz: udpated

PR #1150, opened 2016-10-06
GEOMESA-1358 Configurable precision for XZ indices

elahrvivaz:

### DIFF
--- a/docs/user/data_management.rst
+++ b/docs/user/data_management.rst
@@ -186,6 +186,29 @@ user data using the hint ``geomesa.z3.interval``:
 
 See below for alternate ways to set the user data.
 
+.. _customizing_xz_index:
+
+Customizing the XZ-Index
+------------------------
+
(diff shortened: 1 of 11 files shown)


docs/user/data_management.rst
```diff
@@ -186,6 +186,30 @@ user data using the hint ``geomesa.z3.interval``:
 
 See below for alternate ways to set the user data.
 
+.. _customizing_xz_index:
+
+Customizing the XZ-Index
```
jnh5y: Do we explain elsewhere when a 'XZ' index is used?  Is it worth explaining that any XZ2 / XZ3 tables created by will use this config?

jnh5y:
My only comment is about documentation; otherwise, this is good to go.

docs/user/data_management.rst
elahrvivaz: yeah, it's explained at the top of this page: http://www.geomesa.org/documentation/user/data_management.html#index-structure

PR #1152, opened 2016-10-07
GEOMESA-616 Add attributes parameter to explain command

aheyne:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/ExplainCommand.scala
+++ b/geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/ExplainCommand.scala
@@ -15,7 +15,7 @@ import org.geotools.filter.text.ecql.ECQL
 import org.locationtech.geomesa.index.utils.ExplainPrintln
 import org.locationtech.geomesa.tools.accumulo.GeoMesaConnectionParams
 import org.locationtech.geomesa.tools.accumulo.commands.ExplainCommand.ExplainParameters
-import org.locationtech.geomesa.tools.common.{CQLFilterParam, FeatureTypeNameParam}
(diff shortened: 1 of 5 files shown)


geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/ExplainCommand.scala
```diff
@@ -24,6 +24,11 @@ class ExplainCommand(parent: JCommander) extends CommandWithCatalog(parent) with
   override def execute() =
     try {
       val q = new Query(params.featureName, ECQL.toFilter(params.cqlFilter))
+      Option(params.attributes).foreach { attribute =>
+        val splitAttrs = attribute.toString().split("""(?<!\\),""").map(_.trim.replace("\\,", ","))
```
elahrvivaz: don't we use this line somewhere else? might be good to pull it out into a common trait

geomesa-tools-common/src/main/scala/org/locationtech/geomesa/tools/common/CommonParams.scala
```diff
@@ -64,12 +64,12 @@ trait OptionalDTGParam {
 
 trait OptionalAttributesParam {
   @Parameter(names = Array("-a", "--attributes"), description = "Attributes to evaluate (comma-separated)")
-  var attributes: java.util.List[String] = null
+  var attributes: String = null
```
elahrvivaz: is there a reason to change this to a string? i believe (although not 100% sure) keeping it a list lets you specify multiple flags instead of forcing comma-separated, e.g. `-a foo -a bar` vs `-a foo,bar`. But jcommander also does some splitting on a single argument if you specify a list - i think we override that in certain places. anyway, not sure it really matters

aheyne: I've converted these all back to Lists but I had to touch a lot of places to get the typing right.

aheyne:
Closing this pr as another pr had conflicts that needed to be merged. See #1156

PR #1153, opened 2016-10-07
GEOMESA-1471 SimpleFeatureAttribute should inherit from

mdzimmerman:
SimpleNullableAttribute, not SimpleAttribute, which by CQEngine
contract should not return nulls.


mdzimmerman:
This is a blocker for 1.3.0-m1. It causes NPEs in LiveFeatureCacheCQEngine if an SF attribute has (1) a null value and (2) there is an index on that attribute.

jnh5y:
It would be good to see a failing unit test for this.  That said, merging it now!

PR #1154, opened 2016-10-10
GEOMESA-1472 Transformers functionMap needs to be synchronized

anthonyccri:


jnh5y:
@anthonyccri is it possible to get a unit test for this?  If not, that's fine.

anthonyccri:
It would be more trouble than its worth to unit test it.

elahrvivaz:
just curious what problem this is fixing - since it's an object, i didn't think it would need to be synchronized?

anthonyccri:
There's a race condition in initializing the map.  There's the declaration and then there's the initialization.  I was loading up multiple converters in the same JVM and was seeing some cases in which the map was empty and throwing lots of errors.

elahrvivaz:
weird, i thought that the object couldn't be accessed until it had 'initialized'... we probably have a lot of other places this might be an issue too

jnh5y:
Since functionMap is a lazy val, it is null until the Transformers object is initialized. 

I'm fine with this change...

PR #1162, opened 2016-10-13
GEOMESA-1046 Fix progress bar scaling on ingest

aheyne:
Progress bar (just the bar) now scales on ingest. If the info string length increases the bar shrinks dynamically.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/AbstractIngest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/AbstractIngest.scala
@@ -193,15 +193,34 @@ object AbstractIngest {
                     emptyBar: String,
                     replacement: Char,
                     indicator: Char)(progress: Float, start: Long, pass: Long, fail: Long, done: Boolean): Unit = {
-    val numFilled = (emptyBar.length * progress).toInt
-    val bar = if (numFilled < 1) {
-      emptyBar
-    } else if (numFilled >= emptyBar.length) {
-      buildString(replacement, numFilled)
-    } else {
-      s"${buildString(replacement, numFilled - 1)}$indicator${emptyBar.substring(numFilled)}"
-    }
     val percent = f"${(progress * 100).toInt}%3d"
+    val infoStr = s" $percent% complete $pass ingested $fail failed in ${getTime(start)}"
+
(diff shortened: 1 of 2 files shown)


geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/AbstractIngest.scala
```diff
@@ -231,8 +231,9 @@ object AbstractIngest {
   }
 
   private def buildString(c: Char, length: Int) = {
-    val sb = new StringBuilder(length)
-    (0 until length).foreach(_ => sb.append(c))
+    val slength = if (length < 0) 0 else length
```
cne1x: You know how it goes...  The smallest PRs get the most comments/line.  Couldn't we just reduce the object creation by saying something like:

```scala
if (length < 0) return ""
val sb = new StringBuilder(length)
(0 until length).foreach(_ => sb.append(c))
sb.toString()
```

aheyne: Done.

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/AbstractIngest.scala
```diff
@@ -193,24 +193,45 @@ object AbstractIngest {
...
-    val bar = if (numFilled < 1) {
-      emptyBar
-    } else if (numFilled >= emptyBar.length) {
-      buildString(replacement, numFilled)
+    val percent = f"${(progress * 100).toInt}%3d"
+    val infoStr = s" $percent% complete $pass ingested $fail failed in ${getTime(start)}"
+
+    // Figure out if and how much the progress bar should be scaled to accommodate smaller terminals
+    val scaleFactor: Float = try {
+      val tWidth: Float = jline.Terminal.getTerminal.getTerminalWidth.toFloat
```
elahrvivaz: does jline have any better options than using `\r` to replace the current line? Maybe we should have 2 methods - one the enhanced jline version and one the original basic version in case there's no jline.

elahrvivaz:
build is failing - maybe due to calculate cqs?

aheyne:
Yep, CQ check failed.

PR #1164, opened 2016-10-14
GEOMESA-540, GEOMESA-1092, GEOMESA-1473 XML system properties

aheyne:
I'm opening this now to get feedback on this implementation. We're still waiting on some descriptions of system properties before merge.
- Site config xml now in place and documented.
- Command line parameters take precident over config file. Config properties have a final option that when set to true makes the property take precidence over every other source, including command line.
- By default config properties with a null value will be ignored. This can be overwritten using the final tag in order to use the config to clear a system property.
- Config is loaded whenever an accumulo or kafka datastore is instantiated.


jnh5y:
@mdzimmerman can you list the undocumented system properties?

aheyne:
@mdzimmerman These should be the missing ones unless they're documented elsewhere.
geomesa.sft.config.urls
geomesa.batchwriter.memory
geomesa.batchwriter.maxthreads
geomesa.convert.config.urls
geomesa.audit.provider.impl

docs/user/installation_and_configuration.rst
```diff
@@ -658,6 +658,21 @@ After building, extract ``target/geomesa-cassandra-gs-plugin_2.11-$VERSION-insta
 
 Restart GeoServer after the JARs are installed.
 
+Additional Configuration
+------------------------
+
+GeoMesa uses a site xml file to maintain system property configurations. This file can be found
+at ``conf/geomesa-site.xml`` of either the Accumulo or Kafka distributions. The default settings
```
elahrvivaz: is this meant to be the geomesa accumulo distribution, or the actual accumulo install? seems like it should be the geomesa one right?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/ConfigLoader.scala
```diff
@@ -0,0 +1,59 @@
...
+    } else if (! Option(System.getenv(KAFKA_HOME)).getOrElse("").isEmpty) {
+      this.init(KAFKA_HOME)
+    } else {
+      logger.warn("Unable to locate GeoMesa config file.")
+    }
+  }
+
+  def init(envName: String): Unit = {
+    if (! Option(System.getProperty("geomesa.config.file")).getOrElse("").isEmpty){
+      this.loadConfig(System.getProperty("geomesa.config.file"))
```
elahrvivaz: I would think the order for finding the file should be:
1. explicit system property
2. check the current classpath
3. check for GEOMESA_ACCUMULO_HOME, GEOMESA_KAKFA_HOME
@jahhulbert-ccri what do you think?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/ConfigLoader.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  def loadConfig(path: String): Unit = {
+    try {
+      val xml = XML.loadFile(path)
+      logger.info("Using GeoMesa config file found at: " + path)
+      val properties = xml \ "property"
+      properties.foreach { prop =>
+        val key = prop \ "name"
+        val value = prop \ "value"
+        val isfinal: Boolean = (prop \ "final").toString.toBoolean
```
elahrvivaz: will these xml selectors work if the tag is missing?

docs/user/accumulo_datastore.rst
```diff
@@ -34,6 +34,18 @@ A script to bootstrap GeoMesa Accumulo on an Elastic Map Reduce cluster is provi
 Configuration
 -------------
 
+GeoMesa uses a site xml file to maintain system property configurations. This file can be found
+at ``conf/geomesa-site.xml`` of the Accumulo distribution. The default settings for GeoMesa are
+stored in ``conf/geomesa-default.xml``. Do not modify this file directly as it is never read,
```
aheyne: We've been trying to mirror hadoop's and accumulo's implementation and the current naming scheme reflects this.
@jahhulbert-ccri thoughts?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/ConfigLoader.scala
aheyne: Yes, it just returns and empty string.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/knn/KNearestNeighborSearchProcessTest.scala
```diff
@@ -8,6 +8,7 @@
 
 package org.locationtech.geomesa.accumulo.process.knn
 
+import com.typesafe.scalalogging.LazyLogging
```
elahrvivaz: this file can be reverted, right?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaProperties.scala
```diff
@@ -13,27 +13,79 @@ import java.util.Properties
...
+  def GEOMESA_SCAN_RANGES_BATCH         = PropOrDefault("geomesa.scan.ranges.batch")
+  def GEOMESA_SFT_CONFIG_URLS           = PropOrDefault("geomesa.sft.config.urls")
+  def GEOMESA_STATS_COMPACT_MILLIS      = PropOrDefault("geomesa.stats.compact.millis")
+
+  case class PropOrDefault(property: String, dft: String = "") {
+    val default = if (dft.nonEmpty) dft
+                  else Option(props.getProperty(property)).getOrElse(dft)
+    def get: String = {
+      ensureConfig()
+      Option(System.getProperty(property)).getOrElse(default)
```
elahrvivaz: this method should return `null` if the property is not set. I would change it to:
`Option(System.getProperty(property)).filterNot(_.isEmpty).getOrElse(default)` and change the signature to `dft: String = null`
Otherwise it introduces having to check for empty strings where this method is used, which is easily forgotten. Null checks shouldn't be required, as if there is the `option` method instead.

geomesa-tools-common/conf/geomesa-default.xml
```diff
@@ -0,0 +1,206 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elahrvivaz: can we rename this to `geomesa-site.xml.template`? otherwise it's not clear that this file isn't actually used for anything.

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa.properties
```diff
@@ -1,4 +1,39 @@
 geomesa.project.version=${project.version}
 geomesa.build.date=${git.build.time}
 geomesa.build.commit.id=${git.commit.id}
-geomesa.build.branch=${git.branch}
\ No newline at end of file
+geomesa.build.branch=${git.branch}
+
+#- QueryProperties
+geomesa.force.count="false"
```
elahrvivaz: is there a way we can keep these + the ones in the default site.xml in a single place, so that we don't have to modify it two places? Possibly put `geomesa-site.xml.template` in src/main/resources and load it from there, plus copy it into tools for the distribution.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/ConfigLoader.scala
```diff
@@ -0,0 +1,67 @@
...
+  def loadConfig(xml: scala.xml.Elem): Unit = {
+    val properties = xml \\ "configuration" \\ "property"
+    properties.foreach { prop =>
+      try { // Use try/catch here so if we fail on a property the rest can still load
+        val key = (prop \ "name").text
+        val value = (prop \ "value").text
+        val isfinal: Boolean = (prop \ "final").text.toString.toBoolean
+        // Don't overwrite properties, this gives commandline params preference
+        if (Option(System.getProperty(key)).getOrElse("").isEmpty && value.nonEmpty || isfinal) {
+          System.setProperty(key, value)
```
elahrvivaz: instead of setting the properties here, we could have this object load the file into a Map[String, String]. Then when checking for props, we can do something like: `Option(System.getProperty).orElse(ConfigLoader.get).orElse(default)`. That would avoid having to store the state of whether the config is loaded, since it could just be part of the object initialization body.

docs/user/accumulo_datastore.rst
```diff
@@ -34,6 +34,18 @@ A script to bootstrap GeoMesa Accumulo on an Elastic Map Reduce cluster is provi
...
+instead copy over the desired configurations into geomesa-site.xml.
+
+By default, system properties set through command line parameters will take precedent over this
```
mdzimmerman: In general, this phrase should always be "take precedence," not "take precedent." See http://www.dailywritingtips.com/precedent-vs-precedence/

docs/user/accumulo_datastore.rst
```diff
@@ -34,6 +34,18 @@ A script to bootstrap GeoMesa Accumulo on an Elastic Map Reduce cluster is provi
...
+configuration file. If you wish a configuration item to always take precedence, even over command
+line parameters change the ``<final>`` tag to true.
```
mdzimmerman: [picky style] comma after "parameters"

docs/user/installation_and_configuration.rst
```diff
@@ -725,6 +725,21 @@ After building, extract ``target/geomesa-cassandra-gs-plugin_2.11-$VERSION-insta
 
 Restart GeoServer after the JARs are installed.
 
+Additional Configuration
+------------------------
+
+GeoMesa uses a site xml file to maintain system property configurations. This file can be found
+at ``conf/geomesa-site.xml`` of either the GeoMesa-Accumulo or GeoMesa-Kafka distributions. The default settings
```
mdzimmerman: [picky style] We have not been hyphenating the GeoMesa distributions (e.g. "GeoMesa-Accumulo") elsewhere in the docs, and to be consistent I would not do it here either.

docs/user/installation_and_configuration.rst
```diff
@@ -725,6 +725,21 @@ After building, extract ``target/geomesa-cassandra-gs-plugin_2.11-$VERSION-insta
...
+for GeoMesa are stored in ``conf/geomesa-default.xml``. Do not modify this file directly as it is
+never read, instead copy over the desired configurations into geomesa-site.xml.
+
+By default command line parameters will take precedent over this configuration file. If you wish
```
mdzimmerman: "precedent" => "precedence"

docs/user/installation_and_configuration.rst
```diff
@@ -725,6 +725,21 @@ After building, extract ``target/geomesa-cassandra-gs-plugin_2.11-$VERSION-insta
...
+a configuration item to always take precedence, even over command line parameters change the
+``<final>`` tag to true.
+
+By default configuration properties with empty values will not be applied, you can change this
```
mdzimmerman: [picky grammar] To prevent a run-on sentence, I would either change this comma to a semicolon, or add "and".

docs/user/kafka_datastore.rst
```diff
@@ -26,7 +26,23 @@ within that interval.
...
+however, a ``geomesa-site.xml`` configuration file is available for
+setting system properties.
+
+This file can be found at ``conf/geomesa-site.xml`` of the Kafka
+distribution. The default settings for GeoMesa are stored in
+``conf/geomesa-default.xml``. Do not modify this file directly as it
+is never read, instead copy over the desired configurations into
+geomesa-site.xml.
+
+By default command line parameters will take precedent over this
```
mdzimmerman: "precedent" => "precedence"

geomesa-tools-common/conf/geomesa-default.xml
```diff
@@ -0,0 +1,206 @@
...
+        </description>
+        <final>false</final>
+    </property>
+
+    <property>
+        <name>geomesa.force.count</name>
+        <value>false</value>
+        <description>This property controls how GeoMesa calculates the size
+            of a result set (e.g. `FeatureSource.getCount`). By default,
+            eomesa will estimate the size of a result set using statistics.
```
mdzimmerman: "eomesa"?

geomesa-tools-common/conf/geomesa-default.xml
```diff
@@ -0,0 +1,206 @@
...
+        <value></value>
+        <description></description>
+        <final>false</final>
+    </property>
+
+    <property>
+        <name>geomesa.stats.compact.millis</name>
+        <value>3600000</value>
+        <description>This property controls how often the GeoMesa statistics
+            table will be compacted. GeoMesa tracks data stastics during
```
mdzimmerman: Units? (Obvious from the property name?)

elahrvivaz:
oops, need to update references to `geomesa-default.xml`

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/AccumuloRunner.scala
```diff
@@ -86,10 +88,10 @@ object AccumuloRunner extends Runner {
         p.instance = try {
           // This block checks for the same system property which Accumulo uses for Zookeeper timeouts.
           //  If it is set, we use it.  Otherwise, a timeout of 5 seconds is used.
-          val lookupTime: Long =
-            Option(System.getProperty("instance.zookeeper.timeout")).flatMap{s =>
-              Try { java.lang.Long.parseLong(s) }.toOption
-            }.getOrElse(5000L)
+          //  Don't give default to PropOrDefault so .option will throw a None
+          val lookupTime: Long = PropOrDefault("instance.zookeeper.timeout").option.flatMap{ p =>
```
jnh5y: Could we just use a default of 5000 milliseconds?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaProperties.scala
```diff
@@ -13,27 +13,79 @@ import java.util.Properties
...
+  val GEOMESA_CONFIG_FILE_NAME = ConfigLoader.GEOMESA_CONFIG_FILE_NAME
+
+  def GEOMESA_TOOLS_ACCUMULO_SITE_XML: String = {
+    val siteXML = getProperty("geomesa.tools.accumulo.site.xml")
+    if (siteXML.nonEmpty) siteXML
+    else s"${System.getenv("ACCUMULO_HOME")}/conf/accumulo-site.xml"
+  }
+  def GEOMESA_AUDIT_PROVIDER_IMPL       = PropOrDefault("geomesa.audit.provider.impl")
+  def GEOMESA_AUTH_PROVIDER_IMPL        = PropOrDefault("geomesa.auth.provider.impl")
+  def GEOMESA_BATCHWRITER_LATENCY_MILLS = PropOrDefault("geomesa.batchwriter.latency.millis")
```
jnh5y: Ok, where is the default for this value now?

jahhulbert-ccri:
@aheyne looks like there are still some conflicts can you check one more time? @elahrvivaz @jnh5y you good with this one?

aheyne:
Doing a test compile atm and then i'll push it up.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaProperties.scala
```diff
@@ -13,27 +13,81 @@ import java.util.Properties
...
 
   val ProjectVersion = props.getProperty("geomesa.project.version")
   val BuildDate      = props.getProperty("geomesa.build.date")
   val GitCommit      = props.getProperty("geomesa.build.commit.id")
   val GitBranch      = props.getProperty("geomesa.build.branch")
+
+  val GEOMESA_CONFIG_FILE      = ConfigLoader.GEOMESA_CONFIG_FILE
+  val GEOMESA_CONFIG_FILE_NAME = ConfigLoader.GEOMESA_CONFIG_FILE_NAME
+
+  def GEOMESA_TOOLS_ACCUMULO_SITE_XML: String = {
```
jahhulbert-ccri: we need a way to keep accumulo stuff in the accumulo project...probably some composite approach here

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/ConfigLoader.scala
```diff
@@ -0,0 +1,69 @@
...
+import scala.xml.XML
+import java.io.FileNotFoundException
+
+import com.typesafe.scalalogging.LazyLogging
+
+object ConfigLoader extends LazyLogging {
+  val GEOMESA_CONFIG_FILE = "geomesa.config.file"
+  val GEOMESA_CONFIG_FILE_NAME = "geomesa-site.xml"
+  val GEOMESA_EMBEDDED_CONFIG_FILE_PATH = "/org/locationtech/geomesa/geomesa-site.xml.template"
+  var isLoaded: Boolean = false
```
jahhulbert-ccri: probably should be synchronized or atomic? @elahrvivaz what think you?

elahrvivaz: yeah, i was thinking use a singleton object that loads the xml in it's constructor and skip the isLoaded check.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaProperties.scala
aheyne: Any thoughts on how we'd do that? I'm unsure how, since the only copy atm is the .xml.template. Unless we somehow scan that file (?) or something at compile time and build the xml.template.

aheyne:
From my testing this appears to be functional now.

PR #1167, opened 2016-10-17
Remove jline provided dependency for progressbar

aheyne:
Fixed the provided jline dependency


aheyne:
Woops, I merged master after the fix and overwrote it.

jnh5y:
Does removing this mess up / remove the progress bar?  Does it work sensibly if someone doesn't download jline?

aheyne:
If a user doesn't have jline it will work the way it did before. Only if they have jline on the cp and their terminal window is too small will this kick in.

PR #1170, opened 2016-10-19
GEOMESA-1485 Increase Accumulo Instance ID lookup to 5s

jnh5y:
* Additional this change provides feedback to the user regarding setting the instance ID.


geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/AccumuloRunner.scala
```diff
@@ -83,8 +85,17 @@ object AccumuloRunner extends Runner {
       if (p.instance == null) {
         p.instance = try {
           // this will hang for 60+ seconds if it's not configured - so we wrap in a future and only wait 1s
```
mdzimmerman: Update comment?

PR #1176, opened 2016-10-24
GEOMESA-1490 The Z3UuidGenerator should warn (not throw a NPE) when t…

cne1x:
…he given geometry is null

Added unit tests that both recreated the initial error and confirm
to the new expectation.

Signed-off-by: Chris Eichelberger cne1x@ccri.com

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
@@ -64,6 +64,8 @@ object Z3UuidGenerator extends RandomLsbUuidGenerator with LazyLogging {
         .getOrElse(System.currentTimeMillis())
 
     val pt = sf.getAttribute(sft.getGeomIndex)
+    validateGeometry(pt.asInstanceOf[Geometry])
+
     if (sft.isPoints) {
       createUuid(pt.asInstanceOf[Point], time, sft.getZ3Interval)
     } else {
@@ -73,16 +75,14 @@ object Z3UuidGenerator extends RandomLsbUuidGenerator with LazyLogging {
   }
 
   def createUuid(geom: Geometry, time: Long, period: TimePeriod): UUID = {
+    validateGeometry(geom)
+
     import org.locationtech.geomesa.utils.geotools.Conversions.RichGeometry
     createUuid(geom.safeCentroid(), time, period)
   }
 
   def createUuid(pt: Point, time: Long, period: TimePeriod): UUID = {
-    // handle missing geometry
-    if (pt == null) {
-      logger.error("Cannot generate a meaningful UUID for a <NULL> geometry; returning a random UUID instead.")
-      return UUID.randomUUID()
-    }
+    validateGeometry(pt)
 
     // create the random part
     // this uses the same temp array we use later, so be careful with the order this gets called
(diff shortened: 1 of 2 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
```diff
@@ -75,7 +77,13 @@ object Z3UuidGenerator extends RandomLsbUuidGenerator {
     createUuid(geom.safeCentroid(), time, period)
   }
 
-  def createUuid(pt: Point, time: Long, period: TimePeriod) = {
+  def createUuid(pt: Point, time: Long, period: TimePeriod): UUID = {
+    // handle missing geometry
+    if (pt == null) {
```
elahrvivaz: we should probably check for the non-point case on line 71 too

elahrvivaz: oh actually on line 77 too... i guess there are 3 entry points to this class.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
```diff
@@ -75,7 +77,13 @@ object Z3UuidGenerator extends RandomLsbUuidGenerator {
...
+      logger.error("Cannot generate a meaningful UUID for a <NULL> geometry; returning a random UUID instead.")
```
cne1x: I'm of half a mind.  On the one hand, a warning is probably appropriate (as we're returning a valid UUID for this feature, even though it has a null geometry).  On the other hand, that we are enabling users to create a feature with an invalid geometry feels... wrong.  Should this entire PR be retracted?  Was the original behavior correct?  Or should we refactor it to still throw an exception, just one with a nicer error?  (Either way, I like having the additional unit test so that the expectation is clear.)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
cne1x: Good catch.  I'll update the unit test, too.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
elahrvivaz: hmm, well I think it will fail further on when trying to [create mutations](https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z3/Z3WritableIndex.scala#L58-L59). However, in a generic sense having the default id generator work without a geometry does seem appropriate (e.g. for a geometry-less schema).

geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGeneratorTest.scala
```diff
@@ -0,0 +1,94 @@
+/***********************************************************************
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
```
jnh5y: Remove the spaces from the headers:(.  This is making it fail CI.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/uuid/Z3FeatureIdGenerator.scala
```diff
@@ -110,6 +118,11 @@ object Z3UuidGenerator extends RandomLsbUuidGenerator {
     new UUID(mostSigBits, leastSigBits)
   }
 
+  // features with a null geometry-to-index should be rejected
+  private def validateGeometry(geom: Geometry): Unit = if (geom == null) {
+    throw new Exception("Cannot meaningfully index a feature with a NULL geometry")
```
jahhulbert-ccri: perhaps this should be an IllegalArgumentException

PR #1182, opened 2016-11-04
Geomesa 1503,1504: remove asm from compute jar, simplify scala 2.10 build

tkunicki:


pom.xml
```diff
@@ -1445,6 +1445,12 @@
                 <groupId>com.esotericsoftware.kryo</groupId>
                 <artifactId>kryo</artifactId>
                 <version>2.21</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.ow2.asm</groupId>
+                        <artifactId>asm</artifactId>
```
elahrvivaz: is asm still on the classpath at all, even in test scope? Seems like kryo might actually need asm, but maybe not in the way we use it.

tkunicki: Kyro does use asm, but it's shaded and relocated in the dependency we use.

aheyne:
We might need to add the Scala 2.11 profile to the travis build command or set the Scala 2.11 profile to activate on the travis profile.

tkunicki:
2.11 build is active by default

pom.xml
elahrvivaz: nice, can we add a comment to the exclusion block so we remember that in the future?

elahrvivaz:
looks good to me

jnh5y:
@tkunicki two quick things:
Anthony likes commits to start with the line GEOMESA-####.  Note that GeoMesa is all caps and the dash is present.

Second, it looks like this has conflicts now...

tkunicki:
@jnh5y rebased PR.  Don't understand your comments about my commit messages...?

PR #1185, opened 2016-11-07
GEOMESA-1508 Using exiting executor for query thread timeout reaper

elahrvivaz:
~~Note: stacks on top of #1184~~

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
@@ -8,8 +8,9 @@
 
 package org.locationtech.geomesa.index.utils
 
-import java.util.concurrent.{Executors, PriorityBlockingQueue, TimeUnit}
+import java.util.concurrent.{PriorityBlockingQueue, ScheduledThreadPoolExecutor, TimeUnit}
 
+import com.google.common.util.concurrent.MoreExecutors
 import com.typesafe.scalalogging.LazyLogging
 import org.locationtech.geomesa.filter.filterToString
 import org.locationtech.geomesa.index.geotools.GeoMesaFeatureReader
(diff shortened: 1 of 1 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
```diff
@@ -53,9 +54,9 @@ object ThreadManagement extends LazyLogging {
     }
   }
 
-  private val executor = Executors.newSingleThreadScheduledExecutor()
+  private val executor = MoreExecutors.getExitingScheduledExecutorService(new ScheduledThreadPoolExecutor(1))
   executor.scheduleWithFixedDelay(reaper, interval, interval, TimeUnit.MILLISECONDS)
-  sys.addShutdownHook(executor.shutdown())
+  sys.addShutdownHook(executor.shutdownNow())
```
mdzimmerman: This resolves the hanging problem for the Accumulo QuickStart, both with and without a ``-Dgeomesa.query.timeout.millis=<value>`` argument.

jahhulbert-ccri: any reason to keep the shutdown hook here if using ```MoreExecutors.getExitingScheduledExecutorService``` ?

jahhulbert-ccri: i guess it doesn't call shutdown now... only shutdown

elahrvivaz: yeah, i didn't really look into this too closely but figure it couldn't hurt (and we do that with the other exiting executors too)

jahhulbert-ccri: good enough for me

PR #1193, opened 2016-11-08
1.2.6 - GEOMESA-1507 Adding bin filtering to saved query servlet

elahrvivaz:


jahhulbert-ccri:
do we want to cut a 1.2.6.1 with this?

elahrvivaz:
yeah, Britt wants it in there even though it's not technically a bug fix.

PR #1196, opened 2016-11-09
GEOMESA-1339 Authorizations provider docs

elahrvivaz:


docs/user/accumulo_datastore.rst
```diff
@@ -88,6 +88,13 @@ at data store, feature, or individual attribute levels.
 
 See :ref:`accumulo_visibilities` for more information.
 
+Scan Authorizations
+-------------------
+
+The GeoMesa Accumulo data store provides a Java SPI for custom handling of user authorizations per query.
+For simple use cases, authorizations can be set at the data store level through a configuration parameter.
```
mdzimmerman: Which one? I think there should be details here.

docs/user/data_management.rst
```diff
@@ -423,7 +423,7 @@ Accumulo Visibilities
 ---------------------
 
 GeoMesa support Accumulo visibilities for securing data. Visibilities can be set at data store level,
-feature level or individual attribute level.
+feature level or individual attribute level. For information on scan authorizations, see :ref:`accumulo_tutorials_security`.
```
mdzimmerman: This should point to the "Scan Authorizations" section that you added, not the tutorials, shouldn't it?

docs/user/accumulo_datastore.rst
mdzimmerman: Ah, you're talking about the "auths" data store parameter. Sorry, I was overthinking it. I think it might be a good idea to take the text here:

https://github.com/geomesa/geomesa-tutorials/tree/master/geomesa-examples-authorizations#authorizations-in-geomesa

and move/copy it here. Alternately we should put a more precise link to this section of this tutorial rather than linking to the "security" section in the tutorials table of contents.

elahrvivaz: oh, that is a good explanation. I'll copy it over.

docs/user/data_management.rst
elahrvivaz: yeah, the auths section just pointed to the tutorials, but since i'll be filling it out i'll update this link

docs/user/accumulo_datastore.rst
elahrvivaz: updated

docs/user/data_management.rst
elahrvivaz: updated

PR #1197, opened 2016-11-09
GEOMESA-1515 Suffixes helper

jahhulbert-ccri:
so we can use it elsewhere. note i pulled off the shelf some suffixes class i had but we can keep the old mem parsing if desired. i had some logging in it

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/Suffixes.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/Suffixes.scala
@@ -1,0 +1,83 @@
+/***********************************************************************
+* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.utils.text
+
+import java.util.regex.Pattern
+
+import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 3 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/Suffixes.scala
```diff
@@ -0,0 +1,83 @@
...
+      .appendMillis().appendSuffix("ms")
+      .toFormatter
+
+    def duration(s: String): Option[Duration] = {
+      Try[Duration](Period.parse(s.toLowerCase, format).toStandardDuration)
+        .orElse(Try(Duration.millis(s.toLong)))
+      match {
+        case Success(d) => Some(d)
+        case Failure(e) =>
+          logger.error(s"Unable to parse duration $s", e)
```
elahrvivaz: this error might be slightly misleading since the first exception from the period parser is being dropped

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/Suffixes.scala
```diff
@@ -0,0 +1,83 @@
...
+      if (m.matches() && m.groupCount() == 2) {
+        Try {
+          val num: Long = m.group(1).toLong
+          val suf = m.group(2)
+          val mult: Long = suf match {
+            case "k" => 1024l
+            case "m" => 1024l * 1024l
+            case "g" => 1024l * 1024l * 1024l
+            case "t" => 1024l * 1024l * 1024l * 1024l
+            case _ => 1l
```
elahrvivaz: should we fail on this case, and add an explicit match for the empty string?

PR #1200, opened 2016-11-10
WIP GEOMESA-1165 Allow configurable shards in z2 (and z3)

cfkelly:
* Framework for configurable shards in z2 and z3 for Accumulo indexes
* @cne1x and I believe this to be one of the more straightforward approaches to this ticket

### DIFF
--- a/docs/user/data_management.rst
+++ b/docs/user/data_management.rst
@@ -160,6 +160,37 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
         .geometry("geom", default = true)
         .build("mySft")
 
+.. _configuring_z_shards:
+
+Configuring Z-Index Shards
+--------------------------
+
+GeoMesa allows you to specify the number of shards (or splits) to be used for each SimpleFeatureType. This gives
+you more flexibility with SimpleFeatureTypes whether they are large or small. By default, SimpleFeatureTypes are
+created with 4 as the number of shards. 4 is the default because previous versions of GeoMesa hardcoded it in
+the z indices.
+
+When dealing with large SimpleFeatureTypes or large clusters, it may be beneficial to use a larger number of shards.
+In most test cases on single machines, the default is most likely preferable. It is important to note that acceptable
+values for the number of shards are integers in the range ``[1, 127]``.
+
+The number of shards is set when calling ``createSchema``. It may be specified through the simple feature type
+user data using the hint ``geomesa.z.splits``:
+
+.. code-block:: scala
+
+    // set the hint directly
+    SimpleFeatureType sft = ...
+    sft.getUserData().put("geomesa.z.splits", "16")
+
+.. code-block:: scala
+
+    // or through using RichSimpleFeatureType
+    import org.locationtech.geomesa.utils.geomesa.RichSimpleFeatureType.RichSimpleFeatureType
+
+    SimpleFeatureType sft = ...
+    sft.setZShards(16)
+    
 .. _customizing_z_index:
 
 Customizing the Z-Index
--- a/docs/user/data_management.rst
+++ b/docs/user/data_management.rst
@@ -190,7 +190,7 @@ user data using the hint ``geomesa.z.splits``:
 
     SimpleFeatureType sft = ...
     sft.setZShards(16)
-    
+
 .. _customizing_z_index:
 
 Customizing the Z-Index
--- a/docs/user/data_management.rst
+++ b/docs/user/data_management.rst
@@ -167,11 +167,13 @@ Configuring Z-Index Shards
 
 GeoMesa Accumulo permits one to specify the number of shards (or splits) into which the Z2 (or Z3) indices are
 divided. This parameter may be changed individually for each ``SimpleFeatureType``. In previous versions of GeoMesa,
-the numberof shards was "hard-coded" to 4, which remains the default. Valid values for the number of shards range
+the numberof shards was fixed at 4, which remains the default. Valid values for the number of shards range
 from 1 to 127 inclusive.
 
-When using very large data sets or large clusters, it may be of benefit to use a larger number of shards. However, in
-most test cases--especially with single-node or small clusters--the default of 4 should be preferred.
+Shards allow us to pre-split tables, which provides some initial parallelism for reads and writes. As more data is
+written, Accumulo will split tables based on size, thus obviating the need for explicit shards. For small data sets,
+shards are more important as the tables might never split from size. Setting the number of shards too high can reduce
+performance, as it requires more calculations to be performed per query.
 
 The number of shards is set when calling ``createSchema``. It may be specified through the simple feature type
 user data using the hint ``geomesa.z.splits``:
@@ -180,15 +182,7 @@ user data using the hint ``geomesa.z.splits``:
 
     // set the hint directly
     SimpleFeatureType sft = ...
-    sft.getUserData().put("geomesa.z.splits", "16");
-
-.. code-block:: java
-
-    // or through using RichSimpleFeatureType
-    import org.locationtech.geomesa.utils.geomesa.RichSimpleFeatureType.RichSimpleFeatureType;
-
-    SimpleFeatureType sft = ...
-    sft.setZShards(16);
+    sft.getUserData().put("geomesa.z.splits", "4");
 
 .. _customizing_z_index:
 
--- a/docs/user/data_management.rst
+++ b/docs/user/data_management.rst
@@ -165,31 +165,30 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
 Configuring Z-Index Shards
 --------------------------
 
-GeoMesa allows you to specify the number of shards (or splits) to be used for each SimpleFeatureType. This gives
-you more flexibility with SimpleFeatureTypes whether they are large or small. By default, SimpleFeatureTypes are
-created with 4 as the number of shards. 4 is the default because previous versions of GeoMesa hardcoded it in
-the z indices.
+GeoMesa Accumulo permits one to specify the number of shards (or splits) into which the Z2 (or Z3) indices are
+divided. This parameter may be changed individually for each ``SimpleFeatureType``. In previous versions of GeoMesa,
+the numberof shards was "hard-coded" to 4, which remains the default. Valid values for the number of shards range
+from 1 to 127 inclusive.
 
-When dealing with large SimpleFeatureTypes or large clusters, it may be beneficial to use a larger number of shards.
-In most test cases on single machines, the default is most likely preferable. It is important to note that acceptable
-values for the number of shards are integers in the range ``[1, 127]``.
+When using very large data sets or large clusters, it may be of benefit to use a larger number of shards. However, in
+most test cases--especially with single-node or small clusters--the default of 4 should be preferred.
 
 The number of shards is set when calling ``createSchema``. It may be specified through the simple feature type
 user data using the hint ``geomesa.z.splits``:
 
-.. code-block:: scala
+.. code-block:: java
 
     // set the hint directly
     SimpleFeatureType sft = ...
-    sft.getUserData().put("geomesa.z.splits", "16")
+    sft.getUserData().put("geomesa.z.splits", "16");
 
-.. code-block:: scala
+.. code-block:: java
 
     // or through using RichSimpleFeatureType
-    import org.locationtech.geomesa.utils.geomesa.RichSimpleFeatureType.RichSimpleFeatureType
+    import org.locationtech.geomesa.utils.geomesa.RichSimpleFeatureType.RichSimpleFeatureType;
 
     SimpleFeatureType sft = ...
-    sft.setZShards(16)
+    sft.setZShards(16);
 
 .. _customizing_z_index:
 
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/SplitArrays.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/SplitArrays.scala
@@ -16,7 +16,7 @@ object SplitArrays {
     new ConcurrentHashMap[Int, Seq[Array[Byte]]]()
 
   def getSplitArray(numSplits: Int): Seq[Array[Byte]] = {
-    require(numSplits < 128, "only up to 128 splits are supported")
+    require(numSplits > 0 && numSplits < 128, "only up to 128 splits are supported")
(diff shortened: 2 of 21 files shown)


elahrvivaz:
@cfkelly @cne1x could we instead make the number of splits configurable as part of the sft user data, same way we do for the date periodicity? Then the number of splits could be retrieved from the simple feature type with something like `sft.getZShards`. Since the sft is already passed around basically everywhere, the interface changes should be minimal. We could have a static cache somewhere of `numSplits -> shard arrays` so that we don't have to keep re-creating the arrays. What do you guys think?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z2/XZ2Index.scala
```diff
@@ -9,9 +9,34 @@
...
+case class XZ2Config(numSplits: Int = XZ2Index.numSplits) extends IndexConfig {
+  val splitArrays = (0 until numSplits).map(_.toByte).toArray.map(Array(_)).toSeq
+}
+
+case class XZ2Index(conf: IndexConfig = DefaultIndexConfig)
+  extends AccumuloFeatureIndexType with IndexConfig with XZ2WritableIndex with XZ2QueryableIndex {
+
+  override val numSplits: Int = conf.numSplits
+  override val splitArrays: Seq[Array[Byte]] = conf.splitArrays
+
```
jahhulbert-ccri: picky - extra linebreaks

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/SplitArrays.scala
```diff
@@ -0,0 +1,19 @@
...
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.accumulo.index
+
+
+
+object SplitArrays {
+  val splitArraysMap: scala.collection.mutable.HashMap[Int, Seq[Array[Byte]]] =
```
elahrvivaz: for thread safety, should make this a java ConcurrentHashMap and modify your getSplitArray to do a get/if-nulll-put. there is a slight chance that an array will be created more than once, but I believe it is faster than full synchronization.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z2/XZ2WritableIndex.scala
```diff
@@ -43,7 +42,7 @@ trait XZ2WritableIndex extends AccumuloWritableIndex {
     val sfc = XZ2SFC(sft.getXZPrecision)
     val sharing = sft.getTableSharingBytes
     (wf: AccumuloFeature) => {
-      val mutation = new Mutation(getRowKey(sfc, sharing)(wf))
+      val mutation = new Mutation(getRowKey(sfc, sharing, sft)(wf))
```
elahrvivaz: probably makes sense to pull out the calls to split arrays and num splits and pass them into the method directly  (same way we do with table sharing bytes) to save us having to look it up for each feature

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z2/Z2WritableIndex.scala
```diff
@@ -33,18 +33,18 @@ trait Z2WritableIndex extends AccumuloWritableIndex {
   }
 
   // split(1 byte), z value (8 bytes), id (n bytes)
-  protected def getPointRowKey(tableSharing: Array[Byte])(wf: AccumuloFeature): Seq[Array[Byte]] = {
+  protected def getPointRowKey(tableSharing: Array[Byte], sft: SimpleFeatureType)(wf: AccumuloFeature): Seq[Array[Byte]] = {
```
elahrvivaz: ditto comment above about passing in split arrays

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z3/XZ3WritableIndex.scala
```diff
@@ -62,9 +63,10 @@ trait XZ3WritableIndex extends AccumuloWritableIndex {
   }
 
   // table sharing (0-1 byte), split (1 byte), time interval(2 bytes), z value (8 bytes), id (n bytes)
-  private def getRowKey(sfc: XZ3SFC, timeToIndex: TimeToBinnedTime, tableSharing: Array[Byte], dtgIndex: Int)
+  private def getRowKey(sfc: XZ3SFC, timeToIndex: TimeToBinnedTime, tableSharing: Array[Byte], dtgIndex: Int,
```
elahrvivaz: ditto pass in split arrays directly

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -308,6 +308,9 @@ object RichSimpleFeatureType {
     def getTableSplitterOptions: Map[String, String] =
       userData[String](TABLE_SPLITTER_OPTS).map(new KVPairParser().parse).getOrElse(Map.empty)
 
+    def setZShards(splits: Int) : Unit = sft.getUserData.put(NUM_SPLITS_KEY, splits.toString)
+    def getZShards: Int = Option[Int](Try(userData[String](NUM_SPLITS_KEY).get.toInt).getOrElse(4)).get
```
elahrvivaz: could rewrite this as:
```
userData[String](NUM_SPLITS_KEY).map(_.toInt).getOrElse(4)
```
generally we don't bother with catching exceptions in the rest of these methods... maybe we should revisit that but in general these are advanced options that we don't expect to fail.

elahrvivaz:
looks good - i suggested a few minor tweaks. could we get a section added to the 'data management' docs too? http://www.geomesa.org/documentation/user/data_management.html#customizing-the-z-index

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/SplitArrays.scala
```diff
@@ -0,0 +1,19 @@
...
+    new scala.collection.mutable.HashMap[Int, Seq[Array[Byte]]]()
+
+  def getSplitArray(numSplits: Int): Seq[Array[Byte]] =
```
elahrvivaz: i would add a `require(numSplits < 128, "only up to 128 splits are supported")`, otherwise I think we'll overflow our positive byte range.

elahrvivaz:
could we get a unit test too?

elahrvivaz:
It would be great if the changes were added here too for non-accumulo data stores: https://github.com/locationtech/geomesa/tree/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index
maybe we can do that as a follow on ticket

docs/user/data_management.rst
```diff
@@ -160,6 +160,37 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
...
 
+.. _configuring_z_shards:
+
+Configuring Z-Index Shards
+--------------------------
+
+GeoMesa allows you to specify the number of shards (or splits) to be used for each SimpleFeatureType. This gives
+you more flexibility with SimpleFeatureTypes whether they are large or small. By default, SimpleFeatureTypes are
+created with 4 as the number of shards. 4 is the default because previous versions of GeoMesa hardcoded it in
+the z indices.
```
mdzimmerman: How about this:

"GeoMesa Accumulo permits one to specify the number of shards (or splits) into which the Z2 (or Z3) indices are divided. This parameter may be changed individually for each ``SimpleFeatureType``. In previous versions of GeoMesa, the number of shards was "hard-coded" set to 4, which remains the default. Valid values for the number of shards range from 1 to 127 inclusive.

When using very large data sets or large clusters, it may be of benefit to use a larger number of shards. However, in most test cases--especially with single-node or small clusters--the default of 4 should be preferred.

The number of shards is set..." and so on.

docs/user/data_management.rst
```diff
@@ -160,6 +160,36 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
...
         .build("mySft")
 
+.. _configuring_z_shards:
+
+Configuring Z-Index Shards
+--------------------------
+
+GeoMesa Accumulo permits one to specify the number of shards (or splits) into which the Z2 (or Z3) indices are
+divided. This parameter may be changed individually for each ``SimpleFeatureType``. In previous versions of GeoMesa,
+the numberof shards was "hard-coded" to 4, which remains the default. Valid values for the number of shards range
```
elahrvivaz: could we change `"hard-coded" to` to `fixed at`

docs/user/data_management.rst
```diff
@@ -160,6 +160,36 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
...
+from 1 to 127 inclusive.
+
+When using very large data sets or large clusters, it may be of benefit to use a larger number of shards. However, in
+most test cases--especially with single-node or small clusters--the default of 4 should be preferred.
```
elahrvivaz: Can we change this paragraph to:
Shards allow us to pre-split tables, which provides some initial parallelism for reads and writes. As more data is written, Accumulo will split tables based on size, thus obviating the need for explicit shards. For small data sets, shards are more important as the tables might never split from size. Setting the number of shards too high can reduce performance, as it requires more calculations to be performed per query.

docs/user/data_management.rst
```diff
@@ -160,6 +160,36 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
...
+
+The number of shards is set when calling ``createSchema``. It may be specified through the simple feature type
+user data using the hint ``geomesa.z.splits``:
+
+.. code-block:: java
+
+    // set the hint directly
+    SimpleFeatureType sft = ...
+    sft.getUserData().put("geomesa.z.splits", "16");
```
elahrvivaz: let's change this to 4 - I don't think we should implicitly encourage people to add more shards without knowing what they're doing.

docs/user/data_management.rst
```diff
@@ -160,6 +160,36 @@ If you are using the GeoMesa ``SftBuilder``, you may call the overloaded attribu
...
+
+.. code-block:: java
+
+    // or through using RichSimpleFeatureType
+    import org.locationtech.geomesa.utils.geomesa.RichSimpleFeatureType.RichSimpleFeatureType;
+
+    SimpleFeatureType sft = ...
+    sft.setZShards(16);
```
elahrvivaz: Did you try this out? i doubt java would work with the implicit wrappers here. I think we can remove this code block, the implicit conversion is more for internal use.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/SplitArrays.scala
```diff
@@ -0,0 +1,24 @@
...
+
+
+object SplitArrays {
+  val splitArraysMap: ConcurrentHashMap[Int, Seq[Array[Byte]]] =
+    new ConcurrentHashMap[Int, Seq[Array[Byte]]]()
+
+  def getSplitArray(numSplits: Int): Seq[Array[Byte]] = {
+    require(numSplits > 0 && numSplits < 128, "only up to 128 splits are supported")
+    val temp = splitArraysMap.get(numSplits)
+    if (temp == null) splitArraysMap.put(numSplits, (0 until numSplits).map(_.toByte).toArray.map(Array(_)).toSeq)
```
elahrvivaz: can avoid a second 'get' like:
```
if (temp == null) {
  val arrays = ...
  splitArraysMap.put(numSplits, arrays)
  arrays
} else {
  temp
}
```

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z3/XZ3Index.scala
```diff
@@ -9,10 +9,13 @@
 package org.locationtech.geomesa.accumulo.index.z3
 
 import org.locationtech.geomesa.accumulo.AccumuloFeatureIndexType
+import org.locationtech.geomesa.accumulo.index.SplitArrays
 import org.opengis.feature.simple.SimpleFeatureType
 
 case object XZ3Index extends AccumuloFeatureIndexType with XZ3WritableIndex with XZ3QueryableIndex {
 
+  def splitArrays(numSplits: Int): Seq[Array[Byte]] = SplitArrays.getSplitArray(numSplits)
```
elahrvivaz: [picky] we probably don't need all these `def splitArrays` methods in each index class - can just call the `SplitArrays.getSplitArray` directly

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/ConfigureShardsTest.scala
```diff
@@ -0,0 +1,61 @@
...
+      feats.head.getType.getZShards mustEqual 6
+    }
+
+    "configure from code" >> {
+      val sftPrivate = sft
+      sftPrivate.setZShards(8)
+      val feats = features(sftPrivate)
+      addFeatures(feats)
+      feats.head.getType.getZShards mustEqual 8
+    }
```
elahrvivaz: this test could be fleshed out a little - right now you're not really testing the stored features, as you're only testing the features that are being created here. probably would need to examine the row keys that get created or something...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/z3/XZ3Index.scala
cfkelly: yeah, I was beginning to think that was a bit overkill as well

elahrvivaz:
looks good, going to merge

PR #1217, opened 2016-11-17
GEOMESA-1529 KryoBufferSimpleFeature - caching transform offset check

elahrvivaz:


elahrvivaz:
@cne1x correct

PR #1219, opened 2016-11-17
GEOMESA-1530,GEOMESA-717 Remove accumulo classloader initialization

elahrvivaz:


elahrvivaz:
we tested with 1.7 and 1.8 using hdfs classloading and it worked - it might break 1.6 compatibility but @anthonyccri was ok with that. I don't think we've tested with non-hdfs classloading either - I'll do that with 1.7 at least.

elahrvivaz:
tested 1.7 with jar in lib/ext and query, filter, transform still works in the iterators.

jahhulbert-ccri:
@elahrvivaz was this fixed by the upgrade to commons 2.1 ? http://accumulo.apache.org/release/accumulo-1.6.6/

That one perhaps? We can merge either way just looking for guidance for 1.6  If that was the fix that we can perhaps add a note in the docs or something

jahhulbert-ccri:
@anthonyccri FYI 1.6.6 does upgrade to commons vfs2 version 2.1

elahrvivaz:
@jahhulbert-ccri I'm not sure... We tested with 1.7 and 1.8 but not 1.6 with either vfs jar. I'll try it out and see.

elahrvivaz:
@jahhulbert-ccri @anthonyccri tested with 1.6.6 - iterators do not work.

elahrvivaz:
update - 1.6.6 does not ship with the updated commons-vfs jar. When I manually udpated the jar to 2.1, it works with 1.6.6 as well.

PR #1226, opened 2016-11-21
GEOMESA-1542 Cassandra create-schema, remove-schema, and ingest commands

ericstalbot:
Signed-off-by: Eric S. Talbot <etalbot@ccri.com>

### DIFF
--- a/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraBackedMetaData.scala
+++ b/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraBackedMetaData.scala
@@ -15,7 +15,7 @@ import org.locationtech.geomesa.index.utils.GeoMesaMetadata
 
 import scala.collection.JavaConversions._
 
-class CassandraBackedMetaData(session: Session, val catalog: String)
+class CassandraBackedMetaData(session: Session, catalog: String)
     extends GeoMesaMetadata[String]{
 
   def ensureTableExists(): Unit = {
--- a/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraBackedMetaData.scala
+++ b/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraBackedMetaData.scala
@@ -15,7 +15,7 @@ import org.locationtech.geomesa.index.utils.GeoMesaMetadata
 
 import scala.collection.JavaConversions._
 
-class CassandraBackedMetaData(session: Session, catalog: String)
+class CassandraBackedMetaData(session: Session, val catalog: String)
     extends GeoMesaMetadata[String]{
 
   def ensureTableExists(): Unit = {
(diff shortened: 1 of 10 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/data/RemoveSchemaCommand.scala
```diff
@@ -46,7 +46,7 @@ trait RemoveSchemaCommand[DS <: GeoMesaDataStore[_, _, _ ,_]] extends DataStoreC
         } else {
           println(s"Removing '$typeName'")
           ds.removeSchema(typeName)
-          if (ds.getSchema(typeName) != null) {
+          if (ds.getTypeNames.contains(typeName)) {  // NB: For Cassandra this only checks metadata
```
elahrvivaz: does `getSchema` not return null for the cassandra data store? I'd like to keep it as `getSchema` vs `getTypeNames` b/c for accumulo and hbase `getTypeNames` scans the whole metadata table instead of a single row.

ericstalbot: Right, CassandraDataStore throws an exception instead of returning null. I haven't found a better way around this yet. CassandraDataStore extends ContentDataStore which has some final methods that make a simple override impossible.

elahrvivaz: I think in our non-ContentDataStores, we mimicked the behavior of JdbcDataStore, which returns null, even though the javadocs for DataStore seem to indicate it should throw an exception... let's make it:
`if (try { ds.getSchema(typeName) != null } catch { case _: IOException => false })`
that should cover both cases...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/data/CreateSchemaCommand.scala
```diff
@@ -31,9 +33,9 @@ trait CreateSchemaCommand[DS <: GeoMesaDataStore[_, _, _ ,_]] extends DataStoreC
     lazy val sftString = SimpleFeatureTypes.encodeType(sft)
     logger.info(s"Creating '${params.featureName}' with spec '$sftString'. Just a few moments...")
 
-    if (ds.getSchema(sft.getTypeName) == null) {
+    if (try { ds.getSchema(sft.getTypeName) == null } catch { case _: IOException => true }) {
       ds.createSchema(sft)
-      if (ds.getSchema(sft.getTypeName) != null) {
+      if (try { ds.getSchema(sft.getTypeName) != null } catch { case _: IOException => false }) {
```
elahrvivaz: there is a disconnect between how getSchema should handle non-existing type names. In our data stores, we return null - I believe we patterned this after the JDBCDataStore. In the ContentDataStore (which CassandraDataStore extends), it throws an IOException. This seems to be the correct implementation, according to the javadocs for DataStore. This check is to handle both the null return value or the IOException. Not sure if we should update our data stores to throw an exception...

ericstalbot: I think the logging you're asking for is already there. For CreateSchema a message is logged if the schema already exists, and for RemoveSchema a message is logged if the schema does not exist. Both commands also log messages for successful completion and for a failure.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraBackedMetaData.scala
```diff
@@ -15,7 +15,7 @@ import org.locationtech.geomesa.index.utils.GeoMesaMetadata
 
 import scala.collection.JavaConversions._
 
-class CassandraBackedMetaData(session: Session, catalog: String)
+class CassandraBackedMetaData(session: Session, val catalog: String)
```
elahrvivaz: yeah, it's overriding a field in GeoMesaMetadata

PR #1231, opened 2016-11-22
1.2 - GEOMESA-1537 Cache converters in Local Ingest for performance

anthonyccri:
* Created an object pool to reuse converters

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/ConverterIngest.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/ConverterIngest.scala
@@ -13,10 +13,12 @@ import java.util.concurrent.atomic.AtomicLong
 
 import com.typesafe.config.{Config, ConfigRenderOptions}
 import com.typesafe.scalalogging.LazyLogging
+import org.apache.commons.pool2.BasePooledObjectFactory
+import org.apache.commons.pool2.impl.{DefaultPooledObject, GenericObjectPool}
 import org.apache.hadoop.mapreduce.Job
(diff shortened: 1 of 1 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/ConverterIngest.scala
```diff
@@ -51,11 +61,13 @@ class ConverterIngest(dsParams: Map[String, String],
         override def getFailure: Long          = failures.get()
       }
 
-      val converter = SimpleFeatureConverters.build(sft, converterConfig)
+      val converter = converterPool.borrowObject()
```
elahrvivaz: nvm, I guess if you have more files than threads they would get reused

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/ingest/ConverterIngest.scala
```diff
@@ -42,6 +44,14 @@ class ConverterIngest(dsParams: Map[String, String],
     ds.createSchema(sft)
   }
 
+  val factory = new BasePooledObjectFactory[SimpleFeatureConverter[_]] {
+    override def wrap(obj: SimpleFeatureConverter[_]) = new DefaultPooledObject[SimpleFeatureConverter[_]](obj)
+    override def create(): SimpleFeatureConverter[_] = SimpleFeatureConverters.build(sft, converterConfig)
+  }
+
+  private val converterPool =
+    new GenericObjectPool[SimpleFeatureConverter[_]](factory)
```
elahrvivaz: presumably the pool size is unbounded? would there ever be a case where a thread had to wait for a pooled object?

PR #1240, opened 2016-11-28
GEOMESA-1545 Caching FeatureWriter mutators

elahrvivaz:

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
@@ -9,8 +9,10 @@
 package org.locationtech.geomesa.index.geotools
 
 import java.io.{Closeable, Flushable}
+import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicLong
 
+import com.github.benmanes.caffeine.cache.Caffeine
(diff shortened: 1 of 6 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
```diff
@@ -113,12 +145,12 @@ abstract class GeoMesaFeatureWriter[DS <: GeoMesaDataStore[DS, F, W], F <: Wrapp
...
 
   override def flush(): Unit = {
-    mutators.collect { case c: Flushable => c }.foreach(_.flush())
+    mutators.collect { case c: Flushable => c.flush() }
     statUpdater.flush()
   }
 
   override def close(): Unit = {
-    mutators.collect { case c: Closeable => c }.foreach(_.close())
+    mutators.collect { case c: Closeable => c.close() }
```
jahhulbert-ccri: only thing here with flush and close is the possibility one throws an exception the others won't get called so maybe close quietly...maybe its not too much of an issue...wonder if there is a good pattern for that

elahrvivaz: I'll fix

elahrvivaz: updated

PR #1243, opened 2016-11-29
GEOMESA-1550 Improve performance of KryoBufferSimpleFeature

anthonyccri:
* Avoid calling `contains`
* Use a mutable buffer for the offsets and lengths

### DIFF
--- a/geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
+++ b/geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
@@ -126,22 +126,30 @@ class KryoBufferSimpleFeature(sft: SimpleFeatureType,
         case _ => -1
       }
     }
+
+    val shouldReserialize = indices.contains(-1)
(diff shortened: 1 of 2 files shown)


geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
```diff
@@ -126,22 +126,30 @@ class KryoBufferSimpleFeature(sft: SimpleFeatureType,
         case _ => -1
       }
     }
+
+    val shouldReserialize = indices.contains(-1)
+    val mutableOffsetsAndLength = Array.ofDim[(Int,Int)](indices.length)
```
elahrvivaz: we could move this down inside the if on line 135 (but before the function definition) so that we don't create it in the reserialize case

elahrvivaz:
looks good

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
anthonyccri: Done

PR #1235, opened 2016-11-22
GEOMESA-1514 Standardizing tools output/logging

elahrvivaz:
~~Note: stacks on top of #1222~~

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #1247, opened 2016-11-30
GEOMESA-993 Better messages/logging/error handling for Convert Framework and Ingest Tools

aheyne:
Stacks on top of #1235


elahrvivaz:
@aheyne the conflicts with master have been fixed in the base branch - you can merge in the changes from there

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/VersionRemoteCommand.scala
```diff
@@ -19,15 +20,21 @@ trait VersionRemoteCommand[DS <: GeoMesaDataStore[_, _, _ ,_]] extends DataStore
...
+  @throws[ParameterException]
+  def getIterVersion: String = {
     try {
-      val iterVersion = withDataStore(_.getVersion._2)
-      Command.output.info(s"Distributed runtime version: $iterVersion")
+      withDataStore(_.getVersion._2)
     } catch {
-      case NonFatal(e) => Command.user.error("Could not get distributed version:", e)
+      case NonFatal(e) =>
+        Command.user.error("Could not get distributed version: ")
```
elahrvivaz: [minor] remove the colon or include the exception

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/GetTypeNamesCommand.scala
```diff
@@ -8,15 +8,29 @@
...
   override val name = "get-type-names"
 
   override def execute(): Unit = {
-    Command.output.info("Current features types:")
-    withDataStore(_.getTypeNames.foreach(Command.output.info))
+    getTypeNames
+  }
+
+  @throws[ParameterException]
+  def getTypeNames = {
```
elahrvivaz: maybe instead of changing this here, we could make the `withDataStore` method check that the data store isn't null and throw a ParameterException if it is. That way we'd be covered in all the commands. Would handle the VersionRemoteCommand too.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/ConverterConfigResolver.scala
```diff
@@ -33,10 +33,13 @@ object ConverterConfigResolver extends LazyLogging {
...
    * @return the converter config parsed from the args
    */
-  def getConfig(configArg: String): Option[Config] =
-    getLoadedConf(configArg)
+  def getConfig(configArg: String): Either[String, Config] = {
+    val conf = getLoadedConf(configArg)
       .orElse(parseFile(configArg))
       .orElse(parseString(configArg))
+    if (conf.isDefined) Right(conf.get)
+    else Left(s"Unable to parse Converter config from argument $configArg")
```
elahrvivaz: this isn't really offering any additional info - maybe instead we'd have a failure message for each thing we tried to do and show them all? Something like:
`"Unable to parse Converter config from argument $configArg. No configuration exists in the environment with that name. No file exists with the name: ${fileError}. Could not parse as a config string: ${stringError}"`
Ideally we'd maybe be able to detect what they are trying (e.g. name, file or string) and report only the correct message.
Same thing with the sft resolver below.

elahrvivaz:
@aheyne I think another part of this ticket was to show better error message when e.g. parsing an sft string. Right now we get something like:
```
scala> val sft = org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.createType("foo", "bar:baz")
java.lang.IllegalArgumentException: org.parboiled.errors.ParsingException: Invalid spec string: Invalid input 'a', expected 'y' or 'o' (line 1, pos 6):
bar:baz
     ^

  at org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes$.createType(SimpleFeatureTypes.scala:95)
  at org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes$.createType(SimpleFeatureTypes.scala:82)
  ... 65 elided
Caused by: org.parboiled.errors.ParsingException: Invalid spec string: Invalid input 'a', expected 'y' or 'o' (line 1, pos 6):
bar:baz
     ^

  at org.locationtech.geomesa.utils.geotools.SimpleFeatureSpecParser$$anonfun$parse$1.apply(SimpleFeatureSpecParser.scala:37)
  at org.locationtech.geomesa.utils.geotools.SimpleFeatureSpecParser$$anonfun$parse$1.apply(SimpleFeatureSpecParser.scala:37)
  at scala.Option.getOrElse(Option.scala:121)
  at org.locationtech.geomesa.utils.geotools.SimpleFeatureSpecParser$.parse(SimpleFeatureSpecParser.scala:36)
  at org.locationtech.geomesa.utils.geotools.SimpleFeatureSpecParser$.parse(SimpleFeatureSpecParser.scala:25)
  ... 67 more
```
(it expects 'y' or 'o' because it's matching 'bytes' or 'boolean', but no one would know that...)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ConvertCommand.scala
```diff
@@ -38,7 +38,7 @@ class ConvertCommand extends Command {
     import scala.collection.JavaConversions._
 
     val sft = CLArgResolver.getSft(params.spec)
-    logger.info(s"Using SFT definition: ${sft}")
+    logger.info(s"Using SFT definition: $sft")
```
cfkelly: Why does this use neither the Command.user logger nor the Command.output logger?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/GenerateAvroSchemaCommand.scala
```diff
@@ -20,11 +20,12 @@ class GenerateAvroSchemaCommand extends Command {
 
   override def execute(): Unit = {
     val sft = SftArgResolver.getSft(params.spec, params.featureName)
-    if (sft.isDefined) {
-      val schema = AvroSimpleFeatureUtils.generateSchema(sft.get, withUserData = true)
-      Command.output.info(schema.toString(true))
-    } else {
-      Command.user.warn(s"Could not find feature type ${params.spec}")
+    sft match {
```
aheyne: Actually now that I look at this, I'm not sure why this wasn't routed through the CLArgResolver. I've swapped it to that.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ConvertCommand.scala
aheyne: Good catch. Fixed

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/GetTypeNamesCommand.scala
aheyne: done

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/VersionRemoteCommand.scala
aheyne: The exception is thrown on the next line, as a param exception (so concatenated). Is that not a good idea?

elahrvivaz: oh, in that case you probably don't need this log statement at all since it will get printed out by the Runner class...

aheyne:
@elahrvivaz 
Could you take a peek at this when you get a chance? It all seems to be working correctly with the exception of the thrown errors. For some reason it's throwing a FileNotFoundException when it should be throwing a ParameterException. As far as I can tell the only time that FileNotFoundException is thrown is in a Try block, so I don't think that's the cause.

elahrvivaz:
@aheyne can you rebase on master? the other PR it's stacked on was merged. I haven't gotten a chance to look at your changes yet, but after you rebase I'll take a peek.

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/CLArgResolverTest.scala
```diff
@@ -76,32 +75,32 @@ class CLArgResolverTest extends Specification {
...
 
   "CLArgResolver" should {
-    "work with " >> {
-      val sft = CLArgResolver.getSft(sftSpec, featureName)
-      sft.getAttributeCount mustEqual 3
-      sft.getDescriptor(0).getLocalName mustEqual "name"
-      sft.getDescriptor(1).getLocalName mustEqual "age"
-      sft.getDescriptor(2).getLocalName mustEqual "geom"
-    }
+//    "work with " >> {
```
aheyne: Note for myself to uncomment this code.

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/CLArgResolverTest.scala
```diff
@@ -76,32 +75,32 @@ class CLArgResolverTest extends Specification {
...
+//      val sft = CLArgResolver.getSft(sftSpec, featureName)
+//      sft.getAttributeCount mustEqual 3
+//      sft.getDescriptor(0).getLocalName mustEqual "name"
+//      sft.getDescriptor(1).getLocalName mustEqual "age"
+//      sft.getDescriptor(2).getLocalName mustEqual "geom"
+//    }
 
     "fail when given spec and no feature name" >> {
```
aheyne: This is the test that is failing because it's throwing a FileNotFoundException rather than passing that exception to the CLArgResolver and throwing a ParamException.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -51,55 +120,51 @@ object SftArgResolver extends LazyLogging {
...
 
   // parse spec conf file
-  private[SftArgResolver] def parseConfFile(specArg: String, name: String): Option[SimpleFeatureType] = {
-    Try(parseConf(FileUtils.readFileToString(new File(specArg)), name)) match {
-      case Success(sftOpt) => sftOpt
-      case Failure(e) =>
-        logger.debug(s"Unable to parse sft spec from file $specArg as conf with error ${e.getMessage}")
-        None
+  private[SftArgResolver] def parseConfFile(specArg: String, name: String): sftEither =
+    parseConf(FileUtils.readFileToString(new File(specArg)), name) match {
```
elahrvivaz: @aheyne your file exception might be coming from here

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,94 @@
...
+    def apply(msg: String, e: Throwable) = { message = msg; error = e }
+  }
+
+  // Type Names for arg type guesses
+  object ArgTypes extends Enumeration {
+    val  NAME, SPECSTR, CONFSTR, PATH = Value
+  }
+  import ArgTypes._
+
+  type resEither = Either[(String, Throwable, Value), returnType]
```
elahrvivaz: in general you should capitalize type names, as they're class aliases

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,94 @@
...
+  type argTuple = parseMethodArgTuple
```
elahrvivaz: This seems redundant - can we just use the type parameter directly?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  /**
+   * Recursively run through parseMethodList attempting to parse the sft.
+   * The most relevant error message is saved in ErrorData and all error are sent to log.
+   * ErrorData is sent back to the CLArgResolver for display to user.
+   */
+  @tailrec
+  final def parseMethods(args: parseMethodArgTuple,
```
elahrvivaz: this could be private? might be better since the calling syntax seems a little complex

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,94 @@
...
+import scala.annotation.tailrec
+
+/**
+ * Abstract trait for resolving sft/config parameters and handling errors around
+ * parsing.
+ */
+trait AbstractArgResolver[returnType, parseMethodArgTuple] extends LazyLogging {
+
+  // Object for holding relevant error to return
+  object ErrorData {
```
elahrvivaz: instead of making this a global state, you could make it an instance class and pass it in through the `parseMethods` calls. that way, the methods will be thread-safe.
e.g.
```
class ErrorData { var message: String = _ ; var error: Throwable = _ }
...
val error = new ErrorData
parseMethods(args, None, parseMethodList, error))
...
 val e = new Throwable(error.message + "\n" + error.error.getMessage, error.error)

File: geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala [hunk 13]
```diff
@@ -0,0 +1,94 @@
...
+      .setAllowMissing(false)
+      .setClassLoader(null)
+      .setIncluder(null)
+      .setOriginDescription(null)
+      .setSyntax(null)
+
+  // Should return a guess of the argType without using the parsers (e.g. regex, contains, etc)
+  def argType(args: parseMethodArgTuple): ArgTypes.Value
+  // Get the arg from a provider on the cp
+  def parseOption(args: parseMethodArgTuple): Option[returnType]
```
elahrvivaz: is there a reason to separate this out, or could we just include it in `parseMethodList` ?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,100 @@ import scala.util.{Failure, Success, Try}
...
-    getLoadedSft(specArg, featureName)
-      .orElse(parseSpecString(specArg, featureName))
-      .orElse(parseConfStr(specArg, featureName))
-      .orElse(parseSpecStringFile(specArg, featureName))
-      .orElse(parseConfFile(specArg, featureName))
+object SftArgResolver extends AbstractArgResolver[SimpleFeatureType, (String, String)] with LazyLogging {
+
+  import ArgTypes._
+
+  override def argType(args: argTuple) = {
```
elahrvivaz: similarly, if you use a case class you can include that in the method definitions to make it much clearer:
`override def argType(args: SpecAndName)`

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,100 @@ import scala.util.{Failure, Success, Try}
...
-        case Success(sft) => Some(sft)
-        case Failure(e) =>
-          logger.debug(s"Unable to parse sft spec from string $specArg with error ${e.getMessage}")
-          None
-      }
+  private[SftArgResolver] def parseSpecStringFile(args: argTuple): resEither =
+    Try(SimpleFeatureTypes.createType(
+      Option(args._2).getOrElse(throw new Throwable("Null feature name provided. " +
+        "Unable to parse spec string from null value.")),
+      FileUtils.readFileToString(new File(Option(args._1).getOrElse(""))))
```
elahrvivaz: Does it make sense to replace the `getOrElse("")` with an exception like 'No input file specified' ?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,100 @@ import scala.util.{Failure, Success, Try}
...
+    ) match {
+      case Success(sft) => Right(sft)
+      case Failure(e) => Left((s"Unable to parse sft spec from string ${args._1} with error ${e.getMessage}", e, PATH))
```
elahrvivaz: this should be 'unable to parse sft spec from file' instead of 'from string'

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,100 @@ import scala.util.{Failure, Success, Try}
...
+  }
 
   // gets an sft based on a spec conf string
-  private[SftArgResolver] def parseConfStr(specArg: String, name: String): Option[SimpleFeatureType] =
-    Try(parseConf(specArg, name)) match {
-      case Success(sftOpt) => sftOpt
-      case Failure(e) =>
-        logger.debug(s"Unable to parse sft spec from string $specArg as conf with error ${e.getMessage}")
-        None
+  private[SftArgResolver] def parseConfStr(args: argTuple): resEither =
```
elahrvivaz: i think it's clearer to use `Either[Throwable, SimpleFeatureType]` instead of resEither - it's not so complex and long that there is much to gain from the type alias

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,100 @@ import scala.util.{Failure, Success, Try}
...
-      case Failure(e) =>
-        logger.debug(s"Unable to parse sft spec from file $specArg as conf with error ${e.getMessage}")
-        None
+  private[SftArgResolver] def parseConfFile(args: argTuple): resEither =
+    Try {
+      FileUtils.readFileToString(new File(args._1))
+    } match {
+      case Success(file) => parseConf(file, args._2) match {
+          case Right(sftOpt) => Right(sftOpt)
+          case Left(e) => Left((s"Unable to parse sft spec from file ${args._1} as conf.", e, PATH))
```
elahrvivaz: [minor] i think you could use Either.left here instead of the match - that lets you modify the Left if it's a Left but keep the Right unchanged if it's a Right:
`parseConf(file, args._2).left.map(e => (s"Unable to parse sft spec from file ${args._1} as conf.", e, PATH))
there's no functional difference, but that's the idiomatic way to use Either

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: I use argTuple in the classes that inherit this one. For some reason the type parseMethodArgTuple isn't available there.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,94 @@
...
+  // (Ordered) List of parse methods to attempt
+  def parseMethodList: List[parseMethodArgTuple => resEither]
+
+  /**
+   * @return the SFT parsed from the Args
```
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,94 @@
...
+import com.typesafe.config.ConfigParseOptions
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.annotation.tailrec
+
+/**
+ * Abstract trait for resolving sft/config parameters and handling errors around
+ * parsing.
+ */
+trait AbstractArgResolver[returnType, parseMethodArgTuple] extends LazyLogging {
```
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
aheyne: ResEither is actually, Either[(String, Throwable, Value), ReturnType] because we need to return the type that function is looking for so we can preserve that message if it matches our type guess. we could maybe just have Either[(Throwable, Value), ReturnType] and just pull the message from the exception but that doesn't really gain us anything.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,100 @@ import scala.util.{Failure, Success, Try}
...
-  /**
-   * @return the SFT parsed from the Args
-   */
-  def getSft(specArg: String, featureName: String = null): Option[SimpleFeatureType] =
-    getLoadedSft(specArg, featureName)
-      .orElse(parseSpecString(specArg, featureName))
-      .orElse(parseConfStr(specArg, featureName))
-      .orElse(parseSpecStringFile(specArg, featureName))
-      .orElse(parseConfFile(specArg, featureName))
+object SftArgResolver extends AbstractArgResolver[SimpleFeatureType, (String, String)] with LazyLogging {
```
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: Fixed

aheyne:
Changes up

elahrvivaz:
looks good - @jahhulbert-ccri can you take a peek?
@aheyne another part of this ticket was to improve the error messages when e.g. parsing a sft string. I'm going to open a new ticket for that so we don't lose track of it.

elahrvivaz:
@aheyne 933 seems like the wrong ticket - I don't see another one assigned to you that matches. Do you know the right number?

elahrvivaz:
opened follow-on ticket here: https://geomesa.atlassian.net/browse/GEOMESA-1571
fixed title to correct ticket number, make sure the commit message it right when merging

jahhulbert-ccri:
@elahrvivaz  taking a look now

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Command.scala
```diff
@@ -52,8 +52,10 @@ trait DataStoreCommand[DS <: DataStore] extends Command {
 
   def connection: Map[String, String]
 
+  @throws[ParameterException]
   def withDataStore[T](method: (DS) => T): T = {
-    val ds = DataStoreFinder.getDataStore(connection).asInstanceOf[DS]
+    val ds = Option(DataStoreFinder.getDataStore(connection).asInstanceOf[DS])
+      .getOrElse(throw new ParameterException("Unable to retrieve datastore."))
```
jahhulbert-ccri: good idea

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,93 @@
...
+import com.typesafe.config.ConfigParseOptions
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.annotation.tailrec
+
+/**
+ * Abstract trait for resolving sft/config parameters and handling errors around
+ * parsing.
+ */
+trait AbstractArgResolver[ReturnType, ParseMethodArgs] extends LazyLogging {
```
jahhulbert-ccri: maybe not name the trait "Abstract" since its not an abstract class just name it ArgResolver?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  // Object for holding relevant error to return
+  class ErrorData {
```
jahhulbert-ccri: Is there a reason you can't use Exception here since Exception has message and causedBy?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Recursively run through parseMethodList attempting to parse the sft.
+   * The most relevant error message is saved in ErrorData and all error are sent to log.
+   * ErrorData is sent back to the CLArgResolver for display to user.
+   */
+  @tailrec
+  private def parseMethods(args: ParseMethodArgs,
+                         tryMethod: Option[ParseMethodArgs => ResEither] = None,
```
jahhulbert-ccri: picky - indent these method args to match the top one

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,101 @@ import scala.util.{Failure, Success, Try}
...
+     * Here we use rudimentary checking to guess as what kind of specArg was passed in.
+     * We use this to decide which error message to display to the user, since the
+     * parsers fail frequently. The rest of the errors are logged.
+     */
+    val fileNameReg = """([^.]*)\.([^.]*)""" // e.g. "foo.bar"
+    val specStrReg = """^[a-zA-Z0-9]+[:][String|Integer|Double|Point|Date|Map|List].*""" // e.g. "foo:String..."
+    val specStrRegWError = """^[a-zA-Z0-9]+[:][a-zA-Z0-9]+.*""" // e.g. "foo:Sbartring..."
+    args.specArg match {
+      // Order is important here
+      case s if s.contains("geomesa{")
```
jahhulbert-ccri: wondering if this should be something along the lines of having some trait with classes for each that can have some methods like canProces(arg: String), process(arg:String): Try

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SftArgResolver.scala
```diff
@@ -22,84 +23,101 @@ import scala.util.{Failure, Success, Try}
...
-      .orElse(parseConfFile(specArg, featureName))
+  override def argType(args: SftArgs) = {
+    /**
+     * Here we use rudimentary checking to guess as what kind of specArg was passed in.
+     * We use this to decide which error message to display to the user, since the
+     * parsers fail frequently. The rest of the errors are logged.
+     */
+    val fileNameReg = """([^.]*)\.([^.]*)""" // e.g. "foo.bar"
+    val specStrReg = """^[a-zA-Z0-9]+[:][String|Integer|Double|Point|Date|Map|List].*""" // e.g. "foo:String..."
+    val specStrRegWError = """^[a-zA-Z0-9]+[:][a-zA-Z0-9]+.*""" // e.g. "foo:Sbartring..."
```
jahhulbert-ccri: what is WError stand for?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/ConverterConfigResolver.scala
```diff
@@ -10,61 +10,72 @@ package org.locationtech.geomesa.convert
...
 
-  /**
-   * @return the converter config parsed from the args
-   */
-  def getConfig(configArg: String): Option[Config] =
-    getLoadedConf(configArg)
-      .orElse(parseFile(configArg))
-      .orElse(parseString(configArg))
+  override def argType(args: ConfArgs) = {
+    /**
```
jahhulbert-ccri: picky - use /* instead of /** since its not java docing the variable below

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/ConverterConfigResolver.scala
```diff
@@ -10,61 +10,72 @@ package org.locationtech.geomesa.convert
...
 
 import scala.util.{Failure, Success, Try}
 
 /**
  * Attempts to resolve Converter config from arguments as either a string or
  * as a filename containing the converter config
  */
-object ConverterConfigResolver extends LazyLogging {
+case class ConfArgs(config: String)
+object ConverterConfigResolver extends AbstractArgResolver[Config, ConfArgs] with LazyLogging {
```
jahhulbert-ccri: move the javadoc from lines 19-20 to just above the ConverterConfigResolver so it is attached to that rather than the ConfArgs case class

jahhulbert-ccri: wondering if you need a case class for just 1 string

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: You can't set the message after the Exception has been created. We need to be able to update the message, possibly on every pass.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: Changed

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/ConverterConfigResolver.scala
aheyne: I think we do for the abstraction to work. SftArgResolver has 2 params in it's case class.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/ConverterConfigResolver.scala
aheyne: Fixed

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/AbstractArgResolver.scala
aheyne: Fixed

PR #1248, opened 2016-11-30
1.2 - GEOMESA-1552 Fixing check for join query against BIN non-default geometries

elahrvivaz:
* Changing trackId to be an Int in case classes to remove inconsistency with BinAggregatingIterator


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/encoders/BinEncoder.scala
```diff
@@ -36,7 +36,7 @@ class BinEncoder(sft: SimpleFeatureType, trackIdField: String) {
     val (lat, lon) = getLatLon(sf)
     val dtg = getDtg(sf)
     val trackIdVal = sf.getAttribute(trackIdIndex)
-    val trackId = if (trackIdVal == null) "" else trackIdVal.toString
+    val trackId = if (trackIdVal == null) { 0 } else { trackIdVal.hashCode }
```
jnh5y: Does the trackIdVal need to come through?  Is it ever used directly?

PR #1253, opened 2016-12-01
GEOMESA-1119 Catching exceptions in feature listeners in Kafka DataStore

elahrvivaz:


jnh5y:
The only addition which may make sense here is to add a unit test with an 'exceptional' feature listener.  Not necessary.

PR #1255, opened 2016-12-02
GEOMESA-1300 Data jar with just converter and sft definitions

elahrvivaz:
* Available as artifactId 'geomesa-tools_${scala.binary.version}' with classifier 'data'


elahrvivaz:
you can, you just have to import it in the geomesa-tools/conf/reference.conf. Is that ok?

jnh5y:
Is there anyway to avoid the two step process to add sft confs?

elahrvivaz:
yeah, you can just put the file at the top-level of the conf dir. I wanted to keep the classloading in the tools in sync with the jar - otherwise I figured that someone would add a new sft but forget to add it to the jar. But we can also roll back to just searching for sft folders in the conf file.

jnh5y:
I wonder if anyone is using the ability to add SFTs, etc.  

Could we split the difference?  Load up the jar and search the conf directory structures?

elahrvivaz:
yeah, i kept the existing ones in conf instead of a jar to make it easier to edit them if desired. but I could include the jar instead

jnh5y:
Ok, do we already have a tutorial/other documentation on what the GeoMesa converters do?  Maybe we should update that and explain how things change?  @mdzimmerman thoughts?

mdzimmerman:
There is not a tutorial. There is, however, a chapter on the converters library in ``docs/user/convert.rst`` (http://www.geomesa.org/documentation/current/user/convert.html) which I think would be the appropriate place to describe the changes.

elahrvivaz:
I'll update the docs - @jahhulbert-ccri @jnh5y are we satisfied with the current behavior?

mdzimmerman:
In the final section ("Loading Converters and SFTs at Runtime"), there's a reference to gm-data that should probably be removed.

jnh5y:
@elahrvivaz I think the behavior in this PR is sensible.  My main push is to document it.

elahrvivaz:
updated docs

PR #1259, opened 2016-12-02
GEOMESA-1556 System property to disable stat generation

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
+++ b/geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
@@ -193,6 +193,17 @@
     </property>
 
     <property>
+        <name>geomesa.stats.generate</name>
+        <value></value>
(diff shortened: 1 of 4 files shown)


geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -193,6 +193,17 @@
     </property>
 
     <property>
+        <name>geomesa.stats.generate</name>
+        <value></value>
```
aheyne: Do we need a default value here?

elahrvivaz: default for the system property is null - the default for a datastore is handled in the data store configuration parameter 'generateStats'. Figured it would be confusing to have 2 defaults.

jahhulbert-ccri:
yay :+1:

PR #1263, opened 2016-12-05
1.2 - GEOMESA-1558 Not applying transform iterator for identity transforms

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -35,7 +35,7 @@ import org.locationtech.geomesa.utils.index.IndexMode.IndexMode
 import org.locationtech.geomesa.accumulo.index.geohash.GeoHashIndex
 import org.locationtech.geomesa.accumulo.index.id.RecordIndex
 import org.locationtech.geomesa.accumulo.iterators.ProjectVersionIterator
-import org.locationtech.geomesa.accumulo.util.{DistributedLocking, GeoMesaBatchWriterConfig, Releasable}
+import org.locationtech.geomesa.accumulo.util.{DistributedLocking, Releasable}
 import org.locationtech.geomesa.features.SerializationType
 import org.locationtech.geomesa.features.SerializationType.SerializationType
 import org.locationtech.geomesa.index.stats.{GeoMesaStats, HasGeoMesaStats}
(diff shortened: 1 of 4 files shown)


elahrvivaz:
this needs some more work - causes issues in geoserver

elahrvivaz:
this is working in geoserver now. for existing layers, it might require deleting and re-loading the layer in geoserver for GML to work

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -156,21 +151,20 @@ class AccumuloDataStore(val connector: Connector,
...
-   * @see org.geotools.data.DataStore#getSchema(java.lang.String)
-   * @param typeName feature type name
-   * @return feature type, or null if it does not exist
-   */
-  override def getSchema(typeName: String): SimpleFeatureType = getSchema(new NameImpl(typeName))
+    * @see org.geotools.data.DataAccess#getSchema(org.opengis.feature.type.Name)
+    * @param name feature type name
+    * @return feature type, or null if it does not exist
+    */
+  override def getSchema(name: Name): SimpleFeatureType = getSchema(name.getLocalPart)
```
jahhulbert-ccri: haha yeah good change

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
```diff
@@ -324,15 +324,19 @@ object QueryPlanner extends LazyLogging {
...
+    * @param query query
+    * @param sft simple feature type
+    * @return
+    */
+  def setQueryTransforms(query: Query, sft: SimpleFeatureType) = {
+    import scala.collection.JavaConversions._
+    val properties = query.getPropertyNames
+    query.setProperties(Query.ALL_PROPERTIES)
+    if (properties != null && properties.nonEmpty &&
+        properties.toSeq != sft.getAttributeDescriptors.map(_.getLocalName)) {
```
jahhulbert-ccri: Is there any chance these could be ordered differently and cause the equality check to fail? perhaps a set or ordered seq is better

jahhulbert-ccri: or is that part of the identity transformation i guess?

elahrvivaz: i was thinking that if they're ordered differently then that's an actual transformation that we should apply, but now that you mention it I'm not 100% sure...

elahrvivaz: @jnh5y do you know?

jahhulbert-ccri: @elahrvivaz that would make sense. I think an SFT is ordered by definition since you can reference things by index.

jahhulbert-ccri:
```
 * The type of a SimpleFeature.
 * <p>
 * The definition of a "simple feature" can be summed up as the following:
 * <ul>
 * <li>made up of only non-complex attributes, no associations
 * <li>attributes are of multiplicity 1
 * <li>attributes are ordered
 * <li>attribute names are unqualified (namespaceURI == null)
 * </ul>
``` yeah

PR #1270, opened 2016-12-09
GEOMESA-1573 Implement test framework for GeoMesa SparkSQL support.

mdzimmerman:


geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLDataTest.scala
```diff
@@ -0,0 +1,140 @@
...
+
+@RunWith(classOf[JUnitRunner])
+class SparkSQLDataTest extends Specification {
+  val createPoint = JTSFactoryFinder.getGeometryFactory.createPoint(_: Coordinate)
+
+  "sql data tests" should {
+    sequential
+
+    System.setProperty(QueryProperties.SCAN_RANGES_TARGET.property, "1")
+    System.setProperty(AccumuloQueryProperties.SCAN_BATCH_RANGES.property, s"${Int.MaxValue}")
```
elahrvivaz: should probably put these into the 'before' step, and also need to clear them into an 'after' step

geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLDataTest.scala
```diff
@@ -0,0 +1,140 @@
...
+      sdf.createOrReplaceTempView("states")
+
+      broadcast(sdf).createOrReplaceTempView("broadcastStates")
+
+      sdf.collect.length mustEqual 56
+    }
+
+    "basic sql 1" >> {
+      val r = sc.sql("select * from chicago where case_number = 1")
+      r.show()
```
elahrvivaz: does this print to std out? should probably hide this in a debug statement or something if possible, or remove if not

geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLDataTest.scala
```diff
@@ -0,0 +1,140 @@
...
+
+    "ingest geonames" >> {
+      SparkSQLTestUtils.ingestGeoNames(dsParams)
+
+      gndf = spark.read
+        .format("geomesa")
+        .options(dsParams)
+        .option("geomesa.feature", "geonames")
+        .load()
+      gndf.printSchema()
```
elahrvivaz: do these printSchema calls print to std out?

geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLGeometryConstructorsTest.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class SparkSQLGeometryConstructorsTest extends Specification {
+
+  "sql geometry constructors" should {
+    sequential
+
+    System.setProperty(QueryProperties.SCAN_RANGES_TARGET.property, "1")
+    System.setProperty(AccumuloQueryProperties.SCAN_BATCH_RANGES.property, s"${Int.MaxValue}")
```
elahrvivaz: need to set/clear these in steps as mentioned above

geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLSpatialRelationshipsTest.scala
```diff
@@ -0,0 +1,203 @@
...
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class SparkSQLSpatialRelationshipsTest extends Specification {
+  "SQL spatial relationships" should {
+    sequential
+
+    System.setProperty(QueryProperties.SCAN_RANGES_TARGET.property, "1")
+    System.setProperty(AccumuloQueryProperties.SCAN_BATCH_RANGES.property, s"${Int.MaxValue}")
```
elahrvivaz: ditto sys properties

geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLTestUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.locationtech.geomesa.utils.geotools.{GeneralShapefileIngest, SimpleFeatureTypes}
+
+import scala.collection.JavaConversions._
+
+object SparkSQLTestUtils {
+  def setupMiniAccumulo(): MiniAccumuloCluster = {
+    val randomDir = Files.createTempDirectory("mac").toFile
+    val config = new MiniAccumuloConfig(randomDir, "password").setJDWPEnabled(true)
+    val mac = new MiniAccumuloCluster(config)
+    mac.start()
```
elahrvivaz: we should probably shut down the cluster after the tests? Along those lines, it would probably be a good idea to have a globally shared mini cluster that is spun up before running any tests and stopped at the end. I believe that is doable through maven - possibly you could add some hooks so that if the tests are run outside maven they will start up a new cluster.

mdzimmerman: Yeah, apparently the Accumulo integration test suite does something like this. I'm looking into it.

geomesa-spark/geomesa-spark-accumulo/src/test/scala/org/locationtech/geomesa/spark/accumulo/SparkSQLTestUtils.scala
```diff
@@ -18,14 +18,26 @@ import org.geotools.data.simple.SimpleFeatureStore
...
 
 import scala.collection.JavaConversions._
 
 object SparkSQLTestUtils {
+  def setProperties(): Unit = {
+    System.setProperty(QueryProperties.SCAN_RANGES_TARGET.property, "1")
+    System.setProperty(AccumuloQueryProperties.SCAN_BATCH_RANGES.property, s"${Int.MaxValue}")
+  }
+
+  def clearProperties(): Unit = {
```
elahrvivaz: thanks for updating this - now that i think about it some more, i still see 2 potential issues:

1. the tests might conflict with each other - if a test starts running while another one finishes, the properties might be cleared part-way through the second test. I believe tests are run concurrently in our builds, so this might be a problem
2. we set the SCAN_RANGES_TARGET property in our maven opts, so that our accumulo tests don't blow up (mock accumulo doesn't handle lots of ranges very well). So we need to cache the original value for the prop, and re-set it afterwards

There's actually a thread-local value for system properties that you can use instead of setting them globally - you can set and clear this in a 'beforeEachTest' and 'afterEachTest' method (I think in specs 2 you use the @before and @after annotations). You do something like:
```
AccumuloQueryProperties.SCAN_BATCH_RANGES.threadLocalValue.set("1")
...
AccumuloQueryProperties.SCAN_BATCH_RANGES.threadLocalValue.clear()
```

mdzimmerman:
This PR is replaced by: https://github.com/locationtech/geomesa/pull/1276

PR #1271, opened 2016-12-09
GEOMESA-1571 Improved error reporting in simple feature spec parsing

elahrvivaz:


elahrvivaz:
@jahhulbert-ccri thanks - if you come up with any other common mistakes that aren't being reported, let me know

jahhulbert-ccri:
can we also modify this message to ensure the error makes it to the log: https://github.com/locationtech/geomesa/blob/dba6401e103318f29ca369d4d7e9ba3517d07c0f/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/CLArgResolver.scala#L29

Looks like we swallow everything

jahhulbert-ccri:
or a follow-on ticket is ok to make sure the CL tool logging is improved

elahrvivaz:
that should be handled in #1247 i think

PR #1278, opened 2016-12-14
GEOMESA-1579 Fixed reference to function that no longer exists

aheyne:

### DIFF
--- a/geomesa-tools/bin/common-functions.sh
+++ b/geomesa-tools/bin/common-functions.sh
@@ -96,19 +96,14 @@ function containsElement() {
   return 1
 }
 
-# Define %%gmtools.dist.name%%_HOME and update the PATH if necessary.
-if [[ -z "$%%gmtools.dist.name%%_HOME" ]]; then
-  setGeoHome
-else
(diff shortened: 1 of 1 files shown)


geomesa-tools/bin/common-functions.sh
```diff
@@ -98,14 +98,14 @@ function containsElement() {
 
 # Define %%gmtools.dist.name%%_HOME and update the PATH if necessary.
 if [[ -z "$%%gmtools.dist.name%%_HOME" ]]; then
-  setGeoHome
+  geomesaConfigure
```
elahrvivaz: actually I don't think we want to call configure here... that will prompt the user. The home variable is being set in the script for each distro before calling this function:
https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa#L12
so I think we can just remove the 'if' part of this statement

aheyne: Ah, yeah, you're right. Fixed

PR #1284, opened 2016-12-15
GEOMESA-1518 Better instance and ZK arg resolving

aheyne:


geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -93,24 +98,42 @@ object AccumuloRunner extends Runner {
...
-          Command.user.debug("You can specify the Instance Id via the command line or\n" +
-            "change the Zookeeper timeout by setting the system property 'instance.zookeeper.timeout'.")
+        Command.user.debug(s"Looking up Accumulo Instance Id in Zookeeper for $lookupTime milliseconds.")
+        Command.user.debug("You can specify the Instance Id via the command line or\n" +
+          "change the Zookeeper timeout by setting the system property 'instance.zookeeper.timeout'.")
 
-          import scala.concurrent.duration._
-          Await.result(Future(HdfsZooInstance.getInstance().getInstanceName)(ExecutionContext.global),  lookupTime.millis)
+        import scala.concurrent.duration._
+        lazy val ff: Either[Throwable, String] = try {
```
elahrvivaz: instead of an either here, this seems like a good place for a Try - then you can use .get (which throws if the Try is a Failure)

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -93,24 +98,42 @@ object AccumuloRunner extends Runner {
...
+          val i = HdfsZooInstance.getInstance().getInstanceName
+          Right(i)
         } catch {
-          case NonFatal(e) => Command.user.warn(s"Exception getting zoo instance: ${e.toString}"); null
+          case e: Throwable => Left(e)
         }
+        val future = Await.result(Future(ff)(ExecutionContext.global),  lookupTime.millis)
+
+        future match {
+          case Right(instanceId) => if (p.instance == null) p.instance = instanceId
```
elahrvivaz: is there a reason to run this block if `p.instance` is not null? we should put the `if (p.instance == null)` check back up front and skip all this lookup

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -93,24 +98,42 @@ object AccumuloRunner extends Runner {
...
+        }
+      } catch {
+        case e @ ( _: ClassNotFoundException | _: NoClassDefFoundError ) =>
+          throw new ParameterException("Unable to locate Accumulo instance. Please ensure that $ACCUMULO_HOME is set " +
+            "correctly and/or provide the instance name with the parameter --instance.", e)
+        case e: TimeoutException =>
+          throw new ParameterException("Unable to connect to Zookeeper instance " +
+          "within timeout period. Consider increasing the Zookeeper timeout by setting the system property " +
+          "'instance.zookeeper.timeout' (default: 5000ms).", e)
+        case NonFatal(e) => Command.user.warn(s"Exception getting zoo instance: ${e.toString}"); null
```
elahrvivaz: you can remove the `;null` since this block isn't setting any value.
wouldn't an exception here (meaning zookeepers or instance id is not set) be a failure condition? we should throw an exception so that it bubbles back up and stops the command execution

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
aheyne: Yes, because in testing I found that if you specify an instance and it still can't find it, it will later throw errors that are less than helpful. By checking here we can use the same error message.

elahrvivaz: ok, makes sense - could you add some comments to that effect?
Are there any cases where this check will fail, but the actual process will still succeed? And should we check for a mismatch in the instance specified vs the instance retrieved (or is that valid)?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
aheyne: Unfortunately that doesn't work. The way I have it is the only way I could find that would both catch the exception in the Future and when it comes back. Originally it was throwing the stackstrace twice. Using a Try with .get here only catches the stackstace after the future returns.

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
aheyne: Fixed the null and converted the NonFatal to a catchall.

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
aheyne: Added comments. 
I changed the check to look at the class of the params for the command that was invoked. This should mostly prevent unnecessary checks and allow commands that don't use Accumulo resources through. 
Added check and warning if the instance names don't match. I don't really understand the implications/repercussions here so that seems appropriate.

elahrvivaz:
think this change is causing the build to fail on tools unit tests

aheyne:
Should be fixed now

jnh5y:
@elahrvivaz is this good to go for 1.3.0?

elahrvivaz:
@jnh5y @aheyne let's leave this for 1.3.1 since smoke testing is mostly complete and this looks like it requires some testing to verify.

jahhulbert-ccri:
looks pretty good. I'll likely end up rebasing #1335 on it once its in...

elahrvivaz:
looks good, we need to test it out in a variety of situations (site.xml, no site.xml, provided args, no args, etc)

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -90,14 +95,16 @@ object AccumuloRunner extends Runner {
...
-      if (p.zookeepers == null) {
-        p.zookeepers = zookeepers
-      }
-      if (p.instance == null) {
-        p.instance = try {
+      if (! p.mock) {
+        if (p.zookeepers == null) {
+          p.zookeepers = zookeepers
+        }
+
```
jahhulbert-ccri: also on -i ingest fails

```
ERROR Accumulo Site XML was not found, unable to locate zookeepers. Please provide the --zookeepers parameter or ensure the file $ACCUMULO_HOME/conf/accumulo-site.xml exists or optionally, specify a different configuration file with the System Property 'geomesa.tools.accumulo.site.xml'.

```

jnh5y:
@aheyne are the requested changes clear?  Also, can you knock out the merge conflicts?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
aheyne: I'm not really seeing the same behavior here. Here are my test results.

Without ACCUMULO_HOME set or classpath additions:
-z and -i worked
-z only failed
-i only failed

With ACCUMULO_HOME set:
-z and -i worked
-z only worked 
-i only worked

Without ACCUMULO_HOME set but with classpath additions (no accumulo-site.xml):
-z and -i worked
-z only worked
-i only failed

Without ACCUMULO_HOME set but with classpath additions and geomesa.tools.accumulo.site.xml set:
-z and -i worked
-z only worked
-i only worked

These results seem to make sense to me.

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -107,9 +114,33 @@ object AccumuloRunner extends Runner {
...
+            case e: Throwable => Left(e)
+          }
+          val future = Await.result(Future(ff)(ExecutionContext.global),  lookupTime.millis)
+
+          future match {
+            case Right(instanceId) =>
+              if (p.instance == null) p.instance = instanceId
+              else if (instanceId != p.instance) Command.user.warn("The provided instance ID (--instance) does not " +
+                "match the instance ID of the resolved Accumulo instance.")
+            case Left(e) => throw e
```
elahrvivaz: @aheyne it looks like it will throw an exception if it can't load `accumulo-site.xml` regardless of it -i is defined (that also aligns with what @jahhulbert-ccri reported). We want to be able to specify -i and not require `accumulo-site.xml` on the classpath. (`HdfsZooInstance.getInstance().getInstanceName` loads `accumulo-site.xml`)
It seems like loading the zookeepers separately is also pointless, we can just use `HdfsZooInstance.getInstance().getZookeepers`, which does the same thing.
Is there any situation where you would need only one of -i or -z? It seems like the flow should be:
1. load classOf[Connection], catch ClassNotFoundException, throw param exception for jars/classpath
2. If one or both of -i and -z not set? Load HdfsZooInstance, validate any that are set, set to the loaded values. if can't load, throw a ParamException
3. any other case, don't load HdfsZooInstance - we don't need `accumulo-site.xml` on the classpath

aheyne: Preface: this is from an outdated diff. 

As it is now: All of the errors I experienced while testing in my previous [comment](https://github.com/locationtech/geomesa/pull/1284#discussion_r100390347) are listed there.

Line 133: the error thrown here from the future has to do with looking up the instance name in zookeeper. If we're to this point then -i was not provided and getting the instance from the zookeeper was our last resort so if it's not found and error should be thrown. In addition we're only here if the zookeeper has been resolved, either by -z or the accumulo-site.xml (p.zookeepers != null)

The accumulo-site.xml is only needed with -z is not provided.

elahrvivaz:
looks good now

aheyne:
This successfully builds on my machine. Unsure what problem travis is having.

elahrvivaz:
looks like travis is failing on the parameter stuff - it's likely building for you since you have `ACCUMULO_HOME` set... can you try unsetting it and see if it still builds locally?

aheyne:
Going to tackle this one first thing in the morning.

aheyne:
Retesting....
Without ACCUMULO_HOME set or classpath additions:
-z and -i worked
-z only failed with error 
`ERROR Unable to locate Accumulo instance. Please ensure that $ACCUMULO_HOME is set correctly and/or provide the instance name with the parameter --instance.`
-i only failed with error
`ERROR Accumulo Site XML was not found or was unable to be read, unable to locate zookeepers. Please provide the --zookeepers parameter or ensure the file $ACCUMULO_HOME/conf/accumulo-site.xml exists or optionally, specify a different configuration file with the System Property 'geomesa.tools.accumulo.site.xml'.`

With ACCUMULO_HOME set, no classpath additions:
-z and -i worked
-z only worked
-i only worked

Without ACCUMULO_HOME set but with classpath additions (no accumulo-site.xml):
-z and -i worked
-z only worked
-i only failed with error
`ERROR Accumulo Site XML was not found or was unable to be read, unable to locate zookeepers. Please provide the --zookeepers parameter or ensure the file $ACCUMULO_HOME/conf/accumulo-site.xml exists or optionally, specify a different configuration file with the System Property 'geomesa.tools.accumulo.site.xml'.`

Without ACCUMULO_HOME set but with classpath additions and geomesa.tools.accumulo.site.xml set:
-z and -i worked
-z only worked
-i only worked

elahrvivaz:
@aheyne by classpath additions you mean you copied the accumulo jars into lib? if so, that looks good. I don't think we need to worry too much about supporting only -z or -i - expecting both or neither seems reasonable (although bonus if it works with just one)

elahrvivaz:
@aheyne unit tests are still failing in travis - could you try building without ACCUMULO_HOME set? possibly we need to add a check for mock connections

aheyne:
> @aheyne by classpath additions you mean you copied the accumulo jars into lib? if so, that looks good. I don't think we need to worry too much about supporting only -z or -i - expecting both or neither seems reasonable (although bonus if it works with just one)

Yes, "classpath additions" simply means I ran 
`bin/install-hadoop-accumulo.sh lib/`
to dump the hadoop-accumulo jars in the geomesa-accumulo-tools/lib folder

> @aheyne unit tests are still failing in travis - could you try building without ACCUMULO_HOME set? possibly we need to add a check for mock connections

Yeah, you caught me before I pushed up the commit for that test message I just posted. It is compiling for me now without $ACCUMULO_HOME set.

`[INFO] ------------------------------------------------------------------------`
`[INFO] Reactor Summary:`
`[INFO] `
`[INFO] GeoMesa Accumulo Tools ............................. SUCCESS [ 32.038 s]`
`[INFO] GeoMesa Accumulo Distribution ...................... SUCCESS [ 22.405 s]`
`[INFO] ------------------------------------------------------------------------`
`[INFO] BUILD SUCCESS`
`[INFO] ------------------------------------------------------------------------`
`[INFO] Total time: 55.531 s`
`[INFO] Finished at: 2017-03-02T20:37:47+00:00`
`[INFO] Final Memory: 109M/982M`
`[INFO] ------------------------------------------------------------------------`
`aheyne@aheyne:/opt/devel/geomesa$ $ACCUMULO_HOME`
`aheyne@aheyne:/opt/devel/geomesa$ `

jahhulbert-ccri:
builds fine on command line

PR #1286, opened 2016-12-16
GEOMESA-1591 Added GeoMesa SparkSQL accessor functions

Afterglow375:


geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
```diff
@@ -0,0 +1,87 @@
...
+      case geom: LineString =>
+        if (n < 0) {
+          geom.getPointN(n + geom.getLength.toInt)
+        } else {
+          geom.getPointN(n-1)
+        }
+      case _ => null
+    }
+  }
+  val ST_X: Geometry => java.lang.Float = {
```
Afterglow375: Had to use java.lang.Float instead of scala's float in order to be able to return null. Not sure how bad that is.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+          null
+        }
+      case _ => null
+    }
+  }
+  val ST_IsClosed: Geometry => Boolean = {
+    case geom: LineString => geom.isClosed
+    case geom: MultiLineString => geom.isClosed
+    case _ => true
```
Afterglow375: Not sure if the default case is correct: http://www.postgis.net/docs/ST_IsClosed.html

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+  val ST_IsCollection: Geometry => Boolean = geom => geom.isInstanceOf[GeometryCollection]
+  val ST_IsEmpty: Geometry => Boolean = geom => geom.isEmpty
+  val ST_IsRing: Geometry => Boolean = {
+    case geom: LineString => geom.isClosed && geom.isSimple
+    case geom: MultiLineString => geom.isClosed && geom.isSimple
+    case geom => geom.isSimple
```
Afterglow375: Not sure if the default case is correct: http://www.postgis.net/docs/ST_IsRing.html

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
```diff
@@ -0,0 +1,87 @@
+package org.apache.spark.sql
+
+import com.vividsolutions.jts.geom._
+
+object SQLSpatialAccessorFunctions {
+  val ST_Boundary: Geometry => Geometry = geom => geom.getBoundary
+  val ST_CoordDim: Geometry => Int = geom => {
+    val coord = geom.getCoordinate
+    if (coord.z.isNaN) 2 else 3
```
elahrvivaz: i think a better way would be `geom.getCoordinateSequence.getDimension`

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
```diff
@@ -0,0 +1,87 @@
...
+    case geom: Polygon => geom.getExteriorRing
+    case _ => null
+  }
+  val ST_GeometryN: (Geometry, Int) => Geometry = (geom, n) => geom.getGeometryN(n)
+  val ST_GeometryType: Geometry => String = geom => geom.getGeometryType
+  val ST_InteriorRingN: (Geometry, Int) => Geometry = (geom, n) => {
+    geom match {
+      case geom: Polygon =>
+        if (0 < n && n <= geom.getNumInteriorRing) {
+          geom.getInteriorRingN(n-1)
```
elahrvivaz: do we want to switch to a 1-based offset here? seems like it would make sense to keep it 0 based like we do for st_geometryN. Looking at the postgis docs, they seem to use 1 based for both - but maybe we should not follow that convention.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+  val ST_IsSimple: Geometry => Boolean = geom => geom.isSimple
+  val ST_IsValid: Geometry => Boolean = geom => geom.isValid
+  val ST_NumGeometries: Geometry => Int = geom => geom.getNumGeometries
+  val ST_NumPoints: Geometry => Int = geom => geom.getNumPoints
+  val ST_PointN: (Geometry, Int) => Point = (geom, n) => {
+    geom match {
+      case geom: LineString =>
+        if (n < 0) {
+          geom.getPointN(n + geom.getLength.toInt)
```
elahrvivaz: would we not just return null here? seems like we should check upper limit too

pom.xml
```diff
@@ -1578,6 +1578,11 @@
             </dependency>
             <dependency>
                 <groupId>org.locationtech.geomesa</groupId>
+                <artifactId>geomesa-accumulo-compute_${scala.binary.version}</artifactId>
```
elahrvivaz: this is another duplicate entry...

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLTypes.scala
```diff
@@ -23,22 +23,33 @@ object SQLTypes {
...
 
-  val PointType             = new PointUDT
-  val MultiPointType        = new MultiPointUDT
-  val LineStringType        = new LineStringUDT
-  val MultiLineStringType   = new MultiLineStringUDT
-  val PolygonType           = new PolygonUDT
-  val MultipolygonType      = new MultiPolygonUDT
-  val GeometryType          = new GeometryUDT
-  // TODO: Implement GeometryCollectionType
+  val GeometryTypeInstance           = new GeometryUDT
```
jnh5y: Any reason for the name change here?  Mainly curious; not sure what I think about it yet.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
Afterglow375: The `getCoordinateSequence` method isn't in the Geometry class? Is it possible to mix 2d/3d sub-geometries in a Geometry? Maybe in a GeometryCollection? I could see this function failing if that's the case. It's also unclear if this function is supposed to handle empty geometries gracefully (e.g. return a 0 maybe): http://www.postgis.net/docs/ST_CoordDim.html

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
Afterglow375: Not sure, I just tried to model the PostGIS functions as closely as possible. Our implementation of `ST_GeometryN` not being 1-based was unintentional on my part. Meant to make it 1-based.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLSpatialAccessorFunctions.scala
Afterglow375: ST_PointN supports negative indexing: http://www.postgis.net/docs/ST_PointN.html

Yeah, I'm not sure what's supposed to happen when you pass in an index out of bounds (throw exception vs return null).

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLTypes.scala
Afterglow375: GeometryType is a constant number corresponding to WKB values used for case statements, whereas GeometryTypeInstance is an instance of a GeometryUDT.

PR #1291, opened 2016-12-19
GEOMESA-627 GeohashUtils.getUniqueGeohashSubstringsInPolygon should return Try[Seq[String]]

aheyne:


geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/geohash/GeohashUtilsTest.scala
```diff
@@ -198,25 +198,25 @@ class GeohashUtilsTest extends Specification with LazyLogging {
...
-        getUniqueGeohashSubstringsInPolygon(point, 5, 2)
-        getUniqueGeohashSubstringsInPolygon(point, 0, 7)
-      }
-      result.isSuccess must beTrue
+      val point = "POINT(0.0 0.0)"
+      val result = (
+           getUniqueGeohashSubstringsInPolygon(point, 0, 3).isFailure
+        || getUniqueGeohashSubstringsInPolygon(point, 3, 2).isFailure
+        || getUniqueGeohashSubstringsInPolygon(point, 5, 2).isFailure
+        || getUniqueGeohashSubstringsInPolygon(point, 0, 7).isFailure )
```
elahrvivaz: [minor] this block could be more clearly/concisely written as
```
forall(Seq((0, 3), (3, 2), ...) { case (offset, length) => getUniqueGeohashSubstringsInPolygon(point, offset, length) must beASuccessfulTry }
```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geohash/GeohashUtils.scala
```diff
@@ -1065,7 +1065,7 @@ object GeohashUtils
       else Seq(ghMBR.toBinaryString.drop(minBits).take(usedBits)))
 
     // detect overflow
-    if (bitPrefixes.overflowed) return Seq()
+    if (bitPrefixes.overflowed) throw new IllegalStateException("Bit prefixes overflowed while calculating unique geohash substrings in polygon.")
```
cne1x: [picky?] 1) "geohash" -> "Geohash" (as per geohash.org); 2) it might be nice to dump the arguments to this call, so that the log contains enough information to reproduce the problem

aheyne: Fixed

PR #1295, opened 2016-12-20
GEOMESA-731 GeoMesaInputFormat - Allow number of mappers per shard to…

aheyne:
… be configurable

* Added two new geomesa-site.xml config options:
  * geomesa.mapreduce.split.count.absolute set the exact number of splits to use regardless of tserver count or accumulo-site.xml setting
  * geomesa.mapreduce.split.count.tablet set the number of splits per tserver to use.
  Number of splits falls back to default accumulo splits computed by the job context.


geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -96,6 +96,31 @@
...
+        <description>Set the absolute number of shards instead of basing it on the number of tablet servers.
+            Setting this value overrides geomesa.mapreduce.split.count.tablet.
+
+            Making this value final without setting a value greater than zero may cause unexpected issues.
+        </description>
+        <final>false</final>
+    </property>
+
+    <property>
+        <name>geomesa.mapreduce.split.count.tablet</name>
```
jahhulbert-ccri: probably should be geomesa.mapreduce.split.count.tserver instead of tablet wondering if the better property would be geomesa.mapreduce.splits.tserver.max and have it mean "max number of splits per tserver"

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -96,6 +96,31 @@
     <!-- /geomesa.batchwriter -->
 
     <property>
+        <name>geomesa.mapreduce.split.count.absolute</name>
```
jahhulbert-ccri: maybe instead of it being .absolute it should be .max since if the value is less than the absolute we wouldn't want to create that many splits...maybe rename it to "geomesa.mapreduce.splits.max"

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/package.scala
```diff
@@ -37,6 +37,11 @@ package object accumulo {
       val SCAN_BATCH_RANGES    = SystemProperty("geomesa.scan.ranges.batch", "20000")
     }
 
+    object AccumuloMapperProperties {
+      val DESIRED_SPLITS_PER_TSERVER = SystemProperty("geomesa.mapreduce.split.count.tablet")
```
jahhulbert-ccri: see note at bottom about renaming these properties to "geomesa.mapreduce.splits.max" and "geomesa.mapreduce.splits.tserver.max" to be more in line with hadoop style naming

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,55 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+    // 2. Get splits from #tserver locations * AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER (geomesa.mapreduce.split.count.tablet)
+    // 3. Get splits from AccumuloInputFormat.getSplits(context)
+    val grpSplitsAbs: Option[Int] =
+      try {
+        val absSplits = AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.get.toInt
+        if (absSplits > 0) Some(absSplits) else None
+      } catch {
+        case e: java.lang.NumberFormatException =>
+          logger.warn(s"Unable to parse geomesa.mapreduce.split.count.absolute = " +
+            s"${AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.get} is not a valid Int.")
```
aheyne: It shouldn't, DESIRED_SPLITS_PER_TSERVER is a Geomesa system property which will default to null if it doesn't have a value. This error message is in an else statement of a null check on DESIRED_SPLITS_PER_TSERVER so at this point we know DESIRED_SPLITS_PER_TSERVER is not null and it has a value to get, it just can't be parsed as an Int.

aheyne: Another question here would be should this accept floats as well?

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,55 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+      if (AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get == null) None
+      else {
+        val numLocations = accumuloSplits.flatMap(_.getLocations).distinct.head.length
+        val splitsPerTServer =
+          try {
+            AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get.toInt
+          } catch {
+            case e: java.lang.NumberFormatException =>
+              logger.warn(s"Unable to parse geomesa.mapreduce.split.count.tablet = " +
+                s"${AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get} is not a valid Int.")
```
aheyne: My previous note actually applies here. Though if you want this fixed here as well just to be clear that's fine.

It shouldn't, DESIRED_SPLITS_PER_TSERVER is a Geomesa system property which will default to null if it doesn't have a value. This error message is in an else statement of a null check on DESIRED_SPLITS_PER_TSERVER so at this point we know DESIRED_SPLITS_PER_TSERVER is not null and it has a value to get, it just can't be parsed as an Int.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -230,3 +275,44 @@ class GeoMesaRecordReader(sft: SimpleFeatureType,
...
+  // if we're running in spark, we need to load the context classpath before anything else,
+  // otherwise we get classloading and serialization issues
+  sys.env.get(GeoMesaAccumuloInputFormat.SYS_PROP_SPARK_LOAD_CP).filter(_.toBoolean).foreach { _ =>
+    GeoMesaAccumuloInputFormat.ensureSparkClasspath()
+  }
+
+  var location: String = null
+  var splits: ArrayBuffer[RangeInputSplit] = ArrayBuffer.empty
+
+  override def getLength =  splits.foldLeft(0L)((l: Long, r: RangeInputSplit) => l + r.getLength)
```
aheyne: Fixed

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/package.scala
aheyne: It groups the tablets by location, so it's per tserver.

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
aheyne: Renamed

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
aheyne: Renamed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,55 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+
+    lazy val grpSplitsPerTServer: Option[Int] =
+      if (AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get == null) None
+      else {
+        val numLocations = accumuloSplits.flatMap(_.getLocations).distinct.head.length
+        val splitsPerTServer =
+          try {
+            AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get.toInt
+          } catch {
+            case e: java.lang.NumberFormatException =>
```
jahhulbert-ccri: wondering what hadoop and accumulo do if you set this incorrectly...might be OK to let the exception percolate up to the user and just fail.

jahhulbert-ccri:
@aheyne  looks good i'll give it a run through on a cloud local and give it a live test. I was also thinking it could be nice on the command line tools to do something hadoop can do for ingest:

```geomesa ingest -u foo -p bar -Dgeomesa.splits.x.y=3  ...```

not needed for this PR but might be nice one day.

File: geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala [hunk 8]
```diff
@@ -164,13 +167,55 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+        if (maxSplits > 0) Some(maxSplits) else None
+      } catch {
+        case e: java.lang.NumberFormatException =>
+          AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.option.foreach(value =>
+            logger.warn(s"Unable to parse geomesa.mapreduce.splits.max = $value is not a valid Int."))
+          None
+      }
+
+    lazy val grpSplitsPerTServer: Option[Int] =
+      if (AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get == null) None
```
elahrvivaz: maybe change this to a map or a match - `.get == null` seems bad

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,55 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+      else {
+        val numLocations = accumuloSplits.flatMap(_.getLocations).distinct.head.length
```
elahrvivaz: is the `distinct` in the right place? maybe we need another flatten in there? otherwise seems like .head.length is not actually checking all the locations

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -230,3 +275,44 @@ class GeoMesaRecordReader(sft: SimpleFeatureType,
...
+ */
+class GroupedSplit extends InputSplit with Writable {
+
+  // if we're running in spark, we need to load the context classpath before anything else,
+  // otherwise we get classloading and serialization issues
+  sys.env.get(GeoMesaAccumuloInputFormat.SYS_PROP_SPARK_LOAD_CP).filter(_.toBoolean).foreach { _ =>
+    GeoMesaAccumuloInputFormat.ensureSparkClasspath()
+  }
+
+  var location: String = null
```
elahrvivaz: would be good to make these private or at least package private

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,55 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
     val accumuloSplits = delegate.getSplits(context)
-    logger.debug(s"Got ${accumuloSplits.length} splits")
-    accumuloSplits
+    // Get the appropriate number of mapper splits using the following priority
+    // 1. Get splits from AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS (geomesa.mapreduce.splits.max)
+    // 2. Get splits from #tserver locations * AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER (geomesa.mapreduce.splits.tserver.max)
+    // 3. Get splits from AccumuloInputFormat.getSplits(context)
+    val grpSplitsMax: Option[Int] =
+      try {
+        val maxSplits = AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.get.toInt
```
elahrvivaz: this should use the `.option` method instead of `.get`:
```
AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.option.flatMap { prop =>
  try { Some(prop.toInt).filter(_ > 0) } catch { ... }
}
```
it might make sense to pull that out into a function and use it for the splits per tserver as well

aheyne: It's not clear to me how that's logically different but it's more concise so I've swapped it over.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: fixed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
jnh5y: I think line 181 should be changed based on Emilio's comment.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: try removed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: fixed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: I think I've fixed this.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
jahhulbert-ccri: I'll vote protected perhaps? or package private.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,51 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
-    accumuloSplits
+    // Get the appropriate number of mapper splits using the following priority
+    // 1. Get splits from AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS (geomesa.mapreduce.splits.max)
+    // 2. Get splits from #tserver locations * AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER (geomesa.mapreduce.splits.tserver.max)
+    // 3. Get splits from AccumuloInputFormat.getSplits(context)
+    val grpSplitsMax: Option[Int] = AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.option.flatMap { prop =>
+      try { Some(prop.toInt).filter(_ > 0) } catch {
+        case e: java.lang.NumberFormatException =>
+          AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.option.foreach(value =>
+            logger.warn(s"Unable to parse geomesa.mapreduce.splits.max = $value is not a valid Int."))
```
elahrvivaz: couldn't you just use `prop` here?

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,51 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+    }
+
+    lazy val grpSplitsPerTServer: Option[Int] =
+      AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.option match {
+        case Some(desiredSplits) =>
+          val numLocations = accumuloSplits.flatMap(_.getLocations).toArray.distinct.length
+          val splitsPerTServer = try { desiredSplits.toInt } catch {
+            case e: java.lang.NumberFormatException =>
+              logger.warn(s"Unable to parse geomesa.mapreduce.splits.tserver.max = " +
+                s"${AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.get} is not a valid Int.")
```
elahrvivaz: can just use `desiredSplits`

aheyne: fixed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: fixed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: fixed (again)

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,51 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+              1 // Identity, don't split locations
+          }
+          if (numLocations > 0 && splitsPerTServer > 0) Some(numLocations * splitsPerTServer) else None
```
elahrvivaz: this variable is actually an alternate max right? could we name it something else? it's not actually per-tserver is it?

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,51 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+        case None => None
+      }
+
+    grpSplitsMax.orElse(grpSplitsPerTServer) match {
+      case Some(size) => logger.debug(s"Using desired splits with result of ${accumuloSplits.length} splits")
+        accumuloSplits.groupBy(_.getLocations()(0)).flatMap{ case (location, splits) =>
+          splits.grouped(size).map{ group =>
```
elahrvivaz: isn't this logic wrong? size here is the absolute number of splits we want - each group size would depend on total splits / absolute number of splits

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -96,6 +96,31 @@
     <!-- /geomesa.batchwriter -->
 
     <property>
+        <name>geomesa.mapreduce.splits.max</name>
+        <value></value>
+        <description>Set the absolute number of shards instead of basing it on the number of tablet servers.
```
elahrvivaz: typo - s/shards/splits/

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/GeoMesaConfigurator.scala
```diff
@@ -118,4 +118,9 @@ object GeoMesaConfigurator {
 
   val serializationString: String = s"$writableSerialization,$simpleFeatureSerialization"
 
+  // set/get the number of mapper splits
+  def setDesiredSplits(conf: Configuration, splits: Int): Unit = conf.set(desiredSplits, splits.toString)
+  def getDesiredSplits(conf: Configuration): Int = conf.get(desiredSplits).toInt
+  def getDesiredSplits(job: Job): Int = getDesiredSplits(job.getConfiguration)
```
elahrvivaz: doesn't look like these methods are used anywhere - we should add some hooks or remove them

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
aheyne: fixed

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,49 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
-    logger.debug(s"Got ${accumuloSplits.length} splits")
-    accumuloSplits
+    // Get the appropriate number of mapper splits using the following priority
+    // 1. Get splits from AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS (geomesa.mapreduce.splits.max)
+    // 2. Get splits from #tserver locations * AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER (geomesa.mapreduce.splits.tserver.max)
+    // 3. Get splits from AccumuloInputFormat.getSplits(context)
+    val grpSplitsMax: Option[Int] = AccumuloMapperProperties.DESIRED_ABSOLUTE_SPLITS.option.flatMap { prop =>
+      try { Some(prop.toInt).filter(_ > 0) } catch {
+        case e: java.lang.NumberFormatException =>
+          prop.foreach( value => logger.warn(s"Unable to parse geomesa.mapreduce.splits.max = $value is not a valid Int."))
```
jahhulbert-ccri: @elahrvivaz how do you feel about throwing an IllegalArg exception since it's truly misconfigured here?

elahrvivaz: seems reasonable

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,53 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+          val splitsPerTServer = try { desiredSplits.toInt } catch {
+            case e: java.lang.NumberFormatException =>
+              logger.warn(s"Unable to parse geomesa.mapreduce.splits.tserver.max = " + s"$desiredSplits is not a valid Int.")
+              1 // Identity, don't split locations
+          }
+          if (numLocations > 0 && splitsPerTServer > 0) Some(numLocations * splitsPerTServer) else None
+        case None => None
+      }
+
+    grpSplitsMax.orElse(grpSplitsPerTServer) match {
```
aheyne: This is the critical part here, I think if fixed the logic but could someone else give this a good look before merging? Thanks

jahhulbert-ccri:
NOTE:  we need to the ticket number on the commit i changed it in github its 731 not 713

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
jahhulbert-ccri: @elahrvivaz so as I read the ticket: https://geomesa.atlassian.net/browse/GEOMESA-731

I think the logic that we need to ensure here is:
1. If the number of accumulo splits is greater than geomesa.mapreduce.splits.max then combine splits
2. if the number of splits per any tserver is geomesa.mapreduce.splits.tserver.max then combine splits

else we leave them alone? Or perhaps it sounds like we always want to attempt to group them by location?

aheyne: So the way I understood the ticket is we want to be able to either set the max number of splits, set the number of splits per tserver, in which case we use the location as a proxy for the number of tservers, or we just leave it alone and use what Accumulo gives us.

elahrvivaz: I think the idea was to specify number of mappers per shard instead of always defaulting to 2. Along with that, it would be good to use something besides shards (which we don't even have any more since they were for the geohash index) to determine the number of tservers.

elahrvivaz: let me review the logic here again

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -96,6 +96,31 @@
...
+    <property>
+        <name>geomesa.mapreduce.splits.tserver.max</name>
+        <value></value>
+        <description>Set the max number of splits per tablet server when configuring a mapper. By default this value is
+            calculated using Accumulo's AbstractInputFormat.getSplits method which creates a split for each range. In
+            some scenarios this may create an undesirably large number of splits.
+
+            This value is overwritten by geomesa.mapreduce.splits.max if it is set.
+
+            Making this value final without setting a value greater than zero may cause unexpected issues.
```
elahrvivaz: yeah, these make sense to me - can you also add them to the documentation here:
http://www.geomesa.org/documentation/user/accumulo/configuration.html

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -164,13 +167,53 @@ class GeoMesaAccumuloInputFormat extends InputFormat[Text, SimpleFeature] with L
...
+      }
+    }
+
+    lazy val grpSplitsPerTServer: Option[Int] =
+      AccumuloMapperProperties.DESIRED_SPLITS_PER_TSERVER.option match {
+        case Some(desiredSplits) =>
+          val numLocations = accumuloSplits.flatMap(_.getLocations).toArray.distinct.length
+          val splitsPerTServer = try { desiredSplits.toInt } catch {
+            case e: java.lang.NumberFormatException =>
+              logger.warn(s"Unable to parse geomesa.mapreduce.splits.tserver.max = " + s"$desiredSplits is not a valid Int.")
```
elahrvivaz: we should probably be consistent and throw an exception here too - also should check for > 0

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
elahrvivaz: logic here looks good to me

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
aheyne: Changed to throw an error. We check for > 0 later (line 197 now).

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
aheyne: Added

aheyne:
Closing and moving to #1384

PR #1312, opened 2017-01-03
GEOMESA-1452 Update and delete functionality for GeoJSON servlet

elahrvivaz:
* Added methods to get by feature ID
* Added caching of JSON extractors in GeoJSON index
* Added caching of service objects in GeoJSON servlet
* Fixed bug in deleting features from GeoJSON index
* Updated docs

### DIFF
--- a/geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/GeoJsonGtIndex.scala
+++ b/geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/GeoJsonGtIndex.scala
@@ -10,19 +10,21 @@ package org.locationtech.geomesa.geojson
 
 import java.io.Closeable
 
+import com.github.benmanes.caffeine.cache.Caffeine
 import com.typesafe.scalalogging.LazyLogging
 import com.vividsolutions.jts.geom.{Geometry, Point}
-import org.apache.commons.io.IOUtils
 import org.geotools.data.{DataStore, FeatureWriter, Query, Transaction}
 import org.geotools.factory.Hints
 import org.geotools.geojson.geom.GeometryJSON
-import org.json4s._
 import org.json4s.native.JsonMethods._
+import org.json4s.{JObject, _}
 import org.locationtech.geomesa.features.kryo.json.JsonPathParser
 import org.locationtech.geomesa.features.kryo.json.JsonPathParser.PathElement
 import org.locationtech.geomesa.geojson.query.GeoJsonQuery
-import org.locationtech.geomesa.utils.collection.SelfClosingIterator
+import org.locationtech.geomesa.utils.cache.CacheKeyGenerator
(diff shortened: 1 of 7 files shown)


geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/GeoJsonGtIndex.scala
```diff
@@ -245,33 +283,39 @@ object GeoJsonGtIndex {
...
+  private def jsonExtractors(schema: SimpleFeatureType): JsonExtractors = {
+    def load(): JsonExtractors = {
+      import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
+
+      val getGeom: (JObject) => Geometry =
+        if (schema.isPoints) {
+          (feature) => getGeometry(feature) match {
+            case p: Point    => p
+            case g: Geometry => g.getCentroid
+            case null        => null
```
jnh5y: Can we get a null Geometry which would throw an NPE on the previous line?

geomesa-geojson/geomesa-geojson-rest/src/main/scala/org/locationtech/geomesa/geojson/servlet/GeoJsonServlet.scala
```diff
@@ -8,7 +8,11 @@
 
 package org.locationtech.geomesa.geojson.servlet
 
+import java.io.Closeable
+import java.util.concurrent.ConcurrentHashMap
+
 import org.json4s.{DefaultFormats, Formats}
+import org.locationtech.geomesa.accumulo.data.AccumuloDataStore
```
jnh5y: Is the ADS required for the GeoJSON support?  Or will this work with a general DataStore?

jnh5y:
Looks reasonable.  Might should get some testing.

geomesa-geojson/geomesa-geojson-rest/src/main/scala/org/locationtech/geomesa/geojson/servlet/GeoJsonServlet.scala
elahrvivaz: It will work with a general data store, but the rest servlet base classes only work with the accumulo data store.

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/GeoJsonGtIndex.scala
elahrvivaz: I think null wouldn't match the `g: Geometry` line:
```
scala> val s: String = null
s: String = null
scala> s match { case string: String => "string"; case null => null }
res0: String = null
```

elahrvivaz:
@jnh5y do you mean more unit tests? I've manually run through integration tests already

PR #1272, opened 2016-12-09
GEOMESA-1263 Geomesa CLI Tools Autocomplete

aheyne:
* Added hidden parameter to help command to generete auto complete script.
* Added calls in bash scripts to generate the actual function and alert the user of it's availability.

Signed-off-by: Austin Heyne <aheyne@ccri.com>

PR #1313, opened 2017-01-04
GEOMESA-432 Improvements to tools scripts 

aheyne:
Stacks on #1272 

* Stacks on top of autocomplete branch allowing which allows tools to run without Accumulo/Hadoop installed.
* Moved errors related to ACCUMULO_HOME/HADOOP_HOME/etc to the runners to only throw errors when a class is not found. This allows jars to simply be provided. Jars can be added to lib/, GEOMESA_EXTRA_CLASSPATH or CLASSPATH vars.


geomesa-tools/conf/geomesa-env.sh
```diff
@@ -28,6 +28,7 @@ function setvar() {
     export $1=$2
   fi
 }
+GEOMESA_TOOLS_VERSION=%%gmtools.version%%
```
aheyne: I'm unsure if this is the best place for the version control. Let me know if you have a better idea.

jahhulbert-ccri: @aheyne i have a better way that we can do to use a filtered source file to insert it.

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/data/AddIndexCommand.scala
```diff
@@ -33,13 +33,37 @@ import scala.util.control.NonFatal
   */
 class AddIndexCommand extends AccumuloDataStoreCommand {
 
+  override val name = "add-index"
```
jahhulbert-ccri: YAY!

geomesa-kafka/geomesa-kafka-tools/geomesa-kafka-08-tools/pom.xml
```diff
@@ -126,4 +126,36 @@
         </profile>
     </profiles>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.5.0</version>
```
jahhulbert-ccri: i'd move the version up to the top level pom for version management in pluginManagement section

geomesa-hbase/geomesa-bigtable-tools/bin/geomesa-bigtable
```diff
@@ -24,15 +24,7 @@ if [[ -n "$HADOOP_HOME" ]]; then
...
-if [[ -z "$HADOOP_HOME" ]]; then
-  message="Warning: you have not set HADOOP_HOME as an environment variable."
-  message="${message}\nGeoMesa tools will not run without the appropriate Hadoop jars in the tools classpath."
-  message="${message}\nPlease ensure that those jars are present in the classpath by running 'geomesa-bigtable classpath' ."
-  message="${message}\nTo take corrective action, please place the necessary jar files in the lib directory of geomesa-tools."
-  echo >&2 ""
-  echo -e >&2 "$message"
-  echo >&2 ""
-elif [[ -z "$HADOOP_CONF_DIR" ]]; then
+if [[ -z "$HADOOP_CONF_DIR" ]]; then
```
jahhulbert-ccri: technically I think this won't check if its empty, right? It may be useful to use the ```[[ -z "{VAR+x}" ]]``` construct in bash like this:

```
ABC="" bash -c 'if [[ -z "$ABC" ]]; then echo "set"; else echo "empty"; fi '
set
ABC="" bash -c 'if [[ -z "${ABC+x}" ]]; then echo "set"; else echo "empty"; fi '
empty
```

we probably have a lot of plain ```-z``` when we actually want to check two things (is it set and is it non empty) we may want to clean them up at some point.

aheyne: Yeah, there are probably a few cases where care if it's empty vs set as we could give some more helpful messages. Who ever does this will have to be careful with the logic though, I find it can be very confusing. I think for the HADOOP_CONF_DIR it should be `[[ -z "${HADOOP_CONF_DIR:+x}" ]]`
e.g.
`$ HADOOP_CONF_DIR=`
`$ if [[ -z "${HADOOP_CONF_DIR:+x}" ]]; then echo "We need to Set HADOOP_CONF_DIR"; fi`
`We need to Set HADOOP_CONF_DIR`
`$ unset HADOOP_CONF_DIR`
`$ if [[ -z "${HADOOP_CONF_DIR:+x}" ]]; then echo "We need to Set HADOOP_CONF_DIR"; fi`
`We need to Set HADOOP_CONF_DIR`
`$ HADOOP_CONF_DIR=/abc/`
`$ if [[ -z "${HADOOP_CONF_DIR:+x}" ]]; then echo "We need to Set HADOOP_CONF_DIR"; fi`
`$`

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
```diff
@@ -33,17 +33,8 @@ if [[ -n "$HADOOP_HOME" ]]; then
...
-  message="Warning: you have not set ACCUMULO_HOME and/or HADOOP_HOME as environment variables."
-  message="${message}\nGeoMesa tools will not run without the appropriate Accumulo and Hadoop jars in the tools classpath."
-  message="${message}\nPlease ensure that those jars are present in the classpath by running 'geomesa classpath' ."
-  message="${message}\nTo take corrective action, please place the necessary jar files in the lib directory of geomesa-tools."
-  echo >&2 ""
-  echo -e >&2 "$message"
-  echo >&2 ""
-elif [[ (-z "$ACCUMULO_CONF_DIR") || (-z "$HADOOP_CONF_DIR") ]]; then
-  message="Warning: you have not set ACCUMULO_CONF_DIR and/or HADOOP_CONF_DIR as environment variables."
+if [[ (-z "$ACCUMULO_CONF_DIR") || (-z "$HADOOP_CONF_DIR") ]]; then
```
elahrvivaz: does it make sense to skip the warning about accumulo home but keep the one about the conf dir? if home is not set, conf dir won't be either... (unless they have explicitly set only that one var)

geomesa-hbase/geomesa-bigtable-tools/src/main/scala/org/locationtech/geomesa/bigtable/tools/BigtableRunner.scala
```diff
@@ -44,4 +44,13 @@ object BigtableRunner extends Runner {
     new HBaseStatsTopKCommand,
     new HBaseStatsHistogramCommand
   )
+
+  override def environmentErrorInfo(): Unit = {
+    if (sys.env.get("HADOOP_HOME").isEmpty) {
+      println("Warning: you have not set HADOOP_HOME as an environment variable." +
```
elahrvivaz: instead of printlns here we should stick with the appropriate loggers (applies to all Runners)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
```diff
@@ -20,11 +20,16 @@ import scala.util.control.NonFatal
 trait Runner {
 
   def name: String
+  def environmentErrorInfo(): Unit = {}
```
elahrvivaz: we could make this return an Option[String] and control the output in the main Runner (with a logger)

geomesa-tools/bin/common-functions.sh
```diff
@@ -101,19 +101,59 @@ function containsElement() {
...
+      cat ${GEOMESA_HOME}/conf/autocomplete.sh >> ${compFile}
+      echo "Autocomplete function available, to use now run:"
+      echo ". ${compFile}"
+    fi
   fi
 }
 
-echo >&2 "Using %%gmtools.dist.name%%_HOME = $%%gmtools.dist.name%%_HOME"
+# Version check
+GEOMESA_TOOLS_VERSION=%%gmtools.version%%
```
elahrvivaz: are these version checks necessary? seems unlikely someone would mix-and-match...

geomesa-tools/bin/common-functions.sh
```diff
@@ -101,19 +101,59 @@ function containsElement() {
...
+    GEOMESA_HOME_TOOLS_VERSION="`cat ${%%gmtools.dist.name%%_HOME}/conf/geomesa-env.sh | grep GEOMESA_TOOLS_VERSION | sed 's/GEOMESA_TOOLS_VERSION=//'`"
+  else GEOMESA_HOME_TOOLS_VERSION=""; fi # legacy support
+  if [[ $GEOMESA_TOOLS_VERSION != $GEOMESA_HOME_TOOLS_VERSION ]]; then
+    echo "Warning: Using %%gmtools.dist.name%%_HOME = ${%%gmtools.dist.name%%_HOME} does not match Geomesa tools version"
+  fi
+fi
+if [[ -n "$GEOMESA_CLASSPATH" ]]; then
+  # Search the classpath for conflicting versions
+  cpArr=(`echo $GEOMESA_CLASSPATH | sed 's/:/ /g'`)
+  for i in "${cpArr[@]}"; do
```
elahrvivaz: is this in case you have 2 geomesa tools on the classpath? again, not sure all this is needed...

geomesa-accumulo/geomesa-accumulo-tools/pom.xml
```diff
@@ -135,6 +135,34 @@
         </resources>
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.5.0</version>
```
aheyne: I'll fix this when I rebase on #1272

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
aheyne: Could this happen if the user provides the accumulo jars on the classpath and then provides a conf dir for it?

geomesa-hbase/geomesa-bigtable-tools/src/main/scala/org/locationtech/geomesa/bigtable/tools/BigtableRunner.scala
aheyne: Fixed

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
aheyne: Changed

geomesa-tools/bin/common-functions.sh
aheyne: No, this has to do with allowing them to provide jars and config on the classpath without setting ACCUMULO_HOME, HADOOP_HOME and GEOMESA_CONF_DIR. We need to check that the version of the files in the GEOMESA_CONF_DIR are correct.

elahrvivaz: could you explain further?

geomesa-tools/conf/geomesa-env.sh
aheyne: I went back and looked at this and the problem I was trying to solve was if a user has multiple versions of the tools we should be able to detect if they forgot to reset the GEOMESA_CONF_DIR before we run anything.

geomesa-tools/bin/common-functions.sh
aheyne: We're trying to allow users to just provide resources on the classpath without spitting a bunch of warnings at them. We can do that with this pr but that introduces another problem. If the user is using a classpath configuration with either the GEOMESA_CONF_DIR on the classpath, the GEOMESA_HOME dir on the classpath or the geomesa-env.sh on the classpath and then downloads a new version of the tools and trys to run it without resetting the classpath things could get messy. What this does is makes sure that whatever resources are provided on the classpath are the version they should be.

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
jahhulbert-ccri: if you provide -z and -i you never need the conf dir anyways for accumulo. I think we may have a ticket to not warn if the jars are available on on the classpath but that's more complicated.

jahhulbert-ccri:
@aheyne looks like there are a couple of comments from @elahrvivaz that might need to be addressed unless he's good to go

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
elahrvivaz: @aheyne i would remove these warning for the conf dirs - they're going to always be printed out as normally you wouldn't even set that you'd just have ACCUMULO_HOME

geomesa-hbase/geomesa-bigtable-tools/bin/geomesa-bigtable
elahrvivaz: i'd remove these warnings too

geomesa-hbase/geomesa-hbase-tools/bin/geomesa-hbase
```diff
@@ -34,15 +34,7 @@ if [[ -n "$HADOOP_HOME" ]]; then
...
-  message="Warning: you have not set HBASE_HOME and/or HADOOP_HOME as environment variables."
-  message="${message}\nGeoMesa tools will not run without the appropriate HBase and Hadoop jars in the tools classpath."
-  message="${message}\nPlease ensure that those jars are present in the classpath by running 'geomesa-hbase classpath' ."
-  message="${message}\nTo take corrective action, please place the necessary jar files in the lib directory of geomesa-tools."
-  echo >&2 ""
-  echo -e >&2 "$message"
-  echo >&2 ""
-elif [[ (-z "$HBASE_CONF_DIR") || (-z "$HADOOP_CONF_DIR") ]]; then
+if [[ (-z "$HBASE_CONF_DIR") || (-z "$HADOOP_CONF_DIR") ]]; then
   message="Warning: you have not set HBASE_CONF_DIR and/or HADOOP_CONF_DIR as environment variables."
```
elahrvivaz: ditto

geomesa-tools/bin/common-functions.sh
elahrvivaz: i think i'd revert this file - it seems like a lot of bash-fu that is likely to be brittle, and I'm not convinced it's solving a real problem... @jahhulbert-ccri what do you think?

elahrvivaz: maybe instead we can just verify up front that the script they're running is in GEOMESA_HOME/bin (it it's set)

geomesa-tools/bin/common-functions.sh
jahhulbert-ccri: we do have situations where we have old tools installed like 1.2.6 querying 1.2.7 etc

geomesa-tools/bin/common-functions.sh
jahhulbert-ccri: yeah. I'd be ok punting this out of bash. I don't like the fact that we do so much in bash anyways with our tools. I personally would like to move to doing a bootstrap class into scala or java land like nifi and accumulo do where we can actually have good logic that can be tested, etc.

jahhulbert-ccri: no need to add complexity if we're not convinced its the right way long term to go

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
aheyne: I've just removed the Accumulo conf dir warning. The hadoop one should remain, correct?

jahhulbert-ccri:
@aheyne one conflict in the pom and then we can merge

aheyne:
@jahhulbert-ccri merged master

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
```diff
@@ -20,11 +20,16 @@ import scala.util.control.NonFatal
...
 
   def name: String
+  def environmentErrorInfo(): Option[String] = { None }
 
   def main(args: Array[String]): Unit = {
     try {
       parseCommand(args).execute()
     } catch {
+      case e @ (_: ClassNotFoundException | _: NoClassDefFoundError) =>
+        Command.user.error(s"Warning: Missing dependency for command execution: ${e.getMessage}")
```
elahrvivaz: actually, can we also log this to the other log with the full stack trace?

PR #1321, opened 2017-01-09
GEOMESA-1623 Fixing check for BIN export in tools

elahrvivaz:

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportParams.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportParams.scala
@@ -20,7 +20,7 @@ trait FileExportParams {
   @Parameter(names = Array("--gzip"), description = "Level of gzip compression to apply to output, from 1-9")
   var gzip: Integer = null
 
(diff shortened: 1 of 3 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportParams.scala
```diff
@@ -20,7 +20,7 @@ trait FileExportParams {
   @Parameter(names = Array("--gzip"), description = "Level of gzip compression to apply to output, from 1-9")
   var gzip: Integer = null
 
-  @Parameter(names = Array("-F", "--output-Format"), description = "File format of output files (csv|tsv|gml|json|shp|avro)")
+  @Parameter(names = Array("-F", "--output-format"), description = "File format of output files (csv|tsv|gml|json|shp|avro)")
```
aheyne: We're going to have to be careful with changes like this in the future. The autocomplete won't update so if they already have an autocomplete for that command installed and we change the param, tab complete will provide and invalid name.

elahrvivaz: good point - not sure how this ended up capitalized in the first place - all the long-form arguments should be lower case

PR #1323, opened 2017-01-10
GEOMESA-1625 Storm QS Updates

aheyne:
* Updated Docs for Storm Quickstart
* Fixed supported Accumulo versions
* Added Accumulo 1.8 profile to geomesa-accumulo-datastore.pom to shade curator to prevent conflicts with Accumulo 1.8 and Storm

### DIFF
--- a/docs/common.py
+++ b/docs/common.py
@@ -108,7 +108,7 @@ rst_epilog = """
 
 .. |geotools_version| replace:: 15.1
 
-.. |accumulo_version| replace:: 1.6 or 1.7
+.. |accumulo_version| replace:: 1.7 or 1.8
 
 .. |hadoop_version| replace:: 2.2 or better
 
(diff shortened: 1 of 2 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/pom.xml
```diff
@@ -146,6 +146,60 @@
         </dependency>
     </dependencies>
 
+    <profiles>
+        <profile>
+            <id>accumulo-1.8</id>
+            <properties>
+                <accumulo.version>1.8.0</accumulo.version>
+                <thrift.version>0.9.3</thrift.version>
+            </properties>
```
elahrvivaz: are these props not picked up by the profile in the main pom? having the versions in two places isn't ideal.

elahrvivaz:
Would it be more appropriate to do the shading in whatever the storm client jar is? (e.g. in the tutorial, presumably) Seems like we wouldn't necessarily want to shade it here.

elahrvivaz:
maybe we can just add a note to the documentation about the potential conflict

aheyne:
Managing the curator version in the geomesa-tutorials solves this problem for the storm quickstart.

aheyne:
Could someone flag this 1.3.0 please.

PR #1331, opened 2017-01-13
GEOMESA-1636 generate separate module for accumulo spark runtime

tkunicki:
Create accumulo spark assembly as `geomesa-accumulo-spark-runtime-<version>`
- do we need to update docs?
- do we want to shade in the geotools and converter rdd providers?


geomesa-accumulo/geomesa-accumulo-spark-runtime/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+                                    <shadedPattern>org.locationtech.geomesa.shaded.com.google.common</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>org.apache.commons.codec</pattern>
+                                    <shadedPattern>org.locationtech.geomesa.shaded.org.apache.commons.codec</shadedPattern>
+                                </relocation>
+                                <!-- shade thrift because spark depends on 0.9.2 and accumulo 1.8 depends on 0.9.3 -->
+                                <relocation>
+                                    <pattern>org.apache.thrift</pattern>
+                                    <shadedPattern>org.lt.geomesa.shaded.o.a.thrift</shadedPattern>
```
elahrvivaz: we should try to keep a common naming convention for shaded packages

jnh5y:
@tkunicki Yes!  We need to add some documentation to explain that the accumulo-spark-runtime is the One Jar You Need!  @mdzimmerman can you help identify where that should go?

For my money, I'd argue that the GTRDD Provider should be separate (you have to grab it AND your GT datastore(s), so I'm not worried about getting it down to the One Jar).

I am not sure about the converter RDD Provider.  Thoughts?

geomesa-accumulo/geomesa-accumulo-spark-runtime/pom.xml
tkunicki: more than happy to address, thoughts on convention or existing package to adopt convention from?

elahrvivaz: i would just do it the same as the ones right above, thanks

aheyne:
Do we need license headers in these poms?

jnh5y:
@aheyne Nope, no need to put license headers in pom.xml's.

aheyne:
Cool, just wondering since a bunch of ours have it.

PR #1332, opened 2017-01-15
GEOMESA-1637 Correct Extra Classpaths

jahhulbert-ccri:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
+++ b/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
@@ -57,7 +57,7 @@ fi
 GEOMESA_CP="${GEOMESA_CONF_DIR}:$(findJars $GEOMESA_LIB):${ACCUMULO_CONF_DIR}:${HADOOP_CONF_DIR}"
 
 # Add user defined directories to the classpath
-if [[ -z "$GEOMESA_EXTRA_CLASSPATHS"  ]]; then
+if [[ ! -z ${GEOMESA_EXTRA_CLASSPATHS+x}  ]]; then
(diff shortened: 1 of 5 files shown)


geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
```diff
@@ -57,7 +57,7 @@ fi
 GEOMESA_CP="${GEOMESA_CONF_DIR}:$(findJars $GEOMESA_LIB):${ACCUMULO_CONF_DIR}:${HADOOP_CONF_DIR}"
 
 # Add user defined directories to the classpath
-if [[ -z "$GEOMESA_EXTRA_CLASSPATHS"  ]]; then
+if [[ ! -z ${GEOMESA_EXTRA_CLASSPATHS+x}  ]]; then
```
jnh5y: @aheyne does this make sense to you?

aheyne: I guess it works but I don't really understand why you wouldn't just do 
`if [[ -n "${GEOMESA_EXTRA_CLASSPATHS}" ]]; then`
which seems more readable.

jahhulbert-ccri: yup. i can make it that. I just quickly added the ! to fix the logic error

jahhulbert-ccri:
found more places that it was wrong and standardized them to all look the same

jahhulbert-ccri:
if someone else wants to try here's what I'd do...

```
vim reference.conf   # put a converter named remote-foobar in this
jar cvf /tmp/foobar.jar reference.conf

hadoop fs -put somefile.csv /user/you/somefile.csv

GEOMESA_EXTRA_CLASSPATHS=/tmp/foobar.jar geomesa ingest .... hdfs://namenode:9000/user/you/somefile.csv

```

PR #1333, opened 2017-01-15
GEOMESA-1638 Fix tools mock ingest

jahhulbert-ccri:


geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloDataStoreCommand.scala
```diff
@@ -18,14 +18,25 @@ trait AccumuloDataStoreCommand extends DataStoreCommand[AccumuloDataStore] {
...
+      AccumuloDataStoreParams.passwordParam.getName   -> params.password,
+      AccumuloDataStoreParams.tableNameParam.getName  -> params.catalog,
+      AccumuloDataStoreParams.visibilityParam.getName -> params.visibilities,
+      AccumuloDataStoreParams.authsParam.getName      -> params.auths,
+      AccumuloDataStoreParams.mockParam.getName       -> params.mock.toString
+    ).filter(_._2 != null)
+
+    // AccumuloDataStoreFactory requires instance ID to be set but it should not be required if mock is set...so set a
+    // fake one but be careful NOT to add a mock zoo since other code will then think it has a zookeeper but doesn't
+    if (parsedParams.get(AccumuloDataStoreParams.mockParam.getName).exists(_.toBoolean)) {
```
jnh5y: Would it be better to update the canProcess function?

jahhulbert-ccri: for AccumuloDataStore you mean?

jahhulbert-ccri: Though it may be good to remove the user and password modification in case the user expected something regarding that? I seem to remember though that the standard user/pass combination doesn't work correctly with the mock

elahrvivaz: i've noticed that the user/pass has to match if you run mock more than once - possibly tied to the instance, not sure. also not sure where the mock instance is caching the credentials...

elahrvivaz: i think it might be better to use the instance/user/pwd if they are provided, but if not fall back to the defaults

elahrvivaz:
currently it still requires you to pass in `-user` and `-password` - maybe we should just override the instance param

elahrvivaz:
since mock ingest does work if you specify an instance id, going to push this to 1.3.1

PR #1339, opened 2017-01-19
GEOMESA-1646 Attribute index with secondary z index

elahrvivaz:
* AccumuloWritableIndex renamed to AccumuloFeatureIndex
* Current indices updated to use write and query logic from index-api module
* Older index versions moved to 'legacy' package
* Added back-compatibility test for 1.2.7.3


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
```diff
@@ -72,18 +68,17 @@ class AccumuloGeoMesaStats(val ds: AccumuloDataStore, statsTable: String, val ge
...
-    lazy val hasDupes = sft.nonPoints && {
-      val indices = AccumuloFeatureIndex.indices(sft, IndexMode.Read)
-      // TODO check for multivalued attribute indices
-      Seq(Z2IndexV1, Z3IndexV2).exists(indices.contains)
+    lazy val query = {
+      // restrict fields coming back so that we push as little data as possible
+      val props = Array(Option(sft.getGeomField).getOrElse(sft.getDescriptor(0).getLocalName))
+      new Query(sft.getTypeName, filter, props)
     }
+    lazy val hasDupes = ds.getQueryPlan(query).exists(_.hasDuplicates)
```
jnh5y: Do these variables need to be lazy?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
```diff
@@ -10,58 +10,60 @@ package org.locationtech.geomesa.accumulo.index
...
+import scala.util.Try
+import scala.util.control.NonFatal
+
 object AccumuloFeatureIndex extends AccumuloIndexManagerType with LazyLogging {
 
-  private val SpatialIndices        = Seq(Z2Index, XZ2Index, Z2IndexV1)
+  val FullColumnFamily      = new Text("F")
+  val IndexColumnFamily     = new Text("I")
+  val BinColumnFamily       = new Text("B")
+  val AttributeColumnFamily = new Text("A")
```
jnh5y: We just need a "U" and "L" to spell fibula...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
```diff
@@ -10,58 +10,60 @@ package org.locationtech.geomesa.accumulo.index
...
+
+  val EmptyColumnQualifier  = new Text()
+
+  private val SpatialIndices        = Seq(Z2Index, XZ2Index, Z2IndexV2, Z2IndexV1)
   private val SpatioTemporalIndices = Seq(Z3Index, XZ3Index, Z3IndexV2, Z3IndexV1)
-  private val AttributeIndices      = Seq(AttributeIndex, AttributeIndexV2)
+  private val AttributeIndices      = Seq(AttributeIndex, AttributeIndexV3, AttributeIndexV2)
```
jnh5y: Can I get a short description of Attr v2 vs v3?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
```diff
@@ -90,18 +92,22 @@ object AccumuloFeatureIndex extends AccumuloIndexManagerType with LazyLogging {
 
   /**
     * Maps a simple feature type to a set of default indices based on
-    * attributes and when it was created (schema version)
+    * attributes and when it was created (schema version).
+    *
+    * Note that schema version has been deprecated and this method should not be called
+    * except when transitioning from schema version to per-index versions.
```
jnh5y: Can you say a little more about this?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
```diff
@@ -117,14 +123,60 @@ object AccumuloFeatureIndex extends AccumuloIndexManagerType with LazyLogging {
...
+
+    super.configure(sft, ds)
 
+    val table = getTableName(sft.getTypeName, ds)
+
+    // create table if needed
+    AccumuloVersion.ensureTableExists(ds.connector, table)
+
+    // create splits - drop first split, otherwise we get an empty tablet
+    val splitsToAdd = getSplits(sft).map(new Text(_)).toSet -- ds.tableOps.listSplits(table).toSet
```
jnh5y: Could you just call .tail on the output of the .map?  (Or is that naive?)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
```diff
@@ -151,12 +203,17 @@ trait AccumuloWritableIndex extends AccumuloFeatureIndexType {
...
-    */
-  def serializedWithId: Boolean
+  // back compatibility check for old metadata keys
+  abstract override def getTableName(typeName: String, ds: AccumuloDataStore): String = {
+    lazy val oldKey = this match {
+      case i if i.name == RecordIndex.name    => "tables.record.name"
+      case i if i.name == AttributeIndex.name => "tables.idx.attr.name"
+      case i => s"tables.${i.name}.name"
+    }
+    Try(super.getTableName(typeName, ds)).recoverWith {
```
jnh5y: Idly I wonder if it'd be better to move 'oldKey' into the recoverWith and make it a val?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
```diff
@@ -0,0 +1,137 @@
...
+trait AccumuloIndexAdapter extends IndexAdapter[AccumuloDataStore, AccumuloFeature, Mutation, Range] {
+
+  this: AccumuloFeatureIndex =>
+
+  import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
+
+  override protected def createInsert(row: Array[Byte], feature: AccumuloFeature): Mutation = {
+    val mutation = new Mutation(row)
+    feature.fullValues.foreach(v => mutation.put(v.cf, v.cq, v.vis, v.value))
+    if (hasPrecomputedBins) {
```
jnh5y: Crazy question....  Rather than checking 'hasPrecomputedBins' per record, could this be refactored to check once per instantiation?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
```diff
@@ -0,0 +1,137 @@
...
+      feature.binValues.foreach(v => mutation.put(v.cf, v.cq, v.vis, v.value))
+    }
+    mutation
+  }
+
+  override protected def createDelete(row: Array[Byte], feature: AccumuloFeature): Mutation = {
+    val mutation = new Mutation(row)
+    feature.fullValues.foreach(v => mutation.putDelete(v.cf, v.cq, v.vis))
+    if (hasPrecomputedBins) {
```
jnh5y: And if one did something like that, would it save much?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
```diff
@@ -0,0 +1,137 @@
...
+                                  ds: AccumuloDataStore,
+                                  filter: FilterStrategy[AccumuloDataStore, AccumuloFeature, Mutation],
+                                  hints: Hints,
+                                  ranges: Seq[Range],
+                                  ecql: Option[Filter]): QueryPlan[AccumuloDataStore, AccumuloFeature, Mutation] = {
+    if (ranges.isEmpty) { EmptyPlan(filter) } else {
+      val table = getTableName(sft.getTypeName, ds)
+      val numThreads = queryThreads(ds)
+      val dedupe = hasDuplicates(sft, filter.primary)
+      val ScanConfig(iters, cf, eToF, reduce) = scanConfig(sft, hints, ecql, dedupe)
```
jnh5y: eToF is Entry to Feature?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
elahrvivaz: they normally don't get evaluated - the `hasDupes` only needs to be checked if `exact` is true, and the `query` only needs to be checked in `hasDupes` is true. I could move query back down into the if statement - not sure why I moved it up tbh...

elahrvivaz: oh wait, query is used to evaluate hasDupes, and then again inside the if block - so yeah, i guess lazy is good here

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: there are comments on the classes themselves - v3 has the id encoded in the row key, v2 has the date and fid included in the row key, and the main one now has the z value in the row key. v1 was the old original index that didn't have dates - i think we dropped that before we went to the versioned classes approach.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: I just wanted to make sure no one calls this method accidentally - it's only for a particular use case (transitioning an old schema version).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: i should remove that comment - it's not relevant here any more. the `getSplits` method drops the first element already. I believe the implementations do call `.tail` or `.drop(1)` or something

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: i can - thought it was easier to read this way

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
elahrvivaz: we could - since it is just a boolean check i figured it wasn't a big deal. what do you think?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
elahrvivaz: yeah, i thought it would probably be a wash with the overhead of executing a functional val or whatever.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
elahrvivaz: yeah, confusing?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: http://www.fiba.com/

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
jnh5y: Which one are you referring to as the 'main' one?  That'd be the new attr + z3?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
jnh5y: This works for me then.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
jnh5y: It's probably not a big win.  *shrugs*  Maybe not worth it?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
jnh5y: Since I was right, it wasn't confusing to me;).  If the class 'ScanConfig' has some docs or nicely readable bits, I'd say it is no big deal.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
```diff
@@ -0,0 +1,355 @@
...
+    val values = coverage match {
+      case IndexCoverage.FULL => feature.fullValues
+      case IndexCoverage.JOIN => feature.indexValues
+    }
+    values.foreach(v => mutation.putDelete(v.cf, v.cq, v.vis))
+    mutation
+  }
+
+  override protected def createInsert(row: Array[Byte], feature: AccumuloFeature): Mutation =
+    throw new NotImplementedError("Should be using enhanced version with IndexCoverage")
```
jnh5y: Ok, so what does this mean?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
elahrvivaz: it doesn't really have docs, but it does at least spell out the full names :)

elahrvivaz: https://github.com/elahrvivaz/geomesa/blob/2111f2a881ca923d3c097bc8a7cc6607459828f7/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala#L133

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
elahrvivaz: that it should be using the method 20 lines up. this method is part of the generic `GeoMesaFeatureIndex` trait, but it can't account for join/full indices. we've overridden the `writer` and `remover` methods so that it doesn't get called, but it's still part of the API so we can't remove it - this is just a sanity check.

elahrvivaz: i'll add some more comments

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: yeah, the new `AttributeIndex` class

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/AttributeIndexFilteringIteratorTest.scala
```diff
@@ -65,7 +64,7 @@ class AttributeIndexFilteringIteratorTest extends Specification with TestWithDat
       // % should return all features
       val wildCardQuery = new Query(sftName, ff.like(ff.property("name"),"%"))
       checkStrategies(wildCardQuery, AttributeIndex)
-      SelfClosingIterator(fs.getFeatures()) must haveLength(16)
+      SelfClosingIterator(fs.getFeatures(wildCardQuery)) must haveLength(16)
```
jnh5y: Is the wildCardQuery restrictive?  It appears that the previous line didn't filter at all.  If all features match the wildCard, it might be worth adding another test case.

elahrvivaz: i think this was a typo in the original test case - the wildcard filter should return all features, but it wasn't actually tested. `fs.getFeatures()` won't even hit the attribute index...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/AttributeIndex.scala
```diff
@@ -158,75 +109,165 @@ trait AttributeIndex[DS <: GeoMesaDataStore[DS, F, W], F <: WrappedFeature, W, R
...
+
+object AttributeIndex {
+
+  val NullByte: Byte = 0
+  val NullByteArray  = Array(NullByte)
+
+  val typeRegistry   = LexiTypeEncoders.LEXI_TYPES
+
+  // store 2 bytes for the index of the attribute in the sft - this allows up to 32k attributes in the sft.
+  def indexToBytes(i: Int): Array[Byte] = Array((i << 8).asInstanceOf[Byte], i.asInstanceOf[Byte])
```
jnh5y: If there are more than 32k attributes, does this 'bin' things by the first two bytes?

elahrvivaz: no, you'd just get overlapping index values. think we need to worry about a schema with 32k attributes?? (this is fields in the sft, not number of features)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
jnh5y: Are all Attribute indices going to be attr + Z3 going forward or do you have to specify that option?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloIndexAdapter.scala
jnh5y: WorksForMe.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/AttributeIndex.scala
jnh5y: Oh!  32k fields names... not values.  

And no.  Someone would have to make a 'union' sft of all the datasets.  *shakes head*

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloFeatureIndex.scala
elahrvivaz: new schemas will get the latest attribute index, older schemas will continue to use the older ones. note that it will use whatever SFC is compatible with the schema - Z3, XZ3, Z2 or XZ2.

PR #1348, opened 2017-01-27
1.2 - GEOMESA-1653 Not always returning default geometry if alternate is re…

elahrvivaz:
…quested

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
@@ -57,8 +57,6 @@ case class QueryPlanner(sft: SimpleFeatureType, ds: AccumuloDataStore) extends M
 
   import org.locationtech.geomesa.accumulo.index.QueryPlanner._
 
-  private lazy val serializationType = ds.getFeatureEncoding(sft)
-
   /**
    * Plan the query, but don't execute it - used for m/r jobs and explain query
    */
@@ -215,7 +213,7 @@ object QueryPlanner extends LazyLogging {
   }
 
   def setPerThreadQueryHints(hints: Map[AnyRef, AnyRef]): Unit = threadedHints.put(hints)
-  def clearPerThreadQueryHints() = threadedHints.clear()
+  def clearPerThreadQueryHints(): Unit = threadedHints.clear()
 
   /**
    * Configure the query - set hints, transforms, etc.
@@ -330,7 +328,7 @@ object QueryPlanner extends LazyLogging {
     * @param sft simple feature type
     * @return
     */
-  def setQueryTransforms(query: Query, sft: SimpleFeatureType) = {
+  def setQueryTransforms(query: Query, sft: SimpleFeatureType): Unit = {
     import scala.collection.JavaConversions._
     val properties = query.getPropertyNames
     query.setProperties(Query.ALL_PROPERTIES)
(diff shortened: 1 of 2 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
```diff
@@ -215,7 +213,7 @@ object QueryPlanner extends LazyLogging {
   }
 
   def setPerThreadQueryHints(hints: Map[AnyRef, AnyRef]): Unit = threadedHints.put(hints)
-  def clearPerThreadQueryHints() = threadedHints.clear()
+  def clearPerThreadQueryHints(): Unit = threadedHints.clear()
```
jnh5y: Is this type annotation necessary?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
```diff
@@ -330,7 +328,7 @@ object QueryPlanner extends LazyLogging {
     * @param sft simple feature type
     * @return
     */
-  def setQueryTransforms(query: Query, sft: SimpleFeatureType) = {
+  def setQueryTransforms(query: Query, sft: SimpleFeatureType): Unit = {
```
jnh5y: Ditto

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/QueryPlanner.scala
elahrvivaz: haha, intellij has started flagging public methods without explicit return types, so I've started filling them in :P

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTransformsTest.scala
```diff
@@ -240,5 +240,47 @@ class AccumuloDataStoreTransformsTest extends Specification with TestWithMultipl
...
+      "if default geometry is specified" >> {
+        val query = new Query(sftName, Filter.INCLUDE, Array("name", "geom"))
+        val features = SelfClosingIterator(ds.getFeatureSource(sftName).getFeatures(query).features).toSeq
+        features must haveSize(3)
+        forall(features)(_.getAttributeCount mustEqual 2)
+        features.map(_.getAttribute("geom")) must containTheSameElementsAs(reference.map(_.getAttribute("geom")))
+        features.map(_.getAttribute("name")) must containTheSameElementsAs(reference.map(_.getAttribute("name")))
+      }
+
+      "if alternate geometry is specified" >> {
```
jnh5y: Should we have a test for when a geometry attribute is renamed via a transform?

elahrvivaz: i can add one

elahrvivaz: i added a test, but it fails - there isn't really an easy way to determine what is going to be returned from a transform expression. I tried using the `Definition.binding`, but that doesn't get populated. I think for now I'll leave it as is...

jnh5y: Ah, makes sense; is it worth entering a ticket and/or adding the failing test with a pendingUntilFixed?

elahrvivaz: dunno, think we'll ever fix it?

jnh5y: What happens if you use the transform?  Does it work or throw exceptions?

elahrvivaz: it works, but returns the default geometry. i'll add a failing unit test in the 1.3 branch

PR #1353, opened 2017-02-02
GEOMESA-1662 Clarify description of add-attribute-index command

mdzimmerman:
Make it clear that it (effectively) creates separate indexes for each
attribute listed, not a compound index of all attributes listed.


elahrvivaz:
looks good, merging

PR #1357, opened 2017-02-07
GEOMESA-1655 Fix "geomesa explain" command line tool output

mdzimmerman:


mdzimmerman:
This essentially sidesteps the question of what the logging level for the explainer is set to in ``log4j.properties`` and assumes that if somebody explicitly asks for a query plan by invoking ``geomesa explain``, we can go ahead and log it at ``info`` (and print it out to the console).

elahrvivaz:
great, merging

PR #1359, opened 2017-02-08
GEOMESA-1669 Disabling travis CQ check

elahrvivaz:


mdzimmerman:
How much time does this save on average?

elahrvivaz:
not entirely sure - but i've noticed that a lot of builds will complete the maven part successfully in ~45 minutes, then timeout - so probably on the order of 5-10 minutes. Occasionally it seems to run much faster for some reason.

PR #1362, opened 2017-02-09
GEOMESA-1672 Bug in setup namespace

aheyne:
* Fixed bug in setup-namespace that didn't terminate when provided with insufficient parameters.


geomesa-accumulo/geomesa-accumulo-tools/bin/setup-namespace.sh
```diff
@@ -115,6 +115,7 @@ fi
 
 if [[ -n "$ERROR" ]]; then
     help
+    exit
```
jahhulbert-ccri: can you make it "exit 1" so the return code is not zero

PR #1368, opened 2017-02-17
Fix typo in Javadoc comment

dalbani:


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/AvroFileInputFormat.scala
```diff
@@ -19,7 +19,7 @@ import org.opengis.feature.simple.SimpleFeature
 
 /**
  * Class for reading avro files written using
- * <code>org.locationtech.geomesa.features.avro.AvroDataFileWriter</code>.
+ * <code>org.locationtech.geomesa.features.avro.AvroDataFileReader</code>.
```
jnh5y: @dalbani I think we are using the AvroDataFileWriter to create the files.

Perhaps the comment can be made more clear in another way?

jnh5y:
Closing this one out... feel free to re-open if I was too hasty.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/AvroFileInputFormat.scala
dalbani: @jnh5y You're right, I didn't pay attention to what I read.. I jumped to the "wrong conclusions" where I saw `AvroDataFileWriter` in the comment while the code only mentions `AvroDataFileReader`.
I suppose we could reword the comment as something like:
```
/**
 * Class for reading Avro files using
 * <code>org.locationtech.geomesa.features.avro.AvroDataFileReader</code>,
 * which were originally written with
 * <code>org.locationtech.geomesa.features.avro.AvroDataFileWriter</code>.
 */
```
But I'm probably nitpicking here, for such a trivial comment.

PR #1370, opened 2017-02-20
GEOMESA-1683 - Add support for global JSON variables

jnh5y:

### DIFF
--- a/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
+++ b/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
@@ -46,9 +46,9 @@ class JsonSimpleFeatureConverter(jsonConfig: Configuration,
      }.getOrElse(Seq(Array()))
 
   // NB:  Currently the JSON support for Converters parses the entire JSON document into memory.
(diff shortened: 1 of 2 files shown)


geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
```diff
@@ -136,7 +140,18 @@ trait BaseJsonField[T] extends Field {
...
 
-  def evaluateJsonPath(args: Array[Any]): JsonElement = expression.read[JsonElement](args(0), jsonConfig)
+
+  def evaluateJsonPath(args: Array[Any]): JsonElement = {
+    val ret = expression.read[JsonElement](args(0), jsonConfig)
+    // In the event that we are reading from a 'feature path' and global context, we try to read the
+    //  'feature path' which is the first element of the array.
+    //  If it is null, we can read what we can from the global context.
+    if (ret.isJsonNull && args.length > 1) {
+      expression.read[JsonElement](args(1), jsonConfig)
```
elahrvivaz: instead of checking the root json as a fallback, how about adding a 'global-path' element to the HOCON config - you could add a flag to BaseJsonField (plus the impls) that indicates if it should look in `args(0)` (`path`) or `args(1)` (`global-path`). that way the intent is explicit.

PR #1372, opened 2017-02-20
GEOMESA-1685 Fixing NOT ID filters

elahrvivaz:
* Note: requires iterator jars to be updated

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyFilterTransformIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyFilterTransformIterator.scala
@@ -38,14 +38,15 @@ class KryoLazyFilterTransformIterator extends
 
   import KryoLazyFilterTransformIterator._
 
-  var source: SortedKeyValueIterator[Key, Value] = null
+  var source: SortedKeyValueIterator[Key, Value] = _
   val topValue: Value = new Value()
 
-  var sft: SimpleFeatureType = null
-  var filter: (SimpleFeature) => Boolean = null
+  var sft: SimpleFeatureType = _
+  var filter: (SimpleFeature) => Boolean = _
 
-  var reusableSf: KryoBufferSimpleFeature = null
-  var hasTransform: Boolean = false
+  var reusableSf: KryoBufferSimpleFeature = _
+  var setId: () => Unit = _
+  var hasTransform: Boolean = _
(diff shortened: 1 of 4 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoLazyFilterTransformIterator.scala
```diff
@@ -78,6 +79,14 @@ class KryoLazyFilterTransformIterator extends
       case (None, Some(s))    => s
       case (Some(c), Some(s)) => (sf) => c.evaluate(sf) && s(sf)
     }
+
+    setId = if (index.serializedWithId || cql.isEmpty) { () => {} } else {
```
jnh5y: Ok, why does it matter if the cql filter isn't used?

jnh5y: Oh, nevermind.  I'm silly....

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/IdExtractingVisitor.scala
```diff
@@ -42,13 +42,10 @@ class IdExtractingVisitor extends FilterVisitor {
     (andOption(yes).orNull, andOption(no).orNull)
   }
 
-  override def visit(f: Not, data: AnyRef): AnyRef = {
-    val (yes, no) = f.getFilter.accept(this, data).asInstanceOf[(Filter, Filter)]
-    (if (yes == null) null else ff.not(yes), if (no == null) null else ff.not(no))
```
jnh5y: Ok, this is a little bit mind-bendy for me...  What were we doing here?

elahrvivaz: before it was recursing into NOTs and extracting any IDs from there - now if we see a NOT we just stop and return no match. It's possible we might miss some double-negative, but detecting that would be kind of complicated I think.

jnh5y: SimplifyingFilterVisitor?  https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java

I thought we were using it...

PR #1374, opened 2017-02-20
Distributed Ingest For HBase

MutahirKazmi:
Signed-off-by: Mutahir <kazmi.mutahir@gmail.com>

- Resolved missing Jar issues when trying to do a HBase ingest with YARN.


geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseIngestCommand.scala
```diff
@@ -27,6 +27,7 @@ class HBaseIngestCommand extends IngestCommand[HBaseDataStore] with HBaseDataSto
 
   override def libjarsPaths: Iterator[() => Seq[File]] = Iterator(
     () => ClassPathUtils.getJarsFromEnvironment("GEOMESA_HBASE_HOME"),
+    () => ClassPathUtils.getJarsFromEnvironment("GEOMESA_HOME"),
```
elahrvivaz: We have deprecated the raw `GEOMESA_HOME` in favor of distribution specific ones (e.g. `GEOMESA_HBASE_HOME`). Did you need this to get it to work? Possibly your setup was incorrect...

geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/ingest-libjars.list
```diff
@@ -7,6 +7,9 @@ geomesa
 gson
 gt
 hbase
+hbase-client
+hbase-server
+hbase-protocol
```
elahrvivaz: these are fine - the `hbase` prefix should pick them all up if they're in the same directory, but this would cover cases where they aren't.

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseIngestCommand.scala
MutahirKazmi: Point noted and will update it.

PR #1375, opened 2017-02-21
GEOMESA-1688 - Create example of XML converter which reference attrib…

jnh5y:
…utes in feature-path tag


aheyne:
Looks good to me.

PR #1377, opened 2017-02-23
GEOMESA-1690 HBase Distributed Ingest Enabled and Tested

MutahirKazmi:
Signed-off-by: Mutahir <kazmi.mutahir@gmail.com>

Successfully tested ingest with HBase and Yarn. Fixed missing jar issues.

### DIFF
--- a/geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/ingest-libjars.list
+++ b/geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/ingest-libjars.list
@@ -7,6 +7,10 @@ geomesa
 gson
 gt
 hbase
+hbase-client
+hbase-server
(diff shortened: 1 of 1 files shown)


elahrvivaz:
thanks! how did you distribute the hbase-site.xml file? Is that something we can document or facilitate?

MutahirKazmi:
I made no changes to hbase-site.xml. It is the default without any configurations.

elahrvivaz:
is your setup actually distributed? most people have run into issues where the hbase connection can't be instantiated in the mappers because the hbase-site.xml isn't on the hadoop distributed classpath. If you're running on a single node or something, this might not be an issue. I'm going to create a new ticket for this PR and leave the other ticket open until we can sort out the hbase-site.xml distribution.

MutahirKazmi:
My setup has one namenode, one datanode, one nodemanager (yarn) and one resource manager (yarn). I don't have enough resources / computing power to test it with more than that.

PR #1379, opened 2017-02-24
GEOMESA-1691 Route search WPS

elahrvivaz:
* Search for points or lines matching a route
* Standardized BIN encoding between iterator and non-iterators


elahrvivaz:
note: hold off on merging while UI testing is under way

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
```diff
@@ -0,0 +1,349 @@
...
+    * @param headingThreshold threshold to match the heading of the route, in degrees
+    * @param routeGeomField geometry attribute in the route features that contains the route, optional - will use default geometry if not provided
+    * @param geomField geometry attribute in input features to match with the route, optional - will use default geometry if not provided
+    * @param bidirectional consider the direction of the route or just the path of the route
+    * @param headingField heading attribute in input features, required unless input geometries are linestrings
+    * @param bins encode the result in bin format
+    * @param binGeom geometry attribute to encode in the bin format, optional - will use default geometry if not provided
+    * @param binDtg date attribute to encode in the bin format, optional - will use default date if not provided
+    * @param binTrackId track id attribute to encode in the bin format, optional - will use feature id if not provided
+    * @param binLabel label attribute to encode in the bin format, optional
```
jnh5y: Lots of long lines...  Should we wrap 'em?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
```diff
@@ -0,0 +1,349 @@
...
+              @DescribeParameter(name = "binLabel", description = "Label field to use for BIN records", min = 0)
+              binLabel: String
+             ): SimpleFeatureCollection = {
+
+    logger.debug(s"Route searching on collection type ${features.getClass.getName}")
+
+    val bi = Option(bidirectional).getOrElse(java.lang.Boolean.FALSE)
+
+    if (!features.isInstanceOf[AccumuloFeatureCollection]) {
+      logger.warn(s"The provided feature collection type may not support distributed route search: ${features.getClass.getName}")
```
jnh5y: Is route search really happening in a distributed fashion?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
```diff
@@ -0,0 +1,349 @@
...
+  private var resultCalc: RouteResult = _
+
+  private val routeFilter =
+    orFilters(routes.map(ls => ff.dwithin(ff.property(geomAttribute), ff.literal(ls), routeBuffer, "meters")))
+
+  // for manual check, rewrite the filter to handle meters
+  // geotools doesn't handle this, we normally do it in our query planner
+  private lazy val manualRouteFilter = routeFilter match {
+    case f: DWithin => new QueryPlanFilterVisitor(null).visit(f, null).asInstanceOf[Filter]
+    case f: Or      => new QueryPlanFilterVisitor(null).visit(f, null).asInstanceOf[Filter]
```
jnh5y: Ok, I'm confused; seems like we are going to call the QueryPlanFilterVisitor on each output of ff.dwithin.  Would it make more sense to apply the visit to routes.map or routeFilter entirely?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
```diff
@@ -0,0 +1,349 @@
...
+      resultCalc = RouteResult(matchRoutes(manualCollection))
+    }
+
+    if (manualRouteFilter.evaluate(sf)) {
+      manualCollection.add(sf)
+    }
+  }
+
+  // allows us to accept visitors from retyping feature collections
+  override def getExpressions: java.util.List[Expression] = {
```
jnh5y: OH!  This is what you were talking about?  

The thing we can do here additional 'column/attribute' names which can come back?  I.e. for a density, we could add 'result'?

elahrvivaz: not sure actually - this lets it know which columns you want. if you don't implement this interface, the RetypingFeatureCollection will never delegate. If you do implement it, then it will compare the columns you want with the underlying columns and decide if it can delegate or not. Possibly implementing it but returning an empty list would always let it delegate...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
elahrvivaz: the regular filter gets passed into our query planner during execution, and the dwithin is handled there - so we don't want to modify it first. This filter is used directly when manually iterating through the features, so we have to fix the dwithin ourselves. Unfortunately FilterVisitor doesn't have a generic `visit(Filter)` method, so we have to use the match to call the appropriate filter-type-specific method.

elahrvivaz: i'll clarify the comment

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
elahrvivaz: the initial dwithin is, the comparison of routes is not...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
elahrvivaz: wilco

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
```diff
@@ -0,0 +1,354 @@
...
+    * Optimized method to run distributed query. Sets the result, available from `getResult`
+    *
+    * @param source simple feature source
+    * @param query may contain additional filters to apply
+    */
+  def routeSearch(source: SimpleFeatureSource, query: Query): Unit = {
+    logger.debug(s"Visiting source type: ${source.getClass.getName}")
+    val result = if (routes.isEmpty) { source.getFeatures(Filter.EXCLUDE) } else {
+      import org.locationtech.geomesa.filter._
+      val filter = if (query == null || query.getFilter == Filter.INCLUDE) { routeFilter } else {
```
jnh5y: As a corner case, do we need to consider any transforms or other query hints which might be set?

For instance, can I override the CBO in this process?  (And should I be able to?)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
```diff
@@ -0,0 +1,354 @@
...
+      override def initialValue(): GeodeticCalculator = new GeodeticCalculator
+    }
+    val headingCache = {
+      import scala.collection.JavaConverters._
+      new ConcurrentHashMap[(LineString, Int), Double].asScala
+    }
+
+    // gets the heading for an input feature
+    val getFeatureHeading: (SimpleFeature) => Double = headingIndex match {
+      case Some(index) => (sf) => Converters.convert(sf.getAttribute(index), classOf[Double])
```
jnh5y: What kind of cases does the converter call cover that .asInstanceOf[Double] doesn't?  (Mainly just curious)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
elahrvivaz: good question... since we don't do this in any other processes maybe we should address this holistically in another PR

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
elahrvivaz: it does type casting, so if the type was actually a string, or an int, or whatever it would work where the cast would fail. It may be overkill but it's handy

jnh5y: Eh, I'm fine with using big hammers;).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/RouteSearchProcess.scala
jnh5y: Ok, sounds like a follow-on ticket for when it comes up:).

PR #1382, opened 2017-02-27
GEOMESA-1694 point/poly initial version

normanb:
Failing tests are;

https://github.com/TheWeatherCompany/geomesa/blob/877_cassandra/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraAttributeIndex.scala

and

https://github.com/TheWeatherCompany/geomesa/blob/877_cassandra/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraAttributeIndex.scala

https://github.com/TheWeatherCompany/geomesa/blob/877_cassandra/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraAttributeIndex.scala 
Needs work to parse the row byte arrays in AttributeIndex.

Once the AttributeIndex is working I would like to add more tests. Cassandra is particular about limiting columns in a query string so this module needs a large amount of test coverage.

In general I have tried to break apart the row arrays and populate a Cassandra schema so that we can query by partition key appropriately.

It currently does single threaded inserts (for ease of debugging) I will add a commit to make this multi-threaded before this is merged in.


geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraDataStoreFactory.scala
```diff
@@ -33,25 +49,57 @@ class CassandraDataStoreFactory extends AbstractDataStoreFactory {
...
+    val looseBBox = LooseBBoxParam.lookupWithDefault[Boolean](params)
+
+    val cfg = CassandraDataStoreConfig(catalog, generateStats, audit, caching, queryThreads, queryTimeout, looseBBox)
+
+    new CassandraDataStore(session, cfg)
+  }
 
   override def getDisplayName: String = "Cassandra (GeoMesa)"
 
   override def getDescription: String = "GeoMesa Cassandra Data Store"
```
elahrvivaz: [minor] you can delegate these to the constants in the CassandraDataStoreFactory object so that they're standardized (the constants there aren't used for anything else)

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraFeatureWriter.scala
```diff
@@ -10,71 +10,53 @@ package org.locationtech.geomesa.cassandra.data
...
+                               val filter: Filter)
+    extends CassandraFeatureWriterType(sft, ds, indices) with CassandraModifyFeatureWriterType with CassandraFeatureWriter
+
+trait CassandraFeatureWriter extends CassandraFeatureWriterType {
+  private val serializer = new KryoFeatureSerializer(sft, SerializationOptions.withoutId)
+
+  override protected def createMutators(tables: IndexedSeq[String]): IndexedSeq[Any] = {
+    // this is a no-op for cassandra but we need to satisfy the interface
+    // return metadata about table
+    val m = ds.session.getCluster().getMetadata
```
elahrvivaz: could remove this and just return the input seq

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraQueryPlan.scala
```diff
@@ -0,0 +1,72 @@
...
+  override def scan(ds: CassandraDataStore): CloseableIterator[SimpleFeature] = CloseableIterator.empty
+}
+
+case class QueryPlan(filter: CassandraFilterStrategyType,
+                    table: String,
+                    ranges: Seq[Statement],
+                    clientSideFilter: Option[Filter],
+                    entriesToFeatures: Iterator[Row] => Iterator[SimpleFeature]) extends CassandraQueryPlan {
+
+  override def hasDuplicates: Boolean = true
```
elahrvivaz: are there always going to be duplicates? this will trigger client-side deduplication, which should be avoided if not required

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
```diff
@@ -0,0 +1,186 @@
...
+      val z = ByteBuffer.wrap(row, curr, 8).getLong
+      curr += 8
+      val fid = new String(row, curr, row.length - curr, StandardCharsets.UTF_8)
+      (pk, z, fid, sharing)
+    }
+  }
+
+  override def configure(sft: SimpleFeatureType, ds: CassandraDataStore): Unit = {
+    super.configure(sft, ds)
+    tableName = getTableName(sft.getTypeName, ds)
```
elahrvivaz: this method will only get called when a new schema is created, so shouldn't be used to initialize state. also, table name is going to vary if you create different schemas, so shouldn't be stored in the common object. Instead you can pass it through in CassandraFeatureWriter.createMutators

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  override def delete(sft: SimpleFeatureType, ds: CassandraDataStore, shared: Boolean): Unit = {
+    val name = getTableName(sft.getTypeName, ds)
+    ds.session.execute(s"drop table $tableName")
+  }
+
+  override protected def createInsert(row: Array[Byte], cf: CassandraFeature): Statement = {
+    val (pk:java.lang.Integer, z:java.lang.Long, fid:java.lang.String, sharing:java.lang.String) = getRowValues(row)
+    new SimpleStatement(s"INSERT INTO $tableName (pk, z, fid, sharing, sf) values (?, ?, ?, ?, ?)",
```
elahrvivaz: this might need to be changed so return a string for the statement and the binding values, so that the table name can be inserted after it's available in CassandraFeatureWriter

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
```diff
@@ -0,0 +1,186 @@
...
+    select
+  }
+
+  override protected def rangeExact(row: Array[Byte]): Statement = {
+    val (pk, sharing, fid, z) = getRowValues(row)
+    new SimpleStatement(s"""SELECT * FROM $tableName
+          | WHERE pk = $pk
+          | AND sharing = '$sharing'
+          | AND fid = '$fid'
+          | AND z = $z""".stripMargin)
```
elahrvivaz: should maybe use the QueryBuilder to prevent nested quotes being an issue

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraDataStoreFactory.scala
normanb: sorry, I don't understand what you mean here. How can delegate cfg and the ds?

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
normanb: does this apply for getRowValues as well? I set that up for the SFT in configure. Should I call this function on every insert/delete, cf.feature isn't available on range.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
normanb: ok, can do.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
elahrvivaz: yeah, you don't want any state in this class since it's a globally shared object (the exception would be caches keyed by something unique that's accessible in the method, usually the simple feature type). the idea (which may not work quite right with cassandra) is that the actual values being inserted are separate from the table they are being inserted into. so in our accumulo or hbase impls, we create a mutation here, then in the feature writer we apply it to a particular table. the workflow is like this (using hbase as an example):

1. geomesa feature writer is instantiated and creates 'writers' (details in step 2): https://github.com/locationtech/geomesa/blob/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala#L114
2. each index being written to (z3, z2, etc) configures its own writer, which is responsible for creating the insert row key: https://github.com/locationtech/geomesa/blob/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala#L95-L97
4. a mutator is created for each index being written to - for hbase this is a buffered mutator: https://github.com/locationtech/geomesa/blob/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala#L117
https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseFeatureWriter.scala#L39
5. the mutators are called when a features is actually written after being passed in by the user: https://github.com/locationtech/geomesa/blob/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala#L134
https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseFeatureWriter.scala#L43

For C*, you don't really have mutators since everything operates on the session. Possibly when you create the insert row key, you can create a partial statement that has everything except the table defined. Then, when it gets passed to the 'mutators' you could just fill in the table name.

this is probably clear as mud...

jnh5y:
@normanb first, thanks for the contribution!  I'll try and get some time later this week to review in depth.

As really boring bookkeeping, we use the Maven license plugin to make sure that all source code files have a header block.  

I've got a sample commit here: https://github.com/ccri/geomesa/commit/371ace1f37a8ddcf1875cf0954c7c0d7c231233b which adds a new header file.  

If you'd like to add that to your PR as you get closer, feel free.  If you'd like me to merge an IBM header into master, I can do that first, and you can merge it in.

elahrvivaz:
@normanb I've got some time to help out if there's anything I can do

normanb:
Emilio, great, thank you. I am looking to push an update to the PR to take
state out of the configure. This seems ok, the biggest issue I see is with
the AttributeIndex and being able to parse all different types, if you
could look at that it would help. Thanks.

elahrvivaz:
replacing with #1385

PR #1392, opened 2017-03-06
GEOMESA-1706 clean up JavaGeoMesaSpark API to removed need for cast in client code

tkunicki:


tkunicki:
with `Map<String, String> params = ...` allow for `JavaGeoMesaSpark.apply(params)` instead of `JavaGeoMesaSpark.apply((Map)params)`

PR #1393, opened 2017-03-07
GEOMESA-1539 Support Kerberos Authentication for Accumulo

jrs53:
* Initial Kerberos authentication support
* AccumuloDataStoreFactory acceps a keytabPath parameter
* geomesa accumulo command line tool accepts a --keytab parameter
* setup_namespace.sh accepts a -t parameter
* Includes documentation and unit tests


docs/user/accumulo/kerberos.rst
```diff
@@ -0,0 +1,24 @@
+Kerberos
+========
+
+GeoMesa includes initial support for Accumulo clusters which are authenticated using Kerberos. Currently keytabs
+(and not cached TGTs) are supported, apart from initial setup which requires a cached TGT, usually obtained using ``kinit``.
+
+Kerberos functionality should be used as follows:
+
+#. ``setup_namespace.sh`` should be called using with the ``-t`` flag to use a cached TGT
```
elahrvivaz: [typo] using with the

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -89,8 +92,31 @@ object AccumuloDataStoreFactory {
...
+        (ctor.newInstance(user, new java.io.File(keytabPath), java.lang.Boolean.TRUE)).asInstanceOf[AuthenticationToken]
+      } catch {
+        // Remove reflection exception wrapper
+        case ite: InvocationTargetException => throw ite.getCause
+      }
+    } else {
+      // Will get here only if using mock kerberos thanks to parameter validation.
+      // MockInstance doesn't seem to support Kerberos, so make up a password
+      // Blank password is for the root user
+      new PasswordToken("".getBytes("UTF-8"))
```
elahrvivaz: could we switch the logic here slightly, and move the `!useMock` inside the `else if (keytabPath != null)`, then make the final else throw an exception of some kind since no authentication was provided (previously it would throw an NPE on line 93, which wasn't ideal, but this change would mask that)
edit: I realize that it shouldn't be in that state due to canProcess, but I think it would still be better to be defensive here

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -183,8 +217,35 @@ object AccumuloDataStoreFactory {
...
-  def canProcess(params: JMap[String,Serializable]): Boolean =
-    params.containsKey(instanceIdParam.key) || params.containsKey(connParam.key)
+  // Kerberos is only available in Accumulo >= 1.7.
+  // Probe for whether part of the API is available
+  // Note: doesn't confirm whether correctly configured for Kerberos e.g. core-site.xml on CLASSPATH
+  def isKerberosAvailable() = {
+    try {
+      val conf = new org.apache.accumulo.core.client.ClientConfiguration()
+      val withSasl = conf.getClass().getMethod("withSasl", classOf[Boolean])
+      true
```
elahrvivaz: we also have `org.locationtech.geomesa.accumulo.AccumuloVersion#accumuloVersion` which you can check, although this may be more robust in the end

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -89,8 +92,31 @@ object AccumuloDataStoreFactory {
...
     val password = passwordParam.lookup[String](params)
+    val keytabPath = keytabPathParam.lookup[String](params)
+
+    var usingKerberos = false
+
+    // Build authenticaton token according to how we are authenticating
+    val authToken : AuthenticationToken = if (password != null ) {
+      new PasswordToken(password.getBytes("UTF-8"))
+    } else if (keytabPath != null && !useMock) {
+      usingKerberos = true
```
jnh5y: It looks like the only situation where 'usingKerberos' is true is when "keytabPath != null && !useMock" is true.  

It might be worth it to ditch the var, and then check that condition below.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -183,8 +217,35 @@ object AccumuloDataStoreFactory {
...
+    val instanceIdOrConnection = params.containsKey(instanceIdParam.key) || params.containsKey(connParam.key)
+
+    // Is valid password set?
+    val hasPassword = params.containsKey(passwordParam.key) && passwordParam.lookup[String](params) != null
+
+    // Is valid keytabPath set?
+    val hasKeytabPath = params.containsKey(keytabPathParam.key) && keytabPathParam.lookup[String](params) != null
+
+    instanceIdOrConnection && (
+        (!hasPassword && !hasKeytabPath) || // neither set - included for backwards compatibility, if connector provided
```
elahrvivaz: could we split out the cases where instanceId is provided vs connector? If connector is provided, we don't need anything else, but otherwise we would need  user + credentials

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
```diff
@@ -0,0 +1,61 @@
...
+/**
+  * Trait to simplify tests that require a Key Distribution Center (KDC)
+  */
+trait TestWithKerberos extends Specification {
+
+  // File paths, create workDir if doesn't already exist
+  // TODO: verify this path is ok
+  // TODO: should delete workDir afterwards?
+  // TODO: should stop kdc after testing too?
+  val workDir = new File("." + File.separator + "kerberos")
```
elahrvivaz: could use a temp file instead of modifying the working dir

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -89,8 +92,31 @@ object AccumuloDataStoreFactory {
...
+
+    var usingKerberos = false
+
+    // Build authenticaton token according to how we are authenticating
+    val authToken : AuthenticationToken = if (password != null ) {
+      new PasswordToken(password.getBytes("UTF-8"))
+    } else if (keytabPath != null && !useMock) {
+      usingKerberos = true
+      // Get ctor for principal, keytab path and replace current user
+      val ctor = Class.forName("org.apache.accumulo.core.client.security.tokens.KerberosToken")
```
jnh5y: Why did you end up using reflection here?  It'd be great to avoid it if possible, but sometimes we need to do a little...

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
```diff
@@ -0,0 +1,61 @@
...
+  // Write user keytab to file
+  {
+    val keytabFile = new File(keytabFilename)
+    kdc.exportPrincipals(keytabFile)
+    keytabFile.flush()
+  }
+
+  // Hadoop code seems to do this with the comment
+  // Will be fixed in next Kerby version.
+  Thread.sleep(1000)
```
elahrvivaz: it would probably be better to wrap this initialization in a function, and then use `step { startKerberos() }` and `step { stopKerberos() }` in the unit tests that extend this so that it lives in the test lifecycle

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -88,11 +89,27 @@ object AccumuloRunner extends Runner {
...
     params.collect {
-      case p: RequiredCredentialsParams if p.password == null => p
-      case p: OptionalCredentialsParams if p.password == null && p.user != null => p
+      case p: RequiredCredentialsParams if (p.password == null && p.keytab == null) => p
     }.foreach(_.password = Prompt.readPassword())
 
+    // Error if both password and keytab specified
+
+
+//    params.collect {
```
elahrvivaz: [minor] remove commented out code

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
```diff
@@ -0,0 +1,61 @@
...
+  * ************************************************************************/
+
+package org.locationtech.geomesa.accumulo
+
+import org.specs2.mutable.Specification
+
+import org.apache.kerby.kerberos.kerb.server.SimpleKdcServer
+import org.apache.kerby.util.NetworkUtil
+
+import java.io.File
```
jnh5y: As a picky comment, we use IntelliJ's 'optimize imports' to sort imports in alphabetically order (and remove the spaces between the lines).

geomesa-accumulo/geomesa-accumulo-tools/bin/setup-namespace.sh
```diff
@@ -23,21 +24,25 @@ function help() {
...
+  echo "  -t    Login using an existing Kerberos token, else will prompt for password"
   echo "  -g    Path of the GeoMesa distributed runtime JAR, with or without raster support."
   echo "  -d    Directory to create namespace in, default: /accumulo/classpath"
   echo "  -h    HDFS URI e.g. hdfs://localhost:54310"
   echo ""
   exit 0
 }
 
-while getopts ":u:p:n:g:h:d:help" opt; do
+while getopts ":u:p:tn:g:h:d:help" opt; do
```
jnh5y: Should this be "t:n"?  (I'm unsure exactly what this is doing, just asking for sanity sake.)

geomesa-accumulo/geomesa-accumulo-tools/bin/setup-namespace.sh
```diff
@@ -124,11 +129,16 @@ hadoop fs -copyFromLocal -f "$GEOMESA_JAR" "${NAMESPACE_DIR}/${ACCUMULO_NAMESPAC
...
+        ACCUMULO_SHELL_CMD="accumulo shell -u $ACCUMULO_USER -p $ACCUMULO_PASSWORD"
+    else
+        ACCUMULO_SHELL_CMD="accumulo shell -u $ACCUMULO_USER"
+    fi
     echo -e "createnamespace ${ACCUMULO_NAMESPACE}\n" \
       "grant NameSpace.CREATE_TABLE -ns ${ACCUMULO_NAMESPACE} -u $ACCUMULO_USER\n" \
       "config -s general.vfs.context.classpath.${ACCUMULO_NAMESPACE}=${HDFS_URI}${NAMESPACE_DIR}/${ACCUMULO_NAMESPACE}/.*.jar\n" \
       "config -ns ${ACCUMULO_NAMESPACE} -s table.classpath.context=${ACCUMULO_NAMESPACE}\n" \
-      | accumulo shell -u $ACCUMULO_USER -p $ACCUMULO_PASSWORD
+      | ${ACCUMULO_SHELL_CMD}
```
jnh5y: We can configure the namespaces with the keytab?

Awesome!

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloDataStoreCommand.scala
```diff
@@ -36,8 +38,8 @@ trait AccumuloDataStoreCommand extends DataStoreCommand[AccumuloDataStore] {
 
     if (parsedParams.get(AccumuloDataStoreParams.mockParam.getName).exists(_.toBoolean)) {
       val ret = mockDefaults ++ parsedParams // anything passed in will override defaults
-      // MockAccumulo sets the root password to blank so we must use it if user is root
-      if (ret(AccumuloDataStoreParams.userParam.getName) == "root") {
+      // MockAccumulo sets the root password to blank so we must use it if user is root, providing not using keytab instead
+      if (ret(AccumuloDataStoreParams.userParam.getName) == "root" && !(ret contains AccumuloDataStoreParams.keytabPathParam.getName)) {
```
jnh5y: I'd suggest word wrapping this line at the &&.  (We try to avoid long lines.)

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -88,11 +89,27 @@ object AccumuloRunner extends Runner {
 
     val params = Option(command.params)
 
+    // TODO: handle OptionalCredentialsParams
```
jnh5y: As a final pass, we should convert any remaining TODOs into JIRA tickets.

pom.xml
```diff
@@ -2277,6 +2277,7 @@
                             <include>**/*Test.java</include>
                         </includes>
                         <argLine>${maven.test.jvmargs}</argLine>
+
```
jnh5y: [minor] Remove stray new line.

geomesa-accumulo/geomesa-accumulo-tools/bin/setup-namespace.sh
jrs53: From http://stackoverflow.com/questions/18414054/bash-getopts-reading-optarg-for-optional-flags/18414091#18414091 I *thought* flag type parameters (which are either present or absent i.e. don't take a parameter) didn't use a colon. However, that link isn't exactly definitive

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Because those classes don't exist in Accumulo <1.7 (hence won't compile) but the minimum version of Accumulo supported by GeoMesa is 1.6

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: I'm going to have to use reflection anyway because those classes don't exist in Accumulo <1.7?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
jrs53: Yup - sorry meant to do that before PR

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Yup, there's kind of a meta question about whether if the parameters say to use mock *and* Kerberos whether that is really an error condition because mock doesn't support Kerberos?

docs/user/accumulo/kerberos.rst
jrs53: Wilco, thx

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -183,8 +217,35 @@ object AccumuloDataStoreFactory {
...
+    // Is valid password set?
+    val hasPassword = params.containsKey(passwordParam.key) && passwordParam.lookup[String](params) != null
+
+    // Is valid keytabPath set?
+    val hasKeytabPath = params.containsKey(keytabPathParam.key) && keytabPathParam.lookup[String](params) != null
+
+    instanceIdOrConnection && (
+        (!hasPassword && !hasKeytabPath) || // neither set - included for backwards compatibility, if connector provided
+        (hasPassword && !hasKeytabPath) ||  // only password set, all ok
+        ((!hasPassword && hasKeytabPath) && isKerberosAvailable())) // if keytab set, Kerberos must be available
```
jrs53: Yup, I previously had that when I was also trying to implement cached TGT support so there was yet another check. What about logging the cause of the failure to be helpful?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
jrs53: Wilco

geomesa-accumulo/geomesa-accumulo-tools/bin/setup-namespace.sh
jrs53: Not quite sure what you mean? The difference here is only whether "accumulo shell" is called with a username and password (non-Kerberos) or just the username (Kerberos, assuming cached TGT)

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloDataStoreCommand.scala
jrs53: Wilco

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
jrs53: Wilco - since OptionalCredentialsParams doesn't seem to be used anywhere, should I just take this out?

pom.xml
jrs53: Wilco, oops

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
jrs53: Oops, thanks

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Yup, @jnh5y suggested something similar

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
jrs53: Yup, good idea

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
elahrvivaz: that's probably fine... if it's easy, might be safer to add handling for it in case it does get used some time in the future

jrs53: Well I didn't really know what it was intended to do. The Accumulo data store always requires a username (or connection) so I wasn't quite sure what this parameter flavour  could ever be used for. Will just remove the comment I think.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: I think we've dropped official support for 1.6, although we'd like to keep unofficial support. In general if it compiles with 1.7, but will still work with a 1.6 runtime we should be ok. I think that's the case here, since this block of code won't get hit in 1.6. Alternatively, we'd tried to keep all our accumulo versioned code in the `AccumuloVersion` class, which does use reflection - maybe you could move this there.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: As I mentioned above, I think we'd be ok requiring 1.7 for compile, since this code won't get hit in 1.6 at runtime.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: logging is always good :)

docs/user/accumulo/commandline_tools.rst
```diff
@@ -60,33 +60,38 @@ Run ``geomesa`` without any arguments to produce the following usage text::
...
-username and password in each command by using ``-u`` (or ``--username``) and ``-p`` (or
+
+The Accumulo username and either a password or a path to a Kerberos keytab file is required for each command.
+Specify the username and password in each command by using ``-u`` (or ``--username``) and ``-p`` (or
 ``--password``), respectively. One can also only specify the username on the
 command line using ``-u`` or ``--username`` and type the password in an additional
 prompt, where the password will be hidden from the shell history.
+To use Kerberos authentication instead, use ``--keytab`` with a path to a Kerberos keytab file containing an entry for
+the specified user. Since a keytab file allows authentication without any further constraints, it should be protected
+appropriately.
```
jahhulbert-ccri: @jrs53 should we attempt to verify the permissions are 400 or 600 or something like that?

geomesa-accumulo/geomesa-accumulo-datastore/pom.xml
```diff
@@ -144,6 +145,12 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kerby</groupId>
+            <artifactId>kerb-simplekdc</artifactId>
+            <version>1.0.0-RC2</version>
```
jahhulbert-ccri: consider moving the version number up to the root ```pom.xml``` in the ```<dependencyManagement>``` section

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jahhulbert-ccri: hehe yeah I think in general if you have ```--mock``` we just say ```-p``` and ```--keytab``` don't matter. We have gone the route of making sure that they don't prevent the mock from working if they are provided though.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jahhulbert-ccri: +1 I think we'd like to make sure 1.6 does work.

docs/user/accumulo/commandline_tools.rst
jrs53: Could do (c.f. ssh keys), but I'm not 100% sure how to do this cross platform since the client might be on Windows too. Caveat emptor?

geomesa-accumulo/geomesa-accumulo-datastore/pom.xml
jrs53: I've actually removed the dep, since although I went to the trouble of creating a mini KDC server, it wasn't actually required anywhere! (legacy from some previous code trying to get TGTs working)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: I've changed this to look at the type of authToken which I think is cleaner, and not in performance critical code anyway

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Done, agree it is clearer that way (throwing IllegalArgumentException)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Done - no longer using reflection.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Done - no longer using reflection.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Done - much simplified, but also checking other mandatory params

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithKerberos.scala
jrs53: OBE - since my lovely KDC turned out not to be used

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Made clearer as suggested, but no logging added due to the large number of failure modes

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -1,5 +1,6 @@
 /***********************************************************************
 * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* Portions Crown Copyright (c) 2016 Dstl
```
elahrvivaz: I guess it's 2016 because you initially wrote it then?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
```diff
@@ -43,8 +43,11 @@ trait OptionalTypeNameParam extends TypeNameParam {
 }
 
 trait PasswordParams {
-  @Parameter(names = Array("-p", "--password"), description = "Connection password")
+  @Parameter(names = Array("-p", "--password"), description = "Connection password (will prompt if not supplied and not using Kerberos)")
   var password: String = null
+
+  @Parameter(names = Array("--keytab"), description = "Path to Kerberos keytab file")
+  var keytab: String = null
```
elahrvivaz: This needs to be moved into the accumulo-tools module - this trait is also used for the cassandra implementation (and probably others going forward).

jrs53: Yes, I've already done that to avoid the conflict with master. Just fixing up the test which fail due to the stricter Params checking even when using mock

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jrs53: Yup, exactly

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
jrs53: Done and pushed. CI (when it doesn't time out) seems happy

elahrvivaz: @jrs53 I meant we need to move the keytab var out of this trait and into another mixin trait, and update the accumulo tools classes with the new mixin, since the PasswordParams trait is re-used in cassandra and keytabs aren't supported there (yet).

jrs53: Ah yes, I think when I originally wrote it the trait wasn't shared. Just pushed the change.

elahrvivaz:
great, lgtm. @jnh5y do you want to handle the merge?

docs/user/accumulo/commandline_tools.rst
jahhulbert-ccri: hehehe windows. i forgot about that. Yeah too detailed of a thought for sure.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -89,8 +92,24 @@ object AccumuloDataStoreFactory {
     val instance = instanceIdParam.lookup[String](params)
     val user = userParam.lookup[String](params)
     val password = passwordParam.lookup[String](params)
+    val keytabPath = keytabPathParam.lookup[String](params)
+
+    // Build authentication token according to how we are authenticating
+    val authToken : AuthenticationToken = if (password != null && keytabPath == null) {
+      new PasswordToken(password.getBytes("UTF-8"))
+    } else if (password == null && keytabPath != null) {
+      if(!useMock) {
```
jahhulbert-ccri: picky - add a space after if but don't let it delay merge we can update it in the upcoming decade :)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreKerberosTest.scala
```diff
@@ -0,0 +1,110 @@
...
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.accumulo.data
+
+import org.locationtech.geomesa.accumulo.data.AccumuloDataStoreFactory
+
```
jahhulbert-ccri: pick extra whitespace

PR #1398, opened 2017-03-07
GEOMESA-1712 Binary encoding WPS process

elahrvivaz:


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/BinConversionProcess.scala
```diff
@@ -0,0 +1,155 @@
...
+      Some(GeometryAttribute(name, axis.getOrElse(AxisOrder.LonLat))) // note: wps seems to always be lon first
+    }
+    val dtgField   = Option(dtg).orElse(sft.getDtgField)
+    val trackField = Option(track).orElse(sft.getBinTrackId).orElse(Some("id"))
+    val labelField = Option(label)
+
+    // validate inputs
+    (geomField.map(_.geom).toSeq ++ dtgField ++ trackField ++ labelField).foreach { attribute =>
+      if (attribute != "id" && sft.indexOf(attribute) == -1) {
+        throw new IllegalArgumentException(s"Attribute $attribute doesn't exist in $sft")
```
jnh5y: I suppose we could still support transforms since the input feature collection would have the SFT with the new, synthetic arguments?

elahrvivaz: right, you'd just have to make sure the fields you specify exist in the input collection

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/BinaryOutputEncoder.scala
```diff
@@ -59,7 +83,33 @@ object BinaryOutputEncoder extends LazyLogging {
...
+    if (sort) {
+      iter.toList.sorted.foreach(encode)
+    } else {
+      iter.foreach(encode)
+    }
+    // Feature collection has already been closed by SelfClosingIterator
+  }
+
+  private def toValues(sft: SimpleFeatureType, options: EncodingOptions):
+      Either[(SimpleFeature) => Seq[ValuesToEncode], (SimpleFeature) => ValuesToEncode] = {
```
jnh5y: Would it be cleaner to return SF=>Seq[ValuesToEncode] here?  And then above on line 86, just flatmap over the seq?

elahrvivaz: yeah, i was trying to avoid the overhead of creating all the `Seq`s since the normal use case is points and the `Seq` are just for linestrings...

PR #1411, opened 2017-03-10
HBase Server Side Filters

spandanagrawal:
This commit adds a server side filtering mechanism to HBase. 

All system tests have been run and passed, including the HBaseDataStoreTest separately.

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/java/org/locationtech/geomesa/hbase/filters/JSimpleFeatureFilter.java
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/java/org/locationtech/geomesa/hbase/filters/JSimpleFeatureFilter.java
@@ -1,0 +1,102 @@
+package org.locationtech.geomesa.hbase.filters;
+
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.CellUtil;
+import org.apache.hadoop.hbase.exceptions.DeserializationException;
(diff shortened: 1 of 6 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/java/org/locationtech/geomesa/hbase/filters/JSimpleFeatureFilter.java
```diff
@@ -0,0 +1,102 @@
...
+import org.geotools.filter.text.ecql.ECQL;
+import org.locationtech.geomesa.features.kryo.KryoFeatureSerializer;
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.locationtech.geomesa.features.interop.SerializationOptions;
+
+import java.io.IOException;
+
+public class JSimpleFeatureFilter extends FilterBase {
```
MutahirKazmi: You may want to change the name of the Feature Filter maybe.

MutahirKazmi:
This depends upon PR # 1383 which has not been approved yet.

elahrvivaz:
you guys have any interest in re-writing in scala? we'd probably need a java class that extends the scala class to provide the static method required by the hbase interface, but that shouldn't be too bad.

PR #1416, opened 2017-03-13
GEOMESA-1695 Adding unit test for HBase counts

elahrvivaz:
* Note: could not reproduce getCount always returning 0


geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
```diff
@@ -151,6 +152,7 @@ class HBaseDataStoreTest extends Specification with LazyLogging {
         feature.getAttribute(attribute) mustEqual results.find(_.getID == feature.getID).get.getAttribute(attribute)
       }
     }
+    ds.getFeatureSource(typeName).getFeatures(query).size() mustEqual results.length
```
jahhulbert-ccri: may also want to compare against the original expected value incase results is returning the wrong answer though I assume our other unit tests for hbase query would catch that :+1:

elahrvivaz: yeah, it check the correct answers right above this - results is actually the expected results passed into the method, not the query results (confusingly named)

PR #1417, opened 2017-03-13
GEOMESA-1630 Tools export only requested attributes

elahrvivaz:
* 'id' in attribute list can be used to export feature ID


jahhulbert-ccri:
@elahrvivaz do we care about ordering? 
```
ahulbert@doubletrouble:/tmp/geomesa-accumulo_2.11-1.3.2-SNAPSHOT$ bin/geomesa export  -u root -p secret -c geomesa.local -f foobar -a name,id
id,name:String:index=true
26236,Hermione
3233,Severus
23623,Harry
```

elahrvivaz:
yeah, id is always at the beginning - i thought about it but wasn't sure it mattered - do you care?

jahhulbert-ccri:
not really :) might be a little confusing if someone attempts to order it though. as long as it is documented i think its OK.

PR #1419, opened 2017-03-14
GEOMESA-1607 Enabling density iterator against AttributeIndex

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
@@ -93,7 +93,7 @@ object Z2DensityIterator {
                 filter: Option[Filter],
                 hints: Hints,
                 priority: Int = KryoLazyDensityIterator.DEFAULT_PRIORITY): IteratorSetting = {
-    val is = KryoLazyDensityIterator.configure(sft, index, filter, hints, priority)
+    val is = KryoLazyDensityIterator.configure(sft, index, filter, hints, deduplicate = false, priority)
     is.setIteratorClass(classOf[Z2DensityIterator].getName)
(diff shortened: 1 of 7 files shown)


elahrvivaz:
one additional optimization that I didn't do: if the 'weight' attribute is the same as the attribute being queried, we could apply the `KryoAttributeKeyValueIterator` and avoid a join.

jahhulbert-ccri:
@elahrvivaz I think we would definitely want to avoid a join in this case...we need a single scan against that table for the use case. The weight attr will be the attribute indexed.

elahrvivaz:
ok, I'll add that in. Wasn't sure if it was a use case (weight isn't a required attribute, it will default to 1 per record)

elahrvivaz:
added

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
```diff
@@ -93,7 +93,7 @@ object Z2DensityIterator {
                 filter: Option[Filter],
                 hints: Hints,
                 priority: Int = KryoLazyDensityIterator.DEFAULT_PRIORITY): IteratorSetting = {
-    val is = KryoLazyDensityIterator.configure(sft, index, filter, hints, priority)
+    val is = KryoLazyDensityIterator.configure(sft, index, filter, hints, deduplicate = false, priority)
```
jnh5y: Can you just write 'false' instead of 'deduplicate = false'?

elahrvivaz: you can, but intellij yells at you to 'name Boolean parameters' :)

jnh5y: I wonder why... but yeah, merging...

PR #1421, opened 2017-03-15
For Comment: Adding JsonPath attribute extractor CQL Function

aheyne:
This adds the CQL function 'jsonattr' to allow specifying jsonpaths in a CQL filter function, in particular, through WFS. This allows queries like these, and more, to work. See the unit test for more examples.

`jsonattr('$.data.[FooBar]') = 'baz'`
`jsonattr('$.data.[Foo Bar]') = 'baz'`
`jsonattr('$.data.[Foo.Bar]') = 'baz'`

Any thoughts on a better way to do this? The JsonPath escape in FastPropertyName is necessary because it is called when using 
`val path = getExpression(0).evaluate(null).asInstanceOf[String]`
to extract the function parameter (the passed in path), otherwise FastPropertyName will throw exceptions.

### DIFF
--- a/geomesa-features/geomesa-feature-kryo/pom.xml
+++ b/geomesa-features/geomesa-feature-kryo/pom.xml
@@ -39,11 +39,11 @@
             <groupId>org.parboiled</groupId>
             <artifactId>parboiled-scala_${scala.binary.version}</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-security_${scala.binary.version}</artifactId>
-            <scope>test</scope>
-        </dependency>
+        <!--<dependency>-->
+            <!--<groupId>org.locationtech.geomesa</groupId>-->
+            <!--<artifactId>geomesa-security_${scala.binary.version}</artifactId>-->
+            <!--<scope>test</scope>-->
+        <!--</dependency>-->
         <dependency>
             <groupId>org.specs2</groupId>
             <artifactId>specs2_${scala.binary.version}</artifactId>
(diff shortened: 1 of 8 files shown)


elahrvivaz:
can we change the function name to `jsonPath`?

geomesa-features/geomesa-feature-kryo/pom.xml
```diff
@@ -39,11 +39,11 @@
...
-        <dependency>
-            <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-security_${scala.binary.version}</artifactId>
-            <scope>test</scope>
-        </dependency>
+        <!--<dependency>-->
+            <!--<groupId>org.locationtech.geomesa</groupId>-->
+            <!--<artifactId>geomesa-security_${scala.binary.version}</artifactId>-->
+            <!--<scope>test</scope>-->
+        <!--</dependency>-->
```
elahrvivaz: any reason for removing this? we're going to need it to work with security on the classpath

aheyne:
Yeah, I was planning on changing it to something like that, jsonPath it is.

geomesa-features/geomesa-feature-kryo/pom.xml
aheyne: I don't think so, I just haven't cleaned this up yet.

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/json/JsonAttrFunction.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.geotools.filter.capability.FunctionNameImpl
+import org.geotools.filter.capability.FunctionNameImpl.parameter
+import org.geotools.filter.expression.PropertyAccessors
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.opengis.feature.simple.SimpleFeature
+import org.opengis.filter.expression.{PropertyName, VolatileFunction}
+
+class JsonAttrFunction extends FunctionExpressionImpl(
+  new FunctionNameImpl("jsonattr",
+    parameter("propertyName", classOf[PropertyName]),
```
jnh5y: Oh snap... @elahrvivaz is right.  If this is a 'string', we can save on the changes to the FastPropertyName. 

*headdesk*

aheyne: Oohhh

PR #1422, opened 2017-03-15
GEOMESA-1728 Expose Kafka's auto.offset.reset configuration via KafkaDataStore Params

chadphillips:
Exposed Kafka auto.offset.reset config as autoOffsetReset data store param.


elahrvivaz:
Verified CLA, thanks

PR #1425, opened 2017-03-16
GEOMESA-1733 add newlines to META-INF/services files and enforce with checkstyle plugin

tkunicki:
fixed newlines with `find . -path '*/resources/META-INF/services/*' -exec sed -i -e '$a\' {} \;` and added checkstyle rule to enforce


elahrvivaz:
seems like the issue the user was having was just a shade plugin version - do we still care about enforcing newlines?

tkunicki:
EOL at the end of files will never create an issue.  As we've seen the absence of them leaves us exposed to implementation nuances of various tools and plugins ([per POSIX a line ends with a newline](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)).   If we're worried about compilation speed we can unbind the checkstyle plugin or put it behind a profile that runs during CI jobs.

PR #1436, opened 2017-03-22
GEOMESA-1743 Reformat license header to conform to Javadoc

aheyne:
License headers now have space before each body line to conform with javadoc style so pasting with IntelliJ is correct.

### DIFF
--- a/build/copyright/geomesa-header.xml
+++ b/build/copyright/geomesa-header.xml
@@ -2,8 +2,8 @@
 <additionalHeaders>
     <geomesa_java>
         <firstLine>/***********************************************************************</firstLine>
-        <beforeEachLine>* </beforeEachLine>
-        <endLine>*************************************************************************/&#xA;</endLine>
+        <beforeEachLine> * </beforeEachLine>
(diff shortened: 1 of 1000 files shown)


elahrvivaz:
nice, thanks

elahrvivaz:
this is going to cause conflicts with all open prs... @jnh5y how do you want to handle it?

jnh5y:
@aheyne did you find a scriptable or other automated way to make these changes?

If so, let's try and zero out the open PR, document the process for updating and then explain it on any PRs which don't make the jump.

aheyne:
It's not scriptable but it doesn't take too long to do. There are only 5 or so distinct headers, you can have IntelliJ make the regex for both the old format and the new and then run the replace.

elahrvivaz:
possible quick way to re-run this (not tested):
```
for file in $(find . -name *.scala); do sed '1,8{/^\*/ */}' $file > tmp && mv tmp $file; done
```

aheyne:
Closed for #1550

PR #1446, opened 2017-03-24
GEOMESA-1740 Stats histogram errors on certain inputs

aheyne:
* Loss of precision was causing rounding errors when converting between point and z2. This was causing binned array bounds in copyInto to be miscalculated and causing ingest failures.
* Added unit test.

Special thanks to @tkunicki and @elahrvivaz 

### DIFF
--- a/geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/NormalizedDimension.scala
+++ b/geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/NormalizedDimension.scala
@@ -17,7 +17,7 @@ trait NormalizedDimension {
     require(x >= min && x <= max, s"Value out of bounds ([$min $max]): $x")
     math.ceil((x - min) / (max - min) * precision).toInt
   }
-  def denormalize(x: Double): Double = (x / (precision + 1)) * (max - min) + min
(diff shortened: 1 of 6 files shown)


geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/NormalizedDimension.scala
```diff
@@ -17,7 +17,7 @@ trait NormalizedDimension {
     require(x >= min && x <= max, s"Value out of bounds ([$min $max]): $x")
     math.ceil((x - min) / (max - min) * precision).toInt
   }
-  def denormalize(x: Double): Double = (x / precision) * (max - min) + min
+  def denormalize(x: Double): Double = (x / (precision + 1)) * (max - min) + min
```
tkunicki: I checked this, round-tripped, across [0,2^31) and it looks like the `precision +1` should be `precision + 0.5d`

I would recommend that we add unit tests to validate `n == normalize(denormalize(n))` the `0` and `precision`

aheyne: Updated

tkunicki:
@aheyne I submitted a PR against this branch on your fork.  I cleaned up the impl and added some unit tests to explain/characterize the changes.

PR #1448, opened 2017-03-27
HBase kerberos

geoHeil:
conditionally load HBase-site.xml kerberos configuration for geomesa.


pom.xml
```diff
@@ -76,11 +76,11 @@
...
         <accumulo.version>1.7.2</accumulo.version>
-        <zookeeper.version>3.4.5</zookeeper.version>
+        <zookeeper.version>3.4.6</zookeeper.version>
         <thrift.version>0.9.1</thrift.version>
-        <hadoop.version>2.5.2</hadoop.version>
-        <spark.version>2.0.0</spark.version>
-        <hbase.version>1.2.3</hbase.version>
+        <hadoop.version>2.7.3</hadoop.version>
+        <spark.version>2.1.0</spark.version>
+        <hbase.version>1.1.2</hbase.version>
```
elahrvivaz: we'll probably need to roll these back for the official build, as we have to open eclipse CQs whenever a dependency version changes. @jnh5y thoughts?

docs/user/hbase/install.rst
```diff
@@ -114,4 +114,20 @@ The HBase data store requires the configuration file ``hbase-site.xml`` to be on
 be accomplished by placing the file in ``geoserver/WEB-INF/classes`` (you should make the directory if it
 doesn't exist).
 
+.. note::
```
elahrvivaz: we should probably put this in a separate section on kerberos (maybe just a subsection), since it applies to more than just geoserver.

elahrvivaz:
@geoHeil I think you need to sign the eclipse CLA, outlined here:
https://github.com/locationtech/geomesa/blob/master/CONTRIBUTING.md
thanks,

docs/user/hbase/install.rst
```diff
@@ -114,4 +114,20 @@ The HBase data store requires the configuration file ``hbase-site.xml`` to be on
...
+
+    To use geomesa in a kerberized environment add the following ``hbase.geomesa.keytab`` and ``hbase.geomesa.principal`` to ``hbase-site.xml``.
+    .. code::
+    <property>
+         <name>hbase.geomesa.principal.principal</name>
```
mdzimmerman: Extra ".principal" in the property name?

docs/user/hbase/install.rst
mdzimmerman: Yeah. How about a separate section at the end of this page called "Configuration"?

geoHeil:
Your feedback sounds good. I will try to incorporate that soon. Regarding
testing :

Geomesa is already using HBaseTestingUtility here
https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala#L36-L42
but it does not provide org.apache.hadoop.minikdc.MiniKdc (in 1.1.2)
outlined from
https://hbase.apache.org/testapidocs/org/apache/hadoop/hbase/HBaseTestingUtility.html
(2.0.0). What other solution would allow to set up a mini hbase + kdc?

What are your thoughts about this?
Matthew Zimmerman <notifications@github.com> schrieb am Di. 28. März 2017
um 16:25:

pom.xml
geoHeil: I will roll these back. if required but spark, Hadoop, and Zookeeper could be upgraded in the near future.

elahrvivaz:
I wouldn't worry about unit tests for kerberos, unless it's fall-down easy. As long as you have verified that it works in a production setting, I think that's good enough for now.
We really need to set up some integration tests, that bootstrap a real environment. But I don't think we have the resources to maintain something like that at the moment.
@jnh5y correct me if you disagree with any of that

docs/user/hbase/install.rst
geoHeil: You mean `hbase.geomesa.principal.principal`?

pom.xml
elahrvivaz: yeah, we like to have recent versions, but there's a cost and delay to updating due to eclipse licensing. For most things (e.g. hadoop), the actual runtime version doesn't really matter, as it's a `provided` dependency. For hbase it's slightly trickier since we sometimes bundle it into uber-jars.

geoHeil:
@elahrvivaz the ECA should be signed: There is a valid ECA on file for georg.kf.heiler@googlemail.com, but the bot is still failing the ip-validation.

Regarding testing: so far we could only test it in a limited development environment but we will try to perform more testing during the next week.

geoHeil:
In order to merge the conflicting changes I ended up recreating my changes (just did a clean branch). I don't know why the ip-validation is failing. I actually ended up creating 2 eclipse accounts one with my main and one with my GitHub e-mail address. @elahrvivaz please can you tell me what is wrong with my eclipse ip-validation check.

jnh5y:
@geoHeil I'll have to check with Eclipse about why their webhook is failing.  Sorry for the troubles.  As a note, https://projects.eclipse.org/user/cla/validate is saying that  georg.kf.heiler@googlemail.com and georg.kf.heiler@gmail.com are ok.  If you just created an Eclipse account for the second, things may have been out of sync when the webhook ran.  

This kind of race condition can happen the first time or so, and we won't let it prevent us from merging a PR.  That said, we do want to find a little time to review the work.

geoHeil:
I created a clean up version of this here https://github.com/locationtech/geomesa/pull/1504

PR #1456, opened 2017-04-04
GEOMESA-1757 Enable stats on transformed attribute fields

jnh5y:


elahrvivaz:
replaced with #1461

PR #1457, opened 2017-04-04
GEOMESA-1747, GEOMESA-1748 Arrow integration

elahrvivaz:
* Arrow file-based data store implementation
* Arrow accumulo iterators for push-down processing
* Arrow WPS process and query response


elahrvivaz:
single place for arrow work. Currently requires building https://github.com/apache/arrow/pull/452 and https://github.com/ccri/arrow-geometry

build/cqs.tsv
```diff
@@ -66,6 +71,9 @@ io.dropwizard.metrics:metrics-core	3.1.2	compile
 io.dropwizard.metrics:metrics-ganglia	3.1.2	compile
 io.dropwizard.metrics:metrics-graphite	3.1.2	compile
 io.netty:netty-all	4.0.23.Final	compile
+io.netty:netty-buffer	4.0.41.Final	compile
+io.netty:netty-common	4.0.41.Final	compile
+io.netty:netty-handler	4.0.41.Final	compile
```
jnh5y: Any idea if we can manage these versions to something that's IP approved?

build/cqs.tsv
```diff
@@ -45,6 +49,7 @@ com.thoughtworks.paranamer:paranamer	2.6	compile
 com.typesafe:config	1.2.1	compile
 com.typesafe.scala-logging:scala-logging_2.11	3.1.0	compile
 com.vividsolutions:jts	1.13	compile
+com.vlkan:flatbuffers	1.2.0-3f79e055	compile
```
jnh5y: Are these definitely gonna be in the final version?  I'm wondering if we can queue up some CQs.

elahrvivaz: I think so - I'm still working on the dependency version management though

build/cqs.tsv
elahrvivaz: I think I tried to use the same version we already have, but it died... I am planning to go back and check all these again

build/cqs.tsv
```diff
@@ -189,12 +200,7 @@ org.json4s:json4s-native_2.11	3.2.10	compile
 org.locationtech.sfcurve:sfcurve-api_2.11	0.2.0	compile
 org.locationtech.sfcurve:sfcurve-zorder_2.11	0.2.0	compile
 org.locationtech.spatial4j:spatial4j	0.6	compile
-org.mortbay.jetty:jetty	6.1.26	compile
-org.mortbay.jetty:jetty-sslengine	6.1.26	compile
 org.mortbay.jetty:jetty-util	6.1.26	compile
-org.mortbay.jetty:jsp-2.1	6.1.14	compile
-org.mortbay.jetty:jsp-api-2.1	6.1.14	compile
-org.mortbay.jetty:servlet-api-2.5	6.1.14	compile
```
elahrvivaz: turns out we didn't actually need these - bumping the maven dependency plugin version fixed a bug in how dependency trees are calculated

elahrvivaz:
closing in favor of #1481 and #1482

PR #1458, opened 2017-04-04
GEOMESA-1758 Transform simple feature wrapper

elahrvivaz:

### DIFF
--- a/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/TransformSimpleFeature.scala
+++ b/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/TransformSimpleFeature.scala
@@ -1,10 +1,10 @@
-/*******************************************************************************
- * Copyright (c) 2013-2017 Commonwealth Computer Research, Inc.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Apache License, Version 2.0
- * which accompanies this distribution and is available at
- * http://www.opensource.org/licenses/apache2.0.php.
- ******************************************************************************/
+/***********************************************************************
+* Copyright (c) 2013-2017 Commonwealth Computer Research, Inc.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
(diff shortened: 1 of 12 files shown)


elahrvivaz:
Allows us to read transformed attributes lazily - this can be used when the transform result is going to be discarded and not returned.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/TransformSimpleFeature.scala
```diff
@@ -0,0 +1,118 @@
...
+  override def validate() = throw new NotImplementedError
+
+  override def toString = s"TransformSimpleFeature:$getID"
+}
+
+object TransformSimpleFeature {
+
+  import scala.collection.JavaConversions._
+
+  def apply(sft: SimpleFeatureType, transformSchema: SimpleFeatureType, transforms: String): TransformSimpleFeature = {
```
jnh5y: I wonder if it'd make sense to expose an easier constructor where one could pass an SFT and property names + transforms and get out the transform schema, transforms, etc.

elahrvivaz: we could, but i think we'd have to move it into index-api - what do you think?

elahrvivaz: the use-case I had was: https://github.com/elahrvivaz/geomesa/blob/6d9229093f23ba53bb1156df5bf03800f6fbb840/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala#L51-L52

jnh5y: Why would it have to move to index-api?

Meh, if I can figure out a way to sort out the call I'm thinking off, I'll add it to this class.

elahrvivaz: the code to create the transform schema and strings from a query is in index-api

PR #1459, opened 2017-04-04
GEOMESA-1759 Concatenating closeable iterator, close with resources

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/CloseableIterator.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/CloseableIterator.scala
@@ -17,7 +17,7 @@ import org.opengis.feature.`type`.FeatureType
 import org.opengis.feature.simple.SimpleFeature
 
 import scala.annotation.tailrec
-import scala.collection.Iterator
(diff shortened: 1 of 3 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/CloseableIterator.scala
```diff
@@ -62,12 +62,23 @@ trait CloseableIterator[+A] extends Iterator[A] with Closeable {
 
   override def filter(p: A => Boolean): CloseableIterator[A] = CloseableIterator(super.filter(p), self.close())
 
+  override def ++[B >: A](that: => GenTraversableOnce[B]): CloseableIterator[B] = {
```
jnh5y: Ok, hate to do it... can you add a quick unit test to CloseableIteratorTest?  

We could use an intern.

elahrvivaz: sure

elahrvivaz: added - and fixed a bug :)

PR #1460, opened 2017-04-05
GEOMESA-1751 Docs: configuring Zeppelin for Spark/Spark SQL

mdzimmerman:

### DIFF
--- a/docs/user/spark/zeppelin.rst
+++ b/docs/user/spark/zeppelin.rst
@@ -1,0 +1,57 @@
+Deploying GeoMesa Spark with Zeppelin
+=====================================
+
+Apache `Zeppelin`_ is a web-based notebook server for interactive data analytics, which includes built-in support for `Spark`_.
+
+.. note::
+
+    The instructions below have been tested with Zeppelin version |zeppelin_version|.
+
+Installing Zeppelin
+-------------------
+
+Follow the `Zeppelin installation instructions`_, as well as the instructions for `configuring the Zeppelin Spark interpreter`_.
+
+Configuring Zeppelin with GeoMesa
+---------------------------------
+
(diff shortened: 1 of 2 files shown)


docs/user/spark/zeppelin.rst
```diff
@@ -0,0 +1,57 @@
...
+     a. the full local path to the ``geomesa-accumulo-spark-runtime_2.11-$VERSION.jar`` described above, or
+     b. the Maven groupId:artifactId:version coordinates (``org.locationtech.geomesa:geomesa-accumulo-spark-runtime_2.11:$VERSION``)
+#. Click "Save". When prompted by the pop-up, click to restart the Spark interpreter.
+
+It is not necessary to restart Zeppelin.
+
+Data Plotting
+-------------
+
+Zeppelin provides built-in tools for visualizing quantitative data, which can be invoked by prepending "%table\n" to tab-separated output (see the `Zeppelin table display system`_). For example, the following method may be used to print a ``DataFrame`` via this display system:
```
aheyne: What's with the "\n"? It might need to be escaped if it's intended because it's displaying like 
>  "%tablen"

docs/user/spark/zeppelin.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. code-block:: scala
+
+    def printDF(df: DataFrame) = {
+      val dfc = df.collect
+      println("%table")
```
aheyne: Was the "\n" supposed to be here?

mdzimmerman: No, because the ``println()`` will add the newline needed.

docs/user/spark/zeppelin.rst
mdzimmerman: You need to have the string "%table" followed by a newline, so that's a good catch!

PR #1463, opened 2017-04-07
GEOMESA-1763 Implement Descriptive Stats

tkunicki:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/stats/KryoLazyStatsIteratorProcessTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/stats/KryoLazyStatsIteratorProcessTest.scala
@@ -82,19 +82,28 @@ class KryoLazyStatsIteratorProcessTest extends Specification with TestWithDataSt
       forall(0 until 5)(rh.count(_) mustEqual 30)
     }
 
-    "work with the Stats stat" in {
-      val results = statsIteratorProcess.execute(fs.getFeatures(query), "Stats(attr)", encode = true)
+    "work with the DescriptiveStats stat" in {
+      val results = statsIteratorProcess.execute(fs.getFeatures(query), "DescriptiveStats(attr)", encode = true)
       val sf = results.features().next
 
       val rh = decodeStat(sf.getAttribute(0).asInstanceOf[String], sft).asInstanceOf[DescriptiveStats]
-//      rh.length mustEqual 5
-//      forall(0 until 5)(rh.count(_) mustEqual 30)
-      rh.min(0) mustEqual 0
       rh.count mustEqual 150
-      rh.mean(0) must beCloseTo(149.0, 0.1)
-      rh.ssdev(0) must beCloseTo(86.46, 0.5)
-      rh.sskew(0) must beCloseTo(0, 0.2)
-      rh.skurt(0) must beCloseTo(1.8, 0.5)
+      rh.bounds(0) mustEqual (0, 298)
+      rh.mean(0) must beCloseTo(149.0, 1e-9)
+      rh.populationVariance(0) must beCloseTo(7499.666666666667, 1e-9)
+      rh.populationStandardDeviation(0) must beCloseTo(86.60061585616275, 1e-9)
+      rh.populationSkewness(0) must beCloseTo(0.0, 1e-9)
+      rh.populationKurtosis(0) must beCloseTo(1.7998933285923824, 1e-9)
+      rh.populationExcessKurtosis(0) must beCloseTo(-1.2001066714076176, 1e-9)
+      rh.sampleVariance(0) must beCloseTo(7550.0, 1e-9)
+      rh.sampleStandardDeviation(0) must beCloseTo(86.89073598491383, 1e-9)
+      rh.sampleSkewness(0) must beCloseTo(0.0, 1e-9)
+      rh.sampleKurtosis(0) must beCloseTo(1.859889772878795, 1e-9)
+      rh.sampleExcessKurtosis(0) must beCloseTo(-1.140110227121205, 1e-9)
+      rh.populationCovariance(0) must beCloseTo(7499.666666666667, 1e-9)
+      rh.populationCorrelation(0) must beCloseTo(1.0, 1e-9)
+      rh.sampleCovariance(0) must beCloseTo(7550.0, 1e-9)
+      rh.sampleCorrelation(0) must beCloseTo(1.0, 1e-9)
     }
 
     "work with multiple stats at once" in {
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/stats/KryoLazyStatsIteratorProcessTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/stats/KryoLazyStatsIteratorProcessTest.scala
@@ -82,6 +82,21 @@ class KryoLazyStatsIteratorProcessTest extends Specification with TestWithDataSt
       forall(0 until 5)(rh.count(_) mustEqual 30)
     }
 
+    "work with the Stats stat" in {
+      val results = statsIteratorProcess.execute(fs.getFeatures(query), "Stats(attr)", encode = true)
+      val sf = results.features().next
+
+      val rh = decodeStat(sf.getAttribute(0).asInstanceOf[String], sft).asInstanceOf[DescriptiveStats]
+//      rh.length mustEqual 5
+//      forall(0 until 5)(rh.count(_) mustEqual 30)
+      rh.min(0) mustEqual 0
+      rh.count mustEqual 150
+      rh.mean(0) must beCloseTo(149.0, 0.1)
+      rh.ssdev(0) must beCloseTo(86.46, 0.5)
+      rh.sskew(0) must beCloseTo(0, 0.2)
+      rh.skurt(0) must beCloseTo(1.8, 0.5)
+    }
+
     "work with multiple stats at once" in {
       val results = statsIteratorProcess.execute(fs.getFeatures(query),
         "MinMax(attr);IteratorStackCount();Enumeration(an_id);Histogram(an_id,5,10,14)", encode = true)
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
@@ -8,8 +8,6 @@
 
 package org.locationtech.geomesa.utils.stats
 
-import org.ejml.data.DenseMatrix64F
-import org.ejml.ops.CommonOps
 import org.ejml.simple.SimpleMatrix
 import org.locationtech.geomesa.utils.stats.SimpleMatrixUtils._
 import org.opengis.feature.simple.SimpleFeature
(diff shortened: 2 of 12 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatParser.scala
```diff
@@ -102,6 +102,13 @@ private class StatParser extends BasicParser {
     }
   }
 
+  private def stats: Rule1[Stat] = rule {
+    "Stats(" ~ string ~ ")" ~~> { attributes =>
```
elahrvivaz: can we change this to 'DescriptiveStats', and change the associated rule name? 'stats' is pretty generic, since they're all stats...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerializer.scala
```diff
@@ -71,53 +71,94 @@ object KryoStatSerializer {
...
+      case s: MinMax[_]           => output.writeByte(MinMaxByte);          writeMinMax(output, sft, s)
+      case s: EnumerationStat[_]  => output.writeByte(EnumerationByte);     writeEnumeration(output, sft, s)
+      case s: TopK[_]             => output.writeByte(TopKByte);            writeTopK(output, sft, s)
+      case s: Histogram[_]        => output.writeByte(HistogramByte);       writeHistogram(output, sft, s)
+      case s: Frequency[_]        => output.writeByte(FrequencyByte);       writeFrequency(output, sft, s)
+      case s: Z3Histogram         => output.writeByte(Z3HistogramByte);     writeZ3Histogram(output, sft, s)
+      case s: Z3Frequency         => output.writeByte(Z3FrequencyByte);     writeZ3Frequency(output, sft, s)
+      case s: IteratorStackCount  => output.writeByte(IteratorStackByte);   writeIteratorStackCount(output, s)
+      case s: SeqStat             => output.writeByte(SeqStatByte);         writeSeqStat(output, sft, s)
+      case s: DescriptiveStats => output.writeByte(DescriptiveStatByte);   writeDescriptiveStats(output, sft, s)
```
elahrvivaz: [picky] alignment

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerializer.scala
```diff
@@ -71,53 +71,94 @@ object KryoStatSerializer {
...
+    val size = input.readInt(true)
+    val attributes = for(_ <- 0 until size) yield input.readInt(true)
+
+    val stats = if (immutable) {
+      new DescriptiveStats(attributes) with ImmutableStat
+    } else {
+      new DescriptiveStats(attributes)
+    }
+
+    stats._min.getMatrix.setData((for(_ <- 0 until size) yield input.readDouble).toArray)
```
elahrvivaz: how big is size likely to be? might be worthwhile to pre-allocate the arrays and then populate them in a while loop

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -112,6 +114,34 @@ trait Stat {
...
+    def serialize(f: jFloat, t: Type, jsc: JsonSerializationContext): JsonElement = f match {
+      /* NaN check, use null to mirror existing behavior for missing/invalid values */
+      case f if jFloat.isNaN(f) => JsonNull.INSTANCE
+      case f if f == jFloat.NEGATIVE_INFINITY => new JsonPrimitive("Infinity")
+      case f if f == jFloat.POSITIVE_INFINITY => new JsonPrimitive("+Infinity")
+      case _ => new JsonPrimitive(f)
+    }
+  }
+
+  val JSON: Gson = new GsonBuilder()
```
elahrvivaz: just double-checking, Gson is thread-safe right?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/stats/KryoLazyStatsIteratorProcessTest.scala
```diff
@@ -82,6 +82,21 @@ class KryoLazyStatsIteratorProcessTest extends Specification with TestWithDataSt
       forall(0 until 5)(rh.count(_) mustEqual 30)
     }
 
+    "work with the Stats stat" in {
+      val results = statsIteratorProcess.execute(fs.getFeatures(query), "Stats(attr)", encode = true)
+      val sf = results.features().next
+
+      val rh = decodeStat(sf.getAttribute(0).asInstanceOf[String], sft).asInstanceOf[DescriptiveStats]
+//      rh.length mustEqual 5
+//      forall(0 until 5)(rh.count(_) mustEqual 30)
```
elahrvivaz: remove commented out code

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+
+  private[stats] var _count: Long = _
+  private[stats] val _min: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _max: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _sum: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _mean: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _m2n: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _m3n: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _m4n: SimpleMatrix = new SimpleMatrix(size, 1)
+  private[stats] val _c2: SimpleMatrix = new SimpleMatrix(size, size)
```
elahrvivaz: [picky] we don't usually use `_` for private variables

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+    _m2n.set(that._m2n)
+    _m3n.set(that._m3n)
+    _m4n.set(that._m4n)
+    _c2.set(that._c2)
+  }
+
+  def count: Long = _count
+
+  def min: Array[Double] = (if (isEmpty) _max else _min).getMatrix.data.clone()
+  def max: Array[Double] = (if (isEmpty) _min else _max).getMatrix.data.clone()
```
elahrvivaz: does this copy the data array? even though it exposes mutable state might be good to avoid depending how often this is called and how big the data is

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+  def bounds: Array[(Double, Double)] = min.zip(max)
+
+  def sum: Array[Double] = _sum.getMatrix.data.clone()
+
+  def mean: Array[Double] = requireCount(1) { _mean }
+
+  def m2: Array[Double] = requireCount(1) { _m2n / _count }
```
elahrvivaz: are these functions obvious enough that people would know what they mean? maybe we can add some brief scaladocs

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+  private def requireCount(count: Int, length: Int = size)(op: => SimpleMatrix):Array[Double] =
+    if (_count < count) Array.fill(length)(Double.NaN) else op.getMatrix.data.clone()
+
+  override def observe(sf: SimpleFeature): Unit =
+    observe(attributes.map(sf.getAttribute(_).asInstanceOf[Number]).toArray)
+
+  def observe(values: Array[Number]): Unit = {
+    if (values.forall(_ != null)) {
+      val values_d = values.map(_.doubleValue)
+      if (values_d.forall(v => v == v)) {
```
elahrvivaz: i assume this is testing for NaN? Can you add a comment to that effect, or just use `Double.isNaN`

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+  private[stats] val _c2: SimpleMatrix = new SimpleMatrix(size, size)
+
+  clear()
+
+  override def clear(): Unit = {
+    _count = 0
+    _min.set(java.lang.Double.MAX_VALUE)
+    _max.set(0d - java.lang.Double.MAX_VALUE)
+    _sum.set(0d)
+    _mean.set(1d)
```
elahrvivaz: curious if there's a reason mean starts at 1 vs 0?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+
+  def pcov: Array[Double] = requireCount(2, size_squared) { _c2 / _count}
+  def pcor: Array[Double] = requireCount(2, size_squared) {
+    val mn2_sqrt = _m2n ** 0.5; (_c2 / (mn2_sqrt |*| mn2_sqrt.T)).diag(1.0) }
+
+  def scov: Array[Double] = requireCount(2, size_squared) { _c2 / (_count -1 ) }
+  def scor: Array[Double] = pcor /* population and sample calculations are equal w/ df term cancellation */
+
+  private def requireCount(count: Int, length: Int = size)(op: => SimpleMatrix):Array[Double] =
+    if (_count < count) Array.fill(length)(Double.NaN) else op.getMatrix.data.clone()
```
elahrvivaz: so is this effectively saying if we don't have enough data, just return an Array of NaN, otherwise calculate the op?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+  def updateMinMax(mins: SimpleMatrix, maxs: SimpleMatrix, values: SimpleMatrix): Unit = {
+    val n = values.getNumElements
+    var i = 0
+    while (i < n) {
+      val v = values.get(i)
+      val min = mins.get(i)
+      val max = maxs.get(i)
+      if (v < min) {
+        mins.set(i, v)
+      } else if (v > max) {
```
elahrvivaz: since min starts at Double.MaxValue, and max starts at negative max value, should this be two separate ifs instead of an else if? Seems like if we evaluated e.g. [10, 9, 8, 7, 6] max would never get updated.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+
+          val B = values_v - _mean
+          _m3n += A * (B * delta * (n - 2d) - _m2n * 3d)
+          _m2n += delta * B
+
+          _c2 += (delta |*| delta.T * (n_i * r))
+
+        } else {
+          _count = 1
+          _min.set(values_v)
```
tkunicki: @elahrvivaz this is portion that allows min max update to bail if min is met

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatParser.scala
tkunicki: I'll leave this to @jnh5y , he's the DSL owner

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatSerializer.scala
tkunicki: size is determined by number of attributes.  in reality, the arrays are preallocated w/ underlying EJML construct.  This call throws them away and replaces.  I wasn't too concerned considering that serialization isn't something that will happen often.  But, always happy to optimize...

elahrvivaz: oh, if size is just the number of attributes then it probably doesn't matter.
if we ever end up writing these stats to our metadata table we might want to revisit since they get serialized a lot then

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
tkunicki: I'd verified this, from the GSON javadoc ["This is the main class for using Gson. Gson is typically used by first constructing a Gson instance and then invoking toJson(Object) or fromJson(String, Class) methods on it. Gson instances are Thread-safe so you can reuse them freely across multiple threads."](https://google.github.io/gson/apidocs/com/google/gson/Gson.html)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: Previous implementations had accessors and `_` was used to differentiate `var` from accessors, this might not be the case any longer...  can adjust if true

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: it's an algorithm requirement due to how the mean is incrementatly updated

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: it does copy, but only when method is called with should typically be once per aggregate calculation.  i really don't want to leak mutable constructs...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: these protect from outputting data where `count` won't support calculation (and divide by 0)

tkunicki: and to answer your Q, yes.  return NaN w/o enough data...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: i can comment...  it's in the JVM spec!

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: this is avoided w/ the special case in _count < 1 in observe (dropped comment for you where this occurs)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
```diff
@@ -0,0 +1,358 @@
...
+        "population_covariance" -> pcov,
+        "population_correlation" -> pcor,
+        "sample_covariance" -> scov,
+        "sample_correlation" -> scor)
+    }
+    Stat.JSON.toJson(map.asJava)
+  }
+}
+
+object SimpleMatrixUtils {
```
jnh5y: Does this object/class deserve its own file?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/StatParser.scala
jnh5y: Sure, +1 on the name change.  (To be fair to Tom, I'm the one who wired this up initially, so the poor name choice is all me;).)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/DescriptiveStats.scala
tkunicki: it could...  everything but the min/max stuff is reusable w/o confusion.  will others want to use EJML w/ the operator overloads?

jnh5y: We have made it this far without, but on the other hand, as we do more data science...  

We can refactor it later.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/process/stats/KryoLazyStatsIteratorProcessTest.scala
```diff
@@ -82,6 +82,21 @@ class KryoLazyStatsIteratorProcessTest extends Specification with TestWithDataSt
...
+      rh.min(0) mustEqual 0
+      rh.count mustEqual 150
+      rh.mean(0) must beCloseTo(149.0, 0.1)
+      rh.ssdev(0) must beCloseTo(86.46, 0.5)
+      rh.sskew(0) must beCloseTo(0, 0.2)
+      rh.skurt(0) must beCloseTo(1.8, 0.5)
```
jnh5y: I had tossed this test in.  Can the error bounds of 0.5 and 0.2 be tightened up?

PR #1469, opened 2017-04-12
GEOMESA-1771 Pushdown Z3Iterator in HBase

anthonyccri:
* Refactored Z3 iterator to enable both Accumulo and HBase
  push down

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z3Iterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z3Iterator.scala
@@ -8,15 +8,15 @@
 
 package org.locationtech.geomesa.accumulo.iterators
 
-import com.google.common.primitives.{Longs, Shorts}
 import org.apache.accumulo.core.client.IteratorSetting
 import org.apache.accumulo.core.data.{ByteSequence, Key, Value, Range => AccRange}
 import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
 import org.apache.hadoop.io.Text
 import org.locationtech.geomesa.accumulo.index.legacy.z3.Z3IndexV2
 import org.locationtech.geomesa.curve.Z3SFC
+import org.locationtech.geomesa.index.filters.Z3Filter
 import org.locationtech.sfcurve.zorder.Z3
-import org.opengis.feature.simple.SimpleFeatureType
+
 
 class Z3Iterator extends SortedKeyValueIterator[Key, Value] {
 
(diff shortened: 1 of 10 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z3Iterator.scala
```diff
@@ -8,88 +8,76 @@
...
   }
 
+  private val text = new Text
   private def findTop(): Unit = {
     topKey = null
     topValue = null
     while (source.hasTop) {
-      if (inBounds(source.getTopKey)) {
+      val row = source.getTopKey.getRow(text)
+      val bytes = row.getBytes
```
elahrvivaz: looks like you could call `getRowData` which would avoid copying the row array - although it does create a new `ArrayByteSequence` instead. Dunno if that would be better...

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -128,7 +128,7 @@ trait HBaseFeatureIndex extends HBaseFeatureIndexType
     else {
       val table = TableName.valueOf(getTableName(sft.getTypeName, ds))
       val dedupe = hasDuplicates(sft, filter.primary)
-      val ScanConfig(hbaseFilters, toFeatures) = scanConfig(sft, filter, hints, ecql, dedupe, ds.remote)
+      val ScanConfig(hbaseFilters, toFeatures) = scanConfig(sft, filter, hints, ecql, dedupe)
```
elahrvivaz: do we not want to allow for running without remote filters anymore?

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
```diff
@@ -48,12 +48,13 @@ class HBaseDataStoreTest extends Specification with LazyLogging {
     "work with points" in {
       val typeName = "testpoints"
 
+      System.setProperty("geomesa.hbase.remote.filtering", "true")
```
elahrvivaz: should be able to use `SystemProperty("geomesa.hbase.remote.filtering").threadLocal.set("true")` (or something close to that) to avoid affecting other tests

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
anthonyccri: I removed that ability.  I guess it wouldn't hurt if we still had it.  Hmmm.

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
anthonyccri: Actually, since I removed the ability to run without remote, I need to get rid of this.  But see the previous comment.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
elahrvivaz: we're going to still need it for bigtable unless we add some more conditional checks there

anthonyccri:
This is OBE as a more thorough PR is coming

PR #1473, opened 2017-04-14
GEOMESA-1774 Convert all Stat JSON serialization to GSON

tkunicki:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumerationStat.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumerationStat.scala
@@ -10,6 +10,7 @@ package org.locationtech.geomesa.utils.stats
 
 import org.opengis.feature.simple.SimpleFeature
 
+import scala.collection.immutable.ListMap
 import scala.reflect.ClassTag
 
 /**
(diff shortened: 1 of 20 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
```diff
@@ -79,7 +82,14 @@ trait Stat {
...
    */
-  def toJson: String
+  def toJson: String = Stat.JSON.toJson(toJsonObject)
+
+  /**
+    * Returns a representation of the stat to be serialized
+    *
+    * @return stat as a json serializable object
+    */
+  def toJsonObject: Any
```
jnh5y: This might be a bigger question, but is there a 'closer' type bound we can use here?

tkunicki: Yeah, it could be a union of {primitives( and boxed types), Date, and String; Maps and Seq containing previous type or themselves}. Thoughts on trying to express that?  In reality Any is appropriate as long as there's a registered serializer.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumerationStat.scala
```diff
@@ -60,11 +61,7 @@ class EnumerationStat[T](val attribute: Int)(implicit ct: ClassTag[T]) extends S
   override def +=(other: EnumerationStat[T]): Unit =
     other.enumeration.foreach { case (key, count) => enumeration(key) += count }
 
-  override def toJson: String = {
-    if (enumeration.isEmpty) { "{ }" } else {
-      enumeration.toSeq.sortBy(_.toString).map { case (k, v) => s""""${stringify(k)}" : $v""" }.mkString("{ ", ", ", " }")
-    }
-  }
+  override def toJsonObject = if (enumeration.isEmpty) Map.empty else ListMap(enumeration.toSeq.sortBy(_.toString):_*)
```
elahrvivaz: presumably ListMap is used to preserve iteration order?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Stat.scala
elahrvivaz: maybe just add that to the scala docs. Presumably in practice it would have to be either a Map or a List to create a valid json object

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/EnumerationStat.scala
tkunicki: yes

PR #1483, opened 2017-04-19
1.2 - GEOMESA-1779 AttributeIndexJob Copy Visibilities

jahhulbert-ccri:


jahhulbert-ccri:
probably should test the upgrade of indexes too but i'll do that later for 1.2.8 release :)

elahrvivaz:
would be good to add to the tools command too: https://github.com/locationtech/geomesa/blob/1.2/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/accumulo/commands/AddIndexCommand.scala#L100

PR #1485, opened 2017-04-20
GEOMESA-1692 Using second scan option instead of full table scan for …

elahrvivaz:
…join queries in Accumulo spark

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/attribute/AttributeQueryableIndex.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/attribute/AttributeQueryableIndex.scala
@@ -36,6 +36,7 @@ import scala.util.Try
 
 trait AttributeQueryableIndex extends AccumuloFeatureIndex with LazyLogging {
 
+  import AttributeIndex.requiresJoin
   import AttributeQueryableIndex.attributeCheck
   import org.locationtech.geomesa.index.conf.QueryHints.RichHints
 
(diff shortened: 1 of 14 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/attribute/AttributeQueryableIndex.scala
```diff
@@ -235,12 +236,23 @@ trait AttributeQueryableIndex extends AccumuloFeatureIndex with LazyLogging {
...
+      }
+    }
+    indexedAttributes.flatMap { case (attribute, coverage) =>
       val (primary, secondary) = FilterExtractingVisitor(filter, attribute, sft, attributeCheck)
-      if (primary.isDefined) {
+      lazy val joinCheck = coverage match {
+        case IndexCoverage.FULL => true
+        case IndexCoverage.JOIN => AllowJoinPlans.get || !requiresJoin(sft, attribute, secondary, transform)
+      }
+      if (primary.isDefined && joinCheck) {
```
mdzimmerman: Just so I'm clear, ``joinCheck`` should return true IFF the FilterStrategy doesn't involve a join OR if the caller explicitly says join plans are OK? I had to read this a couple of times to get that. (That may be a me problem more than a problem with your PR.)

It would be nice (i.e. more readable?) if the logic for this wasn't split between here and the ``requiresJoin`` method, but I know you also need that functionality in ``getCost``, so this might be the best solution.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/AccumuloJobUtils.scala
```diff
@@ -77,8 +79,8 @@ object AccumuloJobUtils extends LazyLogging {
 
       if (queryPlans.isEmpty) {
         EmptyPlan(FilterStrategy(fallbackIndex, None, Some(Filter.EXCLUDE)))
-      } else if (queryPlans.length > 1 || queryPlans.head.isInstanceOf[JoinPlan]) {
-        // this query has a join or requires multiple scans, which we can't execute from input formats
+      } else if (queryPlans.length > 1) {
+        // this query requires multiple scans, which we can't execute from input formats
```
mdzimmerman: Maybe we should change the comment to "*some* input formats"?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/attribute/AttributeQueryableIndex.scala
elahrvivaz: I'll add some clarifying comments, but yeah I think the logic needs to stay split for the reason you say.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/AccumuloJobUtils.scala
elahrvivaz: wilco

PR #1481, opened 2017-04-19
GEOMESA-1747 Arrow geometry abstractions, data store implementation

elahrvivaz:
Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #1487, opened 2017-04-24
GEOMESA-1747 Arrow geometry abstractions, data store implementation

elahrvivaz:

replaces #1481


elahrvivaz:
List of cqs:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13280
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13281
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13282
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13283
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13284
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13285
http://dev.eclipse.org/ipzilla/show_bug.cgi?id=13318

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowDataStore.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+import org.opengis.filter.Filter
+
+import scala.util.Try
+import scala.util.control.NonFatal
+
+class ArrowDataStore(val url: URL) extends ContentDataStore with FileDataStore
+    with HasGeoMesaMetadata[String] with HasGeoMesaStats {
+
+  import org.locationtech.geomesa.arrow.allocator
```
jnh5y: Unused import.:)

Don't worry about it though...

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowFeatureStore.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  override def canFilter: Boolean = true
+
+  override def canReproject: Boolean = false
+  override def canLimit: Boolean = false
+  override def canSort: Boolean = false
+  override def canRetype: Boolean = false
+  override def canLock: Boolean = false
+  override def canTransact: Boolean = false
```
jnh5y: Two things:

1.  I like seeing all these canOperations in one place; that's slick!
2.  These need to be add to ArrowFeatureStore.

As is, if I understand, if the input file is writable, then you'd get an ArrowFeatureStore and then filtering would be handled by the ContentFeatureSource.

elahrvivaz: good catch, thanks - I think anthony already fixed this in our wip branch actually.

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowFeatureStore.scala
```diff
@@ -0,0 +1,108 @@
...
+      override def next(): SimpleFeature = features.next()
+      override def close(): Unit = reader.close()
+    }
+  }
+
+  override def canFilter: Boolean = true
+
+  override def canReproject: Boolean = false
+  override def canLimit: Boolean = false
+  override def canSort: Boolean = false
```
jnh5y: Since we are talking about providing files which are sorted by *one* of the dimensions, I imagine adding an optimization here is possible.  

Is it worth adding a ticket for that?

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowDataStore.scala
elahrvivaz: I think this is actually used implicitly when creating the writers/readers, like on line 53

jnh5y: IntelliJ seemed to think it was not.  Either way;).

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowFeatureStore.scala
elahrvivaz: hmm, if we change this to true, we'd have to handle every sort case - i think we're using the sort structure to optimize filters but not necessarily to actually sort the results - what do you think?

jnh5y: Here's what I am thinking...

If the file's sort order aligns with the query, we implement the optimization.  Otherwise, we slap on a Sorting FS and let GeoTools do it for us.

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowFeatureStore.scala
elahrvivaz: fixed

geomesa-arrow/geomesa-arrow-datastore/src/main/scala/org/locationtech/geomesa/arrow/data/ArrowFeatureStore.scala
elahrvivaz: Yeah, i guess my thought is that it would be bad to sort unless you know that the file is already sorted, and if it's already sorted then you don't need to sort ;)
I'll open a ticket though

elahrvivaz: https://geomesa.atlassian.net/browse/GEOMESA-1817

PR #1496, opened 2017-04-28
GEOMESA-1793 Converter validation

jahhulbert-ccri:
Needs testing of two features:

1. Converter validators
2. Transaction aka Commit Mode - stream is processed all together or not.

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
@@ -8,13 +8,16 @@
 
 package org.locationtech.geomesa.convert
 
-import java.io.{Closeable, InputStream}
+import java.io.{Closeable, IOException, InputStream}
 import java.nio.charset.StandardCharsets
+import java.util.NoSuchElementException
 
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.commons.io.IOUtils
+import org.locationtech.geomesa.convert.ParseMode.ParseMode
 import org.locationtech.geomesa.convert.Transformers._
+import org.locationtech.geomesa.convert.ValidationMode.ValidationMode
 import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
@@ -30,17 +33,46 @@ trait Field {
 
 case class SimpleField(name: String, transform: Transformers.Expr) extends Field
 
-object StandardOptions {
-  val Validating = "options.validating"
-  val LineMode   = "options.line-mode"
+object StandardOption extends Enumeration {
+  type StandardOption = Value
+  @Deprecated val Validating = Value("validating")
+
+  val ValidatorsOpt     = Value("validators")
+  val ValidationModeOpt = Value("validation-mode")
+  val LineModeOpt       = Value("line-mode")
+  val ParseModeOpt      = Value("parse-mode")
+  val VerboseOpt        = Value("verbose")
+
+  implicit class StandardOptionValue(opt: Value) {
+    def path = s"options.$opt"
+  }
 }
+
+object ParseMode extends Enumeration {
+  type ParseMode = Value
+  val Incremental = Value("incremental")
+  val Batch       = Value("batch")
+  val Default     = Incremental
+}
+
+object ValidationMode extends Enumeration {
+  type ValidationMode = Value
+  val SkipBadRecords = Value("skip-bad-records")
+  val RaiseErrors    = Value("raise-errors")
+  val Default        = SkipBadRecords
+}
+
+case class ConvertParseOpts(parseMode: ParseMode,
+                            validator: SimpleFeatureValidator,
+                            validationMode: ValidationMode,
+                            verbose: Boolean)
 
 trait SimpleFeatureConverterFactory[I] {
   def canProcess(conf: Config): Boolean
   def buildConverter(sft: SimpleFeatureType, conf: Config): SimpleFeatureConverter[I]
 }
 
-abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureConverterFactory[I] {
+abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureConverterFactory[I] with LazyLogging {
 
   override def canProcess(conf: Config): Boolean =
     if (conf.hasPath("type")) conf.getString("type").equals(typeToProcess) else false
@@ -49,8 +81,8 @@ abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureCon
     val idBuilder = buildIdBuilder(conf)
     val fields = buildFields(conf)
     val userDataBuilder = buildUserDataBuilder(conf)
-    val validating = isValidating(conf)
-    buildConverter(sft, conf, idBuilder, fields, userDataBuilder, validating)
+    val parseOpts = getParsingOptions(conf, sft)
+    buildConverter(sft, conf, idBuilder, fields, userDataBuilder, parseOpts)
   }
 
   protected def typeToProcess: String
@@ -60,7 +92,7 @@ abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureCon
                                idBuilder: Expr,
                                fields: IndexedSeq[Field],
                                userDataBuilder: Map[String, Expr],
-                               validating: Boolean): SimpleFeatureConverter[I]
+                               parseOpts: ConvertParseOpts): SimpleFeatureConverter[I]
 
   protected def buildFields(conf: Config): IndexedSeq[Field] =
     conf.getConfigList("fields").map(buildField).toIndexedSeq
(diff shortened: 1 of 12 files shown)


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -31,16 +31,25 @@ trait Field {
...
 object StandardOptions {
-  val Validating = "options.validating"
+  @Deprecated val Validating = "options.validating"
+
+  val Validators = "options.validators"
   val LineMode   = "options.line-mode"
+  val CommitMode = "options.commit-mode"
+}
+
+object CommitMode {
```
elahrvivaz: might be better to make this an enum

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -119,44 +162,31 @@ trait SimpleFeatureConverter[I] extends Closeable {
   override def close(): Unit = {}
 }
 
+trait ValidatingConverter extends LazyLogging {
+
+  def validator: SimpleFeatureValidator
+
+  val validate: (SimpleFeature, EvaluationContext) => Boolean =
```
elahrvivaz: any reason this is a val and not just a function?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def name: String = "z-index"
+
+  override def init(sft: SimpleFeatureType): Unit = {
+    valid = {
+      import org.locationtech.geomesa.utils.geotools.Conversions._
+      import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
+
+      val dtgFn: (SimpleFeature) => Boolean = sft.getDtgIndex match {
+        case Some(dtgIdx) => (sf: SimpleFeature) => sf.getAttribute(dtgIdx) != null
```
elahrvivaz: maybe add a check for dates before the epoch?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,105 @@
...
+object ValidatorLoader {
+  def createValidator(names: Seq[String], sft: SimpleFeatureType): SimpleFeatureValidator = {
+    val validators = names.map {
+      case "none" => NoneValidator
+      case "has-geo" => new HasGeoValidator
+      case "has-dtg" => new HasDtgValidator
+      case "z-index" => new ZIndexValidator
+      case unk       => throw new IllegalArgumentException(s"Unknown validator $unk")
+    }
+    val composite = new CompositeValidator(validators)
```
elahrvivaz: might make things a little clearer if we only create the composite if there is more than 1

elahrvivaz:
docs? ;)

jahhulbert-ccri:
@elahrvivaz i changed it from the batch vs auto bc that's not really what it is

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
jahhulbert-ccri: +1

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
jahhulbert-ccri: ya

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
jahhulbert-ccri: probs

jahhulbert-ccri:
I'm still hacking on it but some example usage is in ValidatorTest.scala

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Date
+
+import org.locationtech.geomesa.curve.BinnedTime
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+
+trait SimpleFeatureValidator {
+  def name: String
+  def init(sft: SimpleFeatureType): Unit
+  def validate(sf: SimpleFeature): Boolean
+  def lastError: String
```
jahhulbert-ccri: I guess i could make this an option instead....

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -234,8 +315,16 @@ trait ToSimpleFeatureConverter[I] extends SimpleFeatureConverter[I] with LazyLog
     new EvaluationContextImpl(names, values, counter)
   }
 
-  override def processInput(is: Iterator[I], ec: EvaluationContext): Iterator[SimpleFeature] =
-    is.flatMap(i =>  processSingleInput(i, ec))
+  override def processInput(is: Iterator[I], ec: EvaluationContext): Iterator[SimpleFeature] = {
+    parseOpts.parseMode match {
+       case ParseMode.Incremental  =>
+         is.flatMap(i => processSingleInput(i, ec))
+       case ParseMode.Batch =>
```
elahrvivaz: maybe it would be good to have an option for batch size in this case?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,173 @@
...
+  override def name: String = "has-dtg"
+
+  override def init(sft: SimpleFeatureType): Unit = {
+    valid = {
+      lastErr = null
+      if (sft.getGeometryDescriptor != null) {
+        (sf: SimpleFeature) =>
+          val res = sf.getDefaultGeometry != null
+          if (!res) {
+            lastErr = "Null date attribute"
```
elahrvivaz: geom

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
elahrvivaz: from a functional perspective seems like the validate method should return the error directly - although maybe that would be worse performance wise. possibly we could throw an exception - try/catch should be costless in the happy path. alternatively return an Option or Try, checking against None would probably be pretty cheap...

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+
+  override def validate(sf: SimpleFeature): Boolean = valid(sf)
+
+  override def lastError: String = lastErr
+}
+
+object Z3IndexValidator {
+  val Z3MinDate = new Date(0)
```
elahrvivaz: it might be good to move this into the BinnedTime class - there was some discussion about supporting dates before the epoch, and we'd probably forget to come back and update this if we ever implemented that.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -86,8 +118,58 @@ abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureCon
...
-    if (conf.hasPath(StandardOptions.Validating)) conf.getBoolean(StandardOptions.Validating) else true
+  protected def getParsingOptions(conf: Config, sft: SimpleFeatureType): ConvertParseOpts = {
+    val verbose = if (conf.hasPath(StandardOption.VerboseOpt.path)) conf.getBoolean(StandardOption.VerboseOpt.path) else false
+    val opts = ConvertParseOpts(getParseMode(conf), getValidator(conf, sft), getValidationMode(conf), verbose = verbose)
+    logger.info(s"Using ParseMode ${opts.parseMode} with validation mode ${opts.validationMode} and validator ${opts.validator.name}")
+    opts
+  }
+  protected def getValidator(conf: Config, sft: SimpleFeatureType): SimpleFeatureValidator = {
+    val validators: Seq[String] =
+      if (conf.hasPath(StandardOption.Validating.path) && conf.hasPath(StandardOption.ValidatorsOpt.path)) {
```
jnh5y: What is going on with two types of configurations here?

jahhulbert-ccri: one is deprecated...but you shouldn't use both. i'll add a comment.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
jahhulbert-ccri: hmmm. that's interesting. perhaps this is better renamed "full file" or maybe we can do batch size of 0 in batch mode means full file? the idea was meant to have it be able to parse the whole file and insert no records. so it can still accomplish that. else you don't need batch

elahrvivaz: (not necessarily for this pr but) for ingestion it could be useful to hook batch size into a flush of the feature writer - there were some requests recently to know which records were actually ingested - this would be a way to do that. (of course, the pre-validation mitigates that to a large extent)

jahhulbert-ccri: yeah that's kinda the core issue with Accumulo writing. No transactions.

PR #1499, opened 2017-04-28
GEOMESA-1798 Fix ReTypingFeatureCollection Issue with some WPS Processors

aheyne:
Fix ReTypingFeatureCollection Issue with some WPS Processors

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
@@ -17,8 +17,6 @@ import org.geotools.feature.DefaultFeatureCollection
 import org.geotools.feature.visitor.{AbstractCalcResult, CalcResult, FeatureAttributeVisitor, FeatureCalc}
 import org.geotools.process.factory.{DescribeParameter, DescribeProcess, DescribeResult}
 import org.geotools.util.NullProgressListener
-import org.locationtech.geomesa.filter.FilterHelper
-import org.locationtech.geomesa.index.api.QueryPlanner
 import org.opengis.feature.Feature
 import org.opengis.feature.simple.SimpleFeature
 import org.opengis.filter.Filter
@@ -103,7 +101,7 @@ class QueryVisitor(features: SimpleFeatureCollection,
   }
 
   override def getExpressions: java.util.List[Expression] = {
-    // We return an empty list here to avoid ReTypingFeatureCollections.
+    // We return an empty list here to avoid ReTypingFeatureCollections. Happy day.
     // Calculating the actual list is difficult given transforms/projections/filters
     List[Expression]()
   }
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
@@ -19,10 +19,13 @@ import org.geotools.feature.DefaultFeatureCollection
 import org.geotools.feature.visitor.{AbstractCalcResult, CalcResult, FeatureAttributeVisitor, FeatureCalc}
 import org.geotools.process.factory.{DescribeParameter, DescribeProcess, DescribeResult}
 import org.geotools.util.NullProgressListener
+import org.locationtech.geomesa.filter.FilterHelper
 import org.opengis.feature.Feature
 import org.opengis.feature.simple.SimpleFeature
 import org.opengis.filter.Filter
 import org.opengis.filter.expression.Expression
+
+import scala.collection.mutable
 
 @DescribeProcess(
   title = "Geomesa Query",
@@ -103,7 +106,7 @@ class QueryVisitor(features: SimpleFeatureCollection,
   }
 
   override def getExpressions: java.util.List[Expression] = {
-    origSft.getAttributeDescriptors.map(ad => ff.property(ad.getLocalName)).toList
+    FilterHelper.propertyNames(filter, origSft).map(ff.property).toList
   }
 }
 
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
@@ -60,17 +60,6 @@ class QueryProcess extends LazyLogging {
 
     val visitor = new QueryVisitor(features, Option(filter).getOrElse(Filter.INCLUDE), arrayString)
     features.accepts(visitor, new NullProgressListener)
-
-    features match {
-      case rtfc: ReTypingFeatureCollection =>
-        if(ReTypingFeatureCollection.isTypeCompatible(visitor, rtfc.getSchema)) {
-          logger.info(s"Retypingfeature collection is type compatible with ${rtfc.getSchema}")
-        } else {
-          logger.info(s"Retypingfeature collection is not type compatible with ${rtfc.getSchema}")
-        }
-      case _ => logger.info("not retypingfeaturecollection")
-    }
-
     visitor.getResult.asInstanceOf[QueryResult].results
   }
 }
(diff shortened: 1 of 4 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
```diff
@@ -98,6 +102,9 @@ class QueryVisitor(features: SimpleFeatureCollection,
     source.getFeatures(query)
   }
 
+  override def getExpressions: java.util.List[Expression] = {
+    origSft.getAttributeDescriptors.map(ad => ff.property(ad.getLocalName)).toList
```
elahrvivaz: if there is a transform passed in we should use that right?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/stats/StatsIteratorProcess.scala
```diff
@@ -136,6 +140,11 @@ class StatsVisitor(features: SimpleFeatureCollection, statString: String, encode
     query.getHints.put(QueryHints.ENCODE_STATS, new java.lang.Boolean(encode))
     source.getFeatures(query)
   }
+
+  override def getExpressions: java.util.List[Expression] ={
+    val ff: FilterFactory = CommonFactoryFinder.getFilterFactory
+    origSft.getAttributeDescriptors.map(ad => ff.property(ad.getLocalName)).toList
```
elahrvivaz: we should pull out the attributes used in the stat string and return those

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/unique/UniqueProcess.scala
```diff
@@ -267,6 +271,10 @@ class AttributeVisitor(val features: SimpleFeatureCollection,
     SelfClosingIterator(source.getFeatures(query).features()).foreach(addValue)
     uniqueValues.toMap
   }
+
+  override def getExpressions: java.util.List[Expression] ={
+    origSft.getAttributeDescriptors.map(ad => ff.property(ad.getLocalName)).toList
```
elahrvivaz: this one should just need the attribute being queried on and any attributes in the filter I think

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
aheyne: I don't think so. If I understand it right, this is just telling GeoServer what properties we'll use so that it knows the feature collection doesn't need retyped.

elahrvivaz: right, but if there's a transform we'd only need the reduced set of attributes?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/unique/UniqueProcess.scala
aheyne: Fixed

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
aheyne: Fixed

aheyne:
Fixes pushed but haven't been tested yet.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
```diff
@@ -98,6 +105,9 @@ class QueryVisitor(features: SimpleFeatureCollection,
     source.getFeatures(query)
   }
 
+  override def getExpressions: java.util.List[Expression] = {
+    FilterHelper.propertyNames(filter, origSft).map(ff.property).toList
```
elahrvivaz: need to add some extra logic here: if `properties` is null, then just return all attributes; else return properties + filter attributes (deduped)

aheyne: If `propertyNames` works correctly with the Include filter than this is return all the attributes but if we include properties here I think we could end up forcing a retyping feature collection. Properties can rename and create attributes which if given I believe would cause a retyping?

elahrvivaz: I think that would cause the sft of the feature collection to be modified as well... so they should match. But you're right that we should check for functions/transforms/etc - might need to call the query planner methods to get the return sft based on the properties - I think that would handle it...

aheyne: Ok, I think it's fixed.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
```diff
@@ -98,6 +102,10 @@ class QueryVisitor(features: SimpleFeatureCollection,
     source.getFeatures(query)
   }
 
+  override def getExpressions: java.util.List[Expression] = {
+    val (_, finalSFT) = QueryPlanner.buildTransformSFT(origSft, properties)
+    FilterHelper.propertyNames(filter, finalSFT).map(ff.property).toList
```
elahrvivaz: this should include both the props from the filter and the props from the sft.
you don't have to pass the finalSFT into the propertyNames function, i'm not sure that would do the right thing (it expects the original sft)

aheyne: Ah, guess I misunderstood, that does make sense though, finalSFT could make it retype.

elahrvivaz: sorry, i meant it should return the props from the filter and the props from the finalSFT - you just don't need to pass finalSFT into the filter helper method

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
```diff
@@ -98,6 +100,11 @@ class QueryVisitor(features: SimpleFeatureCollection,
     source.getFeatures(query)
   }
 
+  override def getExpressions: java.util.List[Expression] = {
+    // We return an empty list here to avoid ReTypingFeatureCollections. Happy day.
```
elahrvivaz: haha, nice

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/process/query/QueryProcess.scala
```diff
@@ -64,11 +65,12 @@ class QueryProcess extends LazyLogging {
...
                    filter: Filter,
                    properties: Array[String] = null)
-  extends FeatureCalc
-          with LazyLogging {
+  extends FeatureCalc with FeatureAttributeVisitor with LazyLogging {
 
+  import scala.collection.JavaConversions._
   val manualVisitResults = new DefaultFeatureCollection(null, features.getSchema)
   val ff  = CommonFactoryFinder.getFilterFactory2
+  val origSft = features.getSchema
```
elahrvivaz: we can remove this now right?

aheyne:
Confirmed to work, I've seen it step through the gt-ReTypinFeatureCollection:isTypeCompatible and return true.

PR #1504, opened 2017-05-03
HBase Kerberos

geoHeil:
Cleanup and fix of the initial pull request from https://github.com/locationtech/geomesa/pull/1448


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala
```diff
@@ -35,7 +36,19 @@ class HBaseDataStoreFactory extends DataStoreFactorySpi with LazyLogging {
 
   // TODO: investigate multiple HBase connections per jvm
   private lazy val globalConnection: Connection = {
-    val ret = ConnectionFactory.createConnection(HBaseConfiguration.create())
+    val conf = HBaseConfiguration.create()
+    val auth = conf.get("hbase.security.authentication")
```
jnh5y: In the documentation, you mentioned that this happens in a kerberized environment.  Does that imply that a block like the following is in the hbase-site.xml?

<property>
   <name>hbase.security.authentication</name>
   <value>kerberos</value>
</property>

Basically, I'm wondering if that's how the configuration key-value pair is or if code sets it elsewhere?  (If it is set in configuration, is it worth explaining that in the documentation?  It may have been handled auto-magically by HDP.)

docs/user/hbase/kerberos.rst
```diff
@@ -0,0 +1,34 @@
...
+- ``hbase.geomesa.principal``
+
+All applications will require access to ``hbase-site.xml`` on their classpath in order to obtain the correct configuration.
+
+Note that Kerberos support is only available in HBase 1.1.2 or newer.
+
+Development & Testing
+---------------------
+
+GeoMesa Kerberos support was developed against Hortonworks Data Platform 2.6 authenticating against an MIT KDC.
```
jnh5y: For your testing on HDP 2.6, do you have a link to the steps you used to Kerberize the cluster?  If so, let'd be great!  I'm hoping to find some time to test this out over the next couple of days.

geoHeil:
To kerberize we use ambari and simply deploy a blueprint where kerberos is
enabled. Ambari will then do most of the hard work, I.e.. Generate users
principals.

Though you manually need to install and configure a kdc.

So regarding your first question: ambari will actually generate a fitting
hbase site xml at least that was the case for us. But you probably are
right. This should be outlines more clearly in the documentation.

Here are some links
https://community.hortonworks.com/questions/38357/enable-kerberos-via-ambari-blueprint.html
and https://github.com/bushnoh/ansible-hadoop-asap
James Hughes <notifications@github.com> schrieb am Do. 4. Mai 2017 um 23:22:

jnh5y:
@geoHeil awesome.  Do you know if those directions work well with the HDP Sandbox?  Or does one need to use Ambari to deploy the cluster?

geoHeil:
Ambari is included in any hdp sandbox by default. Indeed, to deploy a
blueprint you need to use ambari.

The default hdp sandbox doesn't come with a kerberos setup but for example
https://github.com/bushnoh/packer-hadoop-asap builds a suitable base image.
James Hughes <notifications@github.com> schrieb am Fr. 5. Mai 2017 um 15:43:

geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/GeoMesaHBaseInputFormat.scala
```diff
@@ -42,6 +43,17 @@ class GeoMesaHBaseInputFormat extends InputFormat[Text, SimpleFeature] with Lazy
     delegate.setConf(conf)
     // see TableMapReduceUtil.java
     HBaseConfiguration.merge(conf, HBaseConfiguration.create(conf))
+    val auth = conf.get("hbase.security.authentication")
+    auth match{
+      case "kerberos" => {
+        conf.set("hadoop.security.authentication", "Kerberos")
```
jnh5y: Just to ask more about this...  is this line changing the value of "hbase.security.authentication" from "kerberos" to "Kerberos"?

Is this a reason for this change?

geoHeil: actually this line is irrelevant / just double setting the property.

jnh5y: From some testing, I think it might be better to call 'SecurityUtil.setAuthenticationMethod(AuthenticationMethod.KERBEROS, conf)' here.  

After the configuration is built up, I think it might be worth a quick call to 'HBaseAdmin.checkHBaseAvailable(conf)'.  If this method fails, we'd know immediately if the Kerberos bits are not aligned.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala
```diff
@@ -35,7 +36,19 @@ class HBaseDataStoreFactory extends DataStoreFactorySpi with LazyLogging {
...
+    auth match{
+      case "kerberos" => {
+        conf.set("hadoop.security.authentication", "Kerberos")
+        UserGroupInformation.setConfiguration(conf)
+        UserGroupInformation.loginUserFromKeytab(conf.get("hbase.geomesa.principal"), conf.get("hbase.geomesa.keytab"))
```
jnh5y: I'd like to see GeoMesa use a similar approach for Kerberos for Accumulo and HBase.  @jrs53 did some initial work on Accumulo (with docs here: http://www.geomesa.org/documentation/current/user/accumulo/kerberos.html).  

In his approach, there are two GeoTools DataStoreFactory params (username and keytab) which are passed in when the datastore is created.

My preference would be to accept these values as data store params.  I'm not opposed to being able to read them from a geomesa-site.xml for default use cases.

jnh5y:
@geoHeil have you had a chance to test with MapReduce or Spark?  I ask since @jrs53 has reported issues getting Kerberos to work with the AccumuloInputFormat.

jnh5y:
For reference, @jrs53's commit here: https://github.com/locationtech/geomesa/commit/5715b20faefe4338b310b833215f7a302cc89012 shows adding DataStore params and the modifications to the command line tools.

Overall, it might be good to identify how Kerberos should work from the GeoMesa command line tools, GeoServer, Spark, and downstream code.  (For instance, I was surprised when James mentioned that he was having issues with Accumulo's Kerberos support.)

geoHeil:
@jnh5y, not yet, but I would it expect to just work assuming a fitting abase-site.xml is present on the class path. I understand your concerns regarding a unified / extended kerberos API especially with all the different integrations.

geoHeil:
Cleanup again here https://github.com/locationtech/geomesa/compare/master...geoHeil:hbase-kerberos4

jnh5y:
@geoHeil thanks for updating against master.  Do you have any time to address the comments I left above?  I think those should be addressed before we merge this PR.

geoHeil:
@jnh5y > Overall, it might be good to identify how Kerberos should work from the GeoMesa command line tools, GeoServer, Spark, and downstream code. 
is still sort of open / out of scope of this issue - at least that is what I think.

Regarding your question about the command line interface: `val conf = HBaseConfiguration.create()` will expect to create the conflicting according to the default Base java utilizes which expert an `abase-site.xml` on the class path. Therefore for any non CLI I think this best practice to put it into the resources folder. Any CLI only app would only need to put the file into the class path.

Using it from spark is straight forward as `GeoMesaHBaseInputFormat.scala` is edited in this pull request as well. This class is providing spark with HBAse access - so it should work.

jnh5y:
@geoHeil Ok, to scope this first PR, I can agree that sorting out datastore parameters and other details may be too much.

To finish things up, I think we should sort out the details about the ```conf.set("hadoop.security.authentication", "Kerberos")``` bit.  I'd prefer to use that static variable rather than the string.  

Also, there are two things I'd like to see in the documentation:
1.  Let's add the links to point folks to how to setup Kerberos on HDP 2.6. 
2.  It'd be worth it to be explicit that this currently works for the 'HBase' user defined in hbase-site.xml.  

As a general note, given @jrs53's issues seeing Kerberos work in Spark, I'm extremely cautious about any optimism that Kerberos + HBase + MR/Spark will work easily.  The good news is that James seems to be getting really close, and I hope that there are some lessons we can apply to HBase!

geoHeil:
@jnh5y actually it only works for the geodes user defined in `hbase.geomesa.keytab` and `hbase.geomesa.principal`

geoHeil:
@jnh5y I updated the documentation.

geoHeil:
@jnh5y I think your point regarding simplification of kerberos configuration is valid. Therefore, I remove the unnecessary configuration and assume it is already properly set up in the `abase-site.xml` here https://github.com/locationtech/geomesa/compare/master...geoHeil:hbase-kerberos4?expand=1

jnh5y:
Ok, that's looking pretty good.  Can you open a new PR with that branch and have the first line of the comments be:

GEOMESA-1874 Add basic Kerberos Support for HBase

I might have a chance to test things out a little more tomorrow.

PR #1503, opened 2017-05-02
GEOMESA-1783 HBase and Bigtable Improvements

elahrvivaz:
* Push-down iterators for z3 and z2
* Separate modules for Bigtable spark support
* Using MultiRowRangeFilter/BigtableExtendedScan for better performance
* Bumping Bigtable version to 0.9.6.1
    
Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

Note: stacks on top of #1500 and #1501

PR #1505, opened 2017-05-03
GEOMESA-1797 Adding shards to attribute index

elahrvivaz:
* Shards can be set using 'geomesa.attr.splits' and default to 4


elahrvivaz:
note to self: when #1503 gets merged this will need to be updated to include a BigtableAttibuteV2 index

PR #1507, opened 2017-05-06
GEOMESA-1811 GeoMesa classpath command can hang and prints duplicate …

jnh5y:
…files

* The GeoMesa 'classpath' command can print multiple copies of the same file if
it is referenced multiple times on the various Hadoop classpaths.

Additionally, calculating the classpath in Bash can hang in some situations.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
+++ b/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
@@ -109,9 +109,7 @@ if  [[ $1 = configure ]]; then
   geomesaConfigure
 elif [[ $1 = classpath ]]; then
   CLASSPATH="$GEOMESA_CP:$ACCUMULO_CP:$HADOOP_CP:${ZOOKEEPER_JAR}"
-  for element in ${CLASSPATH//:/ } ; do
-    echo ${element}
-  done
+  echo $CLASSPATH | tr ':' '\n' | xargs ls -di | sort -u -k1,1 | awk '{print $2}' | sort
 else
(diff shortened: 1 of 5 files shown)


geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa
```diff
@@ -109,9 +109,7 @@ if  [[ $1 = configure ]]; then
   geomesaConfigure
 elif [[ $1 = classpath ]]; then
   CLASSPATH="$GEOMESA_CP:$ACCUMULO_CP:$HADOOP_CP:${ZOOKEEPER_JAR}"
-  for element in ${CLASSPATH//:/ } ; do
-    echo ${element}
-  done
+  echo $CLASSPATH | tr ':' '\n' | xargs ls -di | sort -u -k1,1 | awk '{print $2}' | sort
```
elahrvivaz: does the pipe through `ls` handle missing files? i.e. do we still see them after piping through the sort, etc

jnh5y: Ah, no.  Actually, my fix is listing the actual files (which is different maybe...)

Eh, I think I'll close this PR.  The ticket might rot for a bit (unless someone else posts on it).

elahrvivaz: what is the scenario that causes it to hang?

elahrvivaz: we could do something like:
`echo $CLASSPATH | tr ':' '\n' | sort | unique`

PR #1516, opened 2017-05-18
GEOMESA-1824 GeoJsonQuery updates

blyncsy-david-lewis:
* Updated GeoJsonQuery to allow parsing from json4s ast objects instead of just strings
* Adding support for DWithin queries through the geo json query parser
* GeoJsonQuery case classes no longer store a Seq[PathAttribute], but instead us a string. That string is verified to be a valid PathAttribute in the case that it starts with "$."

### DIFF
--- a/geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
+++ b/geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
@@ -1,10 +1,10 @@
-/** *********************************************************************
+/***********************************************************************
   * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
   * All rights reserved. This program and the accompanying materials
   * are made available under the terms of the Apache License, Version 2.0
   * which accompanies this distribution and is available at
   * http://www.opensource.org/licenses/apache2.0.php.
-  * ************************************************************************/
+  *************************************************************************/
 
 package org.locationtech.geomesa.geojson.query
 
@@ -14,7 +14,7 @@ import com.vividsolutions.jts.geom.Geometry
 import org.geotools.geojson.geom.GeometryJSON
 import org.json4s.{JArray, JObject, JValue}
 import org.locationtech.geomesa.features.kryo.json.JsonPathParser
-import org.locationtech.geomesa.features.kryo.json.JsonPathParser.{PathAttribute, PathElement}
+import org.locationtech.geomesa.features.kryo.json.JsonPathParser.PathElement
 import org.opengis.filter.Filter
 
 import scala.util.control.NonFatal
--- a/geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
+++ b/geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
@@ -1,10 +1,10 @@
-/***********************************************************************
-* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
-* All rights reserved. This program and the accompanying materials
-* are made available under the terms of the Apache License, Version 2.0
-* which accompanies this distribution and is available at
-* http://www.opensource.org/licenses/apache2.0.php.
-*************************************************************************/
+/** *********************************************************************
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+  * All rights reserved. This program and the accompanying materials
+  * are made available under the terms of the Apache License, Version 2.0
+  * which accompanies this distribution and is available at
+  * http://www.opensource.org/licenses/apache2.0.php.
+  * ************************************************************************/
 
 package org.locationtech.geomesa.geojson.query
 
@@ -16,7 +16,6 @@ import org.json4s.{JArray, JObject, JValue}
 import org.locationtech.geomesa.features.kryo.json.JsonPathParser
 import org.locationtech.geomesa.features.kryo.json.JsonPathParser.{PathAttribute, PathElement}
 import org.opengis.filter.Filter
-import org.opengis.filter.expression.PropertyName
 
 import scala.util.control.NonFatal
 
@@ -53,9 +52,22 @@ object GeoJsonQuery {
 
   import org.locationtech.geomesa.filter.ff
 
-  private [query] val GeoJsonGeometryPath = Seq(PathAttribute("geometry"))
+  private[query] val defaultGeom = "geom"
 
   private val jsonGeometry = new GeometryJSON()
+
+  /**
+    * Parse a query string
+    *
+    * @param jsonValue json query value
+    * @return
+    */
+  def apply(jsonValue: JValue): GeoJsonQuery = {
+    jsonValue match {
+      case j: JObject => evaluate(j)
+      case _ => throw new IllegalArgumentException("Invalid input - expected JSON object")
+    }
+  }
 
   /**
     * Parse a query string
@@ -67,15 +79,22 @@ object GeoJsonQuery {
     import org.json4s._
     import org.json4s.native.JsonMethods._
 
-    if (query == null || query.isEmpty) { Include } else {
+    if (query == null || query.isEmpty) {
+      Include
+    } else {
       try {
-        parse(query) match {
-          case j: JObject => evaluate(j)
-          case _ => throw new IllegalArgumentException("Invalid input - expected JSON object")
-        }
+        apply(parse(query))
       } catch {
         case NonFatal(e) => throw new IllegalArgumentException(s"Invalid query string:\n$query", e)
       }
+    }
+  }
+
+  def checkJsonPaths(str: String): String = {
+    if (str.startsWith("$.")) {
+      JsonPathParser.print(JsonPathParser.parse(str))
+    } else {
+      str
     }
   }
 
@@ -86,10 +105,12 @@ object GeoJsonQuery {
     * @return
     */
   private def evaluate(json: JObject): GeoJsonQuery = {
-    if (json.obj.isEmpty) { Include } else {
+    if (json.obj.isEmpty) {
+      Include
+    } else {
       val predicates = json.obj.map {
         case (prop, v: JObject) =>
-          evaluatePredicate(JsonPathParser.parse(prop), v)
+          evaluatePredicate(checkJsonPaths(prop), v)
 
         case ("$or", v: JArray) =>
           val ors = v.arr.map {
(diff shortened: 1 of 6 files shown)


geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
```diff
@@ -1,10 +1,10 @@
...
-* which accompanies this distribution and is available at
-* http://www.opensource.org/licenses/apache2.0.php.
-*************************************************************************/
+/** *********************************************************************
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+  * All rights reserved. This program and the accompanying materials
+  * are made available under the terms of the Apache License, Version 2.0
+  * which accompanies this distribution and is available at
+  * http://www.opensource.org/licenses/apache2.0.php.
+  * ************************************************************************/
```
jnh5y: So, this is a slightly picky comment.  We use the Maven license plugin to make sure that we apply licenses properly, etc.  Sadly, the plugin is picky and doesn't do fuzzy matching.  If you can revert this header change, that'd be great!

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
```diff
@@ -99,55 +120,64 @@ object GeoJsonQuery {
...
-        Intersects(path, evaluateGeometry(v))
+        Intersects(prop, evaluateGeometry(v))
 
       case Some(("$within", v: JObject)) =>
         // { "$within" : { "$geometry" : { "type" : "Polygon", "coordinates": [ [ [0,0], [3,6], [6,1], [0,0] ] ] } } }
-        Within(path, evaluateGeometry(v))
+        Within(prop, evaluateGeometry(v))
+
+      case Some(("$dwithin", v: JObject)) =>
+        // { "$dwithin" : { "$geometry" : { "type" : "Point", "coordinates" : [30, 10] }, "$dist" : 100.50, "$unit" : "feet" } }
```
jnh5y: Dwithin is rather subtle.  I think the units are either meters or degrees.

jnh5y: (I say that because it is easy for the 'units' to be discarded in the GeoTools layer.)

geomesa-geojson/geomesa-geojson-api/src/test/scala/org/locationtech/geomesa/geojson/query/GeoJsonQueryTest.scala
```diff
@@ -1,10 +1,10 @@
...
-* are made available under the terms of the Apache License, Version 2.0
-* which accompanies this distribution and is available at
-* http://www.opensource.org/licenses/apache2.0.php.
-*************************************************************************/
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+  * All rights reserved. This program and the accompanying materials
+  * are made available under the terms of the Apache License, Version 2.0
+  * which accompanies this distribution and is available at
+  * http://www.opensource.org/licenses/apache2.0.php.
+  *************************************************************************/
```
jnh5y: Ditto with this header.

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
```diff
@@ -219,156 +252,168 @@ object GeoJsonQuery {
...
-      }
+      //      if (idPath.exists(_ == path)) {
+      //        val fids = value match {
+      //          case v: Iterable[_] => v.map(_.toString).toSeq
+      //          case v => Seq(v.toString)
+      //        }
+      //        ff.id(fids.map(ff.featureId): _*)
+      //      } else {
+      //        ff.equals(filterAttribute(path, dtgPath), ff.literal(value))
+      //      }
```
jnh5y: Can the commented out code go?

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
blyncsy-david-lewis: We were mimicking the FilterFactory.dwithin method, which takes the unit. Are you suggesting using a default of "meters" if it's not specified?

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
blyncsy-david-lewis: I think the code should go, and I should have deleted it rather than pushing commented out code. I do wonder about the original intent though, and ask @elahrvivaz if it's something we should keep.

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
jnh5y: My main goal is to point out that (unfortunately) there are dragons here.  Since part of the GeoTools API drops the units and uses the CRS's units, one can pick up degrees.  Since GeoMesa likes meters, some query pathways may re-write dwithin assuming meters using the GeodeticCalculator. 

Effectively, dwithin hits the difference between geometry and geography.

blyncsy-david-lewis: Oh, I see. Perhaps that's why DWithin was not included in the original implementation. We needed to query using DWithin for our purposes and didn't want to have a separate code path. Maybe this can be fixed with proper documentation, or this could be something we pull into Blyncsy's forked repo.

jahhulbert-ccri: meters i think. i actually worked on that and wildcards. I think geotools may ignore what you think. we should look in the unit tests for our implementation for dwithin and have it agree with that perhaps?

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
```diff
@@ -171,21 +201,24 @@ object GeoJsonQuery {
...
+    * { "$geometry" : { "type" : "Point", "coordinates" : [30, 10] } }
     *
-    * @param path json-path
-    * @param dtgPath path of date field in the sft, if any
+    * @param json geometry object
     * @return
     */
-  private def filterAttribute(path: Seq[PathElement], dtgPath: Option[Seq[PathElement]]): PropertyName = {
-    if (path == GeoJsonGeometryPath) {
-      ff.property("geom")
```
elahrvivaz: we need to keep these checks for geometry and date, otherwise we will never use our spatial indices when querying geojson. The geometry and date are extracted during ingestion and used as simple feature attributes, so we have to do the same during query.

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
```diff
@@ -171,21 +201,24 @@ object GeoJsonQuery {
...
-    } else if (dtgPath.exists(_ == path)) {
-      ff.property("dtg")
-    } else {
-      ff.property(JsonPathParser.print(PathAttribute("json") +: path))
```
elahrvivaz: We also need to keep the 'json' prefix for our geojson index, or add it in somewhere else. I think the reason we parse the path is so that we can prepend the json field

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/GeoJsonQuery.scala
elahrvivaz: this should be fine - geomesa's filter processing will handle the units correctly. if you don't use the resulting filter against a geomesa data store you might not get what you expect though

elahrvivaz:
this looks reasonable as a way to make the geojson query parsing more generic - but I think it's going to break our geojson module, which has some assumptions that were baked into the query parsing. Let me look into it some more

elahrvivaz:
@blyncsy-david-lewis yeah, the geojson-api module is failing unit tests now. I think we need a second layer of configuration that preserves the original logic, but allows for hooks to have a generic parsing like you are interested in.

elahrvivaz:
maybe we should separate out the part that parses the json query and the part that converts to a geotools filter

blyncsy-david-lewis:
I think you're right, we should separate the GeoMesa specific json requirements from the GeoJsonQuery parsing because GeoJson is it's own separate spec.

blyncsy-david-lewis:
So it seems like the test expect that every attribute is a json attribute, though that does not conform to the documentation for GeoJson, or GeoMesa's GeoJson documentation: http://www.geomesa.org/documentation/user/geojson.html

elahrvivaz:
Our geojson documentation expects that you do everything through our api. The geojson implementation is basically a translation layer between geojson/json queries and simple feature types/geotools filters. When you ingest a geojson object, it gets turned into a simple feature with an attribute named 'json' where the geojson is stored. Any geometry and/or date is extracted and used as a top-level simple feature attribute for indexing. When parsing the json query, similar logic is applied to match the underlying simple feature.

geomesa-geojson/geomesa-geojson-api/src/test/scala/org/locationtech/geomesa/geojson/GeoJsonGtIndexTest.scala
```diff
@@ -46,15 +46,15 @@ class GeoJsonGtIndexTest extends Specification {
       result must haveLength(3)
       result must contain(f0, f1, f2)
 
-      val nameResult = index.query(name, """{ "properties.name" : "n1" }""").toList
+      val nameResult = index.query(name, """{ "$.properties.name" : "n1" }""").toList
```
elahrvivaz: do we have to add a `$.` to attributes now? i'd like to not change the geojson-api syntax if possible

blyncsy-david-lewis: Oops, I those aren't required. Didn't mean to commit this file, I'll undo it

PR #1520, opened 2017-05-19
WIP: Apparent bug in Fixed Width Converter when converter definition is out of order 

ronq:
Added a janky unit test to illustrate the result when converter defintions are placed out of order in the converter defintion for the fixed-width converter:…values from previous records are used instead. It should be noted that this occurs when validation is not used.

I haven't drilled down to the actual origin of the problem just yet....

### DIFF
--- a/geomesa-convert/geomesa-convert-fixedwidth/src/test/resources/sft_testsft.conf
+++ b/geomesa-convert/geomesa-convert-fixedwidth/src/test/resources/sft_testsft.conf
@@ -5,5 +5,6 @@
     { name = "lat", type = "Double", index = "false" }
     { name = "lon", type = "Double", index = "false" }
     { name = "dtg", type = "Date", index = "false" }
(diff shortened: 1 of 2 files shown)


BrittEvans:
I've hit this problem too, solution was to just arrange the converter in the right order.

ronq:
Of course, but the issue here is that if the user doesn't know or forgets about order being important, then you have the potential for a subtle, rather than obvious bug in the converter definition.  

It seems that the right and proper thing to happen is for the out-of-order fields to be `null` --- that would at least alert the user that they are doing something silly/incorrect.  

I'll also clarify that we hit this when we were doing composition with the converter definitions , rather than putting everything in manually, since the number of attributes tends to ramp up quickly when doing stream processing.

jnh5y:
Merging @jahhulbert-ccri's EC clearing PR.

PR #1524, opened 2017-05-22
GEOMESA-1831 clear evaluation context between records

jahhulbert-ccri:


jahhulbert-ccri:
quick hack to handle clearing previous records. need to add doc that converter fields that reference other fields must be in order. GEOMESA-1833 is the full fix when it can be completed

PR #1527, opened 2017-05-23
GEOMESA-1837 Arrow optimizations

elahrvivaz:
* Arrow version updated to 0.4.0
* New arrow datastore geoserver plugin
* Arrow files encode SimpleFeatureType info in the metadata, instead of in field names
* Arrow files have a date precision option
* Arrow dates are encoded as ints or longs, instead of dates
* Arrow exports can be sorted by a particular field
* Arrow readers can take advantage of sorted arrow files when reading
* Arrow readers use optimized fast filters
* Arrow readers can cache results
* Convert command can output arrow files
* Convert command can handle compressed input files


elahrvivaz:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13568

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeaturePropertyAccessor.scala
```diff
@@ -24,7 +24,7 @@ trait SimpleFeaturePropertyAccessor extends LazyLogging {
     Thread.currentThread.setContextClassLoader(getClass.getClassLoader)
     try {
       import scala.collection.JavaConversions._
-      PropertyAccessors.findPropertyAccessors(sf, name, null, null).find(_.canHandle(sf, name, classOf[AnyRef]))
+      PropertyAccessors.findPropertyAccessors(sf, name, null, null).headOption
```
elahrvivaz: note: `findPropertyAccessors` already calls `canHandle`, so calling it here was redundant

elahrvivaz:
also, arrow 0.4.0 jars are not yet published to maven central.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
```diff
@@ -44,78 +43,132 @@ class ArrowBatchIterator extends KryoLazyAggregatingIterator[ArrowBatchAggregate
...
+
+  import org.locationtech.geomesa.arrow.allocator
+
+  private var index = 0
+  private val features = Array.ofDim[SimpleFeature](batchSize)
+
+  private val vector = SimpleFeatureVector.create(sft, dictionaries, encoding)
+  private val batchWriter = new RecordBatchUnloader(vector)
+
+  private val ordering = new Ordering[SimpleFeature] {
```
jnh5y: How many instance of the enclosing class are going to be around?  Would it make sense to move this ordering evidence to a companion object?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
```diff
@@ -44,78 +43,132 @@ class ArrowBatchIterator extends KryoLazyAggregatingIterator[ArrowBatchAggregate
...
+      val left = x.getAttribute(sortField).asInstanceOf[Comparable[Any]]
+      val right = y.getAttribute(sortField).asInstanceOf[Comparable[Any]]
+      left.compareTo(right)
+    }
+  }
+
+  override def add(sf: SimpleFeature): Unit = {
+    // we have to copy since the feature might be re-used
+    // TODO we could probably optimize this...
+    features(index) = ScalaSimpleFeature.copy(sf)
```
jnh5y: Could this approach just copy the backing byte arrays as an intermediate optimization?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
```diff
@@ -44,78 +43,132 @@ class ArrowBatchIterator extends KryoLazyAggregatingIterator[ArrowBatchAggregate
...
+
+  override def size: Int = index
+
+  override def clear(): Unit = {
+    index = 0
+    vector.clear()
+  }
+
+  override def encode(): Array[Byte] = {
+    java.util.Arrays.sort(features, 0, index, if (reverse) { ordering.reverse } else { ordering })
```
jnh5y: Would it be any faster to build up a priority queue rather than fill up the array and then sort?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
elahrvivaz: the problem comes from transforms, which are lazily evaluated using the `TransformSimpleFeature` in the parent class. need to refactor the parent some

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/features/ArrowSimpleFeature.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  override def getDefaultGeometry: AnyRef = getAttribute(geomIndex)
+  override def getAttributeCount: Int = sft.getAttributeCount
+
+  override def getBounds: BoundingBox = getDefaultGeometry match {
+    case g: Geometry => new ReferencedEnvelope(g.getEnvelopeInternal, sft.getCoordinateReferenceSystem)
+    case _           => new ReferencedEnvelope(sft.getCoordinateReferenceSystem)
+  }
+
+  override def getAttributes: jList[AnyRef] = {
```
jnh5y: Hmmm...  I wonder if getAttributes should return a lazy array...

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
```diff
@@ -0,0 +1,198 @@
...
+object ArrowFilterOptimizer extends LazyLogging {
+
+  import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
+
+  import scala.collection.JavaConversions._
+
+  private val ff: FilterFactory2 = FastFilterFactory.factory
+
+  def rewrite(filter: Filter, sft: SimpleFeatureType, dictionaries: Map[String, ArrowDictionary]): Filter = {
+    FastFilterFactory.sfts.set(sft)
```
jnh5y: Sanity check, this is a thread-local set?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
elahrvivaz: i'm fairly sure it's faster to sort the end array once than to build up a sorted collection, although i'm sure it depends on lots of things... this stack overflow has some discussion: https://stackoverflow.com/questions/3607593/is-it-faster-to-add-to-a-collection-then-sort-it-or-add-to-a-sorted-collection

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
elahrvivaz: yeah, i had to expose it here... not great but it works

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/features/ArrowSimpleFeature.scala
elahrvivaz: that might be tricky, especially since it's converting to a java list on return - what were you thinking exactly?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
```diff
@@ -58,127 +51,186 @@ class SimpleFeatureArrowFileReader(is: InputStream, filter: Filter = Filter.INCL
...
+        val all = Iterator.range(0, vector.reader.getValueCount).map { i => vector.reader.load(i); feature }
+        all.filter(filter.evaluate)
+      }
     }
   }
-}
\ No newline at end of file
+
+  // holder for a skip indicator - this will be toggled if we ever determine there can be no more results
+  private [io] class SkipIndicator(var skip: Boolean = false)
```
jnh5y: Should this be a value class?  (I.e. extend AnyVal)?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
elahrvivaz: there will be one per iterator instance. but it couldn't go in a companion object because it depends on the `sortField` - we could cache it though, what do you think?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
elahrvivaz: no, in this case we want to pass around the outer class, while allowing the inner skip flag to be toggled. it's not very 'functional', but with all the stacked arrow iterators it was a lot simpler to be able to flip this flag way down in the stack and have it available everywhere else.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/CachingSimpleFeatureArrowFileReader.scala
```diff
@@ -0,0 +1,168 @@
...
+  override def sft: SimpleFeatureType = readers.head.sft
+
+  override def dictionaries: Map[String, ArrowDictionary] = readers.head.dictionaries
+
+  override def features(filter: Filter): Iterator[ArrowSimpleFeature] with Closeable = {
+    val skip = new SkipIndicator
+    val nextBatch = SimpleFeatureArrowFileReader.features(sft, filter, skip, sort, dictionaries)
+    new Iterator[ArrowSimpleFeature] with Closeable {
+      // short-circuit if skip is toggled
+      val iter = readers.iterator.takeWhile(_ => !skip.skip).flatMap(_.features(nextBatch, skip))
```
jnh5y: Once skip is set, all further batches are skipped?

elahrvivaz: yeah, the idea is that if we detect that we've reached a value beyond our search space, we don't need to keep checking the rest of the file, since it's sorted. e.g. if we're searching for 12:00-1:00, when we hit a value with 1:01 we can ignore the rest of the file.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/StreamingSimpleFeatureArrowFileReader.scala
```diff
@@ -0,0 +1,143 @@
...
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * Streams features from an input stream - note that a feature may not be valid after a call to `.next()`,
+  * as the underlying data may be reclaimed
+  *
+  * @param is creates a new input stream
+  * @param allocator buffer allocator
+  */
+class StreamingSimpleFeatureArrowFileReader(is: () => InputStream)(implicit allocator: BufferAllocator)
```
jnh5y: Would it be possible to take an InputStream directly so that we needn't read the metadata separate from the rest of the file read?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/CachingSimpleFeatureArrowFileReader.scala
elahrvivaz: although there is some trickiness with the fact that if it's the first time being read, we need to consume the input file iterator up to the next logical file, or the arrow parsing gets out of sync

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/StreamingSimpleFeatureArrowFileReader.scala
```diff
@@ -0,0 +1,143 @@
...
+      override def hasNext: Boolean = {
+        if (done) {
+          false
+        } else if (batch.hasNext) {
+          true
+        } else if (!skip.skip && stream.available() > 0) {
+          // new logical file
+          reader = new StreamingSingleFileReader(stream)
+          readers.append(reader)
+          batch = reader.features(nextBatch, skip)
```
jnh5y: Is it possible to call reader.metadata and then reader.features?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/StreamingSimpleFeatureArrowFileReader.scala
elahrvivaz: the problem with just having a single input stream is that once you read a batch, any previous batches are lost. so you either have to cache the results (as in the caching reader), or re-read a new input stream

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeReader.scala
```diff
@@ -102,26 +114,26 @@ object ArrowAttributeReader {
...
             vector: FieldVector,
             dictionary: Option[ArrowDictionary],
-            precision: GeometryPrecision): ArrowAttributeReader = {
+            encoding: SimpleFeatureEncoding): ArrowAttributeReader = {
     val accessor = vector.getAccessor
     dictionary match {
       case None =>
         bindings.head match {
+          case ObjectType.GEOMETRY => ArrowGeometryReader(vector, classBinding, encoding.geometry)
+          case ObjectType.DATE     => ArrowDateReader(vector, encoding.date)
```
jnh5y: Ok, I'll bite...  What's the difference between the Arrow{String,Float,etc}Reader and the ArrowGeometry/DateReaders?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/StreamingSimpleFeatureArrowFileReader.scala
elahrvivaz: yeah, you could do that - you mean so that we don't have to load the metadata separately? one problem with that is you have to keep the reader open and not close it, or the dictionaries may get invalidated. there are some improvements we could make here (which is why i put the TODO), but they're not entirely trivial. a lot of the underlying clumsiness comes from the fact that the arrow reader can't just read the metadata/dictionaries and stop, it has to load the first batch at the same time. i'd like to fix that in arrow first, then this logic should get easier

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeReader.scala
```diff
@@ -174,56 +194,124 @@ object ArrowAttributeReader {
...
+        dictionary.lookup(holder.value)
+      }
+    }
+
+    override def getEncoded(i: Int): Int = accessor.get(i)
+  }
+
+  object ArrowGeometryReader {
+    def apply(vector: FieldVector, binding: Class[_], precision: EncodingPrecision): ArrowAttributeReader = {
+      if (binding == classOf[Point]) {
```
jnh5y: Idly, I wonder if it'd be 'better' to make ObjectType 'know' about the individual geometry types. 

In the above list, we'd be able to dispatch directly to the Arrow{Point,LineString}Readers, no?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeReader.scala
elahrvivaz: the primitive ones can just read the value directly from the underlying arrow vector - date and geometry are encoded differently and so need special handling. geometries are a bunch of nested vectors that need to be read using the geometry vector classes in geomesa-arrow-jts, dates are stored as longs or ints (for second precision) and need to be converted back to dates on read

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeReader.scala
elahrvivaz: yeah, ObjectType was originally written as a glue between our different serialization schemes. i think in that use case, the exact geometry type wasn't important - might be worth re-evaluating.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyName.scala
```diff
@@ -8,47 +8,22 @@
...
 
-  override def evaluate(obj: AnyRef): AnyRef = {
-    val sf = try {
-      obj.asInstanceOf[SimpleFeature]
-    } catch {
-      case e: Exception => throw new IllegalArgumentException("Only simple features are supported", e)
-    }
-    if (getProperty == null) {
-      val nsIndex = name.indexOf(':')
-      val localName = if (nsIndex == -1) name else name.substring(nsIndex + 1)
```
jnh5y: Ok, where did this admittedly complex nonsense go:)?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
```diff
@@ -8,28 +8,91 @@
...
 
 /**
- * Filter factory that overrides property name
+  * Filter factory that creates optimized filters
+  *
+  * Note: usage expects the sft to be set in FastFilterFactory.sfts
+  * FastFilterFactory.toFilter will handle this for you
  */
-class FastFilterFactory extends org.geotools.filter.FilterFactoryImpl with FilterFactory2 {
+class FastFilterFactory private extends org.geotools.filter.FilterFactoryImpl with FilterFactory2 {
```
jnh5y: *blinks*  when you put 'private' here what does it do?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
```diff
@@ -8,28 +8,91 @@
...
+    val sft = FastFilterFactory.sfts.get
+    val colon = name.indexOf(":")
+    val local = if (colon == -1) { name } else { name.substring(colon + 1) }
+    val index = sft.indexOf(local)
+    if (index != -1) {
+      new FastPropertyName(name, index)
+    } else {
+      val sf = new SimpleFeatureBuilder(sft).buildFeature("")
+      SimpleFeaturePropertyAccessor.getAccessor(sf, name) match {
+        case Some(a) => new FastPropertyNameAccessor(name, a)
```
jnh5y: Ah, the logic moved here!?

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/function/FastPropertyTest.scala
```diff
@@ -49,7 +49,7 @@ class FastPropertyTest extends Specification {
       val sf = new ScalaSimpleFeature("id", sft)
       sf.setAttributes(Array[AnyRef]("myname", "POINT(45 45)"))
 
-      val filter = ECQL.toFilter("testns:name = 'myname'", FastFilterFactory.factory)
+      val filter = FastFilterFactory.toFilter(sft, "testns:name = 'myname'")
```
jnh5y: What happens if someone uses "ECQL.toFilter" with a namespace?

I think that case would match what would happen in GeoServer.

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/GeoJsonGtIndex.scala
```diff
@@ -44,6 +44,12 @@ class GeoJsonGtIndex(ds: DataStore) extends GeoJsonIndex with LazyLogging {
   // TODO GEOMESA-1450 support json-schema validation
   // TODO GEOMESA-1451 optimize serialization for json-schema
 
+  def getTransformer(schema:SimpleFeatureType):PropertyTransformer = {
```
jnh5y: (picky) spaces after the :'s

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/PropertyTransformer.scala
```diff
@@ -0,0 +1,23 @@
+/***********************************************************************
+* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Apache License, Version 2.0
+* which accompanies this distribution and is available at
+* http://www.opensource.org/licenses/apache2.0.php.
+*************************************************************************/
+
+package org.locationtech.geomesa.geojson.query
```
jnh5y: This is from a master merge, right?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
elahrvivaz: :) that is equivalent to having a private constructor - prevents it from being instantiated. Since it relies on the thread local value being set, i wanted to prevent people using it directly without realizinng

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyName.scala
elahrvivaz: it went into the pre-processing in FastFilterFactory: https://github.com/elahrvivaz/geomesa/blob/f70e70b837cf3fd72504b4558ba2f0cd03a1a9be/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala#L32

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
elahrvivaz: oh, yep you found it

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/function/FastPropertyTest.scala
elahrvivaz: if you use plain `ECQL.toFilter` you get the normal filters, instead of our optimized ones. Is that what you mean? I think the namespaces here are at the attribute level, not the feature level, so you'd have to explicitly create them that way. for instance, in the JoinProcess we sometimes use namespaces on attribute names coming back. but in general I don't think they are normally present on the persisted features, even in geoserver.

geomesa-geojson/geomesa-geojson-api/src/main/scala/org/locationtech/geomesa/geojson/query/PropertyTransformer.scala
elahrvivaz: oh, yeah let me merge master back in

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/features/ArrowSimpleFeature.scala
jnh5y: I was just thinking that getAttributes could be called repeatedly and cause recalculation.  Probably not work worrying about...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/ArrowBatchIterator.scala
jnh5y: Meh, I doubt it is worth caching.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/reader/StreamingSimpleFeatureArrowFileReader.scala
jnh5y: +1

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/function/FastPropertyTest.scala
jnh5y: We implemented the namespace logic (which moved around in this PR) to handle queries in GeoServer where namespaced attributes were being used.  I suppose I'm wondering if that pathway will still work...  (Or if the move shifted the code out of the intended pathway.)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/features/ArrowSimpleFeature.scala
elahrvivaz: oh i gotcha - a simple feature instance is re-used though, and `index` gets updated, so we'd have to cache by `index`

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/function/FastPropertyTest.scala
elahrvivaz: i think it should still work - FastPropertyName was only instantiated through FastFilterFactory, so it would get invoked in the same way here, just the logic is calculated before creating the filter instead of after. `ECQL.toFilter` wasn't ever hitting that logic.

jnh5y: Could this test case be duplicated to include the original code?

elahrvivaz: oh, so the key part in this test is that the `FastFilterFactory` was passed in to `ECQL.toFilter` - we're doing the same thing in `FastFilterFactory.toFilter`, it's just that the class has a private constructor now so you can't call it the old way (to enforce setting the thread local sft). But it is testing the same thing still.
https://github.com/elahrvivaz/geomesa/blob/7122424cb4d49343cc1d86177636048595e26fd5/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala#L93

elahrvivaz:
cqs and maven central are gtg now

PR #1531, opened 2017-05-23
GEOMESA-1840 Limit listing hbase/bigtable tables

blyncsy-david-lewis:
* Created DataStoreConnector that handles caching calls to DataStoreFinder.getDataStore
* Created unsafe save method in HBaseSpatialRDDProvider that does not check the schema
* Modified save and rdd methods from HBaseSpatialRDDPRovider to use DataStoreConnector to use cached data stores

### DIFF
--- a/geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
+++ b/geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
@@ -98,19 +98,7 @@ class HBaseSpatialRDDProvider extends SpatialRDDProvider {
       ds.dispose()
     }
 
-    rdd.foreachPartition { iter =>
-      val ds = DataStoreConnector.getOrFindDataStore(writeDataStoreParams).asInstanceOf[HBaseDataStore]
-      val featureWriter = ds.getFeatureWriterAppend(writeTypeName, Transaction.AUTO_COMMIT)
-      try {
-        iter.foreach { rawFeature =>
-          FeatureUtils.copyToWriter(featureWriter, rawFeature, overrideFid = true)
-          featureWriter.write()
-        }
-      } finally {
-        CloseQuietly(featureWriter)
-        ds.dispose()
-      }
-    }
+    unsafeSave(rdd, writeDataStoreParams, writeTypeName)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jnh5y:
@blyncsy-david-lewis by the way, Eclipse wants us to have contributors submit code with the Git Sign-off flag set.  (https://github.com/locationtech/geomesa/blob/master/CONTRIBUTING.md#coding-standards)  That sign-off means that you are responsible for the contribution being good-to-go etc.  Eclipse has an ip-validation webhook which checks each commit for the sign-off bits.  (If you sign-off the next commits, we can fix it when we merge the PR.)

geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
```diff
@@ -98,7 +99,32 @@ class HBaseSpatialRDDProvider extends SpatialRDDProvider {
...
+    * The type must exist in the data store, and all of the features in the RDD must be of this type.
+    * This method assumes that the schema exists.
+    *
+    * @param rdd
+    * @param writeDataStoreParams
+    * @param writeTypeName
+    */
+  def unsafeSave(rdd: RDD[SimpleFeature], writeDataStoreParams: Map[String, String], writeTypeName: String): Unit = {
+    rdd.foreachPartition { iter =>
+      val ds = DataStoreConnector.getOrFindDataStore(writeDataStoreParams).asInstanceOf[HBaseDataStore]
```
jnh5y: I'd like to see us factor out the code here to avoid duplication if possible.

blyncsy-david-lewis: good call, fixed

blyncsy-david-lewis:
I see, I just made a commit with -s

geomesa-spark/geomesa-spark-core/src/main/scala/org/locationtech/geomesa/spark/DataStoreConnector.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object DataStoreConnector {
+  import org.locationtech.geomesa.spark.CaseInsensitiveMapFix._
+
+  private val writeLock = new Object
+  @volatile private var connectionMap: Map[Map[String, String], DataStore] = Map()
+
+  def getOrFindDataStore(writeDataStoreParams: Map[String, String]): DataStore = {
+    var map = connectionMap
+    map.get(writeDataStoreParams) match {
```
jnh5y: We typically use Caffeine loading caches for things like this.  E.g.:

https://github.com/locationtech/geomesa/blob/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala#L174-L177

I'd suggest that since it avoids implementing the careful synchronized code here.  Also, if we find it useful, it'd be possible to add cache configuration to the DataStoreConnector object.  (I.e., connection could timeout after 10 minutes, etc.)

blyncsy-david-lewis: Sounds good, I think I'll make that change in my production code as well.

PR #1533, opened 2017-05-24
GEOMESA-1835 Fixing cleanup of shapefile ingest

elahrvivaz:


elahrvivaz:
the `TypeUpdatingFeatureCollection` seems like it was implemented properly - instead of trying to fix it for a general case, I just moved the type updating logic into the shapefile ingest (as that is the only place using it)

geomesa-accumulo/geomesa-accumulo-dist/src/main/assemblies/component.xml
```diff
@@ -20,6 +20,7 @@
                 <exclude>org.locationtech.geomesa:geomesa-geojson*</exclude>
                 <exclude>org.locationtech.geomesa:geomesa-blobstore*</exclude>
                 <exclude>org.locationtech.geomesa:geomesa-accumulo-compute*</exclude>
+                <exclude>org.locationtech.geomesa:geomesa-accumulo-spark-runtime*</exclude>
```
jnh5y: +1!

aheyne: Need to update the docs here http://www.geomesa.org/documentation/current/user/spark/accumulo_spark_runtime.html?highlight=spark

elahrvivaz: it's still in `dist/spark/`, just not under `lib/` - i think the docs are ok as is?

aheyne: Ah, yeah, good catch then!

PR #1534, opened 2017-05-24
GEOMESA-1842 Move GM Test to geomesa-test module

aheyne:

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -28,53 +28,9 @@
 
     <profiles>
         <profile>
-            <id>html</id>
-            <properties>
-                <target>html</target>
-            </properties>
-        </profile>
-        <profile>
-            <id>singlehtml</id>
-            <properties>
-                <target>singlehtml</target>
-            </properties>
-        </profile>
-        <profile>
-            <id>latex</id>
-            <properties>
-                <target>latex</target>
-            </properties>
-        </profile>
-        <profile>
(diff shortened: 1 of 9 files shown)


docs/pom.xml
```diff
@@ -28,54 +28,10 @@
 
     <profiles>
         <profile>
-            <id>html</id>
```
elahrvivaz: are these docs profiles not used? @mdzimmerman ?

docs/pom.xml
```diff
@@ -105,6 +61,54 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <id>gm-test-auto</id>
```
elahrvivaz: i think i would move these test profiles into the test module pom

aheyne: Gah, ignore this git failed me again...

pom.xml
```diff
@@ -41,6 +41,7 @@
         <module>geomesa-security</module>
         <module>geomesa-spark</module>
         <module>geomesa-stream</module>
+        <module>geomesa-test</module>
```
aheyne: I'm using filtering from the maven assembly plugin to make things smoother.

aheyne: Additionally there was talk about moving these test to jars later to give us better test validation.

PR #1537, opened 2017-05-25
GEOMESA-1844 Reorganization of WPS classes

elahrvivaz:
* GeoMesaProcess/GeoMesaProcessVisitor traits allow for discoverable processes
* Process implementations are in geomesa-process/geomesa-process-vector
* Process factory is a separate jar in geomesa-process/geomesa-process-wps


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureCollection.scala
```diff
@@ -28,16 +22,7 @@ class AccumuloFeatureCollection(source: GeoMesaFeatureSource, query: Query)
...
-      case v: ProximityVisitor => v.setValue(v.proximitySearch(source, query))
-      case v: QueryVisitor     => v.setValue(v.query(source, query))
-      case v: StatsVisitor     => v.setValue(v.query(source, query))
-      case v: SamplingVisitor  => v.setValue(v.sample(source, query))
-      case v: KNNVisitor       => v.setValue(v.kNNSearch(source,query))
-      case v: AttributeVisitor => v.setValue(v.unique(source, query))
-      case v: RouteVisitor     => v.routeSearch(source, query)
-      case v: BinVisitor       => v.binQuery(source, query)
-      case v: ArrowVisitor     => v.arrowQuery(source, query)
+      case v: GeoMesaProcessVisitor => v.execute(source, query)
```
jnh5y: +1.  Austin and I noticed that this api improvement was possible.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/process/tube/TubeSelectProcessTest.scala
```diff
@@ -31,8 +31,7 @@ class TubeSelectProcessTest extends Specification {
 
   sequential
 
-  val dtgField = org.locationtech.geomesa.accumulo.process.tube.DEFAULT_DTG_FIELD
```
jnh5y: Humm... so seems like there are an implementation detail in having a field named "dtg" floating around. 

Previously, this static variable was marking the use and variable name.  I wonder if we should keep it?

elahrvivaz: it was only used in test classes, so i thought it was ok to remove... i think there are already constants for 'dtg' we could use instead, let me check.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
```diff
@@ -142,9 +140,6 @@ class ArrowVisitor(sft: SimpleFeatureType, dictionaryFields: Seq[String], batchS
       unloadManualResults(true)
     }
   }
-
-  // hook to allow delegation from retyping feature collections
-  override def getExpressions: java.util.List[Expression] = Collections.emptyList()
```
jnh5y: Do we still need this for GeoServer RetypingFeatureCollections?

elahrvivaz: I added it to the `GeoMesaProcessVisitor` trait

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/GeoMesaProcess.scala
```diff
@@ -0,0 +1,39 @@
...
+  /**
+    * Optimized method to run distributed processing. Should set the result, available from `getResult`
+    *
+    * @param source simple feature source
+    * @param query may contain additional filters to apply
+    */
+  def execute(source: SimpleFeatureSource, query: Query): Unit
+
+  // hook to allow delegation from retyping feature collections
+  override def getExpressions: java.util.List[Expression] = Collections.emptyList()
```
jnh5y: Thanks for pointing this out. 

Are there any situations we can think of where we do not want the RetypingFC to delegate to our collections?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureCollection.scala
jnh5y: As I was thinking through how having this trait would work, I was wondering how we can handle partially visitor support in HBase.  

Thoughts on that?  For instance, wiring up the StatsProcess would next should be pretty quick.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/GeoMesaProcess.scala
elahrvivaz: hmm, nothing i can think of...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureCollection.scala
```diff
@@ -28,16 +22,7 @@ class AccumuloFeatureCollection(source: GeoMesaFeatureSource, query: Query)
 
   override def accepts(visitor: FeatureVisitor, progress: ProgressListener): Unit =
     visitor match {
-      case v: TubeVisitor      => v.setValue(v.tubeSelect(source, query))
-      case v: ProximityVisitor => v.setValue(v.proximitySearch(source, query))
-      case v: QueryVisitor     => v.setValue(v.query(source, query))
-      case v: StatsVisitor     => v.setValue(v.query(source, query))
-      case v: SamplingVisitor  => v.setValue(v.sample(source, query))
```
jnh5y: Idly, for HBase, the SamplingVisitor could be re-worked to act like the DensityProcess where it just sets some query hints which could be handled by the GeoMesa HBase Filter.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureCollection.scala
elahrvivaz: in the `HBaseFeatureCollection` we could match on the concrete classes that we support, instead of the trait, and delegate the rest.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureCollection.scala
elahrvivaz: yeah, everything goes through query hints - then it's up to the data store impls to respect the hints

elahrvivaz: SamplingProcess already just sets the hints in `invertQuery`

PR #1541, opened 2017-05-29
GEOMESA-1850 - Add gt-main back to the HBase distribution

jnh5y:


aheyne:
Added to #1545

PR #1544, opened 2017-05-30
GEOMESA-1805 Not cloning cell values in HBase z-filters

elahrvivaz:


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseZ2Index.scala
```diff
@@ -24,18 +24,20 @@ case object HBaseZ2Index extends HBaseLikeZ2Index with HBasePlatform {
                                                sft: SimpleFeatureType,
                                                filter: HBaseFilterStrategyType,
                                                transform: Option[(String, SimpleFeatureType)]): Seq[HFilter] = {
+    import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
     val z2Filter = Z2Index.currentProcessingValues.map { case Z2ProcessingValues(_, bounds) =>
-      configureZ2PushDown(bounds)
+      val offset = if (sft.isTableSharing) { 2 } else { 1 } // sharing + shard
```
jnh5y: Just clarify, was it the case that table sharing was not working before this?

elahrvivaz: originally i was wondering that too haha, but no, we set table sharing to always be false for hbase: https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStore.scala#L48
i think it's probably more consistent to keep the table sharing logic here even though it's not used, what do you think?

jnh5y: Blergh.  Having the logic without it working is a confusion to us.  That said, this'll make it *slightly* easier to enable later.  I think it is ok as-is.

PR #1545, opened 2017-05-30
GEOMESA-1853 Allow for autoregistration of coprocessors on table crea…

aheyne:
…tion.

Added in ability to read coprocessor configurtaion from geomesa-site, system property, datastore param or classpath.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
@@ -65,8 +65,8 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
   /**
    * We write the geometry at the center of the zbox that this row represents
    */
-  def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
-    case p: Point => densityUtils.writeNonPoint(p, weight, result)
+  override def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
+    case p: Point => writeNonPoint(p, weight, result)
     case _ =>
       val row = topKey.getRowData
       val zOffset = row.offset() + zPrefix
@@ -77,7 +77,7 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
       }
       val (x, y) = Z2SFC.invert(Z2(Longs.fromByteArray(zBytes)))
       val nWeight = normalizeWeight(weight)
-      densityUtils.writePointToResult(x, y, nWeight, result)
+      writePointToResult(x, y, nWeight, result)
   }
 }
 
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
@@ -66,7 +66,7 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
    * We write the geometry at the center of the zbox that this row represents
    */
   override def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
-    case p: Point => writeNonPoint(p, weight, result)
+    case p: Point => writePointToResult(p, weight, result)
     case _ =>
       val row = topKey.getRowData
       val zOffset = row.offset() + zPrefix
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
@@ -65,8 +65,8 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
   /**
    * We write the geometry at the center of the zbox that this row represents
    */
-  override def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
-    case p: Point => writePointToResult(p, weight, result)
+  def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
+    case p: Point => densityUtils.writeNonPoint(p, weight, result)
     case _ =>
       val row = topKey.getRowData
       val zOffset = row.offset() + zPrefix
@@ -77,7 +77,7 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
       }
       val (x, y) = Z2SFC.invert(Z2(Longs.fromByteArray(zBytes)))
       val nWeight = normalizeWeight(weight)
-      writePointToResult(x, y, nWeight, result)
+      densityUtils.writePointToResult(x, y, nWeight, result)
   }
 }
 
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/KryoLazyDensityCoprocessor.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/KryoLazyDensityCoprocessor.scala
@@ -41,14 +41,14 @@ class KryoLazyDensityCoprocessor extends KryoLazyDensityService with Coprocessor
   import KryoLazyDensityCoprocessor._
 
   private var env: RegionCoprocessorEnvironment = _
-  private var sft: SimpleFeatureType = _
+  //private var sft: SimpleFeatureType = _
 
-  def init(options: Map[String, String], sft: SimpleFeatureType): DensityResult = {
-    this.sft = sft
-    initialize(options, sft)
-  }
+//  def init(options: Map[String, String], sft: SimpleFeatureType): DensityResult = {
+//    //this.sft = sft
+//    initialize(options, sft)
+//  }
 
-  def aggregateResult(sf: SimpleFeature, result: DensityResult): Unit = writeGeom(sf, result)
+  //def aggregateResult(sf: SimpleFeature, result: DensityResult): Unit = writeGeom(sf, result)
 
   @throws[IOException]
   def start(env: CoprocessorEnvironment): Unit = {
@@ -67,6 +67,12 @@ class KryoLazyDensityCoprocessor extends KryoLazyDensityService with Coprocessor
   def getDensity(controller: RpcController,
                  request: KryoLazyDensityProto.DensityRequest,
                  done: RpcCallback[KryoLazyDensityProto.DensityResponse]): Unit = {
+
+    val this_grid = new KryoLazyDensityUtils {}
+    //this_grid.initialize()
+
+    def aggregateResult(sf: SimpleFeature, result: DensityResult): Unit = writeGeom(sf, result)
+
     var scanner : InternalScanner = null
     var filterList : FilterList = new FilterList()
 
@@ -91,7 +97,7 @@ class KryoLazyDensityCoprocessor extends KryoLazyDensityService with Coprocessor
       }
 
       val serializer = new KryoFeatureSerializer(sft, SerializationOptions.withoutId)
-      val densityResult: DensityResult = this.init(options, sft)
+      val densityResult: DensityResult = this_grid.initialize(options, sft)
 
       scanList.foreach(scan => {
         scan.setFilter(filterList)
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/KryoLazyDensityCoprocessor.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/KryoLazyDensityCoprocessor.scala
@@ -1,5 +1,5 @@
 /***********************************************************************
-* Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+* Copyright (c) 2013-2017 Commonwealth Computer Research, Inc.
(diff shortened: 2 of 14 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -56,14 +60,52 @@ trait HBaseFeatureIndex extends HBaseFeatureIndexType
 
   import HBaseFeatureIndex.{DataColumnFamily, DataColumnQualifier}
 
+  private val logger = LoggerFactory.getLogger("org.locationtech.geomesa.hbase.index.HBaseFeatureIndex")
```
elahrvivaz: any reason not to use the LazyLogger trait?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -56,14 +60,52 @@ trait HBaseFeatureIndex extends HBaseFeatureIndexType
...
+
   override def configure(sft: SimpleFeatureType, ds: HBaseDataStore): Unit = {
     super.configure(sft, ds)
+
     val name = TableName.valueOf(getTableName(sft.getTypeName, ds))
     val admin = ds.connection.getAdmin
+    val coproUrl: Option[Path] = if(!ds.config.coprocessorUrl.equals(HBaseDataStoreParams.CoprocessorUrl.sample)) {
```
elahrvivaz: instead of having this logic here, why don't we just not have a default for the param, so it will be null? `HBaseDataStoreConfig` can have an `Option[Path]`

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -56,14 +60,52 @@ trait HBaseFeatureIndex extends HBaseFeatureIndexType
...
+        Some(ds.config.coprocessorUrl)
+      } else {
+        GeoMesaSystemProperties.SystemProperty("geomesa.hbase.coprocessor.path", null).option match {
```
elahrvivaz: instead of a match, this could be more succinctly written as `.option.map(new Path(_))`

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
aheyne: Ah, nope.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
aheyne: Fancy, changed

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
aheyne: Changed

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala
```diff
@@ -114,6 +117,7 @@ class HBaseDataStoreFactory extends DataStoreFactorySpi with LazyLogging {
 object HBaseDataStoreParams {
   val BigTableNameParam    = new Param("bigtable.table.name", classOf[String], "Table name", true)
   val ConnectionParam      = new Param("connection", classOf[Connection], "Connection", false)
+  val CoprocessorUrl       = new Param("coprocessor.url", classOf[Path], "Coprocessor Url", false, Option[Path])
```
elahrvivaz: we probably shouldn't include Option in the param, as it's a java api

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala
```diff
@@ -75,9 +76,10 @@ class HBaseDataStoreFactory extends DataStoreFactorySpi with LazyLogging {
       if (security) {
        Some(HBaseDataStoreFactory.buildAuthsProvider(connection, params))
       } else None
+    val coprocessorUrl = CoprocessorUrl.lookupWithDefault[Path](params)
```
elahrvivaz: can use `.lookupOpt`

elahrvivaz: oh, maybe you are still working on this

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -52,18 +58,49 @@ object HBaseFeatureIndex extends HBaseIndexManagerType {
...
+          // TODO: Warn if the path given is different from paths registered in other coprocessors. This is to warn if another table needs updated.
+          case Some(path) => desc.addCoprocessor(name, path, Coprocessor.PRIORITY_USER, null)
+          case None       => desc.addCoprocessor(name)
+        }
+      }
+    }
+
     try {
       if (!admin.tableExists(name)) {
+        logger.info(s"Creating table $name")
```
elahrvivaz: let's make this debug level

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/KryoLazyDensityUtils.scala
```diff
@@ -18,48 +18,46 @@ import org.locationtech.geomesa.utils.geotools.GridSnap
...
+    val envelope = new Envelope(bounds(0), bounds(1), bounds(2), bounds(3))
+    val Array(width, height) = options(GRID_OPT).split(",").map(_.toInt)
+    new GridSnap(envelope, width, height)
+  }
+  // function to get the weight from the feature - defaults to 1.0 unless an attribute/exp is specified
+  val weightIndex = options.get(WEIGHT_OPT).map(sft.indexOf).getOrElse(-2)
+
+  def newDensityResult: DensityResult = mutable.Map.empty[(Int, Int), Double]
+
+  def weightFn(sf: SimpleFeature): Double = {
```
elahrvivaz: this should stay as a val so we don't e.g. evaluate the expression on line 46 every time

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/KryoLazyDensityUtils.scala
```diff
@@ -18,48 +18,46 @@ import org.locationtech.geomesa.utils.geotools.GridSnap
...
-        getWeightFromExpression(expression)
-      } else if (sft.getDescriptor(weightIndex).getType.getBinding == classOf[java.lang.Double]) {
-        getWeightFromDouble(weightIndex)
-      } else {
-        getWeightFromNonDouble(weightIndex)
-      }
-
-    writeGeom = if (sft.isPoints) {
-      (sf, result) => writePoint(sf, weightFn(sf), result)
+  def writeGeom(sf: SimpleFeature, result: DensityResult): Unit = {
```
elahrvivaz: ditto this should be a val too so we don't evaluate `sft.isPoints` each time

elahrvivaz: in general these changes look fine - eventually it might be cleaner to create a companion object that does the setup and configuration, and then a simpler class implementation (or multiple impls, one for points, non-points, etc)

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala
aheyne: fixed

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala
aheyne: fixed

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
aheyne: fixed

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
```diff
@@ -65,8 +65,8 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
   /**
    * We write the geometry at the center of the zbox that this row represents
    */
-  override def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
-    case p: Point => writePointToResult(p, weight, result)
+  def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
```
elahrvivaz: hmm, are these methods actually getting invoked anymore? seems like it was overriding methods in KryoLazyDensityUtils and now it's not

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/KryoLazyDensityCoprocessor.scala
```diff
@@ -120,18 +128,17 @@ class KryoLazyDensityCoprocessor extends KryoLazyDensityService with Coprocessor
...
         throw dse
-    } finally CloseQuietly(scanner)
+    }
 
     done.run(response)
   }
 }
 
-object KryoLazyDensityCoprocessor extends KryoLazyDensityUtils {
+object KryoLazyDensityCoprocessor {
```
elahrvivaz: i think the problem with the KryoLazyDensityUtils was that it needs to be used in instance classes, not objects. It has mutable state. So possibly just making the class instead of the object extend it would fix the issues you saw. I think the problem with the current changes is that the Z2/3 density iterators are not overriding the methods correctly anymore

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/Z2DensityIterator.scala
```diff
@@ -66,7 +66,7 @@ class Z2DensityIterator extends KryoLazyDensityIterator {
    * We write the geometry at the center of the zbox that this row represents
    */
   override def writeNonPoint(geom: Geometry, weight: Double, result: DensityResult): Unit = geom match {
-    case p: Point => writePointToResult(p, weight, result)
+    case p: Point => writeNonPoint(p, weight, result)
```
elahrvivaz: think this can be reverted, and in z3

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/KryoLazyDensityCoprocessor.scala
```diff
@@ -67,47 +60,61 @@ class KryoLazyDensityCoprocessor extends KryoLazyDensityService with Coprocessor
...
-        filterList = FilterList.parseFrom(Base64.decode(options(FILTER_OPT)))
+      val filterList: FilterList = if (options.containsKey(FILTER_OPT)) {
+        FilterList.parseFrom(Base64.decode(options(FILTER_OPT)))
+      } else {
+        new FilterList()
       }
 
       val serializer = new KryoFeatureSerializer(sft, SerializationOptions.withoutId)
-      val densityResult: DensityResult = this.init(options, sft)
+      val densityResult: DensityResult = initialize(options, sft)
```
elahrvivaz: think we need to instantiate a new KryoLazyDensityUtils to use here - and the main coprocessor class shouldn't extend it any more

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/KryoLazyDensityUtils.scala
```diff
@@ -18,20 +18,22 @@ import org.locationtech.geomesa.utils.geotools.GridSnap
...
+
+  var geomIndex: Int = _
+  var gridSnap: GridSnap = _
+  var writeGeom: (SimpleFeature, DensityResult) => Unit = _
 
   def initialize(options: Map[String, String], sft: SimpleFeatureType): DensityResult = {
-    import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
-    geomIndex = sft.getGeomIndex
+
+    geomIndex = Option(sft.getGeomIndex).getOrElse(-1)
```
elahrvivaz: think we should revert all the changes to this file too - this isn't going to do quite what you expect, since it's an Int and can't be null, so it will get turned to 0 and thus always be Some. The other changes seem like they're not needed any more

PR #1546, opened 2017-05-30
GEOMESA-1833 Fixing order of field evaluation in converters

elahrvivaz:


build/cqs.tsv
```diff
@@ -192,6 +192,7 @@ org.geotools.xsd:gt-xsd-wfs	15.1	compile
 org.hsqldb:hsqldb	2.3.0	compile
 org.jamon:jamon-runtime	2.4.1	compile
 org.jdom:jdom	1.1.3	compile
+org.jgrapht:jgrapht-core	1.0.1	compile
```
elahrvivaz: jgrapht was already rejected once, this might not fly :(

https://dev.eclipse.org/ipzilla/show_bug.cgi?id=8704

elahrvivaz: Looks like some work was done to remove the problematic files but not sure it's hosted anywhere, or got re-merged back to the main branch: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11214

jnh5y:
Ok, let's talk about circular dependencies... what exception do users see?

elahrvivaz:
it's a `StackOverflowError` in `Field.dependenciesOf` - I initially added some logic to detect cycles in the final graph, but it never got to that point. Current code does the same thing, so it's not any worse with this patch

jnh5y:
The current code doesn't compute values, so there wouldn't be a stackoverflow, right?

In the meantime, would it be worth catching the StackOverflow and throwing an error which says what happened?

elahrvivaz:
the current code still checks for which fields are 'required' (i.e. actually used), so it will still throw the stack overflow.
we can hold off merging this; i'm working on detecting the cycles up front

elahrvivaz:
added checks for circular dependencies, will throw a user-friendly exception now

PR #1548, opened 2017-05-31
GEOMESA-1851 Propagate failed json parsing errors in JsonConverter

elahrvivaz:


elahrvivaz:
I couldn't actually reproduce the original ArrayOutOfBounds error (possibly it has been fixed since 1.31), but this seems like a good change even so.

PR #1549, opened 2017-06-01
GEOMESA-1854 Replacing ciFlatMap with flatMap in CloseableIterator

elahrvivaz:


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/CloseableIterator.scala
```diff
@@ -161,12 +155,12 @@ object SelfClosingIterator {
         iterHasNext
       }
       override def next(): A = iter.next()
-      override def close(): Unit = closeIter()
+      override def close(): Unit = closeIter
     }
 
-  def apply[A](iter: Iterator[A] with Closeable): SelfClosingIterator[A] = apply(iter, iter.close)
+  def apply[A](iter: Iterator[A] with Closeable): SelfClosingIterator[A] = apply(iter, iter.close())
```
jnh5y: Hmm.... does this call iter.close() on construction?

elahrvivaz: no, pretty sure not - i changed the apply syntax from `() => Unit` to `=> Unit` so the syntax of calling it is slightly different. i'll add a unit test just to make sure

elahrvivaz: added

jnh5y: Ah, ok, thanks for the unit test.  I missed the both old and new versions are 'productions' with =>.  I feel better now:).

PR #1554, opened 2017-06-02
GEOMESA-1858 Cleanup metadata tables if createSchema fails for any re…

aheyne:
…ason.

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
@@ -140,17 +140,24 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS, F, W], F <: WrappedFe
           // do this before anything else so that any modifications will be in place
           GeoMesaSchemaValidator.validate(sft)
 
-          // write out the metadata to the catalog table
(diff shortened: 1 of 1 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -140,17 +140,24 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS, F, W], F <: WrappedFe
...
               .foreach(k => reloadedSft.getUserData.put(k, sft.getUserData.get(k)))
 
-          // create the tables in accumulo
-          manager.indices(reloadedSft, IndexMode.Any).foreach(_.configure(reloadedSft, this))
+            // create the tables
+            manager.indices(reloadedSft, IndexMode.Any).foreach(_.configure(reloadedSft, this))
+          } catch {
+            case e: Exception =>
+              // If there was an error creating a schema, clean up.
+              metadata.delete(sft.getTypeName)
```
elahrvivaz: i would wrap this in a try/catch as well - in the second catch block you can use `e.addSuppressed(e2)`

aheyne: added

PR #1558, opened 2017-06-05
GEOMESA-1860 Track label process

elahrvivaz:


jnh5y:
What documentation should we add to this?

elahrvivaz:
I'll add to the docs - looks like we're missing some other ones there too.

jnh5y:
Ayup, I just making sure to ask for docs upfront now since we can do that.

PR #1559, opened 2017-06-06
GEOMESA-1857 Pluggable index selection during query planning

elahrvivaz:


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/StrategyDecider.scala
```diff
@@ -0,0 +1,145 @@
...
+      (option, optionCosts.sum, timing.time)
+    }.sortBy(_._2)
+    explain(s"Costs: ${costs.map(c => s"${c._1} (Cost ${c._2} in ${c._3}ms)").mkString("; ")}")
+    costs.head._1
+  }
+}
+
+object StrategyDecider extends MethodProfiling with LazyLogging {
+
+  private val decider: StrategyDecider = SystemProperty("geomesa.strategy.decider").option match {
```
jahhulbert-ccri: [thought] should we make this a ```def``` instead of ```val``` in order to allow for dynamically reloading it? May be interesting if you want to change it on the fly? could always be later i don't particularly have a concrete use case other than my thought that i would want to possibly change it without restarting a server but currently there's no way to do that.

elahrvivaz: yeah, not creating a new one each time was the reason i did it this way - i was thinking at the time that it would be nice to be able to define a strategy per-data store too, which would be kind of related. but figured i'd do the simple thing first...

elahrvivaz: eventually we could tie it to the data store lifecycle, which would let you swap them at runtime in geoserver by re-loading your data store.

jahhulbert-ccri: yeah that's what i was wondering :)

PR #1562, opened 2017-06-06
GEOMESA-1859 Fixing arrow accumulo distributed export

elahrvivaz:
* Relocating fasterxml classes to avoid conflict with hadoop


jnh5y:
Is there a geomesa-test case we could add which would demonstrate that this works properly?

elahrvivaz:
i added an export command to the test package, but verifying the export is kind of tricky b/c it requires the arrow tools...

PR #1573, opened 2017-06-09
One can run HBase unit tests with a ramdisk with something like this:

jnh5y:
mount -f tmpfs -o size=2048M tmpfs /mnt/ramdisk/
sudo mount -t tmpfs -o size=2048M tmpfs /mnt/ramdisk/
./build/mvn test -pl geomesa-index-api -pl geomesa-hbase/geomesa-hbase-datastore -Dtest.build.data.basedirectory=/mnt/ramdisk/


aheyne:
Could we add a profile to the master pom that would override the build directories to a ram disk mount point? Also, I don't have a `-f` option on mount.

jnh5y:
@aheyne good idea; not sure.  Creating a ramdisk requires sudo/root.

PR #1583, opened 2017-06-14
GEOMESA-1879 Fixing configuration of BIN and density queries with transforms

elahrvivaz:


jahhulbert-ccri:
[for my understanding] I assume this is just because we never need transforms with bin queries since they only return dtg and geom basically?

jahhulbert-ccri:
also this is 1.3.2 right?

elahrvivaz:
yeah, we always operate on the full feature but just return geom/dtg/label. in our refactoring we passed the transform in, which broke things - previously it was ignored. I think clients were passing a transform as a failsafe - if the bin ends up falling back to client-side encoding, it would speed things up (although it should always be done server side now).

PR #1587, opened 2017-06-15
GEOMESA-1882 - Unit test for accessing last element of json array in Converter

ddseapy:


aheyne:
Not sure about why this is failing or if it's valid syntax but I have found a workaround. You can use array[start:end] access method like `$.colors[-1:].color` to access the last item in the array.

aheyne:
[Online example](http://jsonpath.herokuapp.com/?path=$..book[-1:])

PR #1590, opened 2017-06-20
GEOMESA-1810 GeoMesa 'classpath' command is not documented

aheyne:
* Documented classpath command
* Documented configure command

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/ClasspathCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/ClasspathCommand.scala
@@ -11,6 +11,10 @@ package org.locationtech.geomesa.tools.status
 import com.beust.jcommander.Parameters
 import org.locationtech.geomesa.tools.Command
 
+/**
+  * Note: this class is a placeholder for the 'classpath' function implemented in the 'geomesa-*' script, to get it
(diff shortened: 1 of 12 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/status/ClasspathCommand.scala
```diff
@@ -0,0 +1,22 @@
...
+ ***********************************************************************/
+
+package org.locationtech.geomesa.tools.status
+
+import com.beust.jcommander.Parameters
+import org.locationtech.geomesa.tools.Command
+
+class ClasspathCommand extends Command {
+
+  override val name = "classpath"
```
jnh5y: Does adding these commands interfere with bits that the bash scripts do?

aheyne: Nope, the bash scripts intercept these before they even get to java. This just allows jcommander to report these options in the usage output.

elahrvivaz: ah, can we add comments to that effect?

elahrvivaz:
going to merge and I'll add comments on the classes

PR #1591, opened 2017-06-21
GEOMESA-1888 - Refactoring and Indexing API clean-up

jnh5y:


pom.xml
```diff
@@ -1,12 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--********************************************************************
```
aheyne: Do we want this here? Looks like we have a mix of poms with and without licences. There are at least 48 others with it and a fair bit without.

PR #1592, opened 2017-06-28
GEOMESA-1890 Lambda data store

elahrvivaz:


geomesa-lambda/geomesa-lambda-tools/pom.xml
```diff
@@ -0,0 +1,194 @@
...
+
+    <build>
+        <resources>
+            <resource>
+                <filtering>true</filtering>
+                <directory>src/main/resources</directory>
+            </resource>
+        </resources>
+        <plugins>
+           <!--<plugin>
```
aheyne: Does this not work? It might cause issues with geomesa configure if it's missing.

docs/user/lambda/install.rst
```diff
@@ -0,0 +1,267 @@
...
+requires Hadoop |hadoop_version|, but if you are using Hadoop 2.5.0 you should use the JARs
+that match the version of Hadoop you are running.
+
+There are scripts in the ``geomesa-lambda_2.11-$VERSION/bin`` directory
+(``install-hadoop-accumulo.sh``, ``install-kafka.sh``) which will install these dependencies to a target directory
+using ``wget`` (requires an internet connection).
+
+.. note::
+
+    You may have to edit ``install-hadoop-accumulo.sh`` and/or ``install-kafka.sh`` to set the
```
elahrvivaz: probably it won't work with any version other than 0.9.x... but the kafka jars aren't bundled anywhere so there's no need to recompile either way

geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/data/LambdaQueryRunner.scala
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      merged
+    }
+  }
+
+  override protected [geomesa] def getReturnSft(sft: SimpleFeatureType, hints: Hints): SimpleFeatureType = {
+    if (hints.isBinQuery) {
+      BinaryOutputEncoder.BinEncodedSft
+    } else if (hints.isArrowQuery) {
+      org.locationtech.geomesa.arrow.ArrowEncodedSft
```
elahrvivaz: it's a val in the package object, so the alternatives aren't great... you could import org.locationtech.geomesa.arrow then make it arrow.ArrowEncodedSft, but I don't really like using partial package names like that. Or you could import the whole thing and leave it unqualified here. i prefer just putting the whole path for clarity

geomesa-lambda/geomesa-lambda-tools/pom.xml
elahrvivaz: i think i just commented it out at first and never re-enabled it - I'll get it working, thanks

geomesa-lambda/geomesa-lambda-tools/src/main/scala/org/locationtech/geomesa/lambda/tools/LambdaRunner.scala
```diff
@@ -0,0 +1,50 @@
...
+    new VersionCommand,
+    new AccumuloGetSftConfigCommand,
+    new GenerateAvroSchemaCommand,
+    new AccumuloStatsAnalyzeCommand,
+    new LambdaStatsBoundsCommand,
+    new LambdaStatsCountCommand,
+    new LambdaStatsTopKCommand,
+    new LambdaStatsHistogramCommand,
+    new AddIndexCommand,
+    new ConvertCommand
```
elahrvivaz: added `Configure` and `Classpath`

docs/tutorials/geomesa-quickstart-lambda.rst
```diff
@@ -0,0 +1,141 @@
+GeoMesa Lambda Quick Start
```
elahrvivaz: added it to the tutorials - didn't want to start over on this file with the sync, so we can iron out any discrepancies with the sync next time

docs/user/lambda/index.rst
```diff
@@ -0,0 +1,23 @@
+Lambda Data Store
+=================
+
+The GeoMesa Lambda data store module is an implementation of the GeoTools ``DataStore`` that
+keeps data in two tiers, a transient tier and a persistent tier. The transient storage is backed by `Apache Kafka`_.
+Any other GeoMesa data store can be used for the persistent storage, although the current module only provides
+configuration for `Apache Accumulo`_. The Lambda data store is found in ``geomesa-lambda`` in the source distribution.
```
jnh5y: I think we should have an example use case or two for the LDS named here.  I kinda see this bit as an advert for the LDS.

elahrvivaz: I gave some loose examples in the 'overview' section. do you think I should also add something here? or expand the examples there?

PR #1594, opened 2017-06-29
GEOMESA-1563 Improves Row to Simple Feature conversions.

atallahhezbor:
Removes need for lookup by name and pattern matching. Significantly reduces time to convert RDD[Row] to an RDD[SimpleFeature] and vice versa


geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSparkSQL.scala
```diff
@@ -273,39 +285,18 @@ object SparkUtils extends LazyLogging {
...
-          case t if t == classOf[com.vividsolutions.jts.geom.MultiPolygon]     => row.getAs[MultiPolygon](name)
-          // JNH: Add Geometry types here.
-
-          case t if      classOf[Geometry].isAssignableFrom(t)    => SQLTypes.GeometryTypeInstance
-          case _                                                  => null
-        }
-        builder.set(name, value)
+  def row2Sf(nameMappings: List[(String, Int)], row: Row, builder: SimpleFeatureBuilder, id: String): SimpleFeature = {
+    nameMappings.foreach{ case (name, index) =>
+      builder.set(name, row.getAs[Object](index))
```
atallahhezbor: Based on some digging and tests I ran, it didn't seem necessary to pattern match and getAs a specific type. Getting the value as Object is sufficient since the SimpleFeatureBuilder will handle conversions.

However, there wasn't a real speed difference between pattern matching and not pattern matching, so if we'd prefer to have it for other reasons I could add it back.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSparkSQL.scala
```diff
@@ -169,11 +169,12 @@ class GeoMesaDataSource extends DataSourceRegister
     val rddToSave: RDD[SimpleFeature] = data.rdd.mapPartitions( iterRow => {
       val innerDS = DataStoreFinder.getDataStore(parameters)
       val sft = innerDS.getSchema(newFeatureName)
+      // AH: I don't believe SimpleFeatureBuilder is serializable
```
jnh5y: Meh, we can remove this comment.

PR #1606, opened 2017-07-07
GEOMESA-1915 ViewParams - using reflection to avoid specifying each QueryHint

elahrvivaz:
* Note: removes 'STRATEGY' as an alias for 'QUERY_INDEX'


jnh5y:
I like it.  Any risk in shipping it in 1.3.2?

elahrvivaz:
probably could ship it - I removed the 'STRATEGY' alias so we would just need to make sure clients aren't still using that (replaced with 'QUERY_INDEX')

jnh5y:
In terms of STRATEGY, etc what shipped in 1.3.1?  I wonder if we should take some care to be compatible, etc?

elahrvivaz:
we've supported strategy going back to the original implementation (1.2.0?) -it's technically been deprecated but probably not marked as such.

elahrvivaz:
since i don't think we're going to be adding any more query hints in 1.3.2, there's not much upside to merging this until 1.3.3/1.4.0

PR #1609, opened 2017-07-11
GEOMESA-1918 Adding sort to arrow WPS

elahrvivaz:


jnh5y:
How bad would it be to add tests for non-null sort field and orders?

elahrvivaz:
added sort tests

PR #1610, opened 2017-07-11
GEOMESA-1919 Periodically renew Kerberos TGT

jrs53:
Spawning thread to periodically renew Kerberos TGT if needed

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -9,9 +9,13 @@
 
 package org.locationtech.geomesa.accumulo.data
 
+import java.io.IOException
+import java.lang.InterruptedException
+
 import org.apache.accumulo.core.client._
 import org.apache.accumulo.core.client.admin.TableOperations
 import org.apache.accumulo.core.security.Authorizations
+import org.apache.hadoop.security.UserGroupInformation
 import org.geotools.data.Query
 import org.locationtech.geomesa.accumulo._
 import org.locationtech.geomesa.accumulo.audit.AccumuloAuditService
@@ -57,6 +61,33 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
 
   private val statsTable = GeoMesaFeatureIndex.formatSharedTableName(config.catalog, "stats")
   override val stats = new AccumuloGeoMesaStats(this, statsTable, config.generateStats)
+
+  // If on a secured cluster, create a thread to periodically renew Kerberos tgt
+  private val kerberosRenewerThread : Option[Thread] = if (UserGroupInformation.isSecurityEnabled) {
+    logger.info("Creating thread to periodically renew TGT from keytab")
+    Some(new Thread(new Runnable {
+      def run() {
+        while (true) {
+          try {
+            logger.info("Checking whether TGT needs renewing...")
+            UserGroupInformation.getCurrentUser.checkTGTAndReloginFromKeytab()
+          } catch {
+            case iox: IOException => logger.warn("Error checking and renewing tgt: " + iox.toString)
+          }
+
+          try {
+            // Sleep for 10 mins
+            Thread.sleep(10*60*1000)
+          } catch {
+            case ie: InterruptedException => logger.info("Kerberos renewer thread stopping"); return
+          }
+        }
+      }
+    }))
+  } else None // not using Kerberos
+
+  // Start thread (if it exists)
+  kerberosRenewerThread.map( _.start() )
 
   // some convenience operations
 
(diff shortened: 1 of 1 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -58,6 +62,33 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
...
+            UserGroupInformation.getCurrentUser.checkTGTAndReloginFromKeytab()
+          } catch {
+            case iox: IOException => logger.warn("Error checking and renewing tgt: " + iox.toString)
+          }
+
+          try {
+            // Sleep for 10 mins
+            Thread.sleep(10*60*1000)
+          } catch {
+            case ie: InterruptedException => logger.info("Kerberos renewer thread stopping"); return
```
jnh5y: Rather than a return, I'd be tempted to toss an atomic boolean in the Runnable and set it to false.  The while(true) becomes while(!stopped).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -236,6 +267,14 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
       lock.release()
     }
   }
+
+  override def dispose(): Unit = {
+    super.dispose()
+    kerberosRenewerThread.map( t => {
```
jnh5y: As a picky/style note, it is a convention to prefer 'foreach' over 'map' when the lambda is being used for its side-effects like here and line 90.

jnh5y:
Looks great!  Let us know how the testing goes!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -58,6 +62,33 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
...
+          try {
+            logger.info("Checking whether TGT needs renewing...")
+            UserGroupInformation.getCurrentUser.checkTGTAndReloginFromKeytab()
+          } catch {
+            case iox: IOException => logger.warn("Error checking and renewing tgt: " + iox.toString)
+          }
+
+          try {
+            // Sleep for 10 mins
+            Thread.sleep(10*60*1000)
```
jnh5y: Hmm... I wonder if there is a quick way to get the remaining time on the lease.  If so, it might be cleaner to 'schedule' a call right before the expiry to try and renew then.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
jrs53: And keep the interrupt? I know it's a bit yuck, but otherwise you'd have to wait up to 10 mins until the boolean gets tested

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
jrs53: Thanks, fill fix

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
jrs53: That's pretty much what the underlying UGI code already does:

https://github.com/apache/hadoop/blob/release-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L981

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
elahrvivaz: Instead of tracking sleeping, running, etc in your thread, you could use a scheduled executor service to run it repeatedly: `val exector = Executors.newSingleThreadScheduledExecutor(); executor.scheduleAtFixedRate(...)`
you'd want to shut it down in the dispose method.
Using an executor is generally preferred over `Thread.start` anyway, although for a simple use case like this it probably doesn't matter too much

jrs53: That looks much cleaner, thanks. Do I need to stop the executor or anything on shutdown? The docs seems a little sparse.

elahrvivaz: yeah, just call `executor.shutdown()` in the dispose method

PR #1612, opened 2017-07-12
GEOMESA-1920 Add in-memory indexing and partitioning to GeoMesa Relation

atallahhezbor:


build/cqs.tsv
```diff
@@ -116,6 +116,7 @@ net.minidev:accessors-smart	1.1	compile
 net.minidev:json-smart	2.2.1	compile
 net.sf.geographiclib:GeographicLib-Java	1.44	compile
 net.sf.trove4j:trove4j	3.0.3	compile
+org.antlr:antlr4-runtime	4.5.3	compile
```
elahrvivaz: can we manage this to match 4.5? otherwise we'll have to open a new cq

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -113,8 +113,19 @@ object SQLRules extends LazyLogging {
...
 
+    private def extractGridId(envelopes: List[Envelope], e: org.apache.spark.sql.catalyst.expressions.Expression): Option[List[Int]] = e match {
+      case And(l, r) => extractGridId(envelopes, l).orElse(extractGridId(envelopes, r))
+      case ScalaUDF(_, _, Seq(_, GeometryLiteral(_, geom)), _) => Some(SparkUtils.gridIdMapper(geom, envelopes))
+      case GeometryLiteral(_,geom) => Some(SparkUtils.gridIdMapper(geom, envelopes))
+      case _ => None
+    }
+
     override def apply(plan: LogicalPlan): LogicalPlan = {
+      println(s"Optimizer sees $plan")
```
elahrvivaz: remove, or change to logger.debug

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/PartitionedJoinExec.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.spark.Partition
+import org.apache.spark.rdd.{CartesianPartition, CartesianRDD, RDD}
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.codegen.GenerateUnsafeRowJoiner
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression, JoinedRow, UnsafeRow}
+import org.apache.spark.sql.execution.{BinaryExecNode, SparkPlan}
+import org.apache.spark.sql.execution.metric.SQLMetrics
+
+/* A more efficient CartesianRDD that only constructs
```
jahhulbert-ccri: picky can these be maybe javadoc syntax? e.g. two stars

```
/** 
  * A More efficient...
  * foobar
  */

**Comment by jahhulbert-ccri** [comment]
will review soon.

**Comment by elahrvivaz** [comment]
@atallahhezbor is this ready to go? There are a few comments that need to be addressed, and I'd like to review once more if you're done

**Comment by atallahhezbor** [comment]
@elahrvivaz It could be ready to go if we decide to stop here. There's still a few features we could add here to speed things up. Particularly in the form of code generation. Also I'm working through a different approach now that may or may not be better.

File: geomesa-spark/geomesa-spark-sql/pom.xml [hunk 4]
```diff
@@ -61,9 +61,14 @@
...
             <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-cqengine_${scala.binary.version}</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
             <artifactId>geomesa-cqengine-datastore_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
-            <scope>test</scope>
+            <scope>compile</scope>
```
jnh5y: @atallahhezbor this might be the issue.  Not sure what the best 'fix' is quite yet...

atallahhezbor: @jnh5y Same build error when changing this to provided.

What is the downside to just adding the GeoCQEngineDataStore dependency to the leaflet module with scope 'provided'? We did this for HBase Spark Runtime.

jnh5y: That sounds like a great thing to try!

atallahhezbor: @jnh5y I have tried it, and it fixes the issue, but I'm unsure if it's the proper solution.

jnh5y: Hmmm... we might have to try deploying things with and without the new memory-caching options.

It may be that to use the in-memory bits, you have to add some jars to the classpath.  If that's the case, careful documentation is our solution...

atallahhezbor: For the record, the solution I'm going for is to pull out the parts of SparkUtils that use GeoCQEngineDataStore into a separate object. That way Leaflet can use functions like sf2row without needing that dependency.

I'm also realizing that the spark-sql and hbase-spark-runtime cant both have cqengine-datastore as scope 'provided'. To handle both this and not updating the CQs, I'm adding an exclusion for antlr in the cqengine-datastore dependency.

Both of these fixes are in my next commit.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSparkSQL.scala
```diff
@@ -213,31 +376,326 @@ case class GeoMesaRelation(sqlContext: SQLContext,
...
+        partitionEnvelopes += new Envelope(xPartitionStart, xPartitionEnd, yPartitionStart, yPartitionEnd)
+      }
+    }
+    partitionEnvelopes.toList
+  }
+
+  def weightedPartitioning(rawRDD: RDD[SimpleFeature], bound: Envelope, numPartitions: Int, sampleSize: Int): List[Envelope] = {
+    implicit val CoordinateOrdering: Ordering[Coordinate] =
+      Ordering.by {
+        case (c: Coordinate) => c.x
```
elahrvivaz: same comment about `case` - since you're using this in multiple places maybe you could put it in a package object or something

geomesa-spark/geomesa-spark-sql/src/test/scala/org/locationtech/geomesa/spark/SparkSQLTestUtils.scala
```diff
@@ -88,5 +88,26 @@ object SparkSQLTestUtils {
     val fs = ds.getFeatureSource(name).asInstanceOf[SimpleFeatureStore]
     fs.addFeatures(features)
   }
+
+  def generatePoints(gf: GeometryFactory, numPoints: Int): Map[String, String] = {
+    (1 until numPoints).map { i =>
+      val x = -180 + 360 * Math.random()
```
elahrvivaz: for all these random numbers, it would be good to create a Random with a defined seed up front so that the test uses the same values each time it's run.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -113,8 +114,78 @@ object SQLRules extends LazyLogging {
       }
     }
 
+    private def extractGridId(envelopes: List[Envelope], e: org.apache.spark.sql.catalyst.expressions.Expression): Option[List[Int]] = e match {
+      case And(l, r) => extractGridId(envelopes, l).orElse(extractGridId(envelopes, r))
```
atallahhezbor: Ah you're right. If there are multiple predicates joined with `And`, then only the first one will be used. This would return the wrong results in some cases. 
I can change it to union the two grid ids instead and we wouldn't have this problem.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -113,8 +114,78 @@ object SQLRules extends LazyLogging {
...
+      case ScalaUDF(_, _, Seq(_, GeometryLiteral(_, geom)), _) => Some(RelationUtils.gridIdMapper(geom, envelopes))
+      case GeometryLiteral(_,geom) => Some(RelationUtils.gridIdMapper(geom, envelopes))
+      case _ => None
+    }
+
+    // Converts a pair of GeoMesaRelations into one GeoMesaJoinRelation
+    private def alterRelation(left: GeoMesaRelation, right: GeoMesaRelation, cond: Expression): GeoMesaJoinRelation = {
+      val joinedSchema = StructType(left.schema.fields ++ right.schema.fields)
```
atallahhezbor: ```schema.fields ``` is an ```Array[StructField]```. So using `++` should just append to the end.

In my tests there is a common field with the same name and type, and I had no problems. At the Catalyst level, attributes are bound with unique identifiers

PR #1618, opened 2017-07-14
GEOMESA-1861 Checking each server version in version-remote command

elahrvivaz:


elahrvivaz:
tested the happy path and it worked, haven't tried different jars on different tservers

jahhulbert-ccri:
ooooooo different jars on different tservers that happens alot :) would be nice if folks used vfs all the time.

elahrvivaz:
tested with multiple tservers with different jar versions and worked as expected

PR #1625, opened 2017-07-18
GEOMESA-1941 Lambda data store - reducing offset state memory footprint

elahrvivaz:

Reduces memory footprint of cache to a hashmap and a priority queue. I ran overnight and it seemed fine.

@jnh5y what do you think?


jnh5y:
Looks reasonable to me.  Honestly, some of the fine details are a little hazy to me, but I like the direction things are going.

PR #1626, opened 2017-07-18
GEOMESA-1942 Lambda data store - using deque instead of sorted queue …

elahrvivaz:
…for in-memory state

* Messages are guaranteed to come in order


geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/stream/kafka/KafkaFeatureCache.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  override def get(id: String): SimpleFeature = features.get(id)
+
+  override def all(): Iterator[SimpleFeature] = {
+    import scala.collection.JavaConverters._
+    features.values.iterator.asScala
+  }
+
+  override def add(feature: SimpleFeature, partition: Int, offset: Long, created: Long): Unit = {
+    if (offsets(partition).get < offset) {
```
jnh5y: Should we check the expiry ttl to see if we should ignore this feature?

elahrvivaz: we'd still want to put the feature in the cache if the ttl has passed, as otherwise it won't be picked up by the persistence run and written to accumulo

elahrvivaz: if the offset check fails, that means that it's *already* been written to accumulo and we don't need it in the cache

PR #1632, opened 2017-07-19
GEOMESA-1950 Lambda ds - shuffle partition processing

elahrvivaz:


geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/stream/kafka/DataStorePersistence.scala
```diff
@@ -110,10 +115,14 @@ class DataStorePersistence(ds: DataStore,
...
               toPersist.remove(updated.getID)
             }
+            count += 1
           }
+          logger.debug(s"Wrote $count updated features to persistent storage in ${System.currentTimeMillis() - start}ms")
         }
         // if any weren't updates, add them as inserts
+        start = System.currentTimeMillis()
         if (toPersist.nonEmpty) {
+          val count = toPersist.size
```
jnh5y: Is this .size 'expensive'?  I'm wondering if we can skip it since it is only used in a debug statement.

elahrvivaz: toPersist is just a scala map, I'm pretty sure size is kept as a constant

PR #1635, opened 2017-07-20
GEOMESA-1953 Conf objects per query

jahhulbert-ccri:


jahhulbert-ccri:
i think this fixes the threading issues afox was seeing

PR #1636, opened 2017-07-20
GEOMESA-1955 Adding default scala version property to support sbt builds

elahrvivaz:
* Fixed a couple scala_2.10 method calls


jnh5y:
I wonder if we should add a small SBT project to geomesa-tutorials to verify that such a project can work correctly.

PR #1637, opened 2017-07-20
GEOMESA-1958 Fixing cluster check for Bigtable data store

elahrvivaz:


jnh5y:
Let's add the BigTable profile to the Travis build...  (Or Hudson.)

elahrvivaz:
I think there was a reason we didn't do that... the jars aren't available or something?

jnh5y:
Let's try it in Travis CI.  Maybe we don't have CQs for the jars...

PR #1649, opened 2017-08-01
GEOMESA-1970 Lazily evaluated simple feature deserialization

elahrvivaz:
* Added `Lazy` option to SerializationOptions
* KryoFeatureSerializer now instantiated with `apply` instead of `new`
* Changed order of ScalaSimpleFeature constructor arguments

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #1650, opened 2017-08-04
GEOMESA-1889 Refactoring BIN code, maintaining order of BIN labels

elahrvivaz:


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
```diff
@@ -72,32 +72,54 @@ class PrecomputedBinAggregatingIterator extends BinAggregatingIterator {
...
-      (_) => {
-        val row = source.getTopKey.getRow
-        sf.setId(getId(row.getBytes, 0, row.getLength))
-        setValuesFromBin(sf, gf)
-        sf
+
+      val callback = new BinaryOutputCallback() {
+        override def apply(trackId: Int, lat: Float, lon: Float, dtg: Long): Unit = {
+          sf.setAttribute(geomIndex, gf.createPoint(new Coordinate(lat, lon)))
+          sf.setAttribute(trackIndex, Int.box(trackId)) // TODO setAttributeNoConvert
```
jnh5y: Ok, so this is a nutty place to ask this question...

I'm assuming we have a KryoSimpleFeature here... is there a sane way above boxing and write directly to memory?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
```diff
@@ -72,32 +72,54 @@ class PrecomputedBinAggregatingIterator extends BinAggregatingIterator {
     // we only need to decode the parts required for the filter/dedupe/sampling check
     // note: we wouldn't be using precomputed if sample by field wasn't the track id
     decodeBin = if (filter) {
-      setDate = if (isDtgArray) {
+      val dtgIndex = encoding.dtgField.get
+      val geomIndex = encoding.geomField.get
+      val trackIndex = encoding.trackIdField.get
```
jnh5y: How often are these Option.get's being called?  Once per request?  (If so, that might be fine.  I'm wondering if it'd make sense to resolve the Option[T]s at class construction.)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
```diff
@@ -72,32 +72,54 @@ class PrecomputedBinAggregatingIterator extends BinAggregatingIterator {
...
-        sf
+
+      val callback = new BinaryOutputCallback() {
+        override def apply(trackId: Int, lat: Float, lon: Float, dtg: Long): Unit = {
+          sf.setAttribute(geomIndex, gf.createPoint(new Coordinate(lat, lon)))
+          sf.setAttribute(trackIndex, Int.box(trackId)) // TODO setAttributeNoConvert
+          setDate(sf, dtg)
+        }
+        override def apply(trackId: Int, lat: Float, lon: Float, dtg: Long, label: Long): Unit =
+          apply(trackId, lat, lon, dtg)
```
jnh5y: Ok, just so I'm clear, this callback is not handling the 'label'?  I'm guessing that's handled elsewhere?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
```diff
@@ -251,20 +256,20 @@ object BinAggregatingIterator extends LazyLogging {
     // don't use return sft from query hints, as it will be bin_sft
     val returnSft = hints.getTransformSchema.getOrElse(sft)
 
-    val trackId = Option(hints.getBinTrackIdField)
-    val geom = hints.getBinGeomField.map(GeometryAttribute(_, AxisOrder.LonLat)) // TODO check this out...
```
jnh5y: Just to double-check, did this TODO get resolved by this PR?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
elahrvivaz: yeah, i wanted to revisit this after #1649 is merged - I think we could add some optimizations

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/BinLineStringTest.scala
```diff
@@ -36,14 +36,14 @@ class BinLineStringTest extends Specification with TestWithDataStore {
       val sf = new ScalaSimpleFeature(s"$i", sft)
       val geom = s"LINESTRING(40 6$i, 40.1 6$i, 40.2 6$i, 40.3 6$i)"
       val dates = new java.util.ArrayList[Date]
-      (0 to 4).map(mm => java.util.Date.from(java.time.LocalDateTime.parse(s"2010-05-07T0$i:0$mm:00.000Z", GeoToolsDateFormat).toInstant(java.time.ZoneOffset.UTC))).foreach(dates.add)
+      (0 until 4).map(mm => java.util.Date.from(java.time.LocalDateTime.parse(s"2010-05-07T0$i:0$mm:00.000Z", GeoToolsDateFormat).toInstant(java.time.ZoneOffset.UTC))).foreach(dates.add)
```
jnh5y: Ok, so the change from 'to' to 'until' drops out the '4' case.  Does this change in this test show the errors that this PR is trying to fix?  (From reading of the test, I'm trying to see what was happening and why it was wrong;).)

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/aggregators/HBaseBinAggregator.scala
```diff
@@ -9,24 +9,20 @@
...
 object HBaseBinAggregator {
-  def bytesToFeatures(bytes : Array[Byte]): SimpleFeature = {
-    val sf = new ScalaSimpleFeature("", BinaryOutputEncoder.BinEncodedSft)
-    sf.setAttribute(1, GeometryUtils.zeroPoint)
-    sf.setAttribute(BIN_ATTRIBUTE_INDEX, bytes)
-    sf
-  }
+
+  def bytesToFeatures(bytes : Array[Byte]): SimpleFeature =
+    new ScalaSimpleFeature("", BinaryOutputEncoder.BinEncodedSft, Array(bytes, GeometryUtils.zeroPoint))
```
jnh5y: Nice.  I don't know if this PR is the right place for it, but did you look for other places where this could be done?

One +/- here is that BIN_ATTRIBUTE_INDEX isn't used explicitly.  If the BIN_SFT is changed, we may not know to change this line...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
elahrvivaz: yeah, once per iterator set-up - they're required as Options in the super class, to pass in to the BinaryOutputEncoder. this class is a special case, b/c it needs to reverse parse them out to filter the row values

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
elahrvivaz: yeah, this class only operates on the precomputed bin values, which only support the non-label value... maybe i will just make this a throws since it would be an error case if it was ever called

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BinAggregatingIterator.scala
elahrvivaz: yeah (hope so :)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/BinLineStringTest.scala
elahrvivaz: this was an oversight in the original test i think - the number of dates is supposed to match the number of points in the line string. I started logging that case in the iterator, so the test started spitting out a bunch of warnings.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/aggregators/HBaseBinAggregator.scala
elahrvivaz: I did a brief check and didn't see any other places we could do this. the old code was hard-coded to set the geom to index 1, so seems like we would have still screwed things up if we change the BIN_SFT. really we should probably make a method in BinaryOutputEncoder so we don't have to construct the attributes anywhere else, but eh...

elahrvivaz:
if we can merge #1649 before this PR, I'll make some simple feature optimizations here

PR #1657, opened 2017-08-14
GEOMESA-1983 Shade deps per datastore

jahhulbert-ccri:


elahrvivaz:
should we shade bigtable too?

jahhulbert-ccri:
i think i shaded it as hbase...but maybe it should be separate?

elahrvivaz:
oh didn't notice - i guess let's make it `bigtable` to be consistent

jahhulbert-ccri:
cool i switched it over

PR #1659, opened 2017-08-18
GEOMESA-1986 TubeSelect Modified to Properly Handle IDL

jmglenn:
TubeSelect will now split LineStrings that cross the IDL to prevent queries from wrapping the wrong direction around the globe in the event that a LineString geometry crosses the IDL.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/process/tube/TubeSelectProcessTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/process/tube/TubeSelectProcessTest.scala
@@ -259,6 +259,50 @@ class TubeSelectProcessTest extends Specification {
 
       results.size mustEqual 20
     }
+
+    "should properly handle geometries crossing the IDL" in {
+      val sftName = "tubeTestType"
+      val sft = SimpleFeatureTypes.createType(sftName, s"type:String,$geotimeAttributes")
+
+      val ds = createStore
+
+      ds.createSchema(sft)
+      val fs = ds.getFeatureSource(sftName)
+
+      val featureCollection = new DefaultFeatureCollection(sftName, sft)
+
+      List("c", "d").foreach { name =>
+        List(1, 2, 3, 4).zip(List(-178, -179, 179, 178)).foreach { case (i, lat) =>
+          val sf = AvroSimpleFeatureFactory.buildAvroFeature(sft, List(), name + i.toString)
+          sf.setDefaultGeometry(WKTUtils.read(f"POINT($lat%d 0)"))
+          sf.setAttribute(DefaultDtgField, new DateTime(f"2011-01-01T00:0$i%d:00Z", DateTimeZone.UTC).toDate)
+          sf.setAttribute("type", name)
+          sf.getUserData()(Hints.USE_PROVIDED_FID) = java.lang.Boolean.TRUE
+          featureCollection.add(sf)
+        }
+      }
+
+      // write the feature to the store
+      val res = fs.addFeatures(featureCollection)
+
+      // tube features
+      val tubeFeatures = fs.getFeatures(CQL.toFilter("type = 'c'"))
+
+      // result set to tube on
+      val features = fs.getFeatures(CQL.toFilter("type <> 'd'"))
+
+      // get back type b from tube
+      val ts = new TubeSelectProcess()
+      val results = ts.execute(tubeFeatures, features, null, 1L, 1L, 0.0, 5, "line")
+
+      val f = results.features()
+      while (f.hasNext) {
+        val sf = f.next
+        sf.getAttribute("type") mustEqual "c"
+      }
+
+      results.size mustEqual 4
+    }
   }
 
   "TubeSelect" should {
--- a/geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
+++ b/geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
@@ -156,6 +156,56 @@ class LineGapFill(tubeFeatures: SimpleFeatureCollection,
 
   def nextId: String = id.getAndIncrement.toString
 
+  /**
+    * This function checks if a segment crosses the IDL.
+    * @param point1 The first point in the segment
+    * @param point2 The second point in the segment
+    * @return boolean representing validity of a segment based on segment length
+    */
+  def crossesIDL(point1:Coordinate,point2:Coordinate): Boolean = {
+    return (Math.abs(point1.x-point2.x) > 350)
+  }
+
+  /**
+    * Calculate the latitude at which the segment intercepts the IDL.
+    * @param point1 The first point in the segment
+    * @param point2 The second point in the segment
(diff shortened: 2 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/process/tube/TubeSelectProcessTest.scala
```diff
@@ -259,6 +259,50 @@ class TubeSelectProcessTest extends Specification {
...
+
+      // tube features
+      val tubeFeatures = fs.getFeatures(CQL.toFilter("type = 'c'"))
+
+      // result set to tube on
+      val features = fs.getFeatures(CQL.toFilter("type <> 'd'"))
+
+      // get back type b from tube
+      val ts = new TubeSelectProcess()
+      val results = ts.execute(tubeFeatures, features, null, 1L, 1L, 0.0, 5, "line")
```
elahrvivaz: what happened previously? i.e. is just successfully returning enough of a check, or should we validate the results a little more?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
```diff
@@ -156,6 +156,56 @@ class LineGapFill(tubeFeatures: SimpleFeatureCollection,
...
+
+  /**
+    * Calculate the latitude at which the segment intercepts the IDL.
+    * @param point1 The first point in the segment
+    * @param point2 The second point in the segment
+    * @return a dobule representing the intercept latitude
+    */
+  def calcIDLIntercept(point1: Coordinate, point2: Coordinate): Double = {
+    if(point2.x > point1.x){
+      return point2.y - (((point2.y-point1.y)/(point2.x-(-180+point1.x)))*point2.x)
```
elahrvivaz: for scala, we prefer to omit the 'return' keyword. in this case, you can just remove it and put the second return in an else branch

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
```diff
@@ -156,6 +156,56 @@ class LineGapFill(tubeFeatures: SimpleFeatureCollection,
...
+    //If the points cross the IDL we must generate two line segments
+    val segments = if(crossesIDL(input1, input2)){
+      //Find the latitude where the segment intercepts the IDL
+      val latIntercept = calcIDLIntercept(input1,input2)
+      val p1 = new Coordinate(-180,latIntercept)
+      val p2 = new Coordinate(180,latIntercept)
+      //This orders the points so that point1 is always the east-most point
+      val (point1, point2) = if(input1.x > 0) (input1, input2) else (input2, input1)
+      val westLine = new LineString(new CoordinateArraySequence(Array(point2, p1)), geoFac)
+      val eastLine = new LineString(new CoordinateArraySequence(Array(point1, p2)), geoFac)
```
elahrvivaz: do we care which direction the lines are pointing? i.e. which point is first and which second. seems like this way they are both pointing at the idl

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
```diff
@@ -156,6 +156,56 @@ class LineGapFill(tubeFeatures: SimpleFeatureCollection,
...
+      Array[LineString](westLine,eastLine)
+    }else{
+      val originalLine = new LineString(new CoordinateArraySequence(Array(input1, input2)), geoFac)
+      Array[LineString](originalLine)
+    }
+    new MultiLineString(segments, geoFac)
```
elahrvivaz: since we already return either points or line strings, would it make sense to only return a multi line string if needed?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
```diff
@@ -156,6 +156,56 @@ class LineGapFill(tubeFeatures: SimpleFeatureCollection,
...
   def nextId: String = id.getAndIncrement.toString
 
+  /**
+    * This function checks if a segment crosses the IDL.
+    * @param point1 The first point in the segment
+    * @param point2 The second point in the segment
+    * @return boolean representing validity of a segment based on segment length
+    */
+  def crossesIDL(point1:Coordinate,point2:Coordinate): Boolean = {
+    return (Math.abs(point1.x-point2.x) > 350)
```
elahrvivaz: this seems somewhat arbitrary... should we maybe allow it to be passed into the tubeselect process?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/process/tube/TubeSelectProcessTest.scala
jmglenn: Addressed this by adding points that would be grabbed if the IDL were handled as it has been in the past. They are now not picked up in the resulting query, proving that the change has had an effect. Previously, if a segment crossed the IDL, the resulting geometry would wrap the wrong way around the globe resulting in a buffered line that wrapped the globe. This would grab any points that occurred in the time between points within that belt. Now, because the geometry is split at the IDL, the geometry properly handles the IDL and does not pick up points outside of the intended area.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
jmglenn: You are correct. That was arbitrary. I updated the way that these IDL crossings are detected.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
jmglenn: All instances of return are now removed, else is used where appropriate. Thank you for the guidance on that.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
jmglenn: It is important, and has now been updated to maintain directionality. Thank you for catching that.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
jmglenn: This function has been updated to return a multilinestring or linestring when appropriate.

jmglenn:
Closing PR to resubmit as a merged PR addressing GEOMESA-1987,GEOMESA-1987, and GEOMESA-1993

elahrvivaz:
replaced with #1665

PR #1660, opened 2017-08-21
GEOMESA-1984,GEOMESA-1973 Kafka data store rewrite

elahrvivaz:
* Standardized data store connection keys for 1.4
* Only a single implementation for both Kafka 0.9/0.10 (dropped 0.8 support)
* Supports multiple consumers per topic
* Removes 'replay' ability


docs/tutorials/geomesa-quickstart-kafka.rst
```diff
@@ -278,21 +220,21 @@ the source code.
 
 The GeoTools API also includes a mechanism to fire off a
 `FeatureEvent <http://docs.geotools.org/stable/javadocs/index.html?org/geotools/data/FeatureEvent.Type.html>`__
-each time there is an event (typically when the data are changed) in a
+each time there is an event (typically when the data is changed) in a
```
jahhulbert-ccri: n2q0o

jahhulbert-ccri:
This seemed to do the trick in KafkaFeatureCacheTest...changing line 158:
```
 "cleanup" >> {
      val ticker = new MockTicker

      foreach(caches(Duration("100ms"), Duration("10ms"))(ticker)) { cache =>
        try {
          cache.put(track0v0)

          cache.size() mustEqual 1
          cache.query(track0v0.getID) must beSome(track0v0)
          cache.query(wholeWorldFilter).toSeq mustEqual Seq(track0v0)

          ticker.millis += 1000L

          cache.query(track0v0.getID) must eventually(40, 100.millis)(beNone)
          cache.query(wholeWorldFilter).toSeq must eventually(40, 100.millis)(beEmpty)
          cache.size() mustEqual 0
        } finally {
          cache.close()
        }
      }
    }
```

jahhulbert-ccri:
I think `bin/install-kafka.sh` needs a flag to install kafka10 deps in addition to kafka 09...also use zookeeper.version.recommended if possible in that script and a configurable nexus server. 

Wondering if we should have a ticket to read that from config? the nexus server path? or at least override it. would be nice to have that in my ENV

jahhulbert-ccri:
we should have `slf4j-log4j12` installed by default in the `lib` folder of the dist to prevent this:
```
ahulbert@doubletrouble:/tmp/geomesa-kafka_2.11-1.4.0-SNAPSHOT$ bin/geomesa-kafka 
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

jahhulbert-ccri:
i verified the command line tools seem to be working OK

elahrvivaz:
thanks, i fixed the unit test

elahrvivaz:
for `install-kafka.sh`, I added commented out versions at the top for kafka 10 - is that good enough? I think generally we haven't had different version flags in the install-* scripts, you just edit the script. I also pulled out the maven server as a variable - and added a ticket here: https://geomesa.atlassian.net/browse/GEOMESA-2038

elahrvivaz:
I think the reason we don't include log4j is that it gets picked up from KAFKA_HOME - maybe we should exclude it from there and just bundle it?
```
$ geomesa-kafka classpath | grep log
/opt/devel/src/geomesa/tools/kafka/lib/minlog-1.2.jar
/opt/devel/src/geomesa/tools/kafka/lib/geomesa-logger_2.11-1.4.0-SNAPSHOT.jar
/opt/devel/src/geomesa/tools/kafka/lib/scala-logging_2.11-3.1.0.jar
/opt/devel/install/kafka/libs/kafka-log4j-appender-0.9.0.1.jar
/opt/devel/install/kafka/libs/log4j-1.2.17.jar
/opt/devel/install/kafka/libs/slf4j-log4j12-1.7.6.jar
```

elahrvivaz:
i excluded the logging jars from kafka and included them in tools

jahhulbert-ccri:
Cool, yeah I would rather exclude logging jars since we aren't running in "their" framework. We are running in ours. If we were deploying code into the system (e.g. a spark job or NiFi) I think we would use their logging implementation then and exclude ours...the install script looks good thanks for that!

One last thing...Getting more test failures here too:
```
[2017-09-20 21:01:56,539] DEBUG FeatureCacheCqEngine: Removing feature track0 due to expiration after 100 milliseconds
[2017-09-20 21:01:56,546] DEBUG FeatureCacheGuava: Removing feature track0 due to expiration after 100 milliseconds

java.lang.Exception: There is 1 failure
Stream(ScalaSimpleFeature:track0:track0|POINT (30 30), ?) is not empty
	at org.locationtech.geomesa.kafka.index.KafkaFeatureCacheTest$$anonfun$1$$anonfun$apply$45.apply(KafkaFeatureCacheTest.scala:123)
	at org.locationtech.geomesa.kafka.index.KafkaFeatureCacheTest$$anonfun$1$$anonfun$apply$45.apply(KafkaFeatureCacheTest.scala:120)
```

I think its a similar thing...I'm thinking its the thread running in the background for the cqengine? if you step through slow it always works. and its only on the filter query not the ID queries. BTW this one was in intellij

elahrvivaz:
hmm, that last test is calling `cleanup` which should be synchronous i think... it calls `cleanup` on the first guava cache, which should trigger the removal listener, which should call `remove` on the cq engine cache...
maybe the removal listener isn't synchronous there - I'll try adding an 'eventually'

PR #1663, opened 2017-08-22
GEOMESA-1989 Handling attribute queries with CQL functions

elahrvivaz:
* Functions are currently handled by full table scans
* Removed some implicit filter methods only used in tests
* Standardizing GeoCQ QueryBuilder with FilterHelper.extract*


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/filter/TestFilters.scala
```diff
@@ -91,7 +89,7 @@ object TestFilters {
       "(attr84 = val84 AND dtg BETWEEN '0000-01-01T00:00:00.000Z' AND '9999-12-31T23:59:59.000Z' AND INTERSECTS(geom, POLYGON ((45 23, 48 23, 48 27, 45 27, 45 23))))",
       "(dtg BETWEEN '0000-01-01T00:00:00.000Z' AND '9999-12-31T23:59:59.000Z' AND INTERSECTS(geom, POLYGON ((45 23, 48 23, 48 27, 45 27, 45 23))) AND dtg BETWEEN '2010-07-01T00:00:00.000Z' AND '2010-07-31T00:00:00.000Z')",
       "(dtg BETWEEN '2010-07-01T00:00:00.000Z' AND '2010-07-31T00:00:00.000Z' AND attr92 = val92 AND INTERSECTS(geom, POLYGON ((45 23, 48 23, 48 27, 45 27, 45 23))))"
-    )
+    ).map(ECQL.toFilter)
```
jnh5y: Ok, so you didn't like the implicit;)?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -462,6 +486,9 @@ object FilterHelper {
           inverted
         } else if (inverted.disjoint) {
           FilterValues(Seq(Bounds.everything[T])) // equivalent to not null
+        } else if (!inverted.precise) {
+          FilterHelperLogger.log.warn(s"Falling back to full table scan for inverted query: '${filterToString(f)}'")
```
jnh5y: For my knowledge, what's an example of a filter which would hit here?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterValues.scala
```diff
@@ -0,0 +1,61 @@
...
+  * Holds values extracted from a filter. Values may be empty, in which case nothing was extracted from
+  * the filter. May be marked as 'disjoint', which means that mutually exclusive values were extracted
+  * from the filter. This may be checked to short-circuit queries that will not result in any hits.
+  *
+  * @param values values extracted from the filter. If nothing was extracted, will be empty
+  * @param precise values exactly match the filter, or may return false positives
+  * @param disjoint mutually exclusive values were extracted, e.g. 'a < 1 && a > 2'
+  * @tparam T type parameter
+  */
+case class FilterValues[T](values: Seq[T], precise: Boolean = true, disjoint: Boolean = false) {
```
jnh5y: (Idly) Should this implement filterMonadic or whatever?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/AttributeFilterStrategy.scala
```diff
@@ -62,7 +62,7 @@ trait AttributeFilterStrategy[DS <: GeoMesaDataStore[DS, F, W], F <: WrappedFeat
         val cost = stats.flatMap(_.getCount(sft, f, exact = false)).getOrElse {
           val binding = sft.getDescriptor(attribute).getType.getBinding
           val bounds = FilterHelper.extractAttributeBounds(f, attribute, binding)
-          if (bounds.forall(_.isExact)) {
+          if (bounds.precise && bounds.forall(_.isExact)) {
```
jnh5y: For context, what's the meaning of precise and exact here?

I suppose I'm curious about how something aould be exact and not precise.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/filter/TestFilters.scala
elahrvivaz: yeah, it was only used in the unit tests so i nuked it...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterValues.scala
elahrvivaz: is there an interface already, or you mean we should create one?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/AttributeFilterStrategy.scala
elahrvivaz: It is a bit confusing... i did doc it in the scala docs though :)
precise means that the ranges exactly match the filter, and so there won't be any false positives returned. isExact means that it's a single equality check, vs a range.
got any suggestions for better names?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
elahrvivaz: basically anything with a function, e.g. `strToUpperCase(name) = 'BILL'`. `precise` differentiates between that case, where we have to scan the whole table but aren't actually returning everything, with `is not null`, where we have to scan the whole table. If the filter isn't precise, we can't really invert it since we don't know which ranges we'd have to scan. let me know if that doesn't make sense, i wrote this a while ago...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterValues.scala
jnh5y: http://www.scala-lang.org/api/2.11.8/index.html#scala.collection.generic.FilterMonadic

Using that might allow for use of Scala's synatical sugar around monads.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/AttributeFilterStrategy.scala
jnh5y: Ok, so precise about whether or not we are querying with 'slop', and exact is about 'Gets' versus 'Scans'?

Naming things is hard.  I was mainly curious.  We can leave it as is...

Are there non-precise 'Gets'?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
jnh5y: Seems reasonable.

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/OrSplittingFilterTest.scala
```diff
@@ -26,6 +26,20 @@ class OrSplittingFilterTest extends Specification {
...
 
+  implicit class RichFilter(val filter: Filter) {
+    def &&(that: Filter) = ff.and(filter, that)
+
+    def ||(that: Filter) = ff.or(filter, that)
+
+    def ! = ff.not(filter)
+  }
+
+  implicit def stringToFilter(s: String): Filter = ECQL.toFilter(s)
```
jnh5y: Is this still used?  Or did all the implicit uses become explicit?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/AttributeFilterStrategy.scala
elahrvivaz: i don't think there are non-precise gets, although technically it is allowed... i was originally going to call exact 'isEquals' or something but that seemed confusing too...

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/OrSplittingFilterTest.scala
elahrvivaz: i copied it into the test instead of making everything explicit - it was pretty heavily used in this particular test

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterValues.scala
elahrvivaz: interesting, i'll check it out

elahrvivaz: i got lost down a rabbit hole of CanBuildFrom, etc - maybe I'll revisit later but for now it's too complicated i think

PR #1664, opened 2017-08-23
GEOMESA-1992 Pom changes for publishing to maven central

elahrvivaz:
* Use the 'central' profile to deploy to ossrh, then manually release from the sonatype UI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,6 +10,14 @@
     <packaging>pom</packaging>
     <version>1.3.4-SNAPSHOT</version>
 
+    <licenses>
+        <license>
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
     <modules>
         <module>geomesa-accumulo</module>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -249,59 +257,104 @@
...
+                                <goals>
+                                    <goal>jar</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+            <distributionManagement>
                 <repository>
```
jnh5y: I wonder if this should remain...  How does this block get used by people/programs?

pom.xml
```diff
@@ -2843,6 +2893,39 @@
         <tag>HEAD</tag>
     </scm>
 
+    <developers>
```
jnh5y: Did you copy this from the geomesa-geoserver bit?  (If so, great!)

elahrvivaz: yep :)

pom.xml
elahrvivaz: if we want to publish to locationtech we should be able to use this profile to do so. or do you mean we should keep it in the default pom?

PR #1666, opened 2017-08-25
GEOMESA-1996 ArrowConversionProcess dictionary and sorting support fo…

elahrvivaz:
…r all feature collections

* Added SimpleFeatureOrdering class


elahrvivaz:
note to self: update https://github.com/geomesa/geomesa-geoserver/blob/master/geomesa-gs-wfs/src/main/scala/org/geomesa/gs/wfs/output/ArrowOutputFormat.scala to use same code

anthonyccri:
On quick read, I didn't notice any unit test changes.  Can you add a unit test for `SimpleFeatureOrdering` or does that not make sense?

elahrvivaz:
Yeah, I'll add one. I removed a `pendingUntilFixed` in the ArrowConversionProcessTest for dictionaries, but I'll add one for sorting there too.

PR #1667, opened 2017-08-25
GEOMESA-1997 Enrichment services in converter framework

anthonyccri:
* Added enrichment caches to converter framework to load
  static data via a lookup key
* Implemented a Redis version of the lookup cache

### DIFF
--- a/build/cqs.tsv
+++ b/build/cqs.tsv
@@ -225,6 +225,7 @@ org.slf4j:slf4j-api	1.7.5	compile
 org.slf4j:slf4j-log4j12	1.7.5	compile
 org.tukaani:xz	1.0	compile
 picocontainer:picocontainer	1.2	compile
+redis.clients:jedis	2.9.0	compile
 
 com.datastax.cassandra:cassandra-driver-core	3.0.0	provided
 com.datastax.cassandra:cassandra-driver-mapping	3.0.0	provided
--- a/docs/user/convert/function_usage.rst
+++ b/docs/user/convert/function_usage.rst
@@ -567,3 +567,16 @@ delimiters for a map:
 ::
 
     { name = "numbers", transform = "parseMap('int -> string', $2, ',', '->')" }
+
+
+Enrichment Functions
+~~~~~~~~~~~~~~~~~~~~
+
+cacheLookup
+^^^^^^^^^^^
+
+Description: Looks up a value from a cache
+
+Usage: ``cacheLookup(<cacheName>, <entityKey>, <attributeKey>)``
+
+Example: ``cacheLookup('test', $id, 'name')``
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
@@ -85,8 +85,9 @@ abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureCon
     val fields = buildFields(conf)
 
     val userDataBuilder = buildUserDataBuilder(conf)
+    val cacheServices = buildCacheService(conf)
     val parseOpts = getParsingOptions(conf, sft)
-    buildConverter(sft, conf, idBuilder, fields, userDataBuilder, parseOpts)
+    buildConverter(sft, conf, idBuilder, fields, userDataBuilder, cacheServices, parseOpts)
   }
 
   protected def typeToProcess: String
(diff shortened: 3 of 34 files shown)


docs/user/convert/function_usage.rst
```diff
@@ -567,3 +567,16 @@ delimiters for a map:
 ::
 
     { name = "numbers", transform = "parseMap('int -> string', $2, ',', '->')" }
+
+
+Enrichment Functions
+~~~~~~~~~~~~~~~~~~~~
+
+cacheLookup
+^^^^^^^^
```
elahrvivaz: carrots need to match length of title to be rendered correctly

geomesa-convert/geomesa-convert-redis-cache/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-convert-common_${scala.binary.version}</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>redis.clients</groupId>
+            <artifactId>jedis</artifactId>
+            <version>2.9.0</version>
```
elahrvivaz: we'll need to create cqs for this - can you run `./build/calculate-cqs.sh` and push up the modified `./build/cqs.tsv`?

geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+    new RedisEnrichmentCache(connBuilder, url, timeout)
+  }
+}
+
+
+object RedisEnrichmentCache {
+
+  case class LookupRedis() extends TransformerFn {
+    override def names: Seq[String] = Seq("lookupRedis")
```
elahrvivaz: doesn't look like this is registered anywhere - is it redundant with 'cacheLookup' and can be removed?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -257,6 +272,7 @@ trait ToSimpleFeatureConverter[I] extends SimpleFeatureConverter[I] with LazyLog
   def inputFields: Seq[Field]
   def idBuilder: Expr
   def userDataBuilder: Map[String, Expr]
+  def caches: Map[String, EnrichmentCache]
```
elahrvivaz: maybe we should provide a default implementation here of `= Map.empty` to not break other `ToSimpleFeatureConverter` impls, or do we not care/want to force them to implement the caching?

elahrvivaz:
looks like your CLA is under `anthonyfox@ccri.com` but you're using `anthony.fox@ccri.com` here

geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
anthonyccri: Yes, I will remove it.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
anthonyccri: Hmm, we could provide a default impl but then if a custom impl went to use the caches, they would have to hunt this down.  I could see it going both ways.

build/cqs.tsv
```diff
@@ -225,6 +225,7 @@ org.slf4j:slf4j-api	1.7.5	compile
 org.slf4j:slf4j-log4j12	1.7.5	compile
 org.tukaani:xz	1.0	compile
 picocontainer:picocontainer	1.2	compile
+redis.clients:jedis	2.9.0	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=14037

jnh5y:
Uhmm... we haven't hit ticket 9997:)...

pom.xml
```diff
@@ -77,7 +77,7 @@
         <thrift.version>0.9.1</thrift.version>
         <hadoop.version>2.5.2</hadoop.version>
         <spark.version>2.0.0</spark.version>
-        <hbase.version>1.2.3</hbase.version>
+        <hbase.version>1.3.1</hbase.version>
```
elahrvivaz: this will probably require cqs... maybe we should have this as a separate pr, and/or hold off on bumping hbase versions until GM 1.4?

jnh5y: I'd suggest CQs 'now', and hold the HBase updates until 1.4.0.  

Tentatively, we could start a '1.4.x' CQ-loose-kicking PR.

PR #1672, opened 2017-08-30
GEOMESA-2003 Fixing close of stats metadata

elahrvivaz:
* Issue manifested in thread exhaustion after creating many AccumuloDataStore instances
* The Accumulo MockInstance doesn't exhibit the behavior, so hard to unit test


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloBackedMetadata.scala
```diff
@@ -82,7 +85,13 @@ trait AccumuloMetadataAdapter extends MetadataAdapter {
     CloseableIterator(scanner.iterator.map(_.getKey.getRow.copyBytes), scanner.close())
   }
 
-  override def close(): Unit = if (writerCreated) { CloseWithLogging(writer) }
+  override def close(): Unit = synchronized {
+    if (writerCreated) {
+      CloseWithLogging(writer)
+    } else {
+      writerCreated = true // indicate that we've closed and don't want to open any more resources
```
jahhulbert-ccri: its a little bit of an awkward reading for the variable name ``writerCreated``...but i can't think of a better name

jnh5y: Yeah, we almost wants a three-state thing.  Uninitialized->Open->Closed.  

Are there any concerns here with calling 'close' multiple times?  This code would allow for CloseWithLogging(writer) to be called multiple times.

elahrvivaz: hmm, i guess if you called close twice, the second time it would create the writer and then close it. i could add another variable to track that if you think it makes sense...

jnh5y: Would it create another writer?  Or would it throw new IllegalStateException("Trying to write using a closed instance")?

elahrvivaz: oh you're right - and if you had actually used the writer it would just close it a second time, which seems reasonable...

PR #1673, opened 2017-09-04
GEOMESA-1995 Resolving classpath issues with spark clearspring classes

elahrvivaz:
* Copied clearspring implementations for CountMinSketch, HyperLogLog and StreamSummary (in scala)
* Exposed underlying data structures for fast serialization, without storing class names
* Removed RichCountMinSketch to avoid classloader issues with package protected access
* Removed relocation of clearspring in spark-runtime jars


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/StatsCombiner.scala
```diff
@@ -44,27 +44,40 @@ class StatsCombiner extends Combiner with LazyLogging {
...
+          return head // couldn't parse anything... return head value and let client deal with it
+        }
+      }
+
       iter.foreach { s =>
-        try {
-          stat += serializer.deserialize(s.get)
-        } catch {
-          case e: Exception => logger.error("Error combining stats:", e)
+        try { stat += serializer.deserialize(s.get) } catch {
```
aheyne: Did you intend to use the new deserialize function here?

aheyne: Did you intend to use the new deserialize function here?

elahrvivaz: no, we'd still need a try/catch around the `+=`, and we'd have to check for null. it's a bit redundant, but was trying not to have extra checks

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/clearspring/CountMinSketch.scala
```diff
@@ -0,0 +1,204 @@
...
+  *
+  * This product includes software developed by AddThis.
+  *
+  * This product also includes code adapted from:
+  *
+  * Apache Solr (http://lucene.apache.org/solr/)
+  * Copyright 2014 The Apache Software Foundation
+  *
+  * Apache Mahout (http://mahout.apache.org/)
+  * Copyright 2014 The Apache Software Foundation
```
jnh5y: I think this requires a CQ.:(

elahrvivaz: I created a CQ here: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=14157
Used 'third party code request' but marked as modified and both source and binary

PR #1671, opened 2017-08-29
GEOMESA-1881 Accumulo option for storing timestamps instead of logica…

elahrvivaz:
…l time

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #1677, opened 2017-09-06
GEOMESA-2006 Tools command for configuring age-off iterator

elahrvivaz:
* Additional option to configure based on ingest time
* Extracted age-off classes into common parents for use with other back-ends

note: stacks on top of #1671


elahrvivaz:
need to update docs

elahrvivaz:
updated docs

jahhulbert-ccri:
how hard do you think it would be to allow ageoff with table sharing on? My concern there is that we default to table sharing on. people ingest. then can't ageoff.

elahrvivaz:
yeah you can force override it - for ingest time based age-off it would not really be a problem, but for attribute based age-off it would involve some more processing, as we'd have to check the simple feature type based on the table sharing prefix, then apply the right deserialization and attribute. not that big a deal, but for something that is running under every scan might not be desirable.
I think probably it's good to make sure the user really wants to apply the age-off to any shared feature types though - they might not realize that it will affect the shared features. by forcing them to explicitly override in a table sharing situation, we're at least making them consider it

elahrvivaz:
it would probably be good to increase the info in the exception that gets thrown to warn about shared types and specify the override

PR #1678, opened 2017-09-07
GEOMESA-2009 - Fix download links in GeoMesa nightly documentation

jnh5y:


docs/common.py
```diff
@@ -15,6 +15,7 @@
 import sys
 import os
 import shlex
+import re
```
elahrvivaz: is this import needed?

PR #1680, opened 2017-09-07
GEOMESA-2010 Arrow queries for old index implementations

elahrvivaz:
* Fixes bug with attribute + index arrow accumulo queries

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
@@ -210,37 +210,18 @@ trait AccumuloAttributeIndex extends AccumuloFeatureIndex with AccumuloIndexAdap
       lazy val dictionaries = ArrowBatchScan.createDictionaries(ds.stats, sft, filter.filter, dictionaryFields,
         providedDictionaries, hints.isArrowCachedDictionaries)
       // check to see if we can execute against the index values
+      // TODO GEOMESA-2011 support attr + value
       if (IteratorTrigger.canUseAttrIdxValues(sft, ecql, transform)) {
         val (iter, reduce, kvsToFeatures) = if (hints.getArrowSort.isDefined ||
             hints.isArrowComputeDictionaries || dictionaryFields.forall(providedDictionaries.contains)) {
           val iter = ArrowBatchIterator.configure(indexSft, this, ecql, dictionaries, hints, dedupe)
-          val reduce = Some(ArrowBatchScan.reduceFeatures(indexSft, hints, dictionaries)(_))
+          val reduce = Some(ArrowBatchScan.reduceFeatures(transform.get, hints, dictionaries)(_))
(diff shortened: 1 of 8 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
```diff
@@ -210,38 +210,19 @@ trait AccumuloAttributeIndex extends AccumuloFeatureIndex with AccumuloIndexAdap
...
-          val reduce = Some(ArrowBatchScan.reduceFeatures(indexSft, hints, dictionaries)(_))
+          val reduce = Some(ArrowBatchScan.reduceFeatures(transform.get, hints, dictionaries)(_))
           (iter, reduce, ArrowBatchIterator.kvsToFeatures())
         } else {
           val iter = ArrowFileIterator.configure(indexSft, this, ecql, dictionaryFields, hints, dedupe)
           (iter, None, ArrowFileIterator.kvsToFeatures())
         }
         val iters = visibilityIter(indexSft) :+ iter
         BatchScanPlan(filter, table, ranges, iters, cfs, kvsToFeatures, reduce, numThreads, hasDuplicates = false)
-      } else if (IteratorTrigger.canUseAttrKeysPlusValues(attribute, sft, ecql, transform)) {
```
jnh5y: Ok, is this a 'reduction' in capability?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/id/RecordQueryableIndex.scala
```diff
@@ -71,6 +74,20 @@ trait RecordQueryableIndex extends AccumuloFeatureIndex
       } else if (hints.isDensityQuery) {
         val iter = KryoLazyDensityIterator.configure(sft, this, filter.secondary, hints, dupes)
         (Seq(iter), KryoLazyDensityIterator.kvsToFeatures(), None)
+      } else if (hints.isArrowQuery) {
```
jnh5y: As a discussion point, should the 'legacy' iterators get updates to handle new capabilities like this?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
elahrvivaz: no, it was broken before but none of the tests were hitting it. this way those queries will work, albeit with a join. I added a ticket to go back and fix it

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/id/RecordQueryableIndex.scala
elahrvivaz: in this case we have customers that can't update their data to the latest format but want arrow functionality...

PR #1685, opened 2017-09-12
GEOMESA-2015 Not registering coprocessors if remote filtering is disabled

elahrvivaz:

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
@@ -18,7 +18,7 @@ import org.apache.hadoop.hbase.util.Bytes
 import org.geotools.factory.Hints
 import org.locationtech.geomesa.hbase._
 import org.locationtech.geomesa.hbase.coprocessor.aggregators._
-import org.locationtech.geomesa.hbase.coprocessor.coprocessorList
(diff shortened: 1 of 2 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -92,16 +90,12 @@ trait HBaseFeatureIndex extends HBaseFeatureIndexType
         logger.debug(s"Creating table $name")
         val descriptor = new HTableDescriptor(name)
         descriptor.addFamily(HBaseFeatureIndex.DataColumnFamilyDescriptor)
-        Option(admin.getConfiguration.get(CoprocessorHost.USER_REGION_COPROCESSOR_CONF_KEY)) match {
+        if (ds.config.remoteFilter) {
```
jnh5y: What's the default for remote filtering?  I'm wondering if the command line tools will default to setting up the coprocessors or not...

jnh5y:
@aheyne thoughts?

aheyne:
Makes sense to me.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
elahrvivaz: default is true. it checks the ds param first, then the system prop, then defaults to true

PR #1690, opened 2017-09-13
1.3 - GEOMESA-2013 Buffering heatmap image generation to avoid edge effects

elahrvivaz:
* Discarding grid snap points outside the image envelope, instead of moving to the nearest edge
* Using approximated raster instead of centroid for geometries with extents

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
@@ -7,7 +7,10 @@
  ***********************************************************************/
 
 package org.locationtech.geomesa.index.iterators
-import com.vividsolutions.jts.geom.{Coordinate, Envelope, Geometry, Point}
+import java.awt.image.BufferedImage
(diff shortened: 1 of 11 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
```diff
@@ -183,26 +191,204 @@ object DensityScan {
   /**
     * Writes a density record from a feature that has a point geometry
     */
-  private def writePoint(sf: SimpleFeature, geomIndex: Int, weight: Double, grid: GridSnap, result: DensityResult): Unit =
+  private def writePoint(geomIndex: Int, grid: GridSnap)
```
anthonyccri: This looks like a method that has two arg groups rather than a method that returns a method.

elahrvivaz: yeah it does have 2 arg groups - it just makes it slightly nicer to partially apply it on lines 155-160

PR #1696, opened 2017-09-14
GEOMESA-339 Newlines in CSV

jahhulbert-ccri:


elahrvivaz:
i thought the original issue was to handle actual newlines, e.g.
```
foo,"ba
r",baz
```

jahhulbert-ccri:
hmmm i'll claim that isn't valid csv in that case then.

jahhulbert-ccri:
but i'll try that out too :)

PR #1703, opened 2017-09-18
GEOMESA-2029 More Geometry Transformers

jahhulbert-ccri:

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
@@ -454,9 +454,16 @@ class DateFunctionFactory extends TransformerFunctionFactory {
 }
 
 class GeometryFunctionFactory extends TransformerFunctionFactory {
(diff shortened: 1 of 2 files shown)


jahhulbert-ccri:
this seems kinda superfluous but meh

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -489,13 +514,32 @@ class GeometryFunctionFactory extends TransformerFunctionFactory {
     }
   }
 
+  val multiPolygonParserFn = TransformerFn("multipolygon") { args =>
```
elahrvivaz: wouldn't these all be handled by the plain 'geometry' function? unless we're doing some special processing like we do with points where we accept (x, y)... maybe instead we should just alias 'geometry' to all the other types. Or if we're planning to eventually expand the parsing options here I'm ok with having them separate.

elahrvivaz: i guess they have the advantage of failing if the geometry type is not the expected one

PR #1706, opened 2017-09-19
GEOMESA-2024 Local ingest of distributed files

elahrvivaz:


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/HadoopDelegate.scala
```diff
@@ -0,0 +1,74 @@
...
+  import org.apache.hadoop.fs.{FileStatus, LocatedFileStatus, Path}
+
+  private val conf = new org.apache.hadoop.conf.Configuration()
+  // use the same property as FileInputFormat
+  private val recursive = conf.getBoolean("mapreduce.input.fileinputformat.input.dir.recursive", false)
+
+  // based on logic from hadoop FileInputFormat
+  override def interpretPath(path: String): Seq[FileHandle] = {
+    val p = new Path(path)
+    // TODO close filesystem?
```
elahrvivaz: @jahhulbert-ccri thoughts?

elahrvivaz:
I've tested with local files and hdfs, still need to test with s3

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/HadoopDelegate.scala
```diff
@@ -0,0 +1,74 @@
...
+    val fs = p.getFileSystem(conf)
```
elahrvivaz: it seems like it should be ok to re-use the conf object, but I'm not 100% sure

jahhulbert-ccri: yeah its OK to use them. Note that they are expensive :)

geomesa-utils/src/main/resources/org/locationtech/geomesa/utils/io/remote-prefixes.list
```diff
@@ -0,0 +1,6 @@
+adl
+hdfs
+s3n
+s3a
```
jahhulbert-ccri: I almost want to just convert s3 to s3a via configuration :)

jahhulbert-ccri: also, should these be loaded via site.xml files?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/HadoopDelegate.scala
elahrvivaz: yeah that's why i was trying to share the single one...

geomesa-utils/src/main/resources/org/locationtech/geomesa/utils/io/remote-prefixes.list
elahrvivaz: hmm, i just copied these from the list we had in code - i figured this way we could at least add new ones without recompiling.

elahrvivaz:
tested with s3a, added remote handling to convert e.g.
```
geomesa convert -s gdelt -C gdelt -m 10 's3a://gdelt-open-data/events/20130530.export.csv'
```

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/HadoopDelegate.scala
jahhulbert-ccri: hmmm i haven't been doing this. I think maybe I should now that I look at it. Seems there is some caching and deleting going on perhaps. We should look into it. I don't think it would hurt. Maybe we can troll hadoop for some more examples of how they use it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
```diff
@@ -190,4 +194,25 @@ trait OptionalIndexParam extends TypeNameParam {
...
+  var mode: RunMode = _
+}
+
+object DistributedRunParam {
+  object RunModes extends Enumeration {
+    type RunMode = Value
+    val Distributed, Local = Value
+  }
+
+  class ModeConverter(name: String) extends BaseConverter[RunMode](name) {
```
jahhulbert-ccri: +1

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
```diff
@@ -53,13 +58,13 @@ abstract class AbstractIngest(val dsParams: Map[String, String],
...
 
   /**
-   * Create a local ingestion converter
-   *
-   * @param file file being operated on
-   * @param failures used to tracks failures
-   * @return local converter
-   */
-  def createLocalConverter(file: File, failures: AtomicLong): LocalIngestConverter
+    * Create a local ingestion converter
```
jahhulbert-ccri: might want to move this back on the spacing.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/HadoopDelegate.scala
elahrvivaz: i might have to refactor things here - we need to keep the file system around to open connections with, so this would have to be a CloseableSeq or something

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
elahrvivaz: this is the 'new' scala-approved spacing, as compared to the old java style...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/PathUtils.scala
```diff
@@ -0,0 +1,88 @@
...
+object PathUtils extends FileSystemDelegate {
+
+  private val uriRegex = Pattern.compile("""\w+://.*""")
+
+  private val hadoopAvailable = Try(Class.forName("org.apache.hadoop.conf.Configuration")).isSuccess
+
+  private val localDelegate = new LocalDelegate
+
+  // delegate allows us to avoid a runtime dependency on hadoop
+  private val hadoopDelegate = if (hadoopAvailable) { new HadoopDelegate } else { null }
```
jahhulbert-ccri: cool

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
jahhulbert-ccri: hahahaha ok :) when did that start!

jahhulbert-ccri:
I did a test as well seemed to work. I think we are going to need to revisit a lot of fs.close() commands for the FSDS as well. I'm going to add some checks for that. A quick glance at the code makes it look like there could still be transfers going on in the background:

```
  @Override
  public void close() throws IOException {
    try {
      super.close();
    } finally {
      if (transfers != null) {
        transfers.shutdownNow(true);
        transfers = null;
      }
    }
  }
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
elahrvivaz: it was a while ago - I'm just a slave to intellij :)

PR #1713, opened 2017-09-22
GEOMESA-2035 Passing index values through functions instead of thread…

elahrvivaz:
…-local to allow for nested queries

* Fixes issue with arrow dictionary queries not applying z-iterator


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AttributeIndex.scala
```diff
@@ -41,7 +41,7 @@ case object AttributeIndex extends AccumuloAttributeIndex {
 }
 
 // secondary z-index
-trait AccumuloAttributeIndex extends AccumuloFeatureIndex with AccumuloIndexAdapter
+trait AccumuloAttributeIndex extends AccumuloFeatureIndex with AccumuloIndexAdapter[Unit]
```
jnh5y: Alright, I'll bite.  What's the point of passing around [Unit] key types?  

I wonder if there's a sane type-level equivalent of option[T].

elahrvivaz: yeah, the functionality isn't used in the class, so I just used Unit as the type binding...

PR #1718, opened 2017-09-29
GEOMESA-2043 Add projection support to Converter RDD Provider

atallahhezbor:

### DIFF
--- a/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala
+++ b/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala
@@ -98,7 +98,7 @@ class ConverterRecordReader extends FileStreamRecordReader with LazyLogging {
 
     import scala.collection.JavaConversions._
     val featureReader = if (retypedSpec != null) {
-      val retypedSft = SimpleFeatureTypes.createType("retyped", retypedSpec)
+      val retypedSft = SimpleFeatureTypes.createType(sft.getTypeName, retypedSpec)
       val reader = new DelegateSimpleFeatureReader(sft, new DelegateSimpleFeatureIterator(iter))
       new ReTypeFeatureReader(reader, retypedSft)
     } else {
--- a/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala
+++ b/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala
@@ -15,9 +15,13 @@ import com.typesafe.scalalogging.LazyLogging
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{Path, Seekable}
 import org.apache.hadoop.mapreduce._
+import org.geotools.data.simple.DelegateSimpleFeatureReader
+import org.geotools.data.ReTypeFeatureReader
+import org.geotools.feature.collection.DelegateSimpleFeatureIterator
 import org.geotools.filter.text.ecql.ECQL
 import org.locationtech.geomesa.convert.SimpleFeatureConverters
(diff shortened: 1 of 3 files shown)


geomesa-spark/geomesa-spark-converter/src/main/scala/org/locationtech/geomesa/spark/converter/ConverterSpatialRDDProvider.scala
```diff
@@ -66,7 +67,9 @@ class ConverterSpatialRDDProvider extends SpatialRDDProvider with LazyLogging {
     conf.set(FileInputFormat.INPUT_DIR, job.getConfiguration.get(FileInputFormat.INPUT_DIR))
 
     if (query.getPropertyNames != null && query.getPropertyNames.length > 0) {
-      logger.warn("Ignoring query transform - modify converter definition instead")
+      logger.info("Query transform retyping results")
```
elahrvivaz: this should probably be debug level

geomesa-spark/geomesa-spark-converter/src/test/scala/org/locationtech/geomesa/spark/converter/ConverterSpatialRDDProviderTest.scala
```diff
@@ -70,5 +71,17 @@ class ConverterSpatialRDDProviderTest extends Specification {
       val rdd = GeoMesaSpark(params).rdd(new Configuration(), sc, params, new Query("example-csv"))
       rdd.count() mustEqual 3l
     }
+
+    "handle projections" in {
+      val params = Map (
+        InputFilesKey    -> getClass.getResource("/example.csv").getPath,
```
elahrvivaz: [very picky] spacing

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala
```diff
@@ -86,13 +96,25 @@ class ConverterRecordReader extends FileStreamRecordReader with LazyLogging {
       case None    => raw
     }
 
+    import scala.collection.JavaConversions._
+    val featureReader = if (retypedSpec != null) {
+      val retypedSft = SimpleFeatureTypes.createType("retyped", retypedSpec)
```
elahrvivaz: should we keep the original sft name here?

atallahhezbor: I wasn't quite sure if doing so would cause a conflict later. Or if I should concat the two.

elahrvivaz: hmm, i think it would be ok... we usually don't cache anything solely by sft name. alternatively we could store the retyped sft type name in the config like we do with the regular sft - but i think in our query planning it just re-uses the same name.

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala
```diff
@@ -86,13 +96,25 @@ class ConverterRecordReader extends FileStreamRecordReader with LazyLogging {
...
         s"type name ${sft.getTypeName} and convert conf $confStr")
 
     new Iterator[SimpleFeature] with Closeable {
-      override def hasNext: Boolean = iter.hasNext
-      override def next(): SimpleFeature = iter.next
-      override def close(): Unit = converter.close()
+      override def hasNext: Boolean = featureReader.hasNext
+      override def next(): SimpleFeature = featureReader.next
+      override def close(): Unit = {
+        featureReader.close()
```
jahhulbert-ccri: I prefer to encase these with CloseWithLogging() or CloseQuietly() from geomesa utils

PR #1728, opened 2017-10-03
GEOMESA-2047 Jedis is not thread safe

anthonyccri:
* Using JedisPool to manage connections


geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
```diff
@@ -49,11 +41,11 @@ class RedisEnrichmentCache(connBuilder: RedisConnectionBuilder, url: String,
     builder
       .build(new CacheLoader[String, KV] {
         override def load(k: String): KV = {
-          val conn = connPool.borrowObject(url)
+          val conn = jedisPool.getConn
           try {
             conn.hgetAll(k)
           } finally {
-            connPool.returnObject(url, conn)
+            conn.close()
```
elahrvivaz: will it be able to re-use connections if you close them here?

jahhulbert-ccri: yeah it doesn't actually close them it actually returns them close method....highly inaccurate

PR #1745, opened 2017-10-11
GEOMESA-2058 Fixing serialization of HBase z-filter

elahrvivaz:
* Moves z filters configuration into index-api
* Includes a breaking change in the Accumulo z-iterator config
* Removes z-iterator from deprecated non-point z2/z3 indices

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
@@ -14,30 +14,25 @@ import com.google.common.primitives.{Bytes, Longs}
 import org.apache.accumulo.core.data.{ByteSequence, Key, Range, Value}
 import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
 import org.apache.hadoop.io.Text
+import org.geotools.filter.text.ecql.ECQL
 import org.junit.runner.RunWith
-import org.locationtech.geomesa.accumulo.index.legacy.z3.Z3IndexV2
-import org.locationtech.geomesa.curve.{LegacyZ3SFC, TimePeriod}
-import org.locationtech.sfcurve.zorder.Z3
+import org.locationtech.geomesa.curve.{TimePeriod, Z3SFC}
+import org.locationtech.geomesa.index.index.z3.Z3IndexKeySpace
+import org.locationtech.geomesa.index.utils.ExplainNull
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
 
-import scala.collection.JavaConversions._
-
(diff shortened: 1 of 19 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
```diff
@@ -55,19 +50,16 @@ class Z3IteratorTest extends Specification {
...
 
-    val sfc = LegacyZ3SFC(TimePeriod.Week)
-
-    "iterate on points" >> {
-      val (xmin, ymin, tmin) = sfc.index(lx, ly, lt).decode
-      val (xmax, ymax, tmax) = sfc.index(ux, uy, ut).decode
-
-      val iter = new Z3Iterator
+    val sft = SimpleFeatureTypes.createType("z3IteratorTest", "dtg:Date,*geom:Point:srid=4326")
+    val filter = ECQL.toFilter("bbox(geom, -78, 38, -75, 40) AND dtg DURING 1970-01-01T00:05:00.000Z/1970-01-01T00:15:00.000Z")
```
jnh5y: Would it make sense to add a case or three to this test which has a duration longer than the 2x the Z3.period?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseQueryPlan.scala
```diff
@@ -63,8 +64,8 @@ object HBaseQueryPlan {
     import scala.collection.JavaConversions._
     filter match {
       case f: FilterList    => f.getFilters.map(filterToString).mkString(", ")
-      case f: Z3HBaseFilter => s"Z3HBaseFilter[xy: (${f.filter.xyvals.map(_.mkString(",")).mkString(")(")}), times ${f.filter.minEpoch} to ${f.filter.maxEpoch}: (${f.filter.tvals.map(_.map(_.mkString(",")).mkString(" ")).mkString(")(")})]"
-      case f: Z2HBaseFilter => s"Z2HBaseFilter[xy: (${f.filter.xyvals.map(_.mkString(",")).mkString(")(")})]"
+      case f: Z3HBaseFilter => s"Z3HBaseFilter[${Z3Filter.serializeToStrings(f.filter).mkString(",")}]"
+      case f: Z2HBaseFilter => s"Z2HBaseFilter[${Z2Filter.serializeToStrings(f.filter).mkString(",")}]"
```
jnh5y: +1

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
elahrvivaz: i'm not crazy about this test because you have to construct everything by hand, which means you have to set everything up correctly. seems like it's tested better in the full query tests, which do already cover that scenario

PR #1748, opened 2017-10-13
GEOMESA-2060 Upgrade to caffeine 2.5.6

elahrvivaz:

### DIFF
--- a/build/cqs.tsv
+++ b/build/cqs.tsv
@@ -9,7 +9,7 @@ com.esotericsoftware:minlog	1.3.0	compile
 com.fasterxml.jackson.core:jackson-annotations	2.6.5	compile
 com.fasterxml.jackson.core:jackson-core	2.6.5	compile
 com.fasterxml.jackson.core:jackson-databind	2.6.5	compile
-com.github.ben-manes.caffeine:caffeine	2.3.3	compile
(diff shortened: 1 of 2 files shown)


build/cqs.tsv
```diff
@@ -9,7 +9,7 @@ com.esotericsoftware:minlog	1.3.0	compile
 com.fasterxml.jackson.core:jackson-annotations	2.6.5	compile
 com.fasterxml.jackson.core:jackson-core	2.6.5	compile
 com.fasterxml.jackson.core:jackson-databind	2.6.5	compile
-com.github.ben-manes.caffeine:caffeine	2.3.3	compile
+com.github.ben-manes.caffeine:caffeine	2.5.6	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=14554

elahrvivaz: approved

jahhulbert-ccri: just for fun...what does caffeine track in terms of deps? do any other libs we use have it?

elahrvivaz: I don't think any other libs we use have it but I haven't checked thoroughly...
and it actually doesn't have any transitive deps :D

PR #1767, opened 2017-11-07
GEOMESA-2064 Updating Bigtable jar versions, fixing dependencies script

elahrvivaz:


elahrvivaz:
we could merge this now if we're planning to do full end-to-end testing before the 1.4 release. I think this is at least less likely to be broken than current master, which has a mix of hbase 1.2 and 1.3 jars.

jnh5y:
@elahrvivaz I merged this since the Awaiting Testing tag was gone:).  

Pre-1.4.0, I imagine we should kick the tires...

PR #1770, opened 2017-11-08
GEOMESA-2080 General improvements to Spark SQL

atallahhezbor:

### DIFF
--- a/geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSparkSQL.scala
+++ b/geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSparkSQL.scala
@@ -172,8 +172,14 @@ class GeoMesaDataSource extends DataSourceRegister
 
 
     val ds = DataStoreFinder.getDataStore(parameters)
-    sft.getUserData.put("override.reserved.words", java.lang.Boolean.TRUE)
-    ds.createSchema(sft)
+    val schemaInDs = ds.getTypeNames.contains(newFeatureName)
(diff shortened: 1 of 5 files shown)


geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSparkSQL.scala
```diff
@@ -172,8 +172,14 @@ class GeoMesaDataSource extends DataSourceRegister
...
 
     val ds = DataStoreFinder.getDataStore(parameters)
-    sft.getUserData.put("override.reserved.words", java.lang.Boolean.TRUE)
-    ds.createSchema(sft)
+    val schemaInDs = ds.getTypeNames.contains(newFeatureName)
+    val schemaMatchesDs = schemaInDs && ds.getSchema(newFeatureName) == sft.getTypes
+
+    if (!schemaMatchesDs) {
+      sft.getUserData.put("override.reserved.words", java.lang.Boolean.TRUE)
+      ds.createSchema(sft)
```
elahrvivaz: i think if the schema exists but doesn't match the rdd, that should be an error condition. or we could try doing something complicated to match the rdd to the schema. but re-creating the schema is usually a no-op if it exists, and then it will be inserting data that doesn't match the schema.

atallahhezbor: Got it. But I should still avoid creating the schema if it exists and matches the rdd, since at the moment we can't extract the partition scheme and that would throw an error.

elahrvivaz: right

PR #1774, opened 2017-11-10
GEOMESA-2087 Command line ingest from stdin

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/LocalDelegate.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/LocalDelegate.scala
@@ -68,13 +68,9 @@ object LocalDelegate {
   }
 
   class StdInHandle extends FileHandle {
-    private val is: CountingInputStream = new CountingInputStream(System.in)
-    private val opened = new AtomicBoolean(false)
     override def path: String = "<stdin>"
-    override def length: Long = is.getCount + Try(is.available()).getOrElse(0) // .available will throw if stream is closed
-    override def open: InputStream = if (opened.compareAndSet(false, true)) { is } else {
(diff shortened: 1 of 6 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/LocalDelegate.scala
```diff
@@ -57,9 +60,27 @@ class LocalDelegate extends FileSystemDelegate {
...
+
   class LocalFileHandle(file: File) extends FileHandle {
     override def path: String = file.getAbsolutePath
     override def length: Long = file.length()
     override def open: InputStream = new FileInputStream(file)
   }
+
+  class StdInHandle extends FileHandle {
+    private val is: CountingInputStream = new CountingInputStream(System.in)
+    private val opened = new AtomicBoolean(false)
```
jahhulbert-ccri: should open default to false if you are opening a new input stream above? seems that you could call length before calling open?

elahrvivaz:
yeah it should work for all tools. i tested it like:
`gunzip gdelt/20160901.export.CSV.gz | geomesa ingest ... `
I'll add something quick to the docs. I'm planning to reorganize some of the tools docs before 1.4, since right now they are heavily accumulo-centric - I'll add some additional information then too.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/fs/LocalDelegate.scala
elahrvivaz: this is a bit of a hack to shoehorn `stdin` to the existing file-based model - in order to track the length read (used for % complete in the ingest command), i wanted to keep a single instance of the CountingInputStream. I think i'll move that logic to the ingest command though, and not have it here, since it doesn't quite fit with the `FileHandle` api.

PR #1776, opened 2017-11-13
GEOMESA-2002,GEOMESA-2088 Standardizing index=true to mean full indic…

elahrvivaz:
…es in AccumuloDataStore

* Added docs on table sharing

### DIFF
--- a/docs/user/accumulo/data_management.rst
+++ b/docs/user/accumulo/data_management.rst
@@ -13,11 +13,6 @@ Accumulo Attribute Indices
 See :ref:`attribute_indices` for an overview of attribute indices. The Accumulo data store extends the
 normal attribute indices with an additional 'join' format that stores less data.
 
-.. warning::
-
-    Accumulo data stores use the reduced 'join' format by default. In order to use 'full' indices,
-    the keyword ``full`` must be used in the ``SimpleFeatureType`` as described below.
-
 Join Indices
 ^^^^^^^^^^^^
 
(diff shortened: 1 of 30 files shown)


docs/user/accumulo/data_management.rst
```diff
@@ -176,105 +155,8 @@ GeoMesa often makes updates to indexing formats to improve query and write perfo
 the index format for a given schema is fixed when it is first created. Updating GeoMesa versions
 will provide bug fixes and new features, but will not update existing data to new index formats.
 
-The following tables show the different indices available for different versions of GeoMesa. If not
-known, the schema version for a feature type can be checked by examining the user data value
-``geomesa.version``, or by scanning the Accumulo catalog table for ``version``. For GeoMesa schemas
-created with 1.2.7 or later, the version of each index is tracked separately and the overall schema
-version is no longer maintained.
-
-Schema version 4 - 1.0.0-rc.7
```
jnh5y: Does this PR deprecate/delete the 1.0 indexing schemes?  

Should we provide some details for someone who might want to move from 1.0 to newer versions?

elahrvivaz: It doesn't change the code, I just thought it wasn't really relevant to keep in the docs anymore...
We haven't provided any upgrade guides for those versions so far, we could always add something if you think it's useful but not sure how applicable it will be.

jnh5y: Let's not fuss with it.  I'm interested in upping our game with migration guides, etc.  

That said, 1.0.x is likely not used much...

PR #1778, opened 2017-11-16
GEOMESA-2085 Support for Kafka 1.0.0

elahrvivaz:

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/EmbeddedKafka.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/EmbeddedKafka.scala
@@ -33,6 +33,11 @@ class EmbeddedKafka extends Closeable {
 
   val brokers = s"127.0.0.1:${server.socketServer.boundPort()}"
 
+  // for kafka 1.0.0:
+  // import org.apache.kafka.common.network.ListenerName
+  // import org.apache.kafka.common.security.auth.SecurityProtocol
+  // val brokers = s"127.0.0.1:${server.socketServer.boundPort(ListenerName.forSecurityProtocol(SecurityProtocol.PLAINTEXT))}"
+
   override def close(): Unit = {
     try { server.shutdown() } catch { case _: Throwable => }
     try { zookeeper.shutdown() } catch { case _: Throwable => }
--- a/pom.xml
+++ b/pom.xml
@@ -91,9 +91,10 @@
         <hbase.guava.version>12.0.1</hbase.guava.version>
         <cassandra.version>3.0.0</cassandra.version>
         <cassandra.guava.version>18.0</cassandra.guava.version>
-        <kafka.08.version>0.8.2.1</kafka.08.version>
-        <kafka.09.version>0.9.0.1</kafka.09.version>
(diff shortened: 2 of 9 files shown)


jnh5y:
Does this change any dependencies?  Also, does anything need to change in GeoMesa-NiFi?

elahrvivaz:
doesn't change any dependencies - you can just plug in kafka 1.0 at runtime now. Nifi shouldn't need any changes either.

elahrvivaz:
oops, it did change one test dependency - pushed the new cq file

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/EmbeddedKafka.scala
```diff
@@ -33,6 +33,11 @@ class EmbeddedKafka extends Closeable {
 
   val brokers = s"127.0.0.1:${server.socketServer.boundPort()}"
 
+  // for kafka 1.0.0:
+  // import org.apache.kafka.common.network.ListenerName
+  // import org.apache.kafka.common.security.auth.SecurityProtocol
+  // val brokers = s"127.0.0.1:${server.socketServer.boundPort(ListenerName.forSecurityProtocol(SecurityProtocol.PLAINTEXT))}"
+
```
jnh5y: Hmm?  The test has to change with 1.0.0?!

pom.xml
```diff
@@ -91,9 +91,10 @@
         <hbase.guava.version>12.0.1</hbase.guava.version>
         <cassandra.version>3.0.0</cassandra.version>
         <cassandra.guava.version>18.0</cassandra.guava.version>
-        <kafka.08.version>0.8.2.1</kafka.08.version>
-        <kafka.09.version>0.9.0.1</kafka.09.version>
-        <kafka.10.version>0.10.0.1</kafka.10.version>
+        <kafka.version>0.9.0.1</kafka.version> <!-- also supports: 0.10.2.1, 1.0.0 -->
+        <zkclient.version>0.7</zkclient.version> <!-- corresponding versions: 0.10, 0.10 -->
+        <kafka.jopt.version>3.2</kafka.jopt.version> <!-- corresponding versions: 4.9, 5.0.4 -->
```
jnh5y: Should we document how developers can manage all the versions?  I *think* I'd know what to do, but I doubt that it is clear to an outside developer.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/EmbeddedKafka.scala
elahrvivaz: yeah, but in general you wouldn't ever need to compile with other versions of kafka - i just left it in there for the future

jnh5y: Roger that.

pom.xml
elahrvivaz: you shouldn't actually need to change the versions, as it's a provided dep you plug in at runtime. but looks like the kakfa datastore docs need some updating, so I'll add it in there as part of the big docs pr

PR #1783, opened 2017-11-22
GEOMESA-1584 Optimized deletes for Filter.INCLUDE

elahrvivaz:

### DIFF
--- a/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
+++ b/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
@@ -60,12 +60,19 @@ trait CassandraFeatureIndex extends CassandraFeatureIndexType with ClientSideFil
     }
   }
 
+  override def removeAll(sft: SimpleFeatureType, ds: CassandraDataStore): Unit = {
+    val tableName = getTableName(sft.getTypeName, ds)
+    val truncate = s"TRUNCATE $tableName"
+    logger.info(truncate)
+    ds.session.execute(truncate)
+  }
+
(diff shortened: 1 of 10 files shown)


geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
```diff
@@ -60,12 +60,19 @@ trait CassandraFeatureIndex extends CassandraFeatureIndexType with ClientSideFil
...
+    val tableName = getTableName(sft.getTypeName, ds)
+    val truncate = s"TRUNCATE $tableName"
+    logger.info(truncate)
+    ds.session.execute(truncate)
+  }
+
   override def delete(sft: SimpleFeatureType, ds: CassandraDataStore, shared: Boolean): Unit = {
     if (shared) {
-      throw new NotImplementedError() // TODO
+      throw new NotImplementedError("Cassandra tables are never shared")
```
jnh5y: What happens if you set SFT table.sharing hint and create that in C* GeoMesa?  Does it throw an exception at schema creation?

jnh5y:
Looks good.  One quick question about C* support.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraFeatureIndex.scala
elahrvivaz: No, current behavior is that we just set table sharing to false - we don't check to see if the user has set anything: https://github.com/locationtech/geomesa/blob/f1901ea0d51febb19f0784cf27265c1a080a52f6/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraDataStore.scala#L44

PR #1784, opened 2017-11-22
GEOMESA-1583 Handle Query.ALL

elahrvivaz:


jahhulbert-ccri:
thinking out loud (clacky keyboard)  - do we really want to reject Query.ALL explicitly? is there a historical reason? I'm thinking that in some cases (like storage of a result set) we may want to.

elahrvivaz:
mainly just b/c we'd have to have special code to handle it sprinkled throughout the code base. the main bugaboos are: it's not editable (generally shouldn't be a problem as we clone the query object), and it doesn't have a type name. if we want to support it, we certainly could.

elahrvivaz:
another point, it is supposed to query the 'default feature type' for a data store, but we don't generally have a default feature type. in some circumstances (like if passed to a feature collection or feature store) the feature type is already defined, and could be used. But for things like DataStore.getFeatureReader it's potentially ambiguous.

jahhulbert-ccri:
cool :+1:  Seems that the proper thing to do will be

PR #1790, opened 2017-12-06
GEOMESA-2099 Converter function for reprojecting geometries

elahrvivaz:


docs/user/convert/function_usage.rst
```diff
@@ -300,6 +300,24 @@ Example: Parsing GeoJson geometry
...
     }
 
+reproject
+^^^^^^^^^
+
+Description: Project a geometry from its native CRS to EPSG:4326
+
+Usage: ``reproject($0,'EPSG:3857')``
+
+Example: Reprojecting a parsed point from EPSG:3857 to EPSG:4326:
```
jnh5y: I think it'd be worth looking at the semantics around other reproject methods.  Since other systems don't awesome 4326, they might do things a little differently; I'm wondering if that difference will cause trouble/confusion.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
```diff
@@ -540,6 +545,25 @@ class GeometryFunctionFactory extends TransformerFunctionFactory {
...
+
+    override val names: Seq[String] = Seq("reproject")
+
+    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = {
+      import org.locationtech.geomesa.utils.geotools.CRS_EPSG_4326
+
+      val geom = args(0).asInstanceOf[Geometry]
+      val epsg = args(1).asInstanceOf[String]
+      val lenient = if (args.length > 2) { java.lang.Boolean.parseBoolean(args(2).toString) } else { true }
+      // assuming here that transforms are not thread safe - can't find anything to say if they are or not
```
jnh5y: Might be worth asking the GeoTools community.

docs/user/convert/function_usage.rst
elahrvivaz: do you know some other projects I could look at? Main thing I've seen is GDAL, but that's command line so the semantics don't really translate well...

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Transformers.scala
elahrvivaz: According to Jody they are thread safe, so I'll update it: http://osgeo-org.1560.x6.nabble.com/Are-MathTransform-implementations-thread-safe-td5129781.html

docs/user/convert/function_usage.rst
jnh5y: From GeoTools there is this: http://docs.geotools.org/stable/userguide/library/referencing/crs.html.  

I suppose I'm most uncomfortable with the fact that for most reprojections, a user has to give a source and target CRS.  Since GeoMesa only does 4326, there is a clear target.  

I suppose I'm looking for a way to reduce a user's surprise at GeoMesa's insistence on 4326.:)

elahrvivaz: maybe we should change the function name to something like `projectFrom` ? I didn't want to allow it to project from/to any CRS, since that would likely just cause more confusion... I can also add a note here that geomesa only supports 4326.

jahhulbert-ccri:
@jnh5y you good on this one?

jnh5y:
@jahhulbert-ccri Sure.  I think the function naming is hard here since we are always projecting to 4326.

PR #1795, opened 2017-12-19
GEOMESA-2109 ObjectType matching for specific geometry bindings

elahrvivaz:
* Removes some unused filter methods


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -205,8 +204,7 @@ object RichSimpleFeatureType {
 
     def getBinTrackId: Option[String] = sft.getAttributeDescriptors.find(_.isBinTrackId()).map(_.getLocalName)
 
-    def getSchemaVersion: Int =
-      userData[String](SCHEMA_VERSION_KEY).map(_.toInt).getOrElse(CURRENT_SCHEMA_VERSION)
+    def getSchemaVersion: Int = userData[String](SCHEMA_VERSION_KEY).map(_.toInt).getOrElse(10)
```
jnh5y: Hmm...  the '10' here is CURRENT_SCHEMA_VERSION, no? 

I'm a bit weary of moving between a well-commented object to a magic number:).

Thoughts?

geomesa-utils/src/main/scala/org/locationtech/geomesa/geomesa.scala
```diff
@@ -1,25 +0,0 @@
...
-  // 2  -> sorted keys in the STIDX table
-  // 3  -> skipped for integration
-  // 4  -> kryo encoded index values
-  // 5  -> z3 index (1.1.0-rc.1 through 1.1.0-rc.2)
-  // 6  -> attribute indices with dates (1.1.0-rc.3 through 1.2.0)
-  // 7  -> z3 polygons (1.2.1)
-  // 8  -> z2 index, deprecating stidx (1.2.2 through 1.2.4)
-  // 9  -> per-attribute-visibilities, not serializing ID in value (not released separately)
-  // 10 -> XZ2, XZ3 (1.2.5+)
-  val CURRENT_SCHEMA_VERSION = 10
```
jnh5y: I wonder if anyone is using this variable to figure out which version of GeoMesa's schema they are working with...

jnh5y:
This is good to go after the bits about CURRENT_SCHEMA_VERSION are discussed/sorted out.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: yeah, 10 was the last ever value for CURRENT_SCHEMA_VERSION - CURRENT_SCHEMA_VERSION stopped being used back 1.3.0-ish time frame, when we switched to tracking the version of each index separately. I should probably just remove this whole method actually, or at least mark it deprecated... let me look into that.

geomesa-utils/src/main/scala/org/locationtech/geomesa/geomesa.scala
elahrvivaz: As mentioned in my other comment, this stopped being using around 1.3.0-ish (maybe earlier given it was last updated in 1.2.5). I am fairly certain no one would be using this, as it's an internal variable - but if they are, then they definitely shouldn't be any more ;)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: For now, I've commented the magic number here and marked the method as deprecated. I opened a ticket to remove it fully: https://geomesa.atlassian.net/browse/GEOMESA-2114. Does that sound ok to you?

PR #1796, opened 2017-12-19
GEOMESA-2110 Method profiling refactoring to expose results of profil…

elahrvivaz:
…ed code


jnh5y:
I'm a little unsure about having the implicit output in the profile signature:
 def profile[R](code: => R)(implicit onComplete: (R, Long) => Unit): R = { ... }

Outside of that concern, this PR looks good.

elahrvivaz:
It makes the profiling code less obtrusive, IMO. you can do things like:
```
def foo(...): Int = {
  implicit def complete(count: Int, time: Long): Unit = logger.trace(s"read $count records in $time ms")
  profile {
    ... // calculate count
  }
}
```

Previously you had to do:
```
def foo(...): Int = {
  implicit val timing: Timing = new Timing
  val count = profile {
    ... // calculate count
  }
  logger.trace(s"read $count records in ${timing.time} ms")
  count
}
```

Are you unsure about anything in particular, or just the general idea?

elahrvivaz:
another example with an explicit anonymous onComplete:
```
profile(...) { (count, time) => logger.trace(s"read $count records in $time ms") }
```

elahrvivaz:
conflicts fixed, thanks

PR #1806, opened 2018-01-04
GEOMESA-2126 Fixing HBase scans for ID queries

elahrvivaz:


elahrvivaz:
note: unit tests passed with the old code, but it didn't work against an actual HBase instance :/

PR #1812, opened 2018-01-08
GEOMESA-1963 Combining convert modules

elahrvivaz:
* New merged module is geomesa-convert-core, replaces geomesa-convert-all and all other individual modules except osm


elahrvivaz:
should i bother to rebase this, or do we want to keep the modules separate? either way is fine with me...

jahhulbert-ccri:
i'll look again.

anthonyccri:
I think I'd prefer not to combine all of these.  Can't you use the `geomesa-convert-all_2.11` module if you want to bring in everything?  I like having these separate so we can limit the transitive dependency impact.

elahrvivaz:
closing as won't fix

PR #1822, opened 2018-01-15
GEOMESA-2142 Fixing --index flag in CLI export/explain

elahrvivaz:

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
@@ -14,13 +14,10 @@ import java.util.regex.Pattern
 
 import com.beust.jcommander.converters.BaseConverter
 import com.beust.jcommander.{Parameter, ParameterException}
-import org.locationtech.geomesa.index.api.{GeoMesaFeatureIndex, WrappedFeature}
-import org.locationtech.geomesa.index.geotools.GeoMesaDataStore
 import org.locationtech.geomesa.tools.DistributedRunParam.ModeConverter
 import org.locationtech.geomesa.tools.DistributedRunParam.RunModes.RunMode
 import org.locationtech.geomesa.tools.utils.DataFormats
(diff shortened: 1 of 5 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
```diff
@@ -169,38 +166,18 @@ trait ConverterConfigParam {
...
-        // just index name
-        indices.find(_.name.equalsIgnoreCase(name))
-      }
-      matched.getOrElse {
-        throw new ParameterException(s"Specified index ' $index' not found. " +
-        s"Available indices are: ${indices.map(_.identifier).mkString(", ")}")
-      }
-    }
-  }
+  var index: String = _
```
jahhulbert-ccri: i'm about to do testing...but wondering what happens now when you pass in an index that doesn't exist?

jahhulbert-ccri: so it doesn't throw an exception if the index does not exist...perhaps it should do that?

elahrvivaz: yeah, in the log you get:
```
2018-01-17 22:59:47,421 ERROR [org.locationtech.geomesa.accumulo.index.AccumuloQueryPlanner] Ignoring invalid strategy name: foo. Valid values are z3, z3:5, z2, z2:4, records, records:2
```
should we just throw an exception there? that would affect other queries too, but arguably that would be a good thing

jahhulbert-ccri: yeah i was thinking throw an exception since it's an input error.

elahrvivaz: changed it to throw an exception

jahhulbert-ccri: yay!

jahhulbert-ccri:
verified it throws exception...

PR #1824, opened 2018-01-15
GEOMESA-2144 Adding attribute type encoding for java.sql.Date

elahrvivaz:


elahrvivaz:
haha, busted. i'll add one

elahrvivaz:
in order to make `encode(create("foo:Timestamp")) == "foo:Timestamp"`, I modified the type binding for `"Timestamp"` to `java.sql.Timestamp`. I believe that should be ok, as we should be doing type binding checks for date based on `.isAssignableFrom`. however, it may introduce some subtle bugs if people are defining types as `Timestamp`.

PR #1827, opened 2018-01-17
Added SIGINT handling for geomesa-fs commands

aheyne:


elahrvivaz:
i'm not sure about this... @jahhulbert-ccri what do you think?

aheyne:
I can agree, I think the real solution would be to:
A. Add a reduce task that does what the driver program does. 
B. Make it so driver program can recover/re-attach
C. Keep more metadata around so all this isn't necessary.

jahhulbert-ccri:
yeah I would rather not have any custom trapping in bash...what was the problem again? The metadata is corrupted because of the job? I think we keep extra metadata around we can roll it back?

aheyne:
The list of files that were compacted and need deleted is kept in memory meaning if the driver program is killed then it never gets removed and the metadata never gets updated.

elahrvivaz:
I'm going to close this - created a ticket to track the underlying issue here: https://geomesa.atlassian.net/browse/GEOMESA-2162

PR #1829, opened 2018-01-18
GEOMESA-2146 Remove namerefs for older bash versions

jahhulbert-ccri:


jahhulbert-ccri:
@elahrvivaz here's an old fashioned version without namerefs. I'll see if I can spin up EMR and test it there but i think this will even work with bash 3.x :)

elahrvivaz:
thanks, I just pulled it down and will test on EMR in a few

PR #1832, opened 2018-01-18
GEOMESA-2095 DateOffset process

jnh5y:
Implements a dateoffset WPS.
This WPS allows for the dates in a feature collection to be modified to aid in further analysis.


geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/DateOffsetProcess.scala
```diff
@@ -0,0 +1,60 @@
...
+  @throws(classOf[ProcessException])
+  @DescribeResult(name = "result", description = "Output collection")
+  def execute(@DescribeParameter(name = "data", description = "Input features")
+              obsFeatures: SimpleFeatureCollection,
+              @DescribeParameter(name = "dateField", description = "The date attribute to modify")
+              dateField: String,
+              @DescribeParameter(name = "timeOffset", description = "Time offset (e.g. P1D)")
+              timeOffset: String): SimpleFeatureCollection = {
+
+    val iter = CloseableIterator(obsFeatures.features())
```
elahrvivaz: the iterator isn't being closed - you could change to `SelfClosingIterator`

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/DateOffsetProcess.scala
```diff
@@ -0,0 +1,60 @@
...
+    // TODO: Throw exception if dateField is not in SFT.
```
elahrvivaz: could you implement this check? should be trivial

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/DateOffsetProcess.scala
```diff
@@ -0,0 +1,60 @@
...
+    val period = Period.parse(timeOffset)
+
+    val iter2 = iter.map { sf =>
+
+      val dtg = sf.getAttribute(dateField).asInstanceOf[Date]
+      val ld = dtg.toInstant.atZone(ZoneId.systemDefault()).toLocalDate
+      val ld2 = ld.plus(period)
+      val newDtg = Date.from(ld2.atStartOfDay(ZoneId.systemDefault()).toInstant)
+
+      val updatedSF = ScalaSimpleFeature.copy(sf)
```
elahrvivaz: seems like good defensive programming, but might be expensive to copy vs just modify the input feature. thoughts?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/DateOffsetProcess.scala
elahrvivaz: an easy optimization would be to get the index of the attribute and use that instead of the name in the map function

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/DateOffsetProcess.scala
```diff
@@ -0,0 +1,60 @@
...
+)
+class DateOffsetProcess extends GeoMesaProcess {
+
+  @throws(classOf[ProcessException])
+  @DescribeResult(name = "result", description = "Output collection")
+  def execute(@DescribeParameter(name = "data", description = "Input features")
+              obsFeatures: SimpleFeatureCollection,
+              @DescribeParameter(name = "dateField", description = "The date attribute to modify")
+              dateField: String,
+              @DescribeParameter(name = "timeOffset", description = "Time offset (e.g. P1D)")
```
jahhulbert-ccri: right now its only doing dates and not times...i bet this will fail because Period.parse() won't handle it.

```
"move a feature collection back in time one hour" in {  
      val collection = new ListFeatureCollection(sft, new Random(-1L).shuffle(features))
      val result = process.execute(collection, "dtg", "P-1H")
      val date = result.features().next.getAttribute("dtg").asInstanceOf[Date]
      val dt = new DateTime(date)
      dt.withSecondOfMinute(0) mustEqual new DateTime("2017-02-19T23:00:00.000Z")
    }
```

maybe we rename this datetimeoffsiteprocess or durationoffset or actually add time support. I think you may find that the java.time.Duration class can parse "combined" date times like "P5dT1H"

I think it may be nice to support offset by hours not just days.

jahhulbert-ccri: more info: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-

jahhulbert-ccri: yes i know its weird i noticed that...but after you write a few ageoff iterators you learn iso8601 i guess...i should likely go test the ageoff itr too.

elahrvivaz:
should be good now

PR #1834, opened 2018-01-18
GEOMESA-2152 Mark Commons-lang3 as provided to prevent version confli…

aheyne:
…cts with Spark 2.2.X


elahrvivaz:
did you verify this doesn't affect gs-plugins and command-line tools? we might want to just make it provided in the spark-runtime modules.

aheyne:
That's a better idea, updated.

PR #1836, opened 2018-01-19
GEOMESA-2153 Fixing arrow queries for line strings

elahrvivaz:
Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #1838, opened 2018-01-22
GEOMESA-2156 Passing query sort to Arrow hints

elahrvivaz:
* Query sort is set as a hint, and set to null


elahrvivaz:
note: stacks on top of #1836

elahrvivaz:
rebased

PR #1847, opened 2018-01-31
GEOMESA-2166 Adding support for Cloudera HBase 1.2

elahrvivaz:

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
@@ -20,6 +20,7 @@ import org.locationtech.geomesa.hbase._
 import org.locationtech.geomesa.hbase.coprocessor.AllCoprocessors
 import org.locationtech.geomesa.hbase.data._
 import org.locationtech.geomesa.hbase.index.legacy._
+import org.locationtech.geomesa.hbase.utils.HBaseVersions
 import org.locationtech.geomesa.index.index.ClientSideFiltering
 import org.locationtech.geomesa.utils.conf.GeoMesaSystemProperties
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.Configs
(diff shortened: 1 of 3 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/index/HBaseFeatureIndex.scala
```diff
@@ -79,10 +80,7 @@ trait HBaseFeatureIndex extends HBaseFeatureIndexType with ClientSideFiltering[R
       if (!desc.getCoprocessors.contains(name)) {
         // TODO: Warn if the path given is different from paths registered in other coprocessors
         // if so, other tables would need updating
-        coprocessorUrl match {
-          case Some(path) => desc.addCoprocessor(name, path, Coprocessor.PRIORITY_USER, null)
```
aheyne: Do we still need this check? Not sure what the reflection is going to do if you don't have a url/path.

elahrvivaz: I moved it to the API shim - it takes an `Option[Path]` and uses default values that correspond to the default values in `HTableDescriptor.addCoprocessor(String)` (which calls `HTableDescriptor.addCoprocessor(String, Path, Int, Map) with PRIORITY_USER and nulls)`.

elahrvivaz: https://github.com/elahrvivaz/geomesa/blob/13f23f98524f0138b1c2da7e9919002d2ea334cb/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/utils/HBaseVersions.scala#L25

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/utils/HBaseVersions.scala
```diff
@@ -0,0 +1,56 @@
...
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.hbase.utils
+
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.hbase.{Coprocessor, HColumnDescriptor, HTableDescriptor}
+
+/**
+  * Reflection wrapper for AdminUtils methods between kafka versions 0.9 and 0.10
```
jahhulbert-ccri: kafka?

elahrvivaz: oops, copy paste fail

elahrvivaz:
@jahhulbert-ccri cloudera has some custom hbase jars that have slightly different methods than the normal ones. This causes `NoSuchMethodError`s to get thrown. So I've added some reflection wrappers that get around the differences in method signature.

jahhulbert-ccri:
ewwwwwww cloudera :)

PR #1853, opened 2018-02-07
GEOMESA-2173 Fix return type of TypedColumn spatial functions.

metasim:
Added testing parity between SQL and DataFrame tests of UDFs.
Also refactored mapping between string names and UDFs so we can
be more DRY and programmatically invoke DataFrame tests along with
SQL tests.

Signed-off-by: Simeon H.K. Fitch <fitch@astraea.io>

### DIFF
--- a/geomesa-spark/geomesa-spark-jts/src/main/scala/org/apache/spark/sql/jts/JTSTypes.scala
+++ b/geomesa-spark/geomesa-spark-jts/src/main/scala/org/apache/spark/sql/jts/JTSTypes.scala
@@ -66,6 +66,7 @@ object JTSTypes {
     SQLGeometricOutputFunctions.registerFunctions(sqlContext)
     SQLRules.registerOptimizations(sqlContext)
   }
+
 }
 
 abstract class AbstractGeometryUDT[T >: Null <: Geometry](override val simpleString: String)(implicit cm: ClassTag[T])
--- a/geomesa-spark/geomesa-spark-jts/src/test/resources/log4j.xml
+++ b/geomesa-spark/geomesa-spark-jts/src/test/resources/log4j.xml
@@ -1,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/" debug="false">
+    <appender name="CONSOLE" class="org.apache.log4j.ConsoleAppender">
+        <layout class="org.apache.log4j.EnhancedPatternLayout">
+            <param name="ConversionPattern" value="[%d{HH:mm:ss,SSS}] %5p %c: %m%n"/>
+        </layout>
+    </appender>
+
+    <category name="org.apache.zookeeper">
+        <priority value="warn"/>
+    </category>
+    <category name="org.apache.hadoop">
+        <priority value="error"/>
+    </category>
+    <category name="org.mortbay.log">
+        <priority value="warn"/>
+    </category>
+    <category name="BlockStateChange">
+        <priority value="warn"/>
+    </category>
+    <category name="SecurityLogger.org.apache.hadoop.hbase.Server">
+        <priority value="warn"/>
+    </category>
+    <category name="hsqldb">
+        <priority value="warn"/>
+    </category>
+
+    <category name="org.apache.spark">
+        <priority value="warn"/>
+    </category>
+
+    <category name="org.spark_project.jetty">
+        <priority value="error"/>
+    </category>
+
+    <!-- un-comment the following line to enable verbose log messages
+     from the Spark SQL query-planner; this can be helpful in debugging
+     Spark SQL query plans -->
+    <!--<category name="org.locationtech.geomesa.spark">-->
+        <!--<priority value="info"/>-->
+    <!--</category>-->
+
+    <!--<category name="org.apache.spark.sql.execution">-->
+        <!--<priority value="trace"/>-->
+    <!--</category>-->
+
+    <!--<category name="org.apache.spark.sql">-->
+        <!--<priority value="debug"/>-->
+    <!--</category>-->
+
+    <!-- un-comment the following line to enable verbose log messages
+         from the index query-planner; this can be helpful in debugging
+         query plans -->
+    <!--<category name="org.locationtech.geomesa.index.utils.Explainer">
+        <priority value="trace"/>
+    </category>-->
+
+    <root>
+        <priority value="info"/>
+        <appender-ref ref="CONSOLE" />
+    </root>
+</log4j:configuration>
+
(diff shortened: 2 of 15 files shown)


geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/BlankDataFrame.scala
```diff
@@ -0,0 +1,25 @@
...
+ ***********************************************************************/
+package org.locationtech.geomesa.spark
+
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
+import org.apache.spark.sql.types.StructType
+
+/**
+ * Mixin providing constructor for creating a DataFrame with a single row and no columns.
+ * Useful for testing the invocation of data constructing UDFs.
+ * @author sfitch
```
elahrvivaz: could you remove the author tag? we generally don't include those

geomesa-spark/geomesa-spark-jts/src/main/scala/org/apache/spark/sql/jts/JTSTypes.scala
```diff
@@ -11,7 +11,7 @@ package org.apache.spark.sql.jts
 import com.vividsolutions.jts.geom._
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.GenericInternalRow
+import org.apache.spark.sql.catalyst.expressions.{GenericInternalRow, Literal}
```
elahrvivaz: does this file need to change? seems like it could be rolled back

metasim: Yep. Cruft from an edit that went away.

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/BlankDataFrame.scala
metasim: Sure... was auto generated by IDE.

FWIW, I had pasted the license file in there but I guess from @vpipkt 's comments you need spaces on either side?

metasim:
I'm going to fix the above, and also add a couple more tests. Then I want to submit another PR with some suggested changes to reduce unnecessary friction in the developer experience. I'll write a ticket to outline those. So I'd suggest holding off on a build until then, if you don't mind.

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/BlankDataFrame.scala
elahrvivaz: ah, yeah possibly - the licence check is pretty strict. You can run `mvn license:check` to make sure they're ok.
BTW, if you want to claim copyright, we can add another license template for you.

elahrvivaz:
we're aiming for Friday to release a milestone build, but the fun thing about milestones is you can have as many as you want :)

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/BlankDataFrame.scala
metasim: Thanks; I/we appreciate that. In this case we should let it drop. Let's revisit if it arises again.

metasim:
@elahrvivaz Not sure why the build is marked as failed, because the tests passed. But after all that, this was reported:

> The job exceeded the maximum time limit for jobs, and has been terminated.

https://travis-ci.org/locationtech/geomesa/builds/338716859

elahrvivaz:
we're pushing right up against the limit of travis - the free builds are only allowed to run for 50 minutes before they're killed. I kicked it to re-run, hopefully it will complete ok this time. Since it got through your module, it's probably ok anyway.

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/SparkSQLSpatialRelationshipsTest.scala
```diff
@@ -0,0 +1,505 @@
...
+import org.junit.runner.RunWith
+import org.locationtech.geomesa.spark.SQLGeometricConstructorFunctions._
+import org.locationtech.geomesa.spark.SQLSpatialFunctions._
+import org.locationtech.geomesa.spark.SparkDefaultEncoders._
+import org.locationtech.geomesa.spark.SpatialEncoders._
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class SparkSQLSpatialRelationshipsTest extends Specification with BlankDataFrame with LazyLogging {
```
atallahhezbor: Are you still doing any logging here?

geomesa-spark/geomesa-spark-jts/pom.xml
```diff
@@ -48,6 +48,10 @@
             <artifactId>specs2_${scala.binary.version}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.typesafe.scala-logging</groupId>
```
atallahhezbor: If not doing any logging, would it be worth removing this dependency?

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/SparkSQLSpatialRelationshipsTest.scala
```diff
@@ -0,0 +1,505 @@
...
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.spark
+
+import java.{lang => jl}
+
+import com.typesafe.scalalogging.LazyLogging
```
atallahhezbor: Same goes with the import

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/SparkSQLSpatialRelationshipsTest.scala
metasim: Looks like cruft from when I copied the test guts from `geomesa-spark-sql`. I'll remove.

geomesa-spark/geomesa-spark-jts/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
atallahhezbor: @metasim Oops, this one too. Sorry I missed it on the first pass

metasim: I suggest keeping this one in there, as it reduces the amount of noise you get when running the tests. Even though the tests may not explicitly log, Spark sure does....

atallahhezbor:
LGTM! I like the new patterns

PR #1856, opened 2018-02-07
GEOMESA-2176 Create python conversion for JTS Geometry instances

tkunicki:


jnh5y:
Is there already some documentation for how to use JTS+Python+Spark?

atallahhezbor:
@elahrvivaz I tried testing if it would fix the [ImportError: No module named 'Non'](https://pastebin.com/Bt6RyjRa), (which I think is what spawned this ticket), but it didn't seem to. I think that error has more to do with how we define the json of UDTs. 
Correct me if I'm wrong though @tkunicki or if there's another way to test this.

tkunicki:
This needs a little more work and testing, from testing by @atallahhezbor it appears we need a way to force registration of pickler in spark driver (only seems to register in executors).  I also need to look at the UDT issue mentioned above

metasim:
@atallahhezbor Interestingly enough, I'm also getting the `ImportError: No module named 'Non'` error in my non-JTS RasterFrames code in a different integration I'm working on. My analysis is that it originates here:

https://github.com/apache/spark/blob/44e20c42254bc6591b594f54cd94ced5fcfadae3/python/pyspark/sql/types.py#L729-L732

IOW, I'm getting the same error you guys are, but in my own Scala-defined UDT, so it's not just JTS related. In my test case it's happening when I read it from a Parquet file and then pass to pyspark. In other cases not involving Parquet I've been able to use the UDTs in pyspark fine.

If I understand the UDT pattern correctly, the only time you're *supposed* to have to define a corresponding "pyUDT" is when you absolutely need a wrapper/mirror class for additional method definition, etc. I thought it was supposed to be OK to have `UserDefinedType.pyUDT` be `null`, but maybe [this](https://github.com/apache/spark/blob/568055da93049c207bb830f244ff9b60c638837c/sql/core/src/main/scala/org/apache/spark/sql/test/ExamplePointUDT.scala#L48) and [this](https://github.com/apache/spark/blob/c338c8cf8253c037ecd4f39bbd58ed5a86581b37/python/pyspark/sql/tests.py#L100-L121) is signaling otherwise.

atallahhezbor:
@metasim Yup. That's what I arrived at. When pyUDT is null, then pyClass ends up being `None`, and since that string has no `"."`, then the index is `-1`, hence the "Non" error, since it splices on that `-1` to get the module name.

Not sure what the pattern should be when there is no corresponding python UDT class needed. I was able to get around it by just setting the pyUDT to `"pyspark.sql.tests.ExamplePointUDT"` that way the import goes through, but I'd be in search of something more permanent

metasim:
@atallahhezbor Do you know if a parquet file was involved in the data chain?

I'll try the setting`pyUTD` trick myself. Kinda lame tho.

geomesa-spark/geomesa-spark-core/src/main/scala/org/locationtech/geomesa/spark/GeoMesaSpark.scala
```diff
@@ -8,9 +8,11 @@
 
 package org.locationtech.geomesa.spark
 
-import java.io.{BufferedWriter, StringWriter}
+import java.io.{BufferedWriter, OutputStream, StringWriter}
 import java.util.ServiceLoader
 
+import com.vividsolutions.jts.geom.Geometry
+import net.razorvine.pickle.{IObjectPickler, Pickler}
```
jnh5y: Was this already on the list of dependencies?

tkunicki: it's transitive through spark

metasim:
@atallahhezbor This just crossed my radar on accident, and reminded me that we'd fixed the `ImportError: No module named 'Non'` in `pyrasterframes`. @bguseman is the guy working on it, and can be conscripted for details.

elahrvivaz:
planning to close in favor of #1936 unless someone objects

jnh5y:
That's probably the right move.  I'll try and talk to @tkunicki about it over the weekend.

tkunicki:
This solves an different issue with python RDD, that i was hoping was related to the UDT issue.

elahrvivaz:
we should probably open a new ticket for one or the other of these 2 prs - i'm confused as to which does what at this point so i'll leave it up to you guys :)

elahrvivaz:
replaced with #1993

PR #1871, opened 2018-02-23
GEOMESA-2191 System property to prevent full-table scans

elahrvivaz:
* Set 'geomesa.scan.block-full-table=true'
* Global behavior can be overridden for individual schemas with 'geomesa.scan.<type-name>.block-full-table'


aheyne:
Do we want to add this to the `geomesa-site.xml.template`?

elahrvivaz:
:+1:

PR #1872, opened 2018-02-26
GEOMESA-2194 Implement export-leaflet command

aheyne:
This PR provides a new formate to the `export` command that generates a leaftlet map from the query results and attempts to open it with the default browser. Generates both a feature and heat map.

Example maps generated from FSDS Quickstart

Features + Grayscale Map

Heatmap + OSM


aheyne:
Note to self check default params printing.

elahrvivaz:
could we just make the leaflet export a format in the normal export command? from a cursory look over, it seems like that would require a lot less code changes. the parameter differences appear to just be related to the default values, which we could override inside the execute method

aheyne:
I've merged it into the export command as format `leaflet`. I've also added handling so if the output file name is *.html it'll fill in the format. Additionally, I've worked out a way to handle slf4j not being configured by teeing the sub-shell's output to another I/O file that is redirected to the parents stdout.

docs/tutorials/geomesa-quickstart-fsds.rst
```diff
@@ -171,12 +171,62 @@ contained in ``org.geomesa.example.data.GDELTData``:
...
+ensure you have the command-line tools installed as described in :ref:`setting_up_fsds_commandline`.
+The ``export`` command provides the ``leaflet`` format which will export the features to a leaflet map
+and attempt to open the resulting HTML in your default browser. To produce the map, run the following
+command from the GeoMesa FSDS tools distribution directory:
+
+.. code:: bash
+
+    bin/geomesa-fs export         \
+        --output-format leaflet   \
+        --instance <instance>     \
```
elahrvivaz: copy-pasta - these are the accumulo params

docs/tutorials/geomesa-quickstart-fsds.rst
```diff
@@ -171,12 +171,62 @@ contained in ``org.geomesa.example.data.GDELTData``:
...
+        --zookeepers <zookeepers> \
+        --user <user>             \
+        --password <password>     \
+        --catalog <table>
+
+
+Where the connection parameters are the same you used above during the quickstart. If you ran the
+command on a computer with a default web browser you should see the map below automatically open in it.
```
elahrvivaz: minor thing - could we remove the `note` above, and then change this line to be:

```
If you ran the command on a computer with a default web browser you should see the map below open automatically. If you are running on a headless server (no GUI), you can copy the ``index.html`` to another computer and open it there.
```

docs/tutorials/geomesa-quickstart-fsds.rst
```diff
@@ -171,12 +171,62 @@ contained in ``org.geomesa.example.data.GDELTData``:
...
+``export`` command of the GeoMesa FSDS tools distribution. For a more production ready example, you can
+alternatively stand up a GeoServer and connect it to your FSDS instance.
+
+Visualize Data With Leaflet
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. warning::
+
+    In order to successfully run this command you must have a computer that is connected to the internet
+    in order to access external leaflet resources.
```
elahrvivaz: minor wording - could we change the wording here to get rid of the first 'in order'.

docs/user/cli/export.rst
```diff
@@ -98,8 +98,13 @@ The ``--output-format`` argument defines the encoding used for export. Currently
 * ``geojson``, ``json``
 * ``gml`` `Geography Markup Language <http://www.opengeospatial.org/standards/gml>`_
 * ``shp`` ESRI Shapefile
+* ``leaflet`` Export data to a leaflet map and open in the default browser, if possible
 * ``null`` suppress output entirely
 
+.. note::
+
+    The leaflet format is intended for testing and small scale data exploration and visualization only. For production map generation it is highly recommended to use GeoServer. Additionally, this command requires the use of an online browser in order to access online resources.
```
elahrvivaz: minor wording - change to "Additionally, this command requires the use of a browser in order to access online resources."

docs/user/convert/premade/index.rst
```diff
@@ -3,7 +3,7 @@
 Prepackaged Converter Definitions
 =================================
 
-GeoMesa Tools distributions contain several prepackaged converters for common, publically
+GeoMesa Tools distributions contain several prepackaged converters for common, publicly
```
elahrvivaz: good catch :)

geomesa-arrow/geomesa-arrow-tools/bin/geomesa-arrow
```diff
@@ -47,6 +47,14 @@ else
       GEOMESA_OPTS="$GEOMESA_OPTS $GEOMESA_DEBUG_OPTS"
       shift 1
     fi
-    java ${CUSTOM_JAVA_OPTS} ${GEOMESA_OPTS} -cp ${CLASSPATH} org.locationtech.geomesa.arrow.tools.ArrowRunner "$@"
+    cmd="java ${CUSTOM_JAVA_OPTS} ${GEOMESA_OPTS} -cp ${CLASSPATH} org.locationtech.geomesa.arrow.tools.ArrowRunner ${@}"
+    if isExportLeaflet $@; then
+      exec 3>&1 # Create a new output file to redirect sub-shell's output here
+      res=$(eval $cmd 2>&1 | tee >(cat - >&3); exit ${PIPESTATUS[0]})
```
elahrvivaz: hmm, i'm still not crazy about all this bash-fu to launch the browser - @jahhulbert-ccri what do you think?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/DataFormats.scala
```diff
@@ -37,7 +37,8 @@ object DataFormats extends Enumeration {
     val extension = FilenameUtils.getExtension(filename).toLowerCase(Locale.US)
     values.find(_.toString.equalsIgnoreCase(extension)) match {
       case Some(f) => Right(f)
-      case None    => Left(extension)
+      case None    => if (extension == "html") Right(Leaflet)
```
elahrvivaz: minor - add curly brackets `{}`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
```diff
@@ -152,6 +153,8 @@ object ExportCommand extends LazyLogging {
 
   def getWriter(params: FileExportParams): Writer = new OutputStreamWriter(createOutputStream(params.file, params.gzip))
 
+  def getWriter(file: File, compress: Integer): Writer = new OutputStreamWriter(createOutputStream(file, compress))
```
elahrvivaz: looks to not be used anymore?

geomesa-tools/src/main/resources/leaflet/index.html
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // Add controls to the map
+    L.control.layers(baseMaps, overlayMaps).addTo(geomesaMap);
+    L.control.scale().addTo(geomesaMap);
+
+</script>
+
+</body>
+</html>
```
elahrvivaz: minor - add newline at end of file

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsDataStoreCommand.scala
```diff
@@ -85,4 +85,9 @@ trait PartitionParam {
   var partitions: java.util.List[String] = new util.ArrayList[String]()
 }
 
+trait OptionalQueryThreads {
```
elahrvivaz: looks to not be used any more?

geomesa-bigtable/geomesa-bigtable-tools/src/main/scala/org/locationtech/geomesa/bigtable/tools/BigtableRunner.scala
```diff
@@ -28,7 +28,8 @@ object BigtableRunner extends Runner {
     new HBaseDescribeSchemaCommand,
     new EnvironmentCommand,
     new HBaseExplainCommand,
-    new HBaseExportCommand,
+    new HBaseFileExportCommand,
+    new HBaseLeafletExportCommand,
```
elahrvivaz: looks like this file needs to be reverted

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+  Option(params.maxFeatures) match {
+    case Some(limit) =>
+      if (limit > 10000) {
+        // Note we have to ask if the user wants to continue like this or the output will be
+        // hidden since this command runs in a sub-shell
+        user.warn("A large number of features might be exported. This can cause performance issues when using the map. For large numbers of features it is recommended to use GeoServer to render a map.\nDo you want to continue? [y/N]")
+        // readLine() could be string, null or empty string, handle all
+        val response = Option(readLine()).getOrElse("").headOption.getOrElse("n").toString
+        if (response.matches("^[nN]")) {
+          sys.exit(1)
```
elahrvivaz: we shouldn't call sys.exit except in the main Runner class - maybe throw an exception instead? or ideally just return gracefully but that might be tricky for this workflow. calling sys.exit is generally frowned upon - for example, if someone tried to unit test this class, it would end up killing the maven process.

edit: see below for an alternate suggestion

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+        }
+      }
+    case None =>
```
elahrvivaz: none would require the same warning right? maybe instead of instead of handling it this way, it would be better to:
1. check for max features > 1000 and print a warning but continue
2. if no max features set, set max features to 1000 and print a warning

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  Option(params.gzip) match {
+    case Some(_) => user.warn("GZip parameter cannot be used with leaflet format, ignoring.")
+    case None =>
+  }
+  Option(params.noHeader) match {
+    case Some(_) => user.warn("NoHeader parameter cannot be used with leaflet format, ignoring.")
```
elahrvivaz: noheader is a primitive, and defaults to false - so this will always end up being logged. could safely change to `if (params.noHeader) { logger.warn(...) }`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+        val response = Option(readLine()).getOrElse("").headOption.getOrElse("n").toString
+        if (response.matches("^[nN]")) {
+          sys.exit(1)
+        }
+      }
+    case None =>
+  }
+
+  Option(params.gzip) match {
+    case Some(_) => user.warn("GZip parameter cannot be used with leaflet format, ignoring.")
```
elahrvivaz: minor - since you're not using the unapply, or the alternate match statements, it might be clearer to just write
`if (params.gzip != null) { logger.warn(...) }`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+    case None =>
+  }
+  Option(params.noHeader) match {
+    case Some(_) => user.warn("NoHeader parameter cannot be used with leaflet format, ignoring.")
+    case None =>
+  }
+
+  val GEOMESA_HOME = SystemProperty("geomesa.home", "/tmp")
```
elahrvivaz: does this get set anywhere? or are you just allowing for a hook to override the destination path?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+  val root = new File(GEOMESA_HOME.get)
```
elahrvivaz: minor - root is only used if params.file is not defined, you could move it inside the match statement (or just inline it there)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+object LeafletMapExporter {
+  def checkDestination(file: File): File = {
+    try {
+      file.mkdir()
+      if (! file.isDirectory) {
+        throw new RuntimeException(s"Output destination ${file.toString} must not exist or must be a directory.")
+      }
+      file
+    } catch {
+      case e: SecurityException => throw new RuntimeException("Unable to create output destination, check permissions.", e)
```
elahrvivaz: would be useful to include the file path here

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+  Option(params.noHeader) match {
+    case Some(_) => user.warn("NoHeader parameter cannot be used with leaflet format, ignoring.")
+    case None =>
+  }
+
+  val GEOMESA_HOME = SystemProperty("geomesa.home", "/tmp")
+  val root = new File(GEOMESA_HOME.get)
+  val dest: File = Option(params.file) match {
+    case Some(file) => checkDestination(file)
```
elahrvivaz: earlier it looked like you were detecting `.html` as a file extension - does it make sense to check the parent folder for the output file instead of the file itself? then you could use the file itself in place of the index.html, below

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  def getFeatureInfo(sft: SimpleFeatureType): String = {
+    val str: StringBuilder = new StringBuilder()
+    str.append("    function onEachFeature(feature, layer) {\n")
+    Option(sft) match {
+      case None    => str.append("\n    };\n\n").toString
+      case Some(_) =>
+        str.append("""        layer.bindPopup("ID: " + feature.id + "<br>" + """)
+        str.append(""""GEOM: " + feature.geometry.type + "[" + feature.geometry.coordinates + "]<br>" + """)
```
elahrvivaz: will this work with non-points?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,185 @@
...
+    coords.map(c => SimpleCoordinate(c.x, c.y)).foreach{ sc =>
+      coordMap.put(sc, 1) match {
+        case Some(count) => coordMap.put(sc, count + 1)
+        case None =>
+      }
+    }
+  }
+
+  def normalizeValues(coordMap: mutable.Map[SimpleCoordinate[Double], Int]): Map[SimpleCoordinate[Double], Float] = {
+    val max: Float = coordMap.maxBy(_._2)._2
```
elahrvivaz: probably need to handle the case here where no features are returned - think this would throw an exception

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
aheyne: It's used, ln 65 of LeafletMapExporter

geomesa-tools/src/main/resources/leaflet/index.html
aheyne: You got me... :)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
aheyne: The problem with #1 is that if we're launching their browser we could cause bad things to happen. Would like to give them a warning and an option to bail. We also can't change the param text because we're tied into the general export command.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
aheyne: Throwing an error seems like a better idea.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
aheyne: I guess that change got lost. It's set in `common-functions.sh` as part of the general GEOMESA_OPTS now.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
aheyne: I reworked all of this. should be fine now.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
aheyne: Yeah, I've tested it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
aheyne: Good catch.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
elahrvivaz: :+1:

geomesa-tools/bin/common-functions.sh
```diff
@@ -344,6 +344,41 @@ function geomesaScalaConsole() {
   exec $scalaCMD ${OPTS} -classpath ${classpath} -i "${GEOMESA_CONF_DIR}/.scala_repl_init"
 }
 
+function isExportLeaflet() {
+  params=($@)
+  for i in ${params[@]}; do
+    if [[ "${i}" == "leaflet" || "${i}" == *html ]]; then
```
elahrvivaz: could we make this check both the flag and the value, to be more robust? this syntax is totally wrong but like: `if [[ "($params[i]" == "leaflet" && ($params[i-1] == "-F" || $params[i-1] == "--format")) || /* check output file */]]`

geomesa-tools/src/main/resources/leaflet/index.html
```diff
@@ -0,0 +1,90 @@
...
+    <title>GeoMesa Leaflet Map Viewer</title>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.3.1/dist/leaflet.css"
+          integrity="sha512-Rksm5RenBEKSKFjgI3a41vrjkw4EVPlJ3+OiI65vTjIdo9brlAacEuKOiQ5OFh7cOI1bkDwLqdLw3Zg0cRJAAQ=="
+          crossorigin=""/>
+    <script src="https://unpkg.com/leaflet@1.3.1/dist/leaflet.js"
+            integrity="sha512-/Nsx9X4HebavoBvEBuyp3I7od5tA0UzAxs+j83KgC8PU0kgB4XiK4Lfe4y4cgBtaRJQEIFCW+oC506aPT2L1zw=="
+            crossorigin=""></script>
+    <script src="https://cdn.rawgit.com/aheyne/Leaflet.heat/gh-pages/dist/leaflet-heat.js"></script>
```
elahrvivaz: could we host this somewhere more permanent? we could probably add it as part of this PR and then reference the github URL (although that wouldn't work until the PR is actually merged)

aheyne: I'm not sure, it's licensed under BSD 2-Clause https://github.com/aheyne/Leaflet.heat/blob/gh-pages/LICENSE Would that cause problems?

elahrvivaz: oh, that would be ok i think... but instead, you could just link to the original github repo instead of your fork. just make sure you link to a particular release tag so that it doesn't change out from under us

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/LeafletMapExporter.scala
```diff
@@ -0,0 +1,189 @@
...
+    // Handle both files and directories that could exist or not
+    if (file.exists()) {
+      if (file.isDirectory) {
+        new File(file, "index.html").ensureWritable
+      } else {
+        if (file.toString.endsWith(".html")) { file.ensureWritable }
+        else { throw new RuntimeException(s"Destination file ${file.toString} must end with '.html'") }
+      }
+    } else {
+      file.mkdir()
```
elahrvivaz: might want to switch this to `.mkdirs()` which will create nested folders if required

jahhulbert-ccri:
i'll take a look soon!

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/ConvertCommandTest.scala
```diff
@@ -68,7 +68,12 @@ class ConvertCommandTest extends Specification with LazyLogging {
         command.params.config = conf
         command.params.spec = conf
         command.params.outputFormat = outFmt
-        command.params.file = File.createTempFile("convertTest", s".${outFmt.toString.toLowerCase}")
+        command.params.file = if (outFmt == DataFormats.Leaflet) {
+          File.createTempFile("convertTest", s".html")
```
jahhulbert-ccri: Lets add a line to clean this up...got a lot in my /tmp dir

geomesa-tools/src/main/resources/leaflet/index.html
jahhulbert-ccri: #important

geomesa-arrow/geomesa-arrow-tools/bin/geomesa-arrow
jahhulbert-ccri: i don't think we should launch it by default

jahhulbert-ccri: i prefer not to have our tools launch any other programs since that's not "unix-y"

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/ConvertCommandTest.scala
aheyne: It's already cleaned up on ln 80-81

docs/tutorials/geomesa-quickstart-accumulo.rst
```diff
@@ -201,12 +201,56 @@ contained in ``org.geomesa.example.data.GDELTData``:
...
+.. warning::
+
+    To successfully run this command you must have a computer that is connected to the internet
+    in order to access external leaflet resources.
+
+
+The ``export`` command is a part of the GeoMesa Accumulo command-line tools. In order to use the
+command, ensure you have the command-line tools installed as described in
+:ref:`setting_up_accumulo_commandline`. The ``export`` command provides the ``leaflet`` format which
+will export the features to a leaflet map and attempt to open the resulting HTML in your default
```
elahrvivaz: [minor] doesn't actually attempt to open any more

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/ConvertCommandTest.scala
elahrvivaz: our tests do seem to still leave tmp files around - we should track those down and fix em at some point

PR #1873, opened 2018-02-26
GEOMESA-2190 Readdress handling of geometric literals in DataFrame DSL.

metasim:
Removes column functions where literal variant is non-sensical or of questionable usefulness. Added utility functions for generating geometry literals of specific types. Added tests to increase code coverage those remaining.

### DIFF
--- a/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
+++ b/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
@@ -9,11 +9,14 @@
 package org.locationtech.geomesa.spark.jts
 
 import com.vividsolutions.jts.geom._
-import org.apache.spark.sql.{Column, TypedColumn}
+import org.apache.spark.sql.{Column, Encoder, TypedColumn}
 import org.locationtech.geomesa.spark.jts.encoders.SpatialEncoders
 import org.locationtech.geomesa.spark.jts.util.SQLFunctionHelper._
 import java.{lang => jl}
 
+import org.apache.spark.sql.catalyst.expressions.Literal
+import org.apache.spark.sql.functions.{lit, array}
+import org.apache.spark.sql.jts._
 
 
 /**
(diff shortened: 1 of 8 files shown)


geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/WKUtils.scala
```diff
@@ -11,6 +11,7 @@ package org.locationtech.geomesa.spark.jts.util
 import com.vividsolutions.jts.geom.Geometry
 import com.vividsolutions.jts.io._
 
+// $COVERAGE-OFF$
```
elahrvivaz: does scoverage fail here?

elahrvivaz: haha, ok :see_no_evil:

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
```diff
@@ -39,8 +39,7 @@ object SpatialRelationFunctions {
   val ST_Touches:    (Geometry, Geometry) => jl.Boolean = nullableUDF((geom1, geom2) => geom1.touches(geom2))
   val ST_Within:     (Geometry, Geometry) => jl.Boolean = nullableUDF((geom1, geom2) => geom1.within(geom2))
   val ST_Relate:     (Geometry, Geometry) => String = nullableUDF((geom1, geom2) => geom1.relate(geom2).toString)
-  val ST_RelateBool: (Geometry, Geometry, String) => Boolean =
-    nullableUDF((geom1, geom2, pattern) => geom1.relate(geom2, pattern))
+  val ST_RelateBool: (Geometry, Geometry, String) => jl.Boolean = nullableUDF((geom1, geom2, pattern) => geom1.relate(geom2, pattern))
```
metasim: @elahrvivaz I slipped this change in there, which was the main impetus for [issue 2174](https://geomesa.atlassian.net/browse/GEOMESA-2174). This change makes it boxed, the opposite of what I was advocating for in the issue text. Still want to fine-tooth the code for any more.

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctionsTest.scala
```diff
@@ -485,6 +485,25 @@ class SpatialRelationFunctionsTest extends Specification with TestEnvironment {
...
+      val p1 = points("int")
+      val p2 = points("edge")
+      val dist = dfBlank.select(st_aggregateDistanceSpheroid(array(st_geomFromWKT(p1), st_geomFromWKT(p2)))).first
+      dist.toDouble must beGreaterThan(550000.0)
+    }
+
+    "st_lengthSpheroid" >> {
+      val line = "LINESTRING(1 2, 11 2)"
+      val dist = dfBlank.select(st_lengthSpheroid(st_castToLineString(st_geomFromWKT(line)))).first
+      dist.toDouble must beGreaterThan(1110000.0)
```
metasim: This and the other "magic number" above were extracted from printing the results, assuming that the math behind them is correct. Test case is just to exercise the call path.

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
```diff
@@ -29,90 +32,112 @@ object DataFrameFunctions extends SpatialEncoders {
...
     def st_makePolygon(lineString: LineString): TypedColumn[Any, Polygon] =
-      udfToColumnLiterals(ST_MakePolygon, constructorNames, lineString)
+      st_polygon(lineString)
 
     def st_makePoint(x: Column, y: Column): TypedColumn[Any, Point] =
-      udfToColumn(ST_MakePoint, constructorNames, x, y)
+      st_point(x, y)
     def st_makePoint(x: Double, y: Double): TypedColumn[Any, Point] =
-      udfToColumnLiterals(ST_MakePoint, constructorNames, x, y)
+      st_point(x, y)
```
atallahhezbor: Why don't you need `lit` here?

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctionsTest.scala
elahrvivaz: hmm, instead of using the magic number that isn't verified, maybe just change to something like
`dfBlank.select(st_lengthSpheroid(st_castToLineString(st_geomFromWKT(line)))).first must not(throwAn[Exception])`
that way it's clear that we aren't actually validating the output.

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
metasim: `st_point` already has a `st_point(Double, Double)` form. But I could change it if it's confusing.

PR #1878, opened 2018-03-01
GEOMESA-2198 Checking for iterators on classpath during createSchema

elahrvivaz:

### DIFF
--- a/geomesa-web/geomesa-web-data/pom.xml
+++ b/geomesa-web/geomesa-web-data/pom.xml
@@ -44,7 +44,7 @@
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-vfs2</artifactId>
-                <version>2.0</version>
+                <version>2.1</version>
                 <scope>provided</scope>
             </dependency>
             <dependency>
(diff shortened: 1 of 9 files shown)


geomesa-web/geomesa-web-data/pom.xml
```diff
@@ -44,7 +44,7 @@
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-vfs2</artifactId>
-                <version>2.0</version>
+                <version>2.1</version>
```
jnh5y: Should we externalize the a commons-vfs2.version?

elahrvivaz: we should probably just get rid of the spark endpoint tbh. i think that's the only reason this is here... it was causing unit test failures in this module with the older version.

jnh5y:
Let's nuke it!

geomesa-web/geomesa-web-data/pom.xml
jnh5y: Let's nuke it!

elahrvivaz: opened a ticket here: https://geomesa.atlassian.net/browse/GEOMESA-2210

PR #1879, opened 2018-03-01
GEOMESA-2200 TubeSelect Interpolation Bug Fix

jmglenn:
Added check ensuring that two features do not share a time in interpolated TubeBuilder


geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/tube/TubeBuilder.scala
```diff
@@ -238,10 +238,10 @@ class InterpolatedGapFill(tubeFeatures: SimpleFeatureCollection,
...
         //So that no segment is larger than the buffer. This ensures that each segment has an
-        //times and distance.
-        if (dist > bufferDistance) {
+        //times and distance. Also ensure that features do not share a time value.
+        val timeDiffMillis = t2.toInstant.toEpochMilli - t1.toInstant.toEpochMilli
+        if (dist > bufferDistance && timeDiffMillis > 0) {
           val heading = calc.getAzimuth
-          val timeDiffMillis = t2.toInstant.toEpochMilli - t1.toInstant.toEpochMilli
           val segCount = (dist / bufferDistance).toInt
           val segDuration = timeDiffMillis / segCount
```
elahrvivaz: it would probably be good to validate the segDuration here - it seems like it could still be zero.

elahrvivaz: on line 248 (which i can't comment on directly), it's also possible that the combination of the time difference and the segDuration will result in a single item list, which will cause an exception on line 257

jnh5y:
@jmglenn have you had a chance to work on this PR more?

elahrvivaz:
@jahhulbert-ccri @jnh5y could one of you review my changes before merging?

PR #1886, opened 2018-03-05
GEOMESA-2204 Fixing Kafka consumer sleep while more data to read

elahrvivaz:
Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #1889, opened 2018-03-06
GEOMESA-2206 Kafka - don't index features during initial bulk load

elahrvivaz:
* If 'kafka.consumer.from-beginning', features will not be available
  to query until consumers have caught up, although listeners will be
  called as normal

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
@@ -10,18 +10,20 @@ package org.locationtech.geomesa.kafka.data
 
 import java.io.Closeable
 import java.util.Collections
-import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{ConcurrentHashMap, Executors}
 
+import com.github.benmanes.caffeine.cache.{Cache, Caffeine}
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.kafka.clients.consumer.{Consumer, ConsumerRecord}
 import org.geotools.data.simple.SimpleFeatureSource
 import org.geotools.data.{FeatureEvent, FeatureListener}
+import org.locationtech.geomesa.kafka.KafkaConsumerVersions
 import org.locationtech.geomesa.kafka.consumer.ThreadedConsumer
 import org.locationtech.geomesa.kafka.index.KafkaFeatureCache
 import org.locationtech.geomesa.kafka.utils.GeoMessage.{Change, Clear, Delete}
 import org.locationtech.geomesa.kafka.utils.{GeoMessageSerializer, KafkaFeatureEvent}
-import org.locationtech.geomesa.utils.conf.GeoMesaSystemProperties.SystemProperty
-import org.opengis.feature.simple.SimpleFeatureType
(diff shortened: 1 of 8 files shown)


elahrvivaz:
stacks on top of #1886

geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  override def close(): Unit = {
+    closed.set(true)
+    consumers.foreach(_.wakeup())
+    executor.shutdownNow()
+    executor.awaitTermination(1, TimeUnit.SECONDS)
+  }
+
+  class ConsumerRunnable(val consumer: Consumer[Array[Byte], Array[Byte]], frequency: Long) extends Runnable {
```
elahrvivaz: @jahhulbert-ccri TODO for me - add id for each consumer and use it in logging

geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
```diff
@@ -0,0 +1,103 @@
...
+            }
+          } catch {
+            case e: WakeupException => throw e
+            case e: InterruptedException => throw e
+            case NonFatal(e) =>
+              logger.warn(s"Error receiving message from topic $topic:", e)
+              errorCount += 1
+              if (errorCount < 300) { Thread.sleep(1000) } else {
+                logger.error("Shutting down due to too many errors")
+                throw new InterruptedException("Too many errors")
```
elahrvivaz: TODO this will get caught and suppressed in the outer loop - we want it to propagate out so that it's obvious the thread died

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
```diff
@@ -113,5 +123,90 @@ object KafkaCacheLoader {
...
+        if (offsets.getOrDefault(record.partition, Long.MaxValue) <= record.offset) {
+          offsets.remove(record.partition)
+        }
+      }
+    }
+
+    override def run(): Unit = {
+      import scala.collection.JavaConverters._
+
+      val partitions = consumers.head.partitionsFor(topic).asScala.map(_.partition)
```
jnh5y: Sanity check:  each of the consumers has the same number of partitions?

jnh5y:
Hmm...  @elahrvivaz can you add some brief docs?  Maybe just a paragraph or so explaining this new feature?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
elahrvivaz: this is just a metadata operation for the topic - it doesn't necessarily mean that the consumer is assigned a particular partition. there is another method `assignment` that does that.

elahrvivaz:
I added a snippet in the usage page - this change should be mostly transparent to the end user, they don't have to do anything different.

jnh5y:
Thanks for the small update.  I missed that we already had this option documented.  :shipit:

PR #1891, opened 2018-03-09
GEOMESA-2199 Moves and updates spatial4j and geohash UDFs to spark-jts module

atallahhezbor:


geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/GeometricProcessingFunctions.scala
```diff
@@ -42,23 +40,37 @@ object SQLGeometryProcessingFunctions extends LazyLogging {
...
   val ST_antimeridianSafeGeom: Geometry => Geometry = nullableUDF(geom => {
-    GeohashUtils.getInternationalDateLineSafeGeometry(geom) match {
-      case Success(g) => g
-      case Failure(e) =>
-        logger.warn(s"Error splitting geometry on anti-meridian for $geom", e)
-        geom
+    def degreesToTranslate(x: Double): Double = (((x + 180) / 360.0).floor * -360).toInt
+    if (geom.getEnvelopeInternal.getMinX < -180 || geom.getEnvelopeInternal.getMaxX > 180) {
+
+      geom.apply(new CoordinateSequenceFilter() {
```
elahrvivaz: would it be better to create a new geometry, and not mutate the one that is passed in?

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/GeometricProcessingFunctions.scala
```diff
@@ -42,23 +40,37 @@ object SQLGeometryProcessingFunctions extends LazyLogging {
   }
 
   val ST_antimeridianSafeGeom: Geometry => Geometry = nullableUDF(geom => {
-    GeohashUtils.getInternationalDateLineSafeGeometry(geom) match {
-      case Success(g) => g
-      case Failure(e) =>
-        logger.warn(s"Error splitting geometry on anti-meridian for $geom", e)
-        geom
+    def degreesToTranslate(x: Double): Double = (((x + 180) / 360.0).floor * -360).toInt
+    if (geom.getEnvelopeInternal.getMinX < -180 || geom.getEnvelopeInternal.getMaxX > 180) {
```
elahrvivaz: i think the logic of this method has changed - is that intentional? the `getInternationalDateLineSafeGeometry` method will split any x span > 180 degrees, now it is only checking for x coordinates outside the 'world'

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/GeoHashUtils.scala
```diff
@@ -0,0 +1,165 @@
...
+package org.locationtech.geomesa.spark.jts.util
+
+import com.vividsolutions.jts.geom.{Envelope, Geometry, GeometryFactory, PrecisionModel}
+import scala.collection.BitSet
+import scala.collection.immutable.{BitSet => IBitSet}
+
+object GeoHashUtils {
+
+  def encode(geom: Geometry, precision: Int): String = {
+    GeoHash.encode(geom.getCentroid.getX, geom.getCentroid.getY, precision)
```
elahrvivaz: centroid can sometimes return NaN - see https://github.com/locationtech/geomesa/blob/geomesa_2.11-2.0.0-m.1/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala#L45
Also, probably better to only call `.getCentroid` once, as i'm not sure if the result is cached or re-calculated. (if you can determine it's cached, add a comment to that effect)

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/GeoHashUtils.scala
```diff
@@ -0,0 +1,165 @@
...
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.spark.jts.util
+
+import com.vividsolutions.jts.geom.{Envelope, Geometry, GeometryFactory, PrecisionModel}
+import scala.collection.BitSet
+import scala.collection.immutable.{BitSet => IBitSet}
+
+object GeoHashUtils {
```
elahrvivaz: should add some comments differentiating this from the other GeoHashUtils in geomesa-util, e.g. why this one, differences between the two, etc

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/GeoHashUtils.scala
```diff
@@ -0,0 +1,165 @@
...
+    GeoHash.decode(geohash, Some(precision))
+  }
+}
+
+case class Bounds(low: Double,
+                  high: Double) {
+  lazy val mid = (low+high)/2.0
+}
+
+object GeoHash {
```
elahrvivaz: did you code this up from scratch? if not, we should add a comment indicating where it came from, and might need to include copyright headers

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/GeometricProcessingFunctions.scala
atallahhezbor: My intention was to keep the logic the same. I thought spatial4j handled splitting the geometry? https://github.com/locationtech/spatial4j/blob/spatial4j-0.6/src/main/java/org/locationtech/spatial4j/shape/jts/JtsGeometry.java#L421

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/GeoHashUtils.scala
atallahhezbor: the GeoHash object is just pickings from from utils.geohash.GeoHashUtils.
the GeoHashUtils object is from scratch. I'll make a note, but I'm not sure what headers to need.

elahrvivaz: if it's copied from inside geomesa, then just make sure that you copy any non-ccri copyrights from the class you copied from. i wasn't sure where it was from, or if you had coded it up, since i didn't see some of the methods in geomesa-utils, like `powersOf2Map`.

elahrvivaz: but figured you probably didn't re-implement it given that there are a bunch of implementations already out there

atallahhezbor: Got it. I also borrowed some bits from utils.geohash.GeoHash but it's all from inside geomesa. I'll look through the comments in those two files

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/GeometricProcessingFunctions.scala
elahrvivaz: ah, you may be right. do we have some unit tests that correspond between both methods to validate they do the same thing?

atallahhezbor: The moved the [unit tests](https://github.com/locationtech/geomesa/pull/1891/files/de805b70d713818aff63d81369e8e18ff2132129#diff-1e74d13f3bee54406d98d9faad0152f3R70) for the UDFs that used to use the old implementation 

There's this unit test for GeoHashUtils. https://github.com/locationtech/geomesa/blob/master/geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/geohash/GeohashUtilsTest.scala#L215
I can double check that it still works here, but I don't think it's worth copying over too?

elahrvivaz: that sounds like a good plan, thanks

PR #1892, opened 2018-03-14
GEOMESA-2215 Restore SparkSQL predicate pushdown rules

atallahhezbor:

### DIFF
--- a/geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
+++ b/geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
@@ -276,6 +276,11 @@ object SQLRules extends LazyLogging {
 
   def registerOptimizations(sqlContext: SQLContext): Unit = {
 
+    Seq(STContainsRule).foreach { r =>
+      if(!sqlContext.experimental.extraOptimizations.contains(r))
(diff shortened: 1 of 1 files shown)


geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -276,6 +276,11 @@ object SQLRules extends LazyLogging {
 
   def registerOptimizations(sqlContext: SQLContext): Unit = {
 
+    Seq(STContainsRule).foreach { r =>
```
elahrvivaz: what about all the other predicates here: https://github.com/ccri/geomesa/blob/ff5a0d5b223c9f25bd3829c7e8091067b06cef7c/geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala#L59-L71

elahrvivaz: i guess those are defined as rules - but maybe they should be? doesn't have to be part of this pr

atallahhezbor: Right. This was a point I was gonna bring up with you. The rule is poorly named. It handles all predicates, (and also some spatial join optimizations) but it kept it's original name through all of that

PR #1894, opened 2018-03-19
GEOMESA-2221 Push down Spark filters in Catalyst where possible 

atallahhezbor:

### DIFF
--- a/geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
+++ b/geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
@@ -134,16 +134,7 @@ object SQLRules extends LazyLogging {
       case _ => None
     }
 
-    private def extractScalaUDFs(f: Expression) = {
-      splitConjunctivePredicates(f).partition {
-        // TODO: Add guard which checks to see if the function can be pushed down
-        case ScalaUDF(_, _, _, _, _) => true
-        case _ => false
-      }
-    }
-
     private def extractGridId(envelopes: List[Envelope], e: org.apache.spark.sql.catalyst.expressions.Expression): Option[List[Int]] = e match {
-      case And(l, r) => Some(extractGridId(envelopes, l).getOrElse(List()) ++  extractGridId(envelopes, r).getOrElse(List()))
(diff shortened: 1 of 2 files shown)


geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -196,10 +234,11 @@ object SQLRules extends LazyLogging {
 
           // split up conjunctive predicates and extract the st_contains variable
           val (scalaUDFs: Seq[Expression], otherFilters: Seq[Expression]) = extractScalaUDFs(f)
+          val sparkFilters: Seq[Expression] = otherFilters ++ scalaUDFs
 
-          val (gtFilters: Seq[GTFilter], sFilters: Seq[Expression]) = scalaUDFs.foldLeft((Seq[GTFilter](), otherFilters)) {
+          val (gtFilters: Seq[GTFilter], sFilters: Seq[Expression]) = sparkFilters.foldLeft((Seq[GTFilter](), Seq[Expression]())) {
```
jnh5y: If we aren't operating on just the scalaUDFs anymore, do we need to split the Filters into UDFs and not-UDFs?  

I.e., can we skip calling extractScalaUDFs?  Preferring instead to call only splitConjunctivePredicates(f)?

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -73,12 +72,51 @@ object SQLRules extends LazyLogging {
...
+        val rightExpr = sparkExprToGTExpr(right)
+        if (leftExpr.isEmpty || rightExpr.isEmpty) {
+          None
+        } else {
+          binaryComp match {
+            case eq:  EqualTo => Some(ff.equals(leftExpr.get, rightExpr.get))
+            case lt:  LessThan => Some(ff.less(leftExpr.get, rightExpr.get))
+            case lte: LessThanOrEqual => Some(ff.lessOrEqual(leftExpr.get, rightExpr.get))
+            case gt:  GreaterThan => Some(ff.greater(leftExpr.get, rightExpr.get))
+            case gte: GreaterThanOrEqual => Some(ff.greaterOrEqual(leftExpr.get, rightExpr.get))
```
jnh5y: Would you have time to add tests to cover each of these tests?

PR #1895, opened 2018-03-19
GEOMESA-2223 Use Provided FID in Converters

anthonyccri:
* Converters now check for a non-null id and set Hints.USE_PROVIDED_ID
  if found

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
@@ -15,6 +15,7 @@ import java.util.NoSuchElementException
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.commons.io.IOUtils
+import org.geotools.factory.Hints
 import org.locationtech.geomesa.convert.ParseMode.ParseMode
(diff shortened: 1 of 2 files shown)


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -348,13 +349,24 @@ trait ToSimpleFeatureConverter[I] extends SimpleFeatureConverter[I] with LazyLog
...
-    val sf = new ScalaSimpleFeature(targetSFT, id, sfValues)
-    userDataBuilder.foreach { case (k, v) => sf.getUserData.put(k, v.eval(t)(ec).asInstanceOf[AnyRef]) }
+    val sf = buildFeature(t, sfValues, ec)
 
     validate(sf, ec)
   }
 
+  private def buildFeature(t: Array[Any], sfValues: Array[AnyRef], ec: EvaluationContext): SimpleFeature = {
+    val sf = idBuilder match {
+      case LitNull => new ScalaSimpleFeature(targetSFT, null, sfValues)
```
elahrvivaz: hmm, looking at the code, this will actually throw an NPE in `org.geotools.filter.identity.FeatureIdImpl`. so even though `id-field` is optional in the factory class, everything would fail if it's left out. we could just set this to an empty string here instead, I think that would work.

PR #1898, opened 2018-03-21
GEOMESA-2220 Filter optimization - evaluate static filter functions o…

jnh5y:
…nce and replace with a literal

* Implements Expression evaluation in the QueryPlanFilterVisitor
* Implements optionally period offset parameter for CurrentDateFunction

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
@@ -28,7 +28,7 @@ class CurrentDateFunction extends FunctionExpressionImpl(CurrentDateFunction.Nam
     if (offset.isEmpty) {
       Date.from(now)
     } else {
-      Date.from(now.minus(Duration.parse(offset)))
+      Date.from(now.plus(Duration.parse(offset)))
     }
   }
 }
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
@@ -20,7 +20,6 @@ class CurrentDateFunction extends FunctionExpressionImpl(CurrentDateFunction.Nam
 
   override def evaluate(o: AnyRef): AnyRef = {
     if (offset.isEmpty && getParameters.size > 0) {
-
       offset = getExpression(0).evaluate(null).asInstanceOf[String]
     }
 
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
@@ -8,16 +8,34 @@
 
 package org.locationtech.geomesa.filter.function
 
-import java.time.{ZoneOffset, ZonedDateTime}
+import java.time.{Duration, ZoneOffset, ZonedDateTime}
 import java.util.Date
 
+import org.geotools.data.Parameter
 import org.geotools.filter.FunctionExpressionImpl
 import org.geotools.filter.capability.FunctionNameImpl
 
 class CurrentDateFunction extends FunctionExpressionImpl(CurrentDateFunction.Name) {
-  override def evaluate(o: AnyRef): AnyRef = Date.from(ZonedDateTime.now(ZoneOffset.UTC).toInstant)
+  private var offset = ""
+
+  override def evaluate(o: AnyRef): AnyRef = {
+    if (offset.isEmpty && getParameters.size > 0) {
+
+      offset = getExpression(0).evaluate(null).asInstanceOf[String]
+    }
+
+    val now = ZonedDateTime.now(ZoneOffset.UTC).toInstant
+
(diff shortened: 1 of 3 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -112,6 +114,28 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
   }
 
   private def includeEquivalent(f: Filter): Boolean = f == Filter.INCLUDE || isFilterWholeWorld(f)
+
+  override def visit(expression: Expression, extraData: scala.Any): Expression = {
+    val visitedExpression = super.visit(expression, extraData)
+
+    visitedExpression match {
+            case exp: Expression =>
```
elahrvivaz: is this match necessary? it will always be an expression, right?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -112,6 +114,28 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
...
+        if (exp.accept(IsStaticExpressionVisitor.VISITOR, null).asInstanceOf[Boolean]) {
+          val eval = Try(exp.evaluate(null))
+
+          val ret = eval match {
+            case Success(value) if value != null =>
+              new LiteralExpressionImpl(value)
+            case _ =>
+              visitedExpression
+          }
+          ret
```
elahrvivaz: this block might be more clearly/succinctly written as:
```
Try(exp.evaluate(null)).filter(_ != null).map(new LiteralExpressionImpl(_)).getOrElse(exp)
```

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
```diff
@@ -8,16 +8,34 @@
...
 import org.geotools.filter.FunctionExpressionImpl
 import org.geotools.filter.capability.FunctionNameImpl
 
 class CurrentDateFunction extends FunctionExpressionImpl(CurrentDateFunction.Name) {
-  override def evaluate(o: AnyRef): AnyRef = Date.from(ZonedDateTime.now(ZoneOffset.UTC).toInstant)
+  private var offset = ""
+
+  override def evaluate(o: AnyRef): AnyRef = {
+    if (offset.isEmpty && getParameters.size > 0) {
+
```
elahrvivaz: [picky] extra empty line

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
```diff
@@ -36,6 +39,32 @@ class FilterHelperTest extends Specification {
...
+      val updated = updateFilter(filter)
+      val intervals = FilterHelper.extractIntervals(updated, "dtg", handleExclusiveBounds = true)
+      // JNH: I wonder if this is stable enough for a unit test?:)
+      Period.between(intervals.values(0).lower.value.get.toLocalDate, intervals.values(0).upper.value.get.toLocalDate).toString mustEqual "P1D"
+    }
+
+    "evaluate functions with math" >> {
+      val filter = ECQL.toFilter("number < 1+2")
+      val updated = updateFilter(filter)
+      updated.asInstanceOf[IsLessThenImpl].getExpression2.isInstanceOf[AddImpl] mustEqual false
```
elahrvivaz: [minor] there's a `beAnInstanceOf` matcher you could use here

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
jnh5y: Hmm... I was thinking that the return from the super.visit could potentially be something other than an Expression... 

I'm not sure though...

elahrvivaz: the method signature is `Expression` - could be null maybe...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/function/CurrentDateFunction.scala
```diff
@@ -8,16 +8,33 @@
...
+    if (offset.isEmpty && getParameters.size > 0) {
+      offset = getExpression(0).evaluate(null).asInstanceOf[String]
+    }
+
+    val now = ZonedDateTime.now(ZoneOffset.UTC).toInstant
+
+    if (offset.isEmpty) {
+      Date.from(now)
+    } else {
+      Date.from(now.minus(Duration.parse(offset)))
```
elahrvivaz: it might make sense to use `plus` here instead of `minus` - that way you could do `currentDate('P1D')` for future dates, and `currentDate('-P1D')` for past dates. otherwise it would be the inverse, which makes some sense if you never consider future dates but seems a bit confusing when you do

PR #1899, opened 2018-03-22
GEOMESA-2227 Creating range planning values before converting to bytes

elahrvivaz:
* Core key/range logic is moved into IndexKeySpace implementations
* Allows for native range execution in non-byte oriented databases (i.e. Cassandra, Kudu)
* New AttributeIndex implemenation that standardizes shard byte to front of row


geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraDataStoreFactory.scala
```diff
@@ -133,20 +132,15 @@ object CassandraDataStoreFactory {
...
+
+    override protected def looseBBoxDefault = false
+
     val ContactPointParam = new GeoMesaParam[String]("cassandra.contact.point", "HOST:PORT to Cassandra", optional = false, deprecatedKeys = Seq("geomesa.cassandra.contact.point"))
     val KeySpaceParam     = new GeoMesaParam[String]("cassandra.keyspace", "Cassandra Keyspace", optional = false, deprecatedKeys = Seq("geomesa.cassandra.keyspace"))
     val CatalogParam      = new GeoMesaParam[String]("cassandra.catalog", "Name of GeoMesa catalog table", optional = false, deprecatedKeys = Seq("geomesa.cassandra.catalog.table"))
     val UserNameParam     = new GeoMesaParam[String]("cassandra.username", "Username to connect with", deprecatedKeys = Seq("geomesa.cassandra.username"))
     val PasswordParam     = new GeoMesaParam[String]("cassandra.password", "Password to connect with", password = true, deprecatedKeys = Seq("geomesa.cassandra.password"))
 
-    override val LooseBBoxParam: GeoMesaParam[java.lang.Boolean] = {
```
jnh5y: Just so that I know for later, can you elaborate/remind me about the C* story here?

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraAttributeIndex.scala
```diff
@@ -9,13 +9,106 @@
...
+
+  private val Index     = NamedColumn("attrIdx",   0, "smallint", classOf[Short],  partition = true)
+  private val Value     = NamedColumn("attrVal",   1, "text",     classOf[String])
+  private val Secondary = NamedColumn("secondary", 2, "blob",     classOf[ByteBuffer])
+  private val FeatureId = CassandraFeatureIndex.featureIdColumn(3)
+  private val Feature   = CassandraFeatureIndex.featureColumn(4)
+
+  override val name: String = AttributeIndex.Name
+
   override val version: Int = 2
```
jnh5y: Just to confirm, V2 is correct, right?

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraDataStoreFactory.scala
elahrvivaz: with our other data stores, we use the z-filter for our loose bbox queries, so they're still fairly accurate. with c* we don't do that, so they tend to be pretty inaccurate. so i defaulted c* to not use loose bboxes a little while ago - this change just makes it easier for other datastores to do the same thing without the ugly copying of the param

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/index/CassandraAttributeIndex.scala
elahrvivaz: yeah, the change in the attribute index is that the shard byte and the attribute identifier bytes were flipped position. since c* already had those as separate columns, i didn't need to change the data format.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/utils/HBaseBatchScan.scala
```diff
@@ -15,9 +15,10 @@ import org.apache.hadoop.hbase.client._
 import org.locationtech.geomesa.index.utils.AbstractBatchScan
 
 class HBaseBatchScan(connection: Connection, tableName: TableName, ranges: Seq[Scan], threads: Int, buffer: Int)
-    extends AbstractBatchScan[Scan, Result](ranges, threads, buffer) {
-
-  private lazy val table = connection.getTable(tableName)
+    extends {
+      // use early initialization to ensure table is open before scans kick off
+      private val table = connection.getTable(tableName)
```
jnh5y: Wow.  Nice!

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreFactory.scala
```diff
@@ -28,6 +28,7 @@ object GeoMesaDataStoreFactory {
   val QueryThreadsParam  = new GeoMesaParam[Integer]("geomesa.query.threads", "The number of threads to use per query", default = Int.box(8), deprecatedKeys = Seq("queryThreads", "accumulo.queryThreads"))
   val QueryTimeoutParam  = new GeoMesaParam[Duration]("geomesa.query.timeout", "The max time a query will be allowed to run before being killed, e.g. '60 seconds'", deprecatedParams = Seq(DeprecatedTimeout, DeprecatedAccumuloTimeout), systemProperty = Some(TimeoutSysParam))
   val LooseBBoxParam     = new GeoMesaParam[java.lang.Boolean]("geomesa.query.loose-bounding-box", "Use loose bounding boxes - queries will be faster but may return extraneous results", default = true, deprecatedKeys = Seq("looseBoundingBox"))
+  val StrictBBoxParam    = new GeoMesaParam[java.lang.Boolean]("geomesa.query.loose-bounding-box", "Use loose bounding boxes - queries will be faster but may return extraneous results", default = false, deprecatedKeys = Seq("looseBoundingBox"))
```
jnh5y: Hmm...  is there an error if one tries to see both of the bbox params to true:)?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/BaseFeatureIndex.scala
```diff
@@ -24,41 +25,57 @@ import org.opengis.filter.Filter
...
 
   import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 
-  protected def keySpace: IndexKeySpace[K]
+  /**
+    * Primary key space used by this index
+    *
+    * @return
+    */
+  protected def keySpace: IndexKeySpace[V, K]
```
jnh5y: What do the types V and K here mean?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreFactory.scala
elahrvivaz: they shouldn't both be used by a single data store - the mixin trait `GeoMesaDataStoreParams` (below) picks one based on a config value

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/BaseFeatureIndex.scala
elahrvivaz: i think it was value and key... i'll add a scaladoc comment on the trait

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/BaseTieredFeatureIndex.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
+
+  /**
+    * Tiered key space beyond the primary one, if any
+    *
+    * @param sft simple feature type
+    * @return
+    */
+  protected def tieredKeySpace(sft: SimpleFeatureType): Option[IndexKeySpace[_, _]]
```
jnh5y: Why did this need to be optional?

elahrvivaz: the tiered keyspace available might not support the simple feature type - e.g. we could have an SFT without a geometry, in which case none of the secondary index options would be valid. kind of a fringe case, but...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
```diff
@@ -59,4 +63,99 @@ trait IndexKeySpace[T] {
...
+
+object IndexKeySpace {
+
+  type ToIndexKeyBytes = (Seq[Array[Byte]], SimpleFeature, Array[Byte]) => Seq[Array[Byte]]
+
+  sealed trait ByteRange
+
+  case class BoundedByteRange(lower: Array[Byte], upper: Array[Byte]) extends ByteRange
+  case class SingleRowByteRange(row: Array[Byte]) extends ByteRange
+  case class TieredByteRange(lower: Array[Byte],
```
jnh5y: Ok, how's a TieredByteRange different from a BoundedByteRange?

Do you just need to pass around the pair of Booleans to indicated whether or not key ranges can be prepended and/or appended?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
```diff
@@ -59,4 +63,99 @@ trait IndexKeySpace[T] {
...
+                             upper: Array[Byte],
+                             lowerTierable: Boolean = false,
+                             upperTierable: Boolean = false) extends ByteRange
+
+  object ByteRange {
+
+    import org.locationtech.geomesa.utils.conversions.ScalaImplicits.RichTraversableOnce
+
+    implicit val ByteOrdering: Ordering[Array[Byte]] =
+      Ordering.comparatorToOrdering(UnsignedBytes.lexicographicalComparator)
```
jnh5y: "Who signed off on this?"

jnh5y: Wait no... better:  "Who signed off on this *order*?"

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
```diff
@@ -59,4 +63,99 @@ trait IndexKeySpace[T] {
...
+
+    val UnboundedLowerRange: Array[Byte] = Array.empty
+    val UnboundedUpperRange: Array[Byte] = Array.fill(3)(ByteArrays.MaxByte)
```
jnh5y: (Sanity checks): Is the empty Array of bytes the smallest possible Array[Byte]?  

On the other end, what would have to happen to have the first three bytes of a GeoMesa key be MaxByte?  If we had 255 shards.... and ...  (I suppose I'm wondering how often such a corner case could come up...)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
```diff
@@ -59,4 +63,99 @@ trait IndexKeySpace[T] {
...
+
+    def ranges(prefixes: Seq[Array[Byte]], row: Array[Byte]): Iterator[ByteRange] = {
```
jnh5y: Just to clarify... the prefixes here are actually *distinct* prefixes correct?  (I.e., not a sequence of byte array to combine into one prefix?)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
elahrvivaz: yeah - the bounded byte range is assumed to be both prependable and appendable, otherwise you'd use this one. I'll add some more scaladocs... :)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/attribute/AttributeIndexKey.scala
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  }
+}
+
+object AttributeIndexKey {
+
+  private val typeRegistry = LexiTypeEncoders.LEXI_TYPES
+
+  // store 2 bytes for the index of the attribute in the sft - this allows up to 32k attributes in the sft.
```
jnh5y: Just 32k columns?!  We might should document that somewhere.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/attribute/AttributeIndexKeySpace.scala
```diff
@@ -0,0 +1,281 @@
...
+  override def getRanges(values: AttributeIndexValues[Any]): Iterator[ScanRange[AttributeIndexKey]] = {
+
+    import AttributeIndexKey.encodeForQuery
+    import org.locationtech.geomesa.filter.WILDCARD_SUFFIX
+
+    val AttributeIndexValues(_, i, fb, binding) = values
+
+    if (fb.isEmpty) {
+      // we have an attribute, but weren't able to extract any bounds... scan all values
+      Iterator.single(UnboundedRange(AttributeIndexKey(i.toShort, null, inclusive = false)))
```
jnh5y: When would we want to do a full-table scan on the attribute index?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/attribute/AttributeIndexKeySpace.scala
```diff
@@ -0,0 +1,281 @@
...
+      case LowerBoundedRange(lo) =>
+        TieredByteRange(lower(lo), upper(AttributeIndexKey(lo.i, null)), lowerTierable = true)
+
+      case UpperBoundedRange(hi) =>
+        tieredUpper(hi) match {
+          case None     => TieredByteRange(lower(AttributeIndexKey(hi.i, null)), upper(hi))
+          case Some(up) => TieredByteRange(lower(AttributeIndexKey(hi.i, null)), up, upperTierable = true)
+        }
+
+
```
jnh5y: (Picky) Extra white space.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
elahrvivaz: yeah, an empty array sorts before any non-empty array, so it should be good as a lower range. for the upper range, it doesn't involve shards or table sharing prefixes, so it's just the raw value. for our z-indices, we don't use the first 1-2 bits of the z value, so this should always sort after any valid value (z3 uses 2 bytes for the bin, then the z value, which serendipitously lines up with the 3 byte size i randomly chose). for attribute indices, it's possible but seems unlikely, since it's not a printable char...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
elahrvivaz: right - i'll add more scaladocs :)

elahrvivaz: actually it doesn't seem that these are actually used - i'm going to remove them

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/attribute/AttributeIndexKeySpace.scala
elahrvivaz: we probably wouldn't... but we could if we wanted to! :)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/attribute/AttributeIndexKeySpace.scala
elahrvivaz: thanks, fixed

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreFactory.scala
jnh5y: I see...  what do the override options look like?  When do I have a choice to change the behavior?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
jnh5y: For fun, could you Z3-decode the 3 MaxBytes;)?

jnh5y:
Ok, at a really, really high level.... how does one configure an attribute + {whichever (X)Z{2,3} index} ?

Is that well documented?  (If not, it can be a follow-on ticket.)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreFactory.scala
elahrvivaz: it's defined by the extending class, like so: https://github.com/elahrvivaz/geomesa/blob/fcr_index_keyspace/geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraDataStoreFactory.scala#L134-L136

elahrvivaz:
there's not any configuration that you can do, actually - that behavior isn't changed in this PR. it will pick the first of Z3,Z2,XZ3,XZ2 that supports the simple feature type, and use that as the secondary index.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
elahrvivaz: hmm, let me try encoding the max z3 instead. The max bytes won't really work as a z-value, as i think having the sign bit set causes math errors. and the first thing it does when decoding is to apply a bitmask: https://github.com/locationtech/sfcurve/blob/master/zorder/src/main/scala/org/locationtech/sfcurve/zorder/Z3.scala#L83

elahrvivaz: here's z3:
```
scala> val sfc = Z3SFC(TimePeriod.Week)
sfc: org.locationtech.geomesa.curve.Z3SFC = org.locationtech.geomesa.curve.Z3SFC@6fe595dc
scala> val z = sfc.index(sfc.lon.max, sfc.lat.max, sfc.time.max.toLong).z
z: Long = 9223372036854775807
scala> ByteArrays.toBytes(Short.MaxValue, z).map(ByteArrays.toHex).mkString(":")
res3: String = 7f:ff:7f:ff:ff:ff:ff:ff:ff:ff
scala> ByteArrays.toHex(ByteArrays.MaxByte)
res4: String = ff
```

elahrvivaz: and here's z2:
```
scala> val sfc = org.locationtech.geomesa.curve.Z2SFC
sfc: org.locationtech.geomesa.curve.Z2SFC.type = org.locationtech.geomesa.curve.Z2SFC$@76ff68c5
scala> val z = sfc.index(sfc.lon.max, sfc.lat.max).z
z: Long = 4611686018427387903
scala> ByteArrays.toBytes(z).map(ByteArrays.toHex).mkString(":")
res0: String = 3f:ff:ff:ff:ff:ff:ff:ff
```

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/IndexKeySpace.scala
jnh5y: I think the puns are quite clear....;)  Unsigned bytes?  Ordering?

PR #1903, opened 2018-03-23
GEOMESA-2230 Remove GeoTools dependencies in geomesa-spark-jts

atallahhezbor:

### DIFF
--- a/build/cqs.tsv
+++ b/build/cqs.tsv
@@ -48,6 +48,7 @@ com.thoughtworks.paranamer:paranamer	2.6	compile
 com.typesafe:config	1.2.1	compile
 com.typesafe.scala-logging:scala-logging_2.11	3.1.0	compile
 com.vividsolutions:jts-core	1.14.0	compile
+com.vividsolutions:jts-io	1.14.0	compile
 com.vlkan:flatbuffers	1.2.0-3f79e055	compile
 com.yammer.metrics:metrics-core	2.2.0	compile
 de.topobyte:adt-geo	0.0.2	compile
--- a/geomesa-spark/geomesa-spark-jts/pom.xml
+++ b/geomesa-spark/geomesa-spark-jts/pom.xml
@@ -31,15 +31,13 @@
         </dependency>
         <dependency>
             <groupId>com.vividsolutions</groupId>
+            <artifactId>jts-io</artifactId>
+            <version>${jts.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vividsolutions</groupId>
             <artifactId>jts-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-data</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-geojson</artifactId>
+            <version>${jts.version}</version>
         </dependency>
         <dependency>
             <groupId>org.locationtech.spatial4j</groupId>
--- a/geomesa-spark/geomesa-spark-jts/pom.xml
+++ b/geomesa-spark/geomesa-spark-jts/pom.xml
@@ -32,7 +32,6 @@
         <dependency>
             <groupId>com.vividsolutions</groupId>
             <artifactId>jts-io</artifactId>
-            <version>${jts.version}</version>
         </dependency>
         <dependency>
             <groupId>com.vividsolutions</groupId>
--- a/geomesa-spark/geomesa-spark-jts/pom.xml
+++ b/geomesa-spark/geomesa-spark-jts/pom.xml
@@ -37,7 +37,6 @@
         <dependency>
             <groupId>com.vividsolutions</groupId>
             <artifactId>jts-core</artifactId>
-            <version>${jts.version}</version>
         </dependency>
         <dependency>
             <groupId>org.locationtech.spatial4j</groupId>
--- a/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
+++ b/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
@@ -8,15 +8,12 @@
 
 package org.locationtech.geomesa.spark.jts.udf
 
-import java.awt.geom.AffineTransform
-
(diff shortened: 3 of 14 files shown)


geomesa-spark/geomesa-spark-jts/pom.xml
```diff
@@ -31,15 +31,13 @@
         </dependency>
         <dependency>
             <groupId>com.vividsolutions</groupId>
-            <artifactId>jts-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-data</artifactId>
+            <artifactId>jts-io</artifactId>
```
elahrvivaz: we don't currently pull in jts-io - looks like this will need a cq

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
```diff
@@ -129,14 +128,13 @@ object SpatialRelationFunctions {
 
   def fastDistance(c1: Coordinate, c2: Coordinate): Double = {
     val calc = geoCalcs.get()
-    calc.setStartingGeographicPoint(c1.x, c1.y)
-    calc.setDestinationGeographicPoint(c2.x, c2.y)
-    calc.getOrthodromicDistance
+    val startPoint = spatialContext.getShapeFactory.pointXY(c1.x, c1.y)
+    DistanceUtils.DEG_TO_KM * calc.distance(startPoint, c2.x, c2.y) * 1000
```
elahrvivaz: might be worth checking to see if this class is thread-safe - if so, we don't need the threadlocal wrapper any more

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricConstructorFunctionsTest.scala
```diff
@@ -199,7 +200,7 @@ class GeometricConstructorFunctionsTest extends Specification with TestEnvironme
           |                    st_castToPoint(st_geomFromWKT('POINT(2 2)')))
         """.stripMargin
       )
-      val expected = WKTUtils.read("POLYGON((0.0 0.0, 2.0 0.0, 2.0 2.0, 0.0 2.0, 0.0 0.0))")
+      val expected = WKTUtils.read("POLYGON((0.0 0.0, 0.0 2.0, 2.0 2.0, 2.0 0.0, 0.0 0.0))")
```
elahrvivaz: did axis order change here? that might be jarring to end users...

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
atallahhezbor: The [SpatialContext](https://locationtech.github.io/spatial4j/apidocs/com/spatial4j/core/context/SpatialContext.html) is thread safe, but I'm not sure if that still holds for the DistanceCalculator if multiple threads can call `getDistCalc`

geomesa-spark/geomesa-spark-jts/pom.xml
atallahhezbor: Sure! I also plan on using that module for the Row to GeoJSON ticket

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricConstructorFunctionsTest.scala
elahrvivaz: @jnh5y what do you think? these axis order issues are always tricky. i'd say we should keep behavior consistent with longitude first, like we normally do with `Hints.putSystemDefault(Hints.FORCE_LONGITUDE_FIRST_AXIS_ORDER, true)`

build/cqs.tsv
```diff
@@ -48,6 +48,7 @@ com.thoughtworks.paranamer:paranamer	2.6	compile
 com.typesafe:config	1.2.1	compile
 com.typesafe.scala-logging:scala-logging_2.11	3.1.0	compile
 com.vividsolutions:jts-core	1.14.0	compile
+com.vividsolutions:jts-io	1.14.0	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16011

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricConstructorFunctionsTest.scala
elahrvivaz: @jnh5y thoughts?

geomesa-spark/geomesa-spark-jts/pom.xml
```diff
@@ -31,15 +31,13 @@
...
+            <version>${jts.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-geojson</artifactId>
+            <groupId>com.vividsolutions</groupId>
+            <artifactId>jts-core</artifactId>
+            <version>${jts.version}</version>
```
jahhulbert-ccri: prob should moved these managed versions to the room pom

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricConstructorFunctionsTest.scala
jnh5y: This is a box:), so axis order and polygon winding order can be confused/conflated.

@tkunicki thoughts?

Also, can someone take it from the top for me?

elahrvivaz: good point, maybe just the point order changed, not the axis order. maybe we could re-write the test with different lengths for x and y so we can tell? we'd have to run the original as well

PR #1907, opened 2018-03-27
GEOMESA-2231 Add Row to GeoJSON converter class

atallahhezbor:


docs/user/spark/spark_jts.rst
```diff
@@ -132,6 +132,38 @@ A complete list of the implemented UDFs is given in the next section (:doc:`./sp
...
 .. _OpenGIS Simple feature access SQL option: http://www.opengeospatial.org/standards/sfs
 
+GeoJSON Output
+^^^^^^^^^^^^^^
+
+The Spark JTS module also provides a means of exporting a ``DataFrame`` to a `GeoJSON <http://geojson.org/>`__ string.
+This allows for quick visualization of the data in many front-end mapping libraries that support GeoJSON input such as
+Leaflet or Open Layers.
+
+To convert a DataFrame, each partition can be map its rows to GeoJSON by instantiating a ``RowGeoJSON`` converter.
```
elahrvivaz: grammar typo: 'each partition can be map its rows'

docs/user/spark/spark_jts.rst
```diff
@@ -132,6 +132,38 @@ A complete list of the implemented UDFs is given in the next section (:doc:`./sp
...
+
+.. code-block:: scala
+
+    import org.locationtech.geomesa.spark.jts.util.RowGeoJSON
```
elahrvivaz: thanks for adding the imports - every time i try to execute one of these snippets i have to track down the imports :)

docs/user/spark/spark_jts.rst
```diff
@@ -132,6 +132,38 @@ A complete list of the implemented UDFs is given in the next section (:doc:`./sp
...
+    val schema = df.schema
+    val geojsonDf = df.mapPartitions { iter =>
+        val row2GeoJSON = new RowGeoJSON(schema)
+        iter.map { r => rowJSON.toString(r) }
+    }
+
+If the result can fit in memory, it can then be collected on the driver and written to a file. If not, each executor can
+write to a distributed file system like HDFS.
+
+.. code::
```
elahrvivaz: can add 'scala' typing here

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/RowGeoJSON.scala
```diff
@@ -0,0 +1,75 @@
...
+    val geometry = row.getAs[Geometry](geomOrdinal.get)
+    if (geometry != null) {
+      sb.append(geomJson.write(row.getAs[Geometry](geomOrdinal.get))) // write geometry
+    } else {
+      sb.append(""" "null" """)
+    }
+
+    sb.append(""", "properties":{ """) // start properties
+
+    var i = 0 //1 // start at 1 to skip fid
```
elahrvivaz: should this be 1?

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/RowGeoJSON.scala
```diff
@@ -0,0 +1,75 @@
...
+
+import java.{lang => jl}
+
+import com.vividsolutions.jts.geom.Geometry
+import com.vividsolutions.jts.io.geojson.GeoJsonWriter
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.jts.JTSTypes
+import org.apache.spark.sql.types._
+
+class RowGeoJSON(structType: StructType, var geomOrdinal: Option[Int] = None) {
```
aheyne: To be picky here, I'd put `setGeometryOrdinal()` into a `RowGeoJSON` companion object and call the function as the default. e.g. 
```
import RowGeoJSON.defaultGeometryOrdinal
class RowGeoJSON(structType: StructType, geomOrdinal: Int = defaultGeometryOrdinal) {...}
object RowGeoJSON { 
  def defaultGeometryOrdinal: Int = ...
}
```
I see calling option.get later as a bad thing in general. But that may just be personal preference.

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/util/RowGeoJSONTest.scala
```diff
@@ -0,0 +1,161 @@
...
+import org.junit.runner.RunWith
+import org.locationtech.geomesa.spark.jts._
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class RowGeoJSONTest extends Specification with TestEnvironment {
+
+  "Row to GeoJSON converter" should {
+    sequential
```
aheyne: (Also picky, hey, I have to do something to feel good about myself) 
Does this really need to be sequential?

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/RowGeoJSON.scala
atallahhezbor: Yeah I felt like there was a better pattern out there when writing this. Was kinda hoping someone would point one out ;)

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/RowGeoJSON.scala
atallahhezbor: Oops, that's left over from when I was assuming the rows would always be built from simple features.

atallahhezbor:
@aheyne Re: extending the Row class, I don't think we could easily extend `Row` without making it logically different from `Row`, at which point we couldn't use it with DataFrames.
We could add an extension to DataFrames with implicits, at which point we could do `df.toGeoJSON`. It would just be hiding the same code, but it may make for a better experience.

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/util/RowGeoJSON.scala
elahrvivaz: yeah, it's probably better to move the logic in `setGeometryOrdinal` into an `apply` method in a companion object. you could then remove the default value for geomOrdinal.

elahrvivaz: basically what @aheyne said, but since you need to pass in the structType you won't be able to use it as a default value (unless you create a secondary argument list, but eh)

aheyne:
@atallahhezbor I meant more like making an Implicit class to add the function to the row. e.g.
```
object RowGeoJson {
  implicit class GeoJsonPrintableRow(row: Row) {
    def toGeoJson(): String = { ... }
  }
} 
```
Which would give you something like 
```
import RowGeoJson._
df.mapPartitions { iter => iter.map(_.toGeoJson) }
```
I'd defer to @elahrvivaz 's opinion as that can make the API a little less clear to learn but easier to use.

aheyne:
Oh, also can you run buildcqs so we can get the new dependency in the list?

elahrvivaz:
this needs build/cqs to be run, although that may wash out with the other jts-io PR.

geomesa-spark/geomesa-spark-jts/pom.xml
```diff
@@ -29,6 +29,11 @@
             <version>${spark.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.vividsolutions</groupId>
+            <artifactId>jts-io</artifactId>
+            <version>${jts.version}</version>
```
elahrvivaz: as noted in the other PR, this version should be managed in the main pom

PR #1908, opened 2018-03-28
GEOMESA-2239 Delete gmhadoopinfrastructure.png

nstires-boundless:
Remove duplicate, used image that causes havoc on some OSes. The file of the same name with upper case is called in the docs.


jnh5y:
@nstires-boundless Can you be more specific about the problem?  Is it just the case-sensitivity of the filename?

Also, have you filled out an ECA and can you sign-off on your commits?  https://github.com/locationtech/geomesa/blob/master/CONTRIBUTING.md

nstires-boundless:
@jnh5y Thanks for looking at this. I have created an Eclipse account and signed the Contributor Agreement.

This has to to with handling of case sensitivity within volumes. Operating Systems handle it differently, and having two files of the same name, with different cases, should be avoided. It throws error on some Git clients when it comes across this scenario, making it difficult to work with.

Here's an example: https://stackoverflow.com/questions/40488777/git-remove-same-file-with-different-capitalization

Looking at the source, the file that's called is the version of mixed case, so the all lowercase version should be removed.

elahrvivaz:
@nstires-boundless thanks for catching that

PR #1910, opened 2018-03-30
GEOMESA-2243 Upgrade surefire plugin

jahhulbert-ccri:


jahhulbert-ccri:
didn't trigger any CQs which I guess is good...i'm going to review additional plugins at some point but we'll hold off for now.

PR #1911, opened 2018-03-30
WIP Initial Kudu DataStore implementation

elahrvivaz:
* Point and non-point geometries
* Push-down predicates
* Split configuration and dynamic range splits where possible
* No unit tests - Kudu doesn't have an embedded testing framework


geomesa-kudu/geomesa-kudu-datastore/src/main/scala/org/locationtech/geomesa/kudu/index/KuduFeatureIndex.scala
```diff
@@ -0,0 +1,295 @@
+/***********************************************************************
+ * Copyright (c) 2017-2018 IBM
```
jnh5y: Did we copy GeoMesa-C* code here?

elahrvivaz: oops, yeah haha

PR #1912, opened 2018-03-31
GEOMESA-2244 Pull the zookeepers param from the hbase-site.xml

aheyne:
hbase-site.xml in the conf dir has been tested for distributed ingest.

### DIFF
--- a/geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/HBaseDataStoreCommand.scala
+++ b/geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/HBaseDataStoreCommand.scala
@@ -12,6 +12,9 @@ import com.beust.jcommander.Parameter
 import org.locationtech.geomesa.hbase.data.{HBaseDataStore, HBaseDataStoreParams}
 import org.locationtech.geomesa.hbase.tools.HBaseDataStoreCommand.HBaseParams
 import org.locationtech.geomesa.tools.{CatalogParam, DataStoreCommand, OptionalZookeepersParam}
+import org.locationtech.geomesa.utils.conf.GeoMesaSystemProperties.SystemProperty
+
+import scala.xml.{Node, XML}
 
 /**
  * Abstract class for commands that have a pre-existing catalog
@@ -21,9 +24,35 @@ trait HBaseDataStoreCommand extends DataStoreCommand[HBaseDataStore] {
   override def params: HBaseParams
 
   override def connection: Map[String, String] = {
+
+    val HBASE_SITE_FILE_NAME = "hbase-site.xml"
+
+    def extractZookeepers(site: Node): String = {
+      val props = site \\ "configuration" \\ "property"
+      props.find(prop => (prop \ "name").text == "hbase.zookeeper.quorum") match {
+        case Some(prop) => (prop \ "value").text
+        case None => null
+      }
+    }
+
+    val zkParam: String = if (params.zookeepers != null) {
+      // Command line takes precedence
+      params.zookeepers
+    } else {
+      // Pull from System Properties first then look for the hbase-site on the classpath
+      SystemProperty("geomesa.hbase.site.xml.path").option match {
+        case Some(path) => extractZookeepers(XML.loadFile(path))
+        case None =>
+          Option(getClass.getClassLoader.getResourceAsStream(HBASE_SITE_FILE_NAME)) match {
+            case Some(stream) => extractZookeepers(XML.load(stream))
+            case None => null
+          }
+      }
+    }
+
     Map(
       HBaseDataStoreParams.HBaseCatalogParam.getName    -> params.catalog,
-      HBaseDataStoreParams.ZookeeperParam.getName       -> params.zookeepers,
+      HBaseDataStoreParams.ZookeeperParam.getName       -> zkParam,
       HBaseDataStoreParams.RemoteFilteringParam.getName -> (!params.noRemote).toString
     ).filter(_._2 != null)
   }
--- a/geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
+++ b/geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
@@ -315,4 +315,12 @@
         <final>false</final>
     </property>
 
+    <property>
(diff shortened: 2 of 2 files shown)


geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/HBaseDataStoreCommand.scala
```diff
@@ -21,9 +24,35 @@ trait HBaseDataStoreCommand extends DataStoreCommand[HBaseDataStore] {
...
+    val zkParam: String = if (params.zookeepers != null) {
+      // Command line takes precedence
+      params.zookeepers
+    } else {
+      // Pull from System Properties first then look for the hbase-site on the classpath
+      SystemProperty("geomesa.hbase.site.xml.path").option match {
+        case Some(path) => extractZookeepers(XML.loadFile(path))
+        case None =>
+          Option(getClass.getClassLoader.getResourceAsStream(HBASE_SITE_FILE_NAME)) match {
+            case Some(stream) => extractZookeepers(XML.load(stream))
```
elahrvivaz: does this handle closing the stream?

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
```diff
@@ -315,4 +315,12 @@
         <final>false</final>
     </property>
 
+    <property>
+        <name>geomesa.hbase.site.xml.path</name>
```
elahrvivaz: could you add a snippet in the docs about this param?

elahrvivaz:
Nice, this should reduce a lot of the friction around m/r and the hbase-site.xml. I'm not entirely sure that this will work in all cases - i.e. if the hbase-site.xml contains other settings that could mess up creating a connection (the same is true for the zk param in general though...)

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/HBaseDataStoreCommand.scala
```diff
@@ -21,9 +24,35 @@ trait HBaseDataStoreCommand extends DataStoreCommand[HBaseDataStore] {
   override def params: HBaseParams
 
   override def connection: Map[String, String] = {
+
+    val HBASE_SITE_FILE_NAME = "hbase-site.xml"
+
+    def extractZookeepers(site: Node): String = {
```
elahrvivaz: would it make sense to extract this with a hadoop Configuration object instead? I believe you could do something like:
```
val conf = new Configuration()
conf.addResources(foo)
conf.getString("hbase.zookeeper.quorom")
```
That should handle property substitution as well (if there are any)
@jahhulbert-ccri does that seem reasonable?

aheyne: ah, yeah that might be the more canonical method.

jahhulbert-ccri: yeah we shouldn't hard code hbase-site.xml bc that can be renamed and the real thing is that the param should be in the Configuration option...we already have some code around this here that needs to be respected when "creating" confs: https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala#L128

Also the line ```val conf = new Configuration()``` can take up to 1-2 seconds to do because it runs through the entire classpath (or used to at least). so you don't want to be creating new ones. I think we should figure out the interaction between this pr, the existing HbaseConnectionPool withPaths method, and the ticket https://geomesa.atlassian.net/browse/GEOMESA-2236

we are breaking in spark for the precise reason we call new Configuration() and lose the old one.

geomesa-utils/src/main/resources/org/locationtech/geomesa/geomesa-site.xml.template
jahhulbert-ccri: is this a duplicate of https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreFactory.scala#L193

jahhulbert-ccri: we may just need better doc?

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/HBaseDataStoreCommand.scala
```diff
@@ -21,9 +25,26 @@ trait HBaseDataStoreCommand extends DataStoreCommand[HBaseDataStore] {
...
   override def connection: Map[String, String] = {
+
+    val HBASE_SITE_FILE_NAME = "hbase-site.xml"
+    val HBASE_ZOOKEEPER_QUORUM = "hbase.zookeeper.quorum"
+
+    val zkParam: String = if (params.zookeepers != null) {
+      // Command line takes precedence
+      params.zookeepers
+    } else {
+      val conf = new Configuration()
```
jahhulbert-ccri: still need to look at how this interacts with the other resources...i'll comment soon.

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/HBaseDataStoreCommand.scala
```diff
@@ -21,9 +25,26 @@ trait HBaseDataStoreCommand extends DataStoreCommand[HBaseDataStore] {
...
+      // Pull from System Properties first then look for the hbase-site on the classpath
+      SystemProperty("geomesa.hbase.site.xml.path").option match {
```
jahhulbert-ccri: either way still should replace this with the existing param i think so we can pass in multiple resources and they all make it into the remote conf for the MR job?

elahrvivaz:
should we close this PR?

PR #1914, opened 2018-04-02
GEOMESA-2242 Add filtered conf to dist

jahhulbert-ccri:


jahhulbert-ccri:
I went with a new directory so we don't have to add each individual filtered conf file in the future...we could just filter the whole conf dir if we think that's better?

PR #1916, opened 2018-04-02
GEOMESA-2218 Create printf converter transform

jnh5y:
* Adds format/printf converter functions.


docs/user/convert/function_usage.rst
```diff
@@ -130,6 +130,19 @@ Usage: ``emptyToNull($0)``
 
 Example: ``emptyToNull('') = null``
 
+printf
+^^^^^^^^^^^
```
elahrvivaz: carats need to align with the header or it won't render correctly

elahrvivaz:
:+1:

PR #1917, opened 2018-04-02
GEOMESA-2210 Remove spark sql web endpoint from geomesa-web-data

jnh5y:


jnh5y:
Ok, so let's hope that the third time is a charm;).  Let's go, Travis!

PR #1920, opened 2018-04-03
GEOMESA-2232 Update examples with new executable

atallahhezbor:

### DIFF
--- a/geomesa-hbase/geomesa-hbase-tools/examples/ingest/avro/ingest_avro.sh
+++ b/geomesa-hbase/geomesa-hbase-tools/examples/ingest/avro/ingest_avro.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
 # type and converter example-xml is registered in $GEOMESA_HOME/conf/application.conf
-bin/geomesa-habse ingest -c catalog -s example-avro -C example-avro-no-header example_no_header.avro
+bin/geomesa-hbase ingest -c catalog -s example-avro -C example-avro-no-header example_no_header.avro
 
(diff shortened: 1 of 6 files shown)


geomesa-hbase/geomesa-hbase-tools/examples/ingest/avro/ingest_avro.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
 # type and converter example-xml is registered in $GEOMESA_HOME/conf/application.conf
-bin/geomesa-habse ingest -c catalog -s example-avro -C example-avro-no-header example_no_header.avro
+bin/geomesa-hbase ingest -c catalog -s example-avro -C example-avro-no-header example_no_header.avro
 
-bin/geomesa-habse ingest -c catalog -s example-avro -C example-avro-header example_with_header.avro
+bin/geomesa-hbase ingest -c catalog -s example-avro -C example-avro-header example_with_header.avro
```
elahrvivaz: ah thanks, i noticed this this other day but then forgot about ti

jnh5y:
This is a good start.  There are lots of other examples of 'geomesa' as a command by itself...  For instance, https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-tools/README.md.

I'd suggest looking for examples with 'git grep geomesa\ '.

elahrvivaz:
we should probably just get rid of all the sub-modules readmes - i don't think they have been kept up to date

elahrvivaz:
I opened a new ticket here to track the doc updates: https://geomesa.atlassian.net/browse/GEOMESA-2246

PR #1927, opened 2018-04-04
GEOMESA-2250 Distribute the hbase-site configuration to distributed jobs

aheyne:

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
@@ -21,18 +21,19 @@ import org.apache.hadoop.hbase.security.User
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.hadoop.security.UserGroupInformation.AuthenticationMethod
 import org.locationtech.geomesa.hbase.data.HBaseDataStoreFactory.{HBaseGeoMesaKeyTab, HBaseGeoMesaPrincipal}
-import org.locationtech.geomesa.hbase.data.HBaseDataStoreParams.{ConfigPathsParam, ConnectionParam, ZookeeperParam}
+import org.locationtech.geomesa.hbase.data.HBaseDataStoreParams.{ConfigPathsParam, ConnectionParam, ZookeeperParam, HadoopConfigurationParam}
 
 object HBaseConnectionPool extends LazyLogging {
 
   private var userCheck: Option[User] = _
 
+  private val zkConfiguration = withPaths(HBaseConfiguration.create(), HBaseDataStoreFactory.ConfigPathProperty.option)
+
   private val zkConfigCache = Caffeine.newBuilder().build(
     new CacheLoader[(Option[String], Option[String]), Configuration] {
       override def load(key: (Option[String], Option[String])): Configuration = {
         val (zookeepers, paths) = key
-        val configuration = withPaths(HBaseConfiguration.create(), HBaseDataStoreFactory.ConfigPathProperty.option)
-        val conf = withPaths(configuration, paths)
+        val conf = withPaths(zkConfiguration, paths)
         zookeepers.foreach(zk => conf.set(HConstants.ZOOKEEPER_QUORUM, zk))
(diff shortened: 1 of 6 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
```diff
@@ -27,24 +27,33 @@ object HBaseConnectionPool extends LazyLogging {
...
 
-  // add common resources from system property
-  private val configuration = withPaths(HBaseConfiguration.create(), HBaseDataStoreFactory.ConfigPathProperty.option)
-
-  private val configCache = Caffeine.newBuilder().build(
+  private val zkConfigCache = Caffeine.newBuilder().build(
     new CacheLoader[(Option[String], Option[String]), Configuration] {
       override def load(key: (Option[String], Option[String])): Configuration = {
         val (zookeepers, paths) = key
+        val configuration = withPaths(HBaseConfiguration.create(), HBaseDataStoreFactory.ConfigPathProperty.option)
```
elahrvivaz: can we move this back outside the cache, or will that mess something up? that way we don't have to re-load the base config for different params.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
```diff
@@ -76,10 +85,14 @@ object HBaseConnectionPool extends LazyLogging {
   def getConnection(params: java.util.Map[String, Serializable], validate: Boolean): Connection = {
     if (ConnectionParam.exists(params)) {
       ConnectionParam.lookup(params)
+    } else if (params.containsKey("hadoop.configuration")) {
+      val conf = params.get("hadoop.configuration").asInstanceOf[Configuration]
```
elahrvivaz: let's make this an actual data store parameter - but don't add it to the 'getParamterInfo' method, as we don't need it in geoserver

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/ConverterIngest.scala
```diff
@@ -75,7 +78,8 @@ class ConverterIngest(sft: SimpleFeatureType,
     new LocalIngestConverterImpl(sft, path, converters, failures)
 
   override def runDistributedJob(statusCallback: StatusCallback): (Long, Long) = {
-    new ConverterIngestJob(dsParams, sft, converterConfig, inputs, libjarsFile, libjarsPaths).run(statusCallback)
+    val conf = configuration.flatMap(conf => Some(new Configuration(conf)))
```
elahrvivaz: this could just be a regular .map, without using a Some

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
elahrvivaz: hmm, i guess hbase isn't in scope where it's being used in geomesa-jobs... and hadoop isn't in scope in geomesa-index-api

elahrvivaz: i guess just make the key a constant in GeoMesaDataStoreFactory

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
aheyne: We can move it out only for the zkConfigCache, the confConfigCache has to hand it the incoming configuration. I moved it in here since I thought it made it less confusing but we can move it. out.

elahrvivaz:
looks like there are some compile errors

elahrvivaz:
going to close this, we can re-open when we get back to working on it. I linked this PR to the ticket in JIRA so we don't lose track of it.

PR #1932, opened 2018-04-10
GEOMESA-2256 HBase - Feature ID is garbled for queries against attrib…

elahrvivaz:
…ute index when not returning a geometry

* Also fixed input length in KryoBufferSimpleFeature.setBuffer

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
@@ -72,7 +72,7 @@ class HBaseDataStoreTest extends HBaseTest with LazyLogging {
       val ids = fs.addFeatures(new ListFeatureCollection(sft, toAdd))
       ids.asScala.map(_.getID) must containTheSameElementsAs((0 until 10).map(_.toString))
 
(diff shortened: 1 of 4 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
```diff
@@ -72,7 +72,7 @@ class HBaseDataStoreTest extends HBaseTest with LazyLogging {
       val ids = fs.addFeatures(new ListFeatureCollection(sft, toAdd))
       ids.asScala.map(_.getID) must containTheSameElementsAs((0 until 10).map(_.toString))
 
-      val transformsList = Seq(null, Array("geom"), Array("geom", "dtg"), Array("geom", "name"))
+      val transformsList = Seq(null, Array("geom"), Array("geom", "dtg"), Array("name"))
```
jnh5y: Despite the longer runtimes, would it be better to add instead of replacing a test case?

elahrvivaz: i'd say it's ok to remove it... this originally always included geom b/c we always returned geom, but now we don't have to.

jnh5y: Shipping it!

PR #1936, opened 2018-04-19
GEOMESA-2176 Create python conversion for JTS Geometry instances

aheyne:
Replaces #1856

### DIFF
--- a/geomesa-spark/geomesa_pyspark/src/main/python/setup.py
+++ b/geomesa-spark/geomesa_pyspark/src/main/python/setup.py
@@ -5,5 +5,5 @@ setup(
     version='${python.version}',
     url='http://www.geomesa.org',
     packages=find_packages(),
-    install_requires=['pytz','pyspark']
+    install_requires=['pytz','pyspark>=2.1.1,<2.3.0']
 )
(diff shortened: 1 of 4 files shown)


geomesa-spark/geomesa_pyspark/src/main/python/setup.py
```diff
@@ -5,5 +5,5 @@
     version='${python.version}',
     url='http://www.geomesa.org',
     packages=find_packages(),
-    install_requires=['pytz','pyspark']
+    install_requires=['pytz','pyspark>=2.1.1,<2.3.0']
```
tkunicki: `pyspark` shouldn't be a dependency at all, it's included w/ Spark and the bootstrapping pulls in the version distributed via `SPARK_HOME` or it's equivalent.  There are differences w/ the pyspark module between versions and they are mated to spark versions, so this is generally dangerous.  It appears `pyspark` was added earlier with a "documentation-only" PR

jnh5y: @aheyne what happens if we remove the dependency on pyspark here?

elahrvivaz:
oops, didn't realize this was against the wrong branch - i forced pushed the rc.1 branch to revert the change and cherry-picked into master

PR #1940, opened 2018-04-24
GEOMESA-2113 Upgrade Arrow to 0.9

elahrvivaz:

TODO:
- [ ] update to arrow version with bugfix
- [ ] open cqs
- [ ] update arrow sample files checked into test directory with new IPC format


elahrvivaz:
Found a bug in arrow that will need to be fixed before this can be merged: https://issues.apache.org/jira/browse/ARROW-2500
Since that will require new cqs, will need to file those afterwards.

elahrvivaz:
- [ ] update to arrow version with bugfix
- [ ] open cqs
- [ ] update arrow sample files checked into test directory with new IPC format

jahhulbert-ccri:
@elahrvivaz could we have an arrow09 mime type?

jahhulbert-ccri:
ah its not backwards compat at all :( sad

elahrvivaz:
bhulette said that the javascript impl seemed to work with different versions, so I'm going to look into it some more. But I tried with a 0.6 distributed runtime and 0.9 client and it threw an arrow message serialization exception. Since they haven't released 1.0 yet, they're not focusing on back-compatibility.

elahrvivaz:
There is an arrow PR up [here](https://github.com/apache/arrow/pull/2008) that fixes the bug. Tests passed with that branch. It should be available in arrow 0.10

elahrvivaz:
replaced with https://github.com/locationtech/geomesa/pull/2018

PR #1942, opened 2018-04-25
GEOMESA-2270 Infer schema and converter from geojson file

atallahhezbor:

### DIFF
--- a/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
+++ b/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
@@ -21,7 +21,7 @@ import org.apache.commons.io.IOUtils
 import org.locationtech.geomesa.convert.LineMode.LineMode
 import org.locationtech.geomesa.convert.Transformers.Expr
 import org.locationtech.geomesa.convert._
-import org.locationtech.geomesa.utils.text.WKTUtils
+import org.locationtech.geomesa.utils.text.{DateParsing, WKTUtils}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.JavaConversions._
@@ -112,7 +112,7 @@ object JsonField {
             jsonType: String,
             pathIsRoot: Boolean): BaseJsonField[_] = jsonType match {
     case "string"           => StringJsonField(name, expression, jsonConfig, transform, pathIsRoot)
-    case "date"             => StringJsonField(name, expression, jsonConfig, transform, pathIsRoot)
+    case "date"             => DateJsonField(name, expression, jsonConfig, transform, pathIsRoot)
     case "float"            => FloatJsonField(name, expression, jsonConfig, transform, pathIsRoot)
     case "double"           => DoubleJsonField(name, expression, jsonConfig, transform, pathIsRoot)
     case "int" | "integer"  => IntJsonField(name, expression, jsonConfig, transform, pathIsRoot)
@@ -186,6 +186,11 @@ case class DoubleJsonField(name: String, expression: JsonPath, jsonConfig: Confi
 case class StringJsonField(name: String, expression: JsonPath, jsonConfig: Configuration, transform: Expr, pathIsRoot: Boolean)
     extends BaseJsonField[java.lang.String] {
   override def getAs(el: JsonElement): String = if (el.isJsonNull) null else el.getAsString
+}
+
+case class DateJsonField(name: String, expression: JsonPath, jsonConfig: Configuration, transform: Expr, pathIsRoot: Boolean)
+  extends BaseJsonField[java.util.Date] {
+  override def getAs(el: JsonElement): java.util.Date = if (el.isJsonNull) null else DateParsing.parseDate(el.getAsString)
 }
 
 case class GeometryJsonField(name: String, expression: JsonPath, jsonConfig: Configuration, transform: Expr, pathIsRoot: Boolean)
--- a/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
+++ b/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
@@ -112,6 +112,7 @@ object JsonField {
             jsonType: String,
             pathIsRoot: Boolean): BaseJsonField[_] = jsonType match {
     case "string"           => StringJsonField(name, expression, jsonConfig, transform, pathIsRoot)
+    case "date"             => StringJsonField(name, expression, jsonConfig, transform, pathIsRoot)
     case "float"            => FloatJsonField(name, expression, jsonConfig, transform, pathIsRoot)
     case "double"           => DoubleJsonField(name, expression, jsonConfig, transform, pathIsRoot)
     case "int" | "integer"  => IntJsonField(name, expression, jsonConfig, transform, pathIsRoot)
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngestJob.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngestJob.scala
@@ -96,7 +96,6 @@ class IngestMapper extends Mapper[LongWritable, SimpleFeature, Text, SimpleFeatu
 
   override def map(key: LongWritable, sf: SimpleFeature, context: Context): Unit = {
     logger.debug(s"map key ${key.toString}, map value ${DataUtilities.encodeFeature(sf)}")
-    sf.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.TRUE)
     context.write(text, sf)
   }
 }
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngestJob.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngestJob.scala
@@ -96,6 +96,7 @@ class IngestMapper extends Mapper[LongWritable, SimpleFeature, Text, SimpleFeatu
 
   override def map(key: LongWritable, sf: SimpleFeature, context: Context): Unit = {
     logger.debug(s"map key ${key.toString}, map value ${DataUtilities.encodeFeature(sf)}")
+    sf.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.TRUE)
     context.write(text, sf)
   }
 }
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
@@ -17,6 +17,7 @@ import org.locationtech.geomesa.tools._
 import org.locationtech.geomesa.tools.utils.{CLArgResolver, DataFormats}
 import org.locationtech.geomesa.utils.io.PathUtils
 import org.opengis.feature.simple.SimpleFeatureType
+import utils.GeoJsonInference
 
 import scala.util.Try
 
@@ -31,12 +32,14 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
   def libjarsPaths: Iterator[() => Seq[File]]
 
   override def execute(): Unit = {
-    import DataFormats.{Avro, Csv, Shp, Tsv}
+    import DataFormats.{Avro, Csv, Shp, Tsv, Json}
 
     ensureSameFs(PathUtils.RemotePrefixes)
 
     val ingest = if (params.fmt == Shp) {
       createShpIngest()
+    } else if (params.spec == null && params.config == null && params.fmt == Json) {
+      createJsonIngest()
     } else if (params.spec == null && params.config == null && Seq(Csv, Tsv, Avro).contains(params.fmt)) {
       // if there is no sft and no converter passed in, try to use the auto ingest which will
       // pick up the schema from the input files themselves
@@ -82,6 +85,13 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
 
   protected def createShpIngest(): Runnable = {
     new ShapefileIngest(connection, Option(params.featureName), params.files, params.threads)
+  }
+
+  protected def createJsonIngest(): Runnable = {
+    val filePath = params.files.head
+    val sft = GeoJsonInference.inferSft(filePath)
+    val conf = GeoJsonInference.inferConfig(sft)
+    createConverterIngest(sft, conf)
   }
 
   def ensureSameFs(prefixes: Seq[String]): Unit = {
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
@@ -89,7 +89,7 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
 
   protected def createJsonIngest(): Runnable = {
     val filePath = params.files.head
-    val sft = GeoJsonInference.inferSft(filePath)
+    val sft = GeoJsonInference.inferSft(filePath, Option(params.featureName).getOrElse("geojson"))
     val conf = GeoJsonInference.inferConfig(sft)
     createConverterIngest(sft, conf)
   }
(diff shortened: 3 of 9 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -84,6 +87,13 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
     new ShapefileIngest(connection, Option(params.featureName), params.files, params.threads)
   }
 
+  protected def createJsonIngest(): Runnable = {
+    val filePath = params.files.head
+    val sft = GeoJsonInference.inferSft(filePath)
```
elahrvivaz: maybe we should require them to specify the simple feature type name? or at least allow them to override the default

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -31,12 +32,14 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
...
 
   override def execute(): Unit = {
-    import DataFormats.{Avro, Csv, Shp, Tsv}
+    import DataFormats.{Avro, Csv, Shp, Tsv, Json}
 
     ensureSameFs(PathUtils.RemotePrefixes)
 
     val ingest = if (params.fmt == Shp) {
       createShpIngest()
+    } else if (params.spec == null && params.config == null && params.fmt == Json) {
```
elahrvivaz: would it make sense to have geojson separate from json as an input format?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+import org.locationtech.geomesa.utils.text.DateParsing
+
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
+import scala.reflect.ClassTag
+
+object GeoJsonInference {
+
+  val FieldsPath = "fields"
+  val OptionsPath = "options"
```
elahrvivaz: these might already be defined as constants somewhere else you could use, although maybe not available on the current classpath...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+import org.locationtech.geomesa.utils.io.fs.FileSystemDelegate
+import org.locationtech.geomesa.utils.geotools.{AttributeSpec, CRS_EPSG_4326}
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureSpecConfig.TypeNamePath
+import org.locationtech.geomesa.utils.text.DateParsing
+
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
+import scala.reflect.ClassTag
+
+object GeoJsonInference {
```
elahrvivaz: it might be useful to put this in geomesa-utils, if the dependencies work out that way

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+  val attributeBuilder = new AttributeTypeBuilder()
+
+  val baseConfig: Config = ConfigFactory.empty()
+    .withValue("type", ConfigValueFactory.fromAnyRef("json"))
+    .withValue("feature-path", ConfigValueFactory.fromAnyRef("$[*]"))
+    .withValue("options", ConfigValueFactory.fromMap(Map(("line-mode", "multi"))))
+
+  def inferSft(filePath: String): SimpleFeatureType = {
+    val fileHandle: FileSystemDelegate.FileHandle = PathUtils.interpretPath(filePath).head
+    inferSft(fileHandle.open)
```
elahrvivaz: probably makes sense to close this here, instead of in the method below. i.e.
`WithClose(fileHandle.open)(inferSft)`
As a general guideline, I try to make the `close` of a resource the responsibility of whatever opened it. Otherwise it's easy to lose track of things not being closed, or have them closed when you don't expect it.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+    parser.nextToken()
+
+    var attributes: Map[String, AttributeDescriptor] = Map()
+    var geomTypes: Array[String] = Array()
+
+    var numRead = 0
+    while (parser.hasCurrentToken && numRead < maxToRead) {
+
+      // Advance to the geometry object and store type
+      while(nextUntil(parser, "geometry")) { parser.nextToken() }
```
elahrvivaz: this method is confusingly named IMO, since it doesn't actually advance the parser, while the overloaded definition that takes an object does.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+      geomTypes = geomTypes :+ getKeyValue(parser, "type")
+
+      // Advance to the properties object
+      while(nextUntil(parser, "properties")) { parser.nextToken() }
```
elahrvivaz: the order of json fields doesn't have any meaning does it? i.e. properties could be before the geometry and it would still be valid.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  def inferSft(is: InputStream, sftName: String = "geojson", maxToRead: Int = 10): SimpleFeatureType = {
+    val sftBuilder = new SimpleFeatureTypeBuilder()
+    val factory = new JsonFactory()
+    val parser = factory.createParser(is)
+    parser.nextToken()
+
+    var attributes: Map[String, AttributeDescriptor] = Map()
+    var geomTypes: Array[String] = Array()
```
elahrvivaz: instead of vars, it would probably make sense to use mutable collection vals (ArrayBuffer and mutable.Map)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+    }
+    is.close()
+
+    // add geometry
+    val mergedGeomType = mergeGeomTypes(geomTypes)
+    val allAttrs = attributes.values.toArray :+ buildGeomAttr(mergedGeomType)
+
+    sftBuilder.addAll(allAttrs)
+    sftBuilder.setName(sftName)
+    sftBuilder.setDefaultGeometry("geometry")
```
elahrvivaz: maybe we should call it "geom" to be consistent with our normal default

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+        (name, attrsB(name))
+      } else {
+        val JDouble = classOf[jl.Double]
+        val JInteger = classOf[jl.Integer]
+        val JLong = classOf[jl.Long]
+        (bindingA, bindingB) match {
+          case (JDouble, JInteger) => (name, attrsA(name))
+          case (JInteger, JDouble) => (name, attrsB(name))
+          case (JLong, JInteger) => (name, attrsA(name))
+          case (JInteger, JLong) => (name, attrsB(name))
```
elahrvivaz: could we also merge longs and doubles into doubles?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+        if (geomTypes.forall(_.equals(geomType))) {
+          geomType
+        } else {
+          "Geometry"
+        }
+      case None => throw new UnsupportedOperationException("Could not locate geometry in geojson file.")
+    }
+  }
+
+  // NB: This method is adapted from Apache Spark's JsonInferSchema class
```
elahrvivaz: we probably need to add the apache copyright here

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,292 @@
...
+      } else {
+        val JDouble = classOf[jl.Double]
+        val JInteger = classOf[jl.Integer]
+        val JLong = classOf[jl.Long]
+        (bindingA, bindingB) match {
+          case (JDouble, JInteger) => (name, attrsA(name))
+          case (JInteger, JDouble) => (name, attrsB(name))
+          case (JLong, JInteger) => (name, attrsA(name))
+          case (JInteger, JLong) => (name, attrsB(name))
+          case _ => (name, attrsA(name))
```
elahrvivaz: do we have any sense of weighted types? e.g. if a field parsed as a date 9 times but as a string the 10th, would we consider that a date?

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonSimpleFeatureConverter.scala
```diff
@@ -112,6 +112,7 @@ object JsonField {
             jsonType: String,
             pathIsRoot: Boolean): BaseJsonField[_] = jsonType match {
     case "string"           => StringJsonField(name, expression, jsonConfig, transform, pathIsRoot)
+    case "date"             => StringJsonField(name, expression, jsonConfig, transform, pathIsRoot)
```
elahrvivaz: we probably don't want to have a `date` type unless we're going to convert it to a Date for them...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
atallahhezbor: Yeah there are a few options that could be user specified I just wasn't sure the best way to allow that. It'd be nice to allow specifying the id field too.

Would it make sense as a command line parameter? Or should we expect them to add it to the json somewhere?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
atallahhezbor: I considered that. Right now it's kind of hidden that they're separate. Non-geo json would ingest normally is the converter is specified, but I can define a geojson input format if that would make it more clear in the code.

Do we infer the format based on the extension? Because I've seen geojson files ending in json, though I guess it's not too much to ask the user to rename.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
atallahhezbor: Mmm very good point. I made a bad assumption about geojson https://tools.ietf.org/html/rfc7946#section-1.2

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: I'm guess I'm just wondering what happens if you pass in a normal json file without any converter or anything - it's probably ok to keep them the same, as long as we handle that case with a decent error message

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: i think for ID we can just use an md5 of the row. but they can already specify the feature name on the command line, so we could just pass that in here. there's `-s` and `-f`, i think `-f` is only used to override the type name...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngestJob.scala
```diff
@@ -96,7 +96,6 @@ class IngestMapper extends Mapper[LongWritable, SimpleFeature, Text, SimpleFeatu
 
   override def map(key: LongWritable, sf: SimpleFeature, context: Context): Unit = {
     logger.debug(s"map key ${key.toString}, map value ${DataUtilities.encodeFeature(sf)}")
-    sf.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.TRUE)
```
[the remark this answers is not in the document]
atallahhezbor: Yeah it seems that AutoIngest for Avro and delimited text would also call out to this. I'm switching to taking the md5 of the json object now so it shouldn't be an issue anymore. I'll add it back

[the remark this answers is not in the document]
elahrvivaz: in general I would like to support not defining a FID, I just want to make sure nothing else gets affected.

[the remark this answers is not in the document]
atallahhezbor: Looking more into those two, it seems like they create simple features with IDs. Delimited text expects the first column to be the id, and the avro reader also reads it out of the file. 
Though the fact that the tests passed either way seems to indicate we need more tests for those two.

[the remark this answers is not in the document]
elahrvivaz: yeah, we could also set the hint in those classes instead of here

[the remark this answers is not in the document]
elahrvivaz: and yeah, test coverage for the CLI tools isn't the best...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
atallahhezbor: Any of the existing submodules make sense to house it? `conf`? Though it looks like we'd have to pull in the jackson json parser at least

elahrvivaz: hmm, if we need jackson then i'd lean towards not moving it - jackson is a pretty common dependency to have classpath conflicts with i think :P
we can re-asses if we want to use it somewhere else

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
atallahhezbor: No notion of weighted types at the moment, though that would be nice to have. There's just a hierarchy of precedence where "empty" is at the bottom. I'll play around with how fun/challenging it would be to add that

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
jnh5y: @elahrvivaz We'll probably need to add a CQ for redistribution of the source code in this file.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
```diff
@@ -0,0 +1,320 @@
...
+  def buildGeomAttr[T: ClassTag](clazz: Class[T]) : AttributeDescriptor = {
+    attributeBuilder.setBinding(clazz)
+    attributeBuilder.setName("geom")
+    attributeBuilder.userData("default", "true")
+    attributeBuilder.crs(CRS_EPSG_4326)
+    val nameType = attributeBuilder.buildGeometryType()
+    attributeBuilder.buildDescriptor("geom", nameType)
+  }
+
+
```
jnh5y: (Picky) Extra white space.

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/utils/GeoJsonInferenceTest.scala
```diff
@@ -0,0 +1,172 @@
...
+        |    "name": "Wuhu Islands",
+        |    "dtg": "2018-04-25T11:00:00"
+        |  }
+        |}]""".stripMargin
+
+      val is = new ByteArrayInputStream(geojson.getBytes())
+      val sft = GeoJsonInference.inferSft(is)
+      sft.getDescriptor("dtg").getType.getBinding mustEqual classOf[java.util.Date]
+    }
+
```
jnh5y: (Picky) Extra whitespace.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/GeoJsonInference.scala
elahrvivaz: does this count as 'redistribution'? we already have a CQ for spark, so I thought adding the copyright would be good enough, as we are 'taking ownership' of this piece of code

elahrvivaz:
replaced with #1981

PR #1946, opened 2018-05-01
GEOMESA-2263 Accurate dwithin filters

elahrvivaz:
* New FastDWithin filter that compares using meters
* DWithin filters are not re-written to degrees
* Requires distributed-runtime jar to be updated


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastDWithin.scala
```diff
@@ -0,0 +1,64 @@
...
+import org.locationtech.geomesa.utils.geotools.GeometryUtils
+import org.opengis.filter.FilterVisitor
+import org.opengis.filter.MultiValuedFilter.MatchAction
+import org.opengis.filter.expression.{Expression, Literal}
+import org.opengis.filter.spatial.DWithin
+
+object FastDWithin {
+
+  private val calculators = new ThreadLocal[GeodeticCalculator] {
+    override def initialValue: GeodeticCalculator = new GeodeticCalculator()
```
jnh5y: I think this duplicates the code/ideas of GeometricDistanceFunctions.fastDistance.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -51,106 +51,59 @@ object FilterHelper {
...
     val geom = prop.literal.evaluate(null, classOf[Geometry])
     if (geom.getUserData == SafeGeomString) {
       op // we've already visited this geom once
     } else {
       // check for null or empty attribute and replace with default geometry name
-      val attribute = Option(prop.name).filterNot(_.isEmpty).getOrElse(if (sft == null) null else sft.getGeomField)
+      val attribute = Option(prop.name).filterNot(_.isEmpty).orElse(Option(sft).map(_.getGeomField)).orNull
       // copy the geometry so we don't modify the original
-      val geomCopy = gf.createGeometry(geom)
+      val geomCopy = geom.getFactory.createGeometry(geom)
```
jnh5y: hehehe.... Does Geometry have a clone method for this kind of use?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastDWithin.scala
elahrvivaz: is that a geotools thing?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
elahrvivaz: hmm, it actually does have a clone() implementation - not sure why we didn't use that to begin with. this change was mainly just to preserve any precision models associated with the geom, probably irrelevant...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastDWithin.scala
elahrvivaz: hmm, that doesn't seem to be fast haha, it just uses the geodetic calculator

jnh5y: https://github.com/locationtech/geomesa/blob/master/geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeometricDistanceFunctions.scala

PR #1955, opened 2018-05-11
GEOMESA-2274 Supporting query reprojections

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geometry/Geometry.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geometry/Geometry.scala
@@ -11,9 +11,13 @@ package org.locationtech.geomesa.utils.geometry
 import com.vividsolutions.jts.geom.Polygon
 import org.geotools.geometry.jts.JTS
 
+@deprecated
 object Geometry {
+
+  @deprecated
   val noPolygon : Polygon  = null
 
+  @deprecated
   implicit class RichPolygon(self: Polygon) {
     def getSafeUnion(other: Polygon): Polygon = {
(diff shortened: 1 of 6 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geometry/Geometry.scala
```diff
@@ -11,9 +11,13 @@ package org.locationtech.geomesa.utils.geometry
 import com.vividsolutions.jts.geom.Polygon
 import org.geotools.geometry.jts.JTS
 
+@deprecated
 object Geometry {
+
+  @deprecated
   val noPolygon : Polygon  = null
 
+  @deprecated
```
jnh5y: Can you add some details about the deprecation?  Is there something preferred in its place?

jnh5y:
Do we have any ContentDataStores left?  If so, how does this interact with 'canReproject'?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geometry/Geometry.scala
elahrvivaz: no, i just noticed that we don't actually use it anywhere. do you think it's worth keeping around?

elahrvivaz:
we still use content data store in arrow-datastore, fsds-datastore, cq-engine-datastore, and stream-datastore, but none of those use the query planner.

jnh5y:
Ok, so since we aren't using the abstract ContentDataStore, then it is up to use to handle the reprojections?  (And hence this fixes that?)

elahrvivaz:
right

PR #1960, opened 2018-05-16
GEOMESA-2212 JSON list converter - handle nulls

elahrvivaz:


jnh5y:
Is there any situation where one would want to have a null in a List?

elahrvivaz:
mmm, good question - i'd say no?

jnh5y:
Lemme merge this as-is.  If we find that we need nulls in lists, then we can have a follow-on PR to sort things out.

PR #1963, opened 2018-05-16
GEOMESA-2133 JsonPath converter treats missing elements as null

elahrvivaz:


jnh5y:
Does this deserve any mention in the docs?  I can see it either way.

elahrvivaz:
i'm not sure we want to get down to that level of detail in the docs... since we don't currently mention it one way or the other, i think it's ok to leave it

PR #1969, opened 2018-05-31
GEOMESA-2291 CQEngine - handle filter functions, etc

jnh5y:
* Handles the cases around (in)equalities.

### DIFF
--- a/geomesa-memory/geomesa-cqengine/src/test/scala/org/locationtech/geomesa/memory/cqengine/utils/SampleFeatures.scala
+++ b/geomesa-memory/geomesa-cqengine/src/test/scala/org/locationtech/geomesa/memory/cqengine/utils/SampleFeatures.scala
@@ -266,6 +266,12 @@ object SampleFilters {
     "WhatLong * WhatDouble < 1.0",
     "WhatLong * WhatDouble >= 1.0",
     "WhatLong * WhatDouble <= 1.0",
-    "WhatLong * WhatDouble <> 1.0"
+    "WhatLong * WhatDouble <> 1.0",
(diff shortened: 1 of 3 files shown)


geomesa-memory/geomesa-cqengine/src/test/scala/org/locationtech/geomesa/memory/cqengine/utils/SampleFeatures.scala
```diff
@@ -259,4 +259,13 @@ object SampleFilters {
     "Who ILIKE 'ADD%'",
     "Who ILIKE '%DA%'"
   )
+
+  val functionPredicates: Seq[Filter] = Seq(
+    "strConcat(Who, What) = 'Addams1'",
```
elahrvivaz: will this work with multiple AND'd predicates?

elahrvivaz: can you add some tests for AND'd predicates of various types?

jnh5y:
I didn't check it in here... but if you want more tests (and some non-determinism to boot!), you can check out this commit: https://github.com/ccri/geomesa/commit/039f486526538f40c78a1203316499e77020bf02.

elahrvivaz:
@jnh5y do those extended tests pass? is the non-determinism an issue, or think it's ok to merge?

jnh5y:
@elahrvivaz the extended tests pass.  The non-determinism comes from the fact that I was generating a random subset of 3-combinations of all the existing filters....;)

This should be good to go.

PR #1973, opened 2018-06-11
geomesa convert function geometry() doesn't support geometry collection

gzqlovemxq:
I post here for there is no issues area in geomesa repo.

And I use the geomesa convert function 'geometry'.
And it is described as below:

Description: Parse a geometry from a WKT string or GeoJson.

Usage: geometry($0)

Example: Parsing WKT as a geometry

But, I found that it doesn't support geometrycollection type.

Does any one get the same problem?

Thank you for any help.


elahrvivaz:
We track our issues in JIRA here: https://geomesa.atlassian.net/browse/
I added a ticket to update the documentation, as the geometry function doesn't work with arbitrary geojson unless you are using a json converter. It should work with WKT though.
If you're seeing errors with WKT, or want further clarification, please send an email to the geomesa users' list, or post in our gitter channel: https://github.com/locationtech/geomesa#join-the-community
Thanks,

PR #1977, opened 2018-06-13
GEOMESA-2021,GEOMESA-339 Use InputStreams in converters

elahrvivaz:
* Deprecated convert.SimpleFeatureConverter[T], replaced with convert2.SimpleFeatureConverter
* Converter instances are wrapped appropriately so old and new API are both picked up
* Converters operate on input streams and not type bound instances (e.g. strings)
** Quoted newlines are supported in CSV parsing
** JSON no longer requires multi-line or single-line option
* Refactored expressions and changed parsing to parboiled
* AbstractConverters are based on pureconfig to allow conversion to/from config


build/cqs.tsv
```diff
@@ -10,6 +11,8 @@ com.fasterxml.jackson.core:jackson-annotations	2.6.5	compile
 com.fasterxml.jackson.core:jackson-core	2.6.5	compile
 com.fasterxml.jackson.core:jackson-databind	2.6.5	compile
 com.github.ben-manes.caffeine:caffeine	2.5.6	compile
+com.github.pureconfig:pureconfig_2.11	0.9.1	compile
+com.github.pureconfig:pureconfig-macros_2.11	0.9.1	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16743

build/cqs.tsv
```diff
@@ -2,6 +2,7 @@ asm:asm	3.1	compile
 com.adobe.xmp:xmpcore	5.1.2	compile
 com.beust:jcommander	1.48	compile
 com.carrotsearch:hppc	0.7.2	compile
+com.chuusai:shapeless_2.11	2.3.2	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16766

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
```diff
@@ -168,7 +182,9 @@ abstract class AbstractSimpleFeatureConverterFactory[I] extends SimpleFeatureCon
       } else {
         Seq("has-geo", "has-dtg")
       }
-    ValidatorLoader.createValidator(validators, sft)
+    val validator = SimpleFeatureValidator(validators)
```
jahhulbert-ccri: wondering if we should enable the z3 validator by default for converter2...

jahhulbert-ccri: or at least perhaps we look at the indexes and if z2 or z3 is present we enable z-index validator followon ticket if you think its a good idea?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -20,154 +21,155 @@ trait SimpleFeatureValidator {
...
+      } else if(geom == null) {
+        lastErr = "Null geometry attribute"
+        false
+      } else if (date.before(minDate)) {
+        lastErr = s"Date is before Z3 Min Date ($minDate)"
+        false
+      } else if (date.after(maxDate)) {
+        lastErr = s"Date is after Z3 Max Date ($maxDate)"
+        false
+      } else  if (!wholeWorldEnvelope.contains(geom.getEnvelopeInternal)) {
```
jahhulbert-ccri: picky - extra space

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -20,154 +21,155 @@ trait SimpleFeatureValidator {
...
-
-      (sf: SimpleFeature) => dtgFn(sf) && geomFn(sf)
+    override def validate(sf: SimpleFeature): Boolean = {
+      lastErr = null
+      val date = sf.getAttribute(dtgIndex).asInstanceOf[Date]
+      val geom = sf.getAttribute(geomIndex).asInstanceOf[Geometry]
+      if (date == null) {
+        lastErr = "Null date attribute"
+        false
+      } else if(geom == null) {
```
jahhulbert-ccri: picky add space :)

elahrvivaz: man, i had the right number of spaces, just in the wrong place ;)

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverterFactory.scala
elahrvivaz: good idea. ticket created here: https://geomesa.atlassian.net/browse/GEOMESA-2318

jahhulbert-ccri:
working on going through this again today
```
[ERROR] Failures: 
[ERROR]   KafkaDataStoreTest.KafkaDataStore should::write/update/read/delete features There is 1 failure
Stream(LazyImmutableSimpleFeature:jo:jones|20|Mon Jan 02 00:00:00 UTC 2017|POINT (-10 -10), LazyImmutableSimpleFeature:sm:smith|30|Sun Jan 01 00:00:00 UTC 2017|POINT (0 0))
  is missing: ScalaSimpleFeature:sm:smith2|32|Sun Jan 01 00:00:02 UTC 2017|POINT (2 2)
  must not contain: LazyImmutableSimpleFeature:sm:smith|30|Sun Jan 01 00:00:00 UTC 2017|POINT (0 0)
```

elahrvivaz:
that's an intermittent failure that has started cropping up in master :/
i kicked the build, let's see if it works this time

jnh5y:
@jahhulbert-ccri how's this one looking?

docs/user/convert/json.rst
```diff
@@ -9,12 +9,10 @@ definition.
 Configuration
 -------------
 
-The JSON converter supports parsing files in single-line mode or in multi-line mode. In single-line mode, each line
-of an input file should be a valid JSON document; in multi-line mode, the entire input file should be a single valid
-JSON document. In order to support JSON path expressions, each JSON document is fully parsed into memory.
-For large documents, this may take considerable time and memory. Thus, it is usually better to use single-line mode
-when possible. Line mode may be specified by ``options.line-mode = "single"`` or ``options.line-mode = "multi"``
-in your converter definition. If nothing is specified, single-line mode is used.
```
jnh5y: Ok, is there still a single-line and multi-line option here?

geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverter.scala
```diff
@@ -0,0 +1,68 @@
...
+
+      override def close(): Unit = {}
+    }
+  }
+}
+
+object AvroConverter {
+
+  case class AvroConfig(`type`: String,
+                        schema: Either[String, String], // left indicates string to parse, right indicates resource
```
jnh5y: Hmm....  I'd be tempted to suggest having a derpy little class for StringToParse and ResourceToRead, and then making a SchemaGettingClass which wraps up the Either....

docs/user/convert/json.rst
elahrvivaz: not for json, since we are reading it as an input stream we can just parse out the elements as they come, regardless of line breaks

geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverterFactory.scala
```diff
@@ -0,0 +1,69 @@
...
+  object AvroConfigConvert extends ConverterConfigConvert[AvroConfig] with OptionConvert {
+
+    override protected def decodeConfig(cur: ConfigObjectCursor,
+                                        `type`: String,
+                                        idField: Option[Expression],
+                                        caches: Map[String, Config],
+                                        userData: Map[String, Expression]): Either[ConfigReaderFailures, AvroConfig] = {
+
+      def schemaOrFile(schema: Option[String],
+                       schemaFile: Option[String]): Either[ConfigReaderFailures, Either[String, String]] = {
```
jnh5y: Hmm...  is it possible to use Try instead of having Either[ConfigReaderFailures, 'Success']?

geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.geotools.filter.expression.PropertyAccessorFactory
```diff
@@ -1 +1 @@
-org.locationtech.geomesa.convert.cql.ArrayPropertyAccessorFactory
+org.locationtech.geomesa.convert2.transforms.CqlFunctionFactory$ArrayPropertyAccessorFactory
```
jnh5y: Are users going to have problems finding the right jars to include?  (I.e., is there a new convert2 jar that folks will need to be educated about?)

geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverter.scala
elahrvivaz: sure, would be clearer

geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverterFactory.scala
elahrvivaz: pureconfig uses Eithers for the parsing, in order to chain failures back up - the left value generally stores more info that just an exception

geomesa-convert/geomesa-convert-common/src/main/resources/META-INF/services/org.geotools.filter.expression.PropertyAccessorFactory
elahrvivaz: no, both packages are in the same geomesa-convert jar (java 9 modules? what?)

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/SimpleFeatureConverter.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  def infer(is: () => InputStream, sft: Option[SimpleFeatureType]): Option[(SimpleFeatureType, Config)] = {
+    factories.foldLeft[Option[(SimpleFeatureType, Config)]](None) { (res, f) =>
+      res.orElse(WithClose(is())(in => f.infer(in, sft)))
+    }
+  }
+
+  // TODO close converter?
```
jnh5y: Ticket for this?

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
```diff
@@ -0,0 +1,71 @@
...
+          geoFac.createLineString(toCoordSeq(el.getAsJsonObject.get(CoordsPath)))
+        case "polygon" =>
+          toPolygon(el.getAsJsonObject.get(CoordsPath))
+        case "multipoint" =>
+          geoFac.createMultiPoint(toCoordSeq(el.getAsJsonObject.get(CoordsPath)))
+        case "multilinestring" =>
+          val coords = el.getAsJsonObject.get(CoordsPath).getAsJsonArray
+              .iterator.asScala.map(c => geoFac.createLineString(toCoordSeq(c))).toArray
+          geoFac.createMultiLineString(coords)
+        case "multipolygon" =>
```
jnh5y: Can GeoJson have GeometryCollections?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/SimpleFeatureConverter.scala
elahrvivaz: probably not going to be able to fix this due to the API differences - i'll update the comment with a better explanation

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
elahrvivaz: indeed, it appears so: https://tools.ietf.org/html/rfc7946#page-9

elahrvivaz: i'll add a follow on ticket

jnh5y:
A few questions...  1.  Do the jars that an end user would use change?

jnh5y:
2.  Can you point to where and how we are using Shapeless?

jnh5y:
3.  Do we have any options around the convert -> convert2 move?  Can you say more about that?

elahrvivaz:
1. jars do not change
2. shapeless is used by pureconfig for auto-converting type classes. we don't use it directly
3. I wanted to re-use class names, so created a new package to hold the new traits. The old ones need to stay unchanged (and deprecated) for semver. I think for 3.0 we should move everything back into the main convert package

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/SimpleFeatureConverter.scala
elahrvivaz: added a logger warning

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
elahrvivaz: https://geomesa.atlassian.net/browse/GEOMESA-2325

geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverter.scala
elahrvivaz: updated

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -20,154 +21,155 @@ trait SimpleFeatureValidator {
...
+        false
+      } else if (!wholeWorldEnvelope.contains(geom.getEnvelopeInternal)) {
+        lastErr = s"Geometry exceeds world bounds ($wholeWorldEnvelope)"
+        false
+      } else {
+        true
+      }
     }
+
+    override def lastError: String = lastErr
```
cne1x: [interesting] This line kind of freaks me out, even though there's nothing wrong with it.  `lastErr` is a mutable variable, but returning it from a method does not (evidently) pass the reference, because changing the value after you call this method does not change the value inside the class instance.  Weird.  Cool.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/AbstractConverter.scala
```diff
@@ -0,0 +1,291 @@
...
+import org.locationtech.geomesa.convert2.transforms.Expression
+import org.locationtech.geomesa.features.ScalaSimpleFeature
+import org.locationtech.geomesa.utils.collection.CloseableIterator
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+
+import scala.collection.mutable.{ArrayBuffer, ListBuffer}
+import scala.util.control.NonFatal
+
+/**
+  * Abstract converter implementation. Typically paired with an AbstractConverterFactory. If so, needs to have
```
cne1x: :+1:

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/AbstractConverter.scala
```diff
@@ -0,0 +1,291 @@
...
+  private val requiredFields: Array[Field] =
+    AbstractConverter.requiredFields(targetSft, fields, config.userData.values.toSeq ++ config.idField.toSeq)
+
+  private val requiredFieldsCount: Int = requiredFields.length
+
+  private val requiredFieldsIndices: Array[Int] = requiredFields.map(f => targetSft.indexOf(f.name))
+
+  /**
+    * Read values for simple features out of the input stream. This should be lazily evaluated,
+    * so that any exceptions occur in the call to `hasNext`. If there is any sense of 'lines',
```
cne1x: Why do we want exceptions on `hasNext`?  Doesn't it make more sense to throw them on `next`, even if the feature extraction is lazy?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/AbstractConverter.scala
```diff
@@ -0,0 +1,291 @@
...
+  protected def read(is: InputStream, ec: EvaluationContext): CloseableIterator[Array[Any]]
+
+  override def createEvaluationContext(globalParams: Map[String, Any],
+                                       caches: Map[String, EnrichmentCache],
+                                       counter: Counter): EvaluationContext = {
+    val globalKeys = globalParams.keys.toSeq
+    val names = requiredFields.map(_.name) ++ globalKeys
+    val values = Array.ofDim[Any](names.length)
+    // note, globalKeys are maintained even through EvaluationContext.clear()
+    globalKeys.zipWithIndex.foreach { case (k, i) => values(requiredFields.length + i) = globalParams(k) }
```
cne1x: Not that it matters much, but what about:

```scala
val (globalKeys, values) = globalParams.toArray.foldLeft((Array[Any](), Array[Any]()))((acc,t) => (acc._1 :+ t._1, acc._2 :+ t._2))
```

This yields two variables of type `Array[Any]` without the trouble of `zipWithIndex`.

geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverter.scala
```diff
@@ -0,0 +1,73 @@
...
+  private val schema = config.schema match {
+    case SchemaString(s) => new Parser().parse(s)
+    case SchemaFile(s)   => new Parser().parse(getClass.getResourceAsStream(s))
+  }
+
+  private val reader = new GenericDatumReader[GenericRecord](schema)
+
+  private var decoder: BinaryDecoder = _
+
+  override protected def read(is: InputStream, ec: EvaluationContext): CloseableIterator[Array[Any]] = {
```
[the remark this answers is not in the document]
elahrvivaz: i decided to pull the anonymous class out into the companion object, so there are 2 lines now :)

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureConverters.scala
```diff
@@ -8,41 +8,99 @@
...
+  import scala.collection.JavaConverters._
+
+  private val factories = ServiceLoader.load(classOf[convert2.SimpleFeatureConverterFactory]).asScala.toList
+
+  private val factoriesV1 = ServiceLoader.load(classOf[SimpleFeatureConverterFactory[_]]).asScala.toList
 
+  logger.debug(s"Found ${factories.size + factoriesV1.size} factories: " +
+      (factories ++ factoriesV1).map(_.getClass.getName).mkString(", "))
+
+  @deprecated("org.locationtech.geomesa.convert2.SimpleFeatureConverter.apply")
```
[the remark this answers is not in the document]
elahrvivaz: i was putting the replacement class in the deprecation warning, but that is obviously confusing :)
I'll update the deprecated messages with a more complete description

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -20,154 +21,155 @@ trait SimpleFeatureValidator {
   def lastError: String
 }
 
-object ValidatorLoader {
-  def createValidator(names: Seq[String], sft: SimpleFeatureType): SimpleFeatureValidator = {
+object SimpleFeatureValidator {
+
```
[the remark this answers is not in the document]
elahrvivaz: i don't know if we have a hard and fast rule... i tend to do both, depending on the contents of the object. If there are methods or classes, I put a space, if it is just constants I usually don't:
```
object Foo {
  val bar = ""
  val baz = ""
}
```

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -20,154 +21,155 @@ trait SimpleFeatureValidator {
...
-      if (sft.getGeometryDescriptor != null) {
-        (sf: SimpleFeature) =>
-          val res = sf.getDefaultGeometry != null
-          if (!res) {
-            lastErr = "Null geom attribute"
-          }
-          res
-      } else {
-        (_) => true
+      i = Option(sft.getGeomField).map(sft.indexOf).getOrElse(-1)
```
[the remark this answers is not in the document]
elahrvivaz: i think we're just using the normal java indexOf convention, where -1 means not found

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -20,154 +21,155 @@ trait SimpleFeatureValidator {
...
+    private var minDate: Date = _
+    private var maxDate: Date = _
+    private var lastErr: String = _
+
+    override def name: String = ZIndexValidator.name
+
+    override def init(sft: SimpleFeatureType): Unit = {
+      lastErr = null
+      dtgIndex = sft.getDtgIndex.getOrElse(-1)
+      geomIndex = Option(sft.getGeomField).map(sft.indexOf).getOrElse(-1)
```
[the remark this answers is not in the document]
elahrvivaz: i updated `Conversions.RichSimpleFeatureType.getGeomIndex` to return -1 instead of throwing an NPE if geom doesn't exist

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
elahrvivaz: yeah, I believe that strings and primitives are always passed by value, while other objects are passed by reference.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/AbstractConverter.scala
elahrvivaz: the main reason is that if there is an error when fetching the next value, then we might hit the end of the iterator. So if we return true from `hasNext`, then catch, handle and swallow an error in `next`, we don't have anything to return and we've broken the iterator contract.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/AbstractConverter.scala
elahrvivaz: i replaced `zipWithIndex.foreach` with `org.locationtech.geomesa.utils.conversions.ScalaImplicits.RichTraversableOnce.foreachIndex`, which uses a while loop for optimal efficiency

PR #1978, opened 2018-06-14
GEOMESA-2303 Infer schema and converter from delimited text inputs

elahrvivaz:
* Ingest command will try available converters to determine if one matches
* TypeInference class determines rolled up types for attributes
* Added additional date format functions for easier date inference
* Removes ability to auto-ingest Avro
* Added System.console abstraction to Prompt for unit testing
* Try it out: `./geomesa ingest -c example ../examples/ingest/csv/example.csv`


elahrvivaz:
note: stacks on top of #1977

elahrvivaz:
I'm going to re-add auto Avro ingest in a subsequent PR

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/TypeInference.scala
```diff
@@ -0,0 +1,413 @@
...
+        } else if (longitudeNames.contains(name)) {
+          lon = t.name
+        }
+        // stop iterating if we find them both
+        lat != null && lon != null
+      }
+
+      if (lat == null || lon == null) {
+        // as a fallback, check for 2 consecutive floats or doubles - assume longitude first
+        // note that this is pretty brittle, but hopefully better than nothing
```
jnh5y: Since an inferencer does have to make a bunch of assumptions, how does it let the user know what it picked?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/TypeInference.scala
```diff
@@ -0,0 +1,413 @@
...
+          lat = t.name
+        } else if (longitudeNames.contains(name)) {
+          lon = t.name
+        }
+        // stop iterating if we find them both
+        lat != null && lon != null
+      }
+
+      if (lat == null || lon == null) {
+        // as a fallback, check for 2 consecutive floats or doubles - assume longitude first
```
jnh5y: Also, around these assumptions, is there a check for the size of the values?  A -100 in one column might provide evidence that something is not a latitude...

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/DateFunctionFactory.scala
```diff
@@ -59,13 +59,34 @@ class DateFunctionFactory extends TransformerFunctionFactory {
           .withZone(ZoneOffset.UTC)
   }
 
+  // TODO even though it is aliased to 'isodate', it doesn't correspond with 'DateTimeFormatter.ISO_DATE',
```
jnh5y: Ticket?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/DateFunctionFactory.scala
```diff
@@ -120,8 +141,6 @@ class DateFunctionFactory extends TransformerFunctionFactory {
             .withZone(ZoneOffset.UTC)
     }
 
-  // TODO add more default dates for type inference
```
jnh5y: Did you Pokemon this and get 'em all?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/TypeInference.scala
elahrvivaz: yeah, that's a good idea - i don't think there are checks like that yet

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
```diff
@@ -0,0 +1,113 @@
...
+      foreach(types.permutations)(t => TypeInference.infer(t).map(_.typed) mustEqual Seq(GEOMETRY))
+    }
+    "merge up null values" in {
+      val values = Seq("a", 1, 1L, 1f, 1d, true, new Date(), Seq[Byte](0), uuid, point, lineString,
+        polygon, multiPoint, multiLineString, multiPolygon, geometryCollection)
+      foreach(values) { value =>
+        TypeInference.infer(Seq(Seq(value), Seq(null))) mustEqual TypeInference.infer(Seq(Seq(value)))
+      }
+    }
+    "create points from lat/lon pairs" in {
```
jnh5y: Are your pairs here lat/lon or lon/lat?  If we assume X first, then the comment might need to be changed.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/TypeInference.scala
elahrvivaz: in the normal use case, it gets printed out by the command line ingest so that the user can confirm or modify it

geomesa-convert/geomesa-convert-text/src/main/scala/org/locationtech/geomesa/convert/text/DelimitedTextConverterFactory.scala
```diff
@@ -31,10 +38,72 @@ class DelimitedTextConverterFactory
...
+    import org.locationtech.geomesa.utils.conversions.ScalaImplicits.{RichIterator, RichTraversableLike}
+
+    import scala.collection.JavaConverters._
+
+    val sampleSize = DelimitedTextConverterFactory.InferSampleSize.toInt.getOrElse {
+      // shouldn't ever happen since the default is a valid int
+      throw new IllegalStateException("Could not determine sample size from system property")
+    }
+    val lines = IOUtils.lineIterator(is, StandardCharsets.UTF_8.displayName).asScala.take(sampleSize).toSeq
+    // we assume the first line is a header
```
jnh5y: What formats do we allow the header to take?

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseBulkIngestCommand.scala
```diff
@@ -56,9 +56,6 @@ class HBaseBulkIngestCommand extends HBaseIngestCommand {
     }
   }
 
-  override protected def createAutoIngest(): Runnable =
```
jnh5y: Auto Ingest works for HBase now?

geomesa-tools/examples/ingest/csv/example.csv
```diff
@@ -1,4 +1,4 @@
-ID,Name,Age,LastSeen,Friends,Lon,Lat,Vis
-23623,Harry,20,2015-05-06,"Will, Mark, Suzan","patronus->10,expelliarmus->9",-100.236523,23,user
+ID,Name,Age,LastSeen,Friends,Skills,Lon,Lat,Vis
+23623,Harry,20,2015-05-06,"Will, Mark, Suzan","patronus->10,expelliarmus->9",-100.2365,23,user
```
jnh5y: The inferencer can figure out maps?!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
```diff
@@ -118,7 +118,6 @@ abstract class AbstractIngest(val dsParams: Map[String, String],
           try {
             val (sft, features) = converter.convert(is)
             if (features.hasNext) {
-              ds.createSchema(sft)
```
jnh5y: Can this actually go?  Seems like it'd be needed to create the schema?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaSystemProperties.scala
```diff
@@ -45,8 +45,17 @@ object GeoMesaSystemProperties extends LazyLogging {
...
+        case Success(v) => Some(v)
+        case Failure(_) =>
+          logger.warn(s"Invalid Int for property $property: $value")
+          Option(default).map(java.lang.Integer.parseInt)
+      }
+    }
+
     def toBoolean: Option[Boolean] = option.flatMap { value =>
-      Try(value.toBoolean) match {
+      Try(java.lang.Boolean.parseBoolean(value)) match {
```
jnh5y: For my education, what's the difference between these two calls?

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/DataStoreRegistration.scala
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+import org.geotools.data.DataAccessFactory.Param
+import org.geotools.data.{DataStore, DataStoreFactorySpi}
+import org.geotools.factory.FactoryIteratorProvider
+
+/**
+  * For registering a data store instance that can be loaded via SPI
+  */
+object DataStoreRegistration {
```
jnh5y: Ok, I'm reading quickly and am stumped.  Why is this useful?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -31,58 +34,77 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
...
+      // try to load the sft, first check for an existing schema, then load from the params/environment
+      var sft: SimpleFeatureType =
+        Option(params.featureName).flatMap(n => Try(withDataStore(_.getSchema(n))).filter(_ != null).toOption)
+          .orElse(Option(params.spec).flatMap(s => Option(CLArgResolver.getSft(s, params.featureName))))
+          .orNull
+
+      var converter: Config = Option(params.config).map(CLArgResolver.getConfig).orNull
+
+      if (converter == null && params.files.nonEmpty) {
+        // if there is no converter passed in, try to infer the schema from the input files themselves
```
jnh5y: What happens if a user has a bunch of files of the same format, ingests some and then ingests some more?  

Does the inferencing re-run?  Or is the SFT/converter definition saved after the first run?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -31,58 +34,77 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] {
...
+            name
+          }
+          sft = SimpleFeatureTypes.renameSft(inferred._1, typeName)
+          Command.user.info(s"Inferred schema: $typeName identified ${SimpleFeatureTypes.encodeType(sft)}")
+        }
+        Command.user.info("Inferred converter:\n" +
+            inferred._2.root().render(ConfigRenderOptions.concise().setFormatted(true)))
+        if (Prompt.confirm("Use inferred converter (y/n)? ")) {
+          Command.user.info("For consistency, please use this converter definition for any future ingest")
+          converter = inferred._2
```
jnh5y: Does the user have a chance to save the converter?  Or would they manually have to do so?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/DateFunctionFactory.scala
elahrvivaz: https://geomesa.atlassian.net/browse/GEOMESA-2326

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/DateFunctionFactory.scala
elahrvivaz: yeah, I added all the predefined ones in the java DateTimeFormatter class

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
elahrvivaz: fixed

geomesa-tools/examples/ingest/csv/example.csv
elahrvivaz: no haha, it just treats it as a string - but the header row in this file didn't match the column count, so the names were off when auto-ingested

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
elahrvivaz: at some point we starting calling createSchema in the 'preIngest' method, so this isn't needed anymore

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conf/GeoMesaSystemProperties.scala
elahrvivaz: I think i was mainly trying to match the syntax of the call to `Integer.parseInt` above... looking into it, there is actually a difference, in that `parseBoolean` will always return a Boolean, while `.toBoolean` will throw an exception if it's not either (case-insensitive) 'true' or 'false', which actually seems better. I'm going to switch it back

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/DataStoreRegistration.scala
elahrvivaz: I added a better comment - it's a test class, used to pass a MemoryDataStore instance through SPI loading. Otherwise the data store is loaded, but any features written are lost b/c they're only kept in memory, so you can't verify anything actually happened.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: the command logs out to the user that you need to save the converter definition if you want to ingest with it again. you can re-run without doing that, but the converter might not end up the same...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: they would have to manually copy it out. i'll add a ticket to write it out for them

elahrvivaz: https://geomesa.atlassian.net/browse/GEOMESA-2327

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/TypeInference.scala
elahrvivaz: opened a ticket https://geomesa.atlassian.net/browse/GEOMESA-2330

PR #1980, opened 2018-06-18
GEOMESA-2304 Infer schema and converter from Avro files

elahrvivaz:
* Avro converter can now handle files with embedded schemas
* avroPath expression will return byte arrays instead of byte buffers
* Schema for Avro user-data fixed to reflect actual data layout


elahrvivaz:
note: stacks on top of #1978

geomesa-convert/geomesa-convert-avro/src/test/scala/org/locationtech/geomesa/convert/avro/AvroSimpleFeatureConverterTest.scala
```diff
@@ -1,95 +0,0 @@
...
-
-import com.typesafe.config.ConfigFactory
-import org.junit.runner.RunWith
-import org.locationtech.geomesa.convert2.SimpleFeatureConverter
-import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
-import org.specs2.mutable.Specification
-import org.specs2.runner.JUnitRunner
-
-@RunWith(classOf[JUnitRunner])
-class AvroSimpleFeatureConverterTest extends Specification with AvroUtils {
```
jnh5y: Le sigh, if you figure out how to get your IDE to use git mv, that'd be awesome:).

elahrvivaz: I looked into git mv, but ultimately it doesn't make a difference. it appears the only way to preserve a file is to make it two separate commits.

PR #1981, opened 2018-06-19
GEOMESA-2270,GEOMESA-2302 Infer schema and converter from geojson files

elahrvivaz:
* Supports geojson parsing of geometry collections


elahrvivaz:
note: stacks on top of #1980

docs/user/cli/ingest.rst
```diff
@@ -49,29 +49,27 @@ The ``--converter`` argument may be any of the following:
 * A converter configuration string
 * The name of a file containing a converter configuration
 
-See :ref:`cli_converter_conf` for more details on specifying the converter.
-
-If a converter is not specified, GeoMesa will attempt to extract a schema from the input files. Currently
-this only works in a very limited fashion - the input must be a Shapefile, self describing Avro, or
-delimited text (TSV, CSV) with a header row that defines the attribute name and type of each column. For example::
```
jnh5y: If one still has a file with a header, what happens under inferencing?  Does it still work as before?

docs/user/cli/ingest.rst
```diff
@@ -90,7 +88,8 @@ and Microsoft's Azure file systems are supported with a few configuration change
 :doc:`/user/cli/filesystems` for details.
 
 Instead of specifying files, input data may be piped directly to the ingest command using `stdin` shell redirection.
-Note that this will only work in local mode, and will only use a single thread for ingestion. Progress indicators
-may not be entirely accurate, as the total size isn't known up front. For example::
+Note that this will only work in local mode, and will only use a single thread for ingestion. Schema inference is
+disabled in this case, and progress indicators may not be entirely accurate, as the total size isn't known up front.
+For example::
```
jnh5y: Is it worth being pedantic and suggesting that users with lots of files can 1) do schema inference on a sample file, 2) persist the config, and 3) run their larger ingest?

docs/user/convert/function_usage.rst
```diff
@@ -175,20 +175,47 @@ Usage: ``dateTime($1)``
 
 Example: ``dateTime('2015-01-01T00:00:00.000Z')``
 
-basicDate
-^^^^^^^^^
+basicIsoDate
+^^^^^^^^^^^^
+
+Description: A date format for ``yyyyMMdd``, equivalent to java.time.format.DateTimeFormatter.BASIC_ISO_DATE.
```
jnh5y: I wonder if a table of function name, format (yyyyMMdd), example would be helpful....

docs/user/cli/ingest.rst
elahrvivaz: no, it won't use the types from the header - but you can pull out the header and pass it as the sft spec and it will use that

docs/user/cli/ingest.rst
elahrvivaz: it's always worth being pedantic ;)

docs/user/convert/function_usage.rst
elahrvivaz: i'll add one

jnh5y: You are my hero!

PR #1986, opened 2018-06-25
GEOMESA-2308 Incremental Accumulo compaction command

elahrvivaz:
* Fixes Z3 UUID byte merging sign errors

### DIFF
--- a/docs/user/accumulo/commandline.rst
+++ b/docs/user/accumulo/commandline.rst
@@ -91,6 +91,31 @@ Argument                 Description
 
 For a description of index coverage, see :ref:`accumulo_attribute_indices`.
 
+.. _compact_command:
+
+``compact``
+^^^^^^^^^^^
+
+Incrementally compact tables for a given feature type. This can be used with data age-off to ensure that expired
+rows are physically deleted from disk - see :ref:`ageoff_accumulo` for more information. It can also be used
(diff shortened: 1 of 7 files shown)


docs/user/accumulo/commandline.rst
```diff
@@ -91,6 +91,31 @@ Argument                 Description
...
 
+.. _compact_command:
+
+``compact``
+^^^^^^^^^^^
+
+Incrementally compact tables for a given feature type. This can be used with data age-off to ensure that expired
+rows are physically deleted from disk - see :ref:`ageoff_accumulo` for more information. It can also be used
+to speed up queries by removing entries that are duplicated or marked for deletion. This may be useful for a
+static data set, which will not be automatically compacted by Accumulo once the size stops growing.
```
jnh5y: It seems that this command is useful for three or so different use cases.  Would it make sense to add details about the use cases: 

E.g.  If you were are applying age-off, call with "-duration 90 days".  If you are trying to accelerate a 'now' static table, call without --from or --duration, etc.  

Also, if there ever a time when one would prefer calling the Accumulo compact command?

elahrvivaz: I added some clarification on the difference between this and the accumulo command, and some explanations of the various scenarios you might want to use it.

PR #1993, opened 2018-07-09
GEOMESA-2315 Map JTS to Shapely types for use with PySpark

tkunicki:
The current use of JTS JVM objects in PySpark via the Py4J bridge is problematic as the Py4J bridge is only available in the PySpark driver (and not executors). A more complete solution would be to map JTS (JVM) <-> Shapely (Python).

### DIFF
--- a/geomesa-spark/geomesa-spark-core/src/main/scala/org/apache/spark/geomesa/api/python/GeoMesaSeDerUtil.scala
+++ b/geomesa-spark/geomesa-spark-core/src/main/scala/org/apache/spark/geomesa/api/python/GeoMesaSeDerUtil.scala
@@ -1,0 +1,51 @@
+/***********************************************************************
+ * Copyright (c) 2013-2018 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.apache.spark.geomesa.api.python
+
+import java.io.OutputStream
+
+import com.typesafe.scalalogging.LazyLogging
+import com.vividsolutions.jts.geom.Geometry
+import net.razorvine.pickle.{IObjectPickler, Pickler}
+import org.apache.spark.api.java.JavaRDD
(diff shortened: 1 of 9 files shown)


tkunicki:
Unit tests are never a not a good thing, will see how to hook python tests into our maven build...

@jbouffard we're hooking into the Spark JVM/Python serialization mechanism, a private API.  In looking at JVM <-> python spark interaction it wasn't clear that arrow was an option for this serialization path (ask me how many spark serialization paths i've come across...).  I _believe_ the class you referenced is for python <-> python serialization (and not python <-> JVM).  Also, the pickler/pyrolite path is JVM <-> python RDD-only (included for completeness, essentially and edge case for luddites like me) and a one time hit to pull the object into Python from the JVM (after that you can specify arrow for python <-> python serialization if you love your RDD)

In the context of DataFrames, operations are JVM only w/ serialization to python occurring when an object is requested in the python driver.  The DataFrame serialization paths don't touch the pickler/pyrolite serialization (referenced the JSON 'Non' errors observed ealier)

So...  the UDT serialization doesn't use the pickler, it's there for RDDs

jbouffard:
@tkunicki That class and all other serializes are used by the wrapped `jrdd` in the Python `RDD` to deserialize its contents into Python https://github.com/apache/spark/blob/master/python/pyspark/rdd.py#L211 We do something similar currently with our `ProtoBufSerializer` when bring the objects over to Python https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/layer.py#L1042

tkunicki:
I can poke, just feel like this is an optimization for non-standard execution path.  e.g. make it work, then make it better...

jbouffard:
@tkunicki The benefit to using Arrow would be that you can then take advantage of other Python libraries since going from Scala to Python and back won't be such an intensive process. However, if that's not a priority for you guys right now, then I'd agree that it's a bit of an unneeded optimization.

metasim:
While I would like to see how to eventually use Apache Arrow for this sort of thing, I'm less worried about the serializer in this case because WKB blobs are less intensive than, say, `Tile`s in GPS. It's basically already been serialized to a byte array, and barring compression, I don't see what other serialization schemes bring to the table.

tkunicki:
added unit tests and maven hooks to execute

PR #1998, opened 2018-07-17
Premade converter for Marine Cadastre AIS

jrs53:
Converter (reference.conf) and associated documentation.


jrs53:
I *think* I've done this - many thanks for the help @elahrvivaz

geomesa-tools/conf/sfts/marinecadastre-ais/README.md
```diff
@@ -0,0 +1,49 @@
+# Marine Cadastre AIS
```
aheyne: Would prefer a link to the documentation but don't let this block.

jrs53: Yeah, I was lazy and just copied the existing converter docs

PR #2000, opened 2018-07-19
GEOMESA-2330 Type inference - verify valid lat/lons when infering poi…

elahrvivaz:
…nt types

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
+++ b/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
@@ -47,12 +47,14 @@ class TypeInferenceTest extends Specification with LazyLogging {
 
   "TypeInference" should {
     "infer simple types" in {
-      val types = TypeInference.infer(Seq(Seq("a", 1, 1L, 1f, 1d, true)))
+      // note: don't put any valid lat/lon pairs next to each other or it will create a geometry type
+      val types = TypeInference.infer(Seq(Seq("a", 1, 200L, 1f, 200d, true)))
       types.map(_.typed) mustEqual Seq(STRING, INT, LONG, FLOAT, DOUBLE, BOOLEAN)
     }
     "infer simple types from strings" in {
-      val types = TypeInference.infer(Seq(Seq("a", "1", s"${Int.MaxValue.toLong + 1L}", "1.1", "1.00000001", "true")))
-      types.map(_.typed) mustEqual Seq(STRING, INT, LONG, FLOAT, DOUBLE, BOOLEAN)
+      // note: don't put any valid lat/lon pairs next to each other or it will create a geometry type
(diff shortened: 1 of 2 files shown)


geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
```diff
@@ -47,12 +47,14 @@ class TypeInferenceTest extends Specification with LazyLogging {
 
   "TypeInference" should {
     "infer simple types" in {
-      val types = TypeInference.infer(Seq(Seq("a", 1, 1L, 1f, 1d, true)))
+      // note: don't put any valid lat/lon pairs next to each other or it will create a geometry type
```
jnh5y: Uhhmmm...  so yeah, type inference on data with lots of 'smallish' numeric fields could go poorly?

elahrvivaz: not really, just for this test I didn't want to get any inferred geometries. if there are numeric fields, it will infer the first 2 as lon/lat for a geometry type (assuming no other geometry field exists). we don't really have any way to distinguish between fields at that point, so the user would have to modify the converter to get the right ones.

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
```diff
@@ -102,6 +104,10 @@ class TypeInferenceTest extends Specification with LazyLogging {
       TypeInference.infer(Seq(Seq(45f, 55f, "POINT (40 50)"))).map(_.typed) mustEqual Seq(FLOAT, FLOAT, POINT)
       TypeInference.infer(Seq(Seq(45d, 55d, "POINT (40 50)"))).map(_.typed) mustEqual Seq(DOUBLE, DOUBLE, POINT)
     }
+    "not create points if values are not valid lat/lons" in {
+      TypeInference.infer(Seq(Seq(145f, 155f, "foo"))).map(_.typed) mustEqual Seq(FLOAT, FLOAT, STRING)
+      TypeInference.infer(Seq(Seq(145d, 155d, "foo"))).map(_.typed) mustEqual Seq(DOUBLE, DOUBLE, STRING)
```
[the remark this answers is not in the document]
elahrvivaz: don't think so, i'll add one

geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/TypeInferenceTest.scala
elahrvivaz: also, if the fields are labelled (e.g from a csv header) as 'lat/lon' or 'latitude/longitude' we'll use those. I added a test to show that as well

PR #2004, opened 2018-07-23
GEOMESA-2328 Shapefile converter

elahrvivaz:
* Shapefile ingest is run through a converter
** Allows for renaming, transforms, functions, etc


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/ExpressionParser.scala
```diff
@@ -60,7 +60,7 @@ private [transforms] class ExpressionParser extends BasicParser {
   }
 
   private def column: Rule1[Expression] = rule("$col") {
-    "$" ~ int ~~> { i => if (i == 0) { WholeRecord } else { Column(i) } }
+    "$" ~ int ~~> { i => Column(i) }
```
jnh5y: Is $0 still the whole record?  Or does this change that?

elahrvivaz: still the whole record, just removed the case object as it's not really needed

jnh5y: Above in the documentation, $0 is the id-field?

elahrvivaz: for shp converters, $0 is the feature ID, for delimited text it's the whole record

elahrvivaz: that was the impetus to change it actually, since whole record is specific to delimited text

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseBulkIngestCommand.scala
```diff
@@ -55,9 +55,6 @@ class HBaseBulkIngestCommand extends HBaseIngestCommand {
         throw new NotImplementedError("Bulk ingest not implemented for local mode")
     }
   }
-
-  override protected def createShpIngest(): Runnable =
-    throw new NotImplementedError("Bulk ShpFile ingest not implemented")
```
jnh5y: With this PR, what happens when people try to ingest shapefiles with MR?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -37,85 +39,92 @@ trait IngestCommand[DS <: DataStore] extends DataStoreCommand[DS] with Interacti
...
+      throw new ParameterException("SimpleFeatureType name and/or specification argument is required")
+    } else if (converter == null) {
+      throw new ParameterException("Converter config argument is required")
+    }
 
-      createConverterIngest(sft, converter).run()
+    if (params.fmt == DataFormats.Shp) {
+      // shapefiles have to be ingested locally, as we need access to the related files
+      if (params.mode == RunModes.Distributed) {
+        Command.user.warn("Forcing run mode to local for shapefile ingestion")
```
jnh5y: Nice.

What are the options around running a distributed ingest with shapefiles?  Could we split up the list of files among workers?

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseBulkIngestCommand.scala
elahrvivaz: it will use local mode regardless of what is specified - if you use the bulk ingest command that will make it throw an IllegalArgumentException I believe. We haven't ever supported MR shapefile ingest, this doesn't change that

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/TextTools.scala
```diff
@@ -33,4 +35,30 @@ object TextTools {
       new String(Array.fill(length)(c))
     }
   }
+
+  /**
+    * Builds a natural word list, e.g. 'foo, bar, baz and blu'
```
jnh5y: Does this code use the Oxford comma or not?

Seem it would output 'foo, bar, baz, and blu'...;)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: I don't think we can realistically split up a shapefile among different workers. the other issue is that we'd have to zip up all the relevant files into a single file, as they have to be processed together.

geomesa-convert/geomesa-convert-shp/src/main/scala/org/locationtech/geomesa/convert/shp/ShapefileConverter.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+    * Iterator for reading a shapefile
+    *
+    * @param ds data store to read from
+    * @param array attributes array
+    * @param counter line counter
+    */
+  class ShapefileIterator(ds: ShapefileDataStore, array: Array[Any], counter: Counter)
```
jnh5y: This class doesn't use the fact that this is a ShapefileDataStore.  Do we happen to have a similar DataStore wrapper elsewhere?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/TextTools.scala
elahrvivaz: nope, no oxford comma here :)

geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/ShpIngestTest.scala
```diff
@@ -42,81 +43,84 @@ class ShpIngestTest extends Specification {
...
     "should support renaming the feature type" in {
-      withShapeFile { file =>
-        val args = baseArgs ++ Array("--feature-name", "changed", file.getAbsolutePath)
+      val args = baseArgs ++ Array("--feature-name", "changed", shpFile.getAbsolutePath)
 
-        val command = AccumuloRunner.parseCommand(args).asInstanceOf[AccumuloDataStoreCommand]
-        command.execute()
+      val command = AccumuloRunner.parseCommand(args).asInstanceOf[AccumuloIngestCommand]
+      command.setConsole(new AnyRef {
+        def readLine(): String = "y" // accept prompt to use inferred schema
```
jnh5y: Are users required to enter 'y'?  I.e., use of the geomesa ingest command around shapefiles is changed by this PR, right?

geomesa-convert/geomesa-convert-shp/src/main/scala/org/locationtech/geomesa/convert/shp/ShapefileConverter.scala
elahrvivaz: i don't think so... it does use the fact that it's a FileDataStore (`ds.getFeatureReader`). it's copying the attributes into an array with a +1 offset, which we don't do anywhere else that i can think of

geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/ShpIngestTest.scala
elahrvivaz: yeah, previously it would just ingest without prompting

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/ExpressionParser.scala
jnh5y: Hmm....  I'm slightly nervous about $0 meaning something different in the two different contexts.

We recommend md5sum($0) as a way to roll a unique ID, and that'd be different between the same data in Shapefile and GeoJson format...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
jnh5y: If all the files are on remote filesystems, I don't think we need to fuss with zipping up the files...

And yes, it'd be one or more shapefiles per task.  Splitting up a shapefile wouldn't make sense.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/text/TextTools.scala
jnh5y: Ah, I was reading da code too quickly.  

That said, I'm a fan of more commas...

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/ExpressionParser.scala
elahrvivaz: my reasoning was that $0 will still be unique, since it's the feature ID. it already means a few different things though - for example in json it's just the result of the json-path expression

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: the reason you'd have to zip it up is b/c you wouldn't get all the files into a single mapper if you don't... i guess we could create some kind of weird input format split that has the path to the rest of the files (which is pretty much what i did to hack shp into converters), but it doesn't really fit with the whole m/r paradigm

PR #2005, opened 2018-07-23
GEOMESA-2336 Spark - getPointN fails for negative values

elahrvivaz:


geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/GeometricAccessorFunctions.scala
```diff
@@ -59,9 +59,9 @@ object GeometricAccessorFunctions {
     geom match {
       case geom: LineString =>
         if (n < 0) {
-          geom.getPointN(n + geom.getLength.toInt)
+          geom.getPointN(n + geom.getNumPoints)
```
jnh5y: Unit test? 

Also, supposing the geometry has 10 points, what should happen if someone asks for -25?

elahrvivaz: ironically, there already is a test, but the length of the linestring is 5.236 and it contains 5 points so it still passed.
if someone asks for an invalid index they get an IndexOutOfBoundsException I assume...

elahrvivaz: i'm not really sure what the edge cases are supposed to be - i just identified the obvious bug

jnh5y: https://postgis.net/docs/ST_PointN.html

In version 2.3, negative index became available.  -1 is the last point in the linestring.

jnh5y: Since that's the behavior, I think there may need to be a little more handling here?

elahrvivaz: hmm, what handling?

jnh5y: There are 10 elements, then -15 is out of bounds and should return null, etc...

elahrvivaz: it says 'Returns NULL if there is no linestring in the geometry', does that mean if the geometry isn't a linestring it's null? It doesn't seem to specify what happens if `n` doesn't exist

jnh5y: Yup, lots and lots of nulls....  And it is 1-indexed to boot.  So yeah, if there are n points in the linestring, then there are 2*n values which are non-null.

PR #2012, opened 2018-07-25
GEOMESA-2345 Allow for auto-reprojection of shapefiles on ingest

jnh5y:

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
+++ b/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
@@ -425,8 +425,8 @@ class TransformersTest extends Specification {
         val transformed = trans.eval(Array("", geom))
         transformed must not(beNull)
         transformed.getClass mustEqual classOf[Point]
-        transformed.asInstanceOf[Point].getX must beCloseTo(15d, 0.001)
-        transformed.asInstanceOf[Point].getY must beCloseTo(10d, 0.001)
+        transformed.asInstanceOf[Point].getX must beCloseTo(10d, 0.001)
+        transformed.asInstanceOf[Point].getY must beCloseTo(15d, 0.001)
       }
 
       "handle identity functions" >> {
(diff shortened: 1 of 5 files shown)


geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/common/TransformersTest.scala
```diff
@@ -425,8 +425,8 @@ class TransformersTest extends Specification {
         val transformed = trans.eval(Array("", geom))
         transformed must not(beNull)
         transformed.getClass mustEqual classOf[Point]
-        transformed.asInstanceOf[Point].getX must beCloseTo(15d, 0.001)
-        transformed.asInstanceOf[Point].getY must beCloseTo(10d, 0.001)
+        transformed.asInstanceOf[Point].getX must beCloseTo(10d, 0.001)
+        transformed.asInstanceOf[Point].getY must beCloseTo(15d, 0.001)
```
elahrvivaz: hmm, why did this test in convert common change?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/package.scala
```diff
@@ -24,7 +24,7 @@ import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 package object geotools {
 
   // use the epsg jar if it's available (e.g. in geoserver), otherwise use the less-rich constant
-  val CRS_EPSG_4326 = try { CRS.decode("EPSG:4326") } catch { case t: Throwable => DefaultGeographicCRS.WGS84 }
+  val CRS_EPSG_4326 = try { CRS.decode("EPSG:4326", true) } catch { case t: Throwable => DefaultGeographicCRS.WGS84 }
```
elahrvivaz: I guess the test changed b/c of this change - is this going to cause any other change in behavior? We should probably document it in the upgrade guide

jnh5y: Ok, so this is *subtle*.  This line over here.... https://github.com/locationtech/geomesa/blob/master/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/MetadataBackedDataStore.scala#L43  can change the way that this executes.

In situations where we have a 'big-data datastore', the other line is likely to be called before memoize the variable CRS_EPSG_4326.  That means, that we get a 4326 which is X/Y.  Otherwise, by default, one gets Y/X.

elahrvivaz: sounds good to me

elahrvivaz:
looks good, think we could get a unit test? ;)

PR #2015, opened 2018-07-27
GEOMESA-2349 Fixing GeoJson ingest type inference

elahrvivaz:

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -62,7 +62,7 @@ $ gpg2 --keyserver hkp://pool.sks-keyservers.net --recv-keys CD24F317
 Then verify the file:
 
 ```bash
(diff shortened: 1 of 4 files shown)


build/README.md
```diff
@@ -62,7 +62,7 @@ $ gpg2 --keyserver hkp://pool.sks-keyservers.net --recv-keys CD24F317
 Then verify the file:
 
 ```bash
-$ gpg2 --verify geomesa-accumulo_2.11-2.0.0-bin.tar.gz.asc geomesa-accumulo_2.11-2.0.0-bin.tar.gz
+$ gpg2 --verify geomesa-accumulo_2.11-2.0.2-bin.tar.gz.asc geomesa-accumulo_2.11-2.0.2-bin.tar.gz
```
jnh5y: Derp...  Is there no way to filter these?

elahrvivaz: good question actually - i was thinking of .rst where code blocks don't do replacements, but that might not be the case here - let me check

PR #2016, opened 2018-07-27
GEOMESA-2348 GeoJson type inference - support 3D coordinates

elahrvivaz:

### DIFF
--- a/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
+++ b/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
@@ -78,7 +78,7 @@ trait GeoJsonParsing {
   def parseGeometry(el: JsonElement): Geometry = el match {
     case o: JsonObject    => parseGeometryObject(o)
     case o: JsonPrimitive => WKTUtils.read(o.getAsString)
-    case o: JsonNull      => null.asInstanceOf[Geometry]
+    case _: JsonNull      => null.asInstanceOf[Geometry]
     case _ => throw new IllegalArgumentException(s"Unknown geometry type: $el")
   }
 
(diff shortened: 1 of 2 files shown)


geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
```diff
@@ -119,8 +119,11 @@ trait GeoJsonParsing {
   }
 
   private def toPointCoords(el: JsonElement): Coordinate = {
-    val Seq(x, y) = el.getAsJsonArray.asScala.map(_.getAsDouble).toSeq
-    new Coordinate(x, y)
+    el.getAsJsonArray.asScala.map(_.getAsDouble).toSeq match {
+      case Seq(x, y)    => new Coordinate(x, y)
+      case Seq(x, y, z) => new Coordinate(x, y, z)
+      case s => throw new IllegalArgumentException(s"Invalid point - expected 2 or 3 values, got ${s.mkString(", ")}")
```
jnh5y: Lightning round question we need for later:  Does GeoJson support Geometries with M's?

jnh5y:
Merge conflicts:(

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/GeoJsonParsing.scala
elahrvivaz: according to the spec:
> A position is an array of numbers.  There MUST be two or more
   elements.  The first two elements are longitude and latitude, or
   easting and northing, precisely in that order and using decimal
   numbers.  Altitude or elevation MAY be included as an optional third
   element.

>   Implementations SHOULD NOT extend positions beyond three elements
   because the semantics of extra elements are unspecified and
   ambiguous.  Historically, some implementations have used a fourth
   element to carry a linear referencing measure (sometimes denoted as
   "M") or a numerical timestamp, but in most situations a parser will
   not be able to properly interpret these values.  The interpretation
   and meaning of additional elements is beyond the scope of this
   specification, and additional elements MAY be ignored by parsers.

elahrvivaz:
conflicts fixed

PR #2025, opened 2018-08-02
GEOMESA-2352 Simplify coprocessor install

elahrvivaz:
* Automatically load coprocessor path from `hbase.dynamic.jars.dir`
* Updated docs on registering coprocessors


jnh5y:
@aheyne @jahhulbert-ccri can you review this?

elahrvivaz:
I tested it with a psuedo-distributed local hbase 1.3.1 instance

elahrvivaz:
The current behavior where you specify the URL via ds param or system prop will still take precedence over the auto-lookup

PR #2033, opened 2018-08-08
GEOMESA-2318 Converters - default to validator based on indices used

elahrvivaz:
* Added SPI loading for extending validators

### DIFF
--- a/docs/user/convert/parsing_and_validation.rst
+++ b/docs/user/convert/parsing_and_validation.rst
@@ -14,25 +14,74 @@ At their core, converters transform input streams into SimpleFeatures. Validator
 of those SimpleFeatures before they are written to GeoMesa. For example, you may want to validate that there is a
 geometry field and that the geometry is valid.
 
-There are currently three validators available for use with GeoMesa converters:
+There are three validators provided for use with GeoMesa converters:
 
-* ``has-geo`` - Ensure that the SimpleFeature has a geometry and it is non-null
-* ``has-dtg`` - Ensure that the SimpleFeature has a date and it is non-null
-* ``z-index`` - Ensure the Geometry and Date are within the space/time bounds of GeoMesa Z-Index implementations
-  (i.e. Z2, Z3, XZ2, XZ3)
+* ``has-geo`` - Ensure that the SimpleFeature has a non-null geometry
+* ``has-dtg`` - Ensure that the SimpleFeature has a non-null date
+* ``index`` - Ensure that the SimpleFeature has a geometry and date that are within the space/time bounds of
+  the relevant GeoMesa Z-Index implementations (i.e. Z2, Z3, XZ2, XZ3)
 
-By default the ``has-geo`` and ``has-dtg`` validators are enabled. To enable other validators add this to the options
-block of your typesafe converter config:
+Additional validators may be loaded through Java SPI by by implementing
+``org.locationtech.geomesa.convert.SimpleFeatureValidator$ValidatorFactory`` and including a special service
+descriptor file. See below for additional information.
 
-::
+By default the ``index`` validator is enabled. This is suitable for most use cases, as it will choose the appropriate
+validator based on the SimpleFeatureType. To enable other validators, specify them in the options block of your
+typesafe converter config::
 
     geomesa.converters.myconverter {
       options {
-        validators = ["has-dtg", "has-geo", "z-index"]
+        validators = [ "has-dtg", "has-geo" ]
       }
     }
 
 Validation can be disabled by setting it to an empty array.
+
+Custom Validators
+^^^^^^^^^^^^^^^^^
+
+Custom validators may be loaded through Java SPI by by implementing
+``org.locationtech.geomesa.convert.SimpleFeatureValidator$ValidatorFactory``, shown below. Note that validators
+must be registered through a special service descriptor file.
+
+.. code-block:: scala
+
+  trait ValidatorFactory {
+    def name: String
+    def validator(sft: SimpleFeatureType, config: Option[String]): Validator
+  }
+
+When specifying validators in a converter config, the ``name`` of the factory must match the ``validators`` string.
+Any additional arguments may be specified in parentheses, which will be passed to the ``validator`` method.
+For example::
+
+    geomesa.converters.myconverter {
+      options {
+        validators = [ "my-custom-validator(optionA,optionB)" ]
+      }
+    }
+
+.. code-block:: scala
+
+  class MyCustomValidator extends ValidatorFactory {
+
+    override val name: String = "my-custom-validator"
+
+    override def validator(sft: SimpleFeatureType, config: Option[String]): Validator = {
+      if (config.exists(_.contains("optionA"))) {
+        // handle option a
+      } else {
+        // handle other options
+      }
+    }
+  }
+
+See the GeoMesa
+`unit tests <https://github.com/locationtech/geomesa/blob/master/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert/SimpleFeatureValidatorTest.scala>`__
+for a sample implementation.
+
+For more details on implementing a service provider, see the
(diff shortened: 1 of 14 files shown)


docs/user/convert/parsing_and_validation.rst
```diff
@@ -117,7 +166,7 @@ incremental parse mode:
 
     geomesa.converters.myconverter {
       options {
-        validators = [ "z-index" ]
+        validators = [ "index" ]
```
jnh5y: Since there's a string change, do we support configs with 'z-index'?  Or will old converters die?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
-      case "has-dtg" => new HasDtgValidator
-      case "z-index" => new ZIndexValidator
-      case unk => throw new IllegalArgumentException(s"Unknown validator $unk")
+      case NoneValidator.name   => (NoneValidator, None)
+      case IndexValidator.name  => (IndexValidator, None)
+      case HasGeoValidator.name => (HasGeoValidator, None)
+      case HasDtgValidator.name => (HasDtgValidator, None)
+
+      case "z-index" =>
+        logger.warn("'z-index' validator is deprecated, using 'index' instead")
```
jnh5y: +1

docs/user/convert/parsing_and_validation.rst
elahrvivaz: it's still supported, looks like you found it already

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
+  def default: SimpleFeatureValidator = apply(Seq(IndexValidator.name))
 
   def apply(names: Seq[String]): SimpleFeatureValidator = {
     val validators = names.map {
-      case "none"    => NoneValidator
-      case "has-geo" => new HasGeoValidator
-      case "has-dtg" => new HasDtgValidator
-      case "z-index" => new ZIndexValidator
-      case unk => throw new IllegalArgumentException(s"Unknown validator $unk")
+      case NoneValidator.name   => (NoneValidator, None)
```
jnh5y: How does the NoneValidator interact with the other validators?  I.e., what happens with "['none', 'index']".

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
       }
-    }
-
-    override def lastError: String = lastErr
   }
 
-  object HasGeoValidator {
-    val name: String = "has-geo"
+  /**
+    * Validator based on the indices used by the feature type. Currently only the z2 and z3 indices have
```
jnh5y: When you say 'z2' and 'z3', does that include the XZn indices?

elahrvivaz: yeah, i forgot to update that comment but it works with XZ as well

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
-    override def name: String = ZIndexValidator.name
-
-    override def init(sft: SimpleFeatureType): Unit = {
-      lastErr = null
-      dtgIndex = sft.getDtgIndex.getOrElse(-1)
-      geomIndex = sft.getGeomIndex
-      minDate = Date.from(BinnedTime.ZMinDate.toInstant)
-      maxDate = Date.from(BinnedTime.maxDate(sft.getZ3Interval).toInstant)
+  private class NullValidator(i: Int, error: String) extends Validator {
+    override def validate(sf: SimpleFeature): String = if (sf.getAttribute(i) == null) { error } else { null}
```
jnh5y: Space after null...

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
elahrvivaz: it would just be an extra call that wouldn't do anything - since the none validator always passes, the end result would be equivalent to just the 'index' validator.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
+    *
+    * @param geom geom index
+    */
+  private class Z2Validator(geom: Int) extends Validator {
+    override def validate(sf: SimpleFeature): String = {
+      val g = sf.getAttribute(geom).asInstanceOf[Geometry]
+      if (g == null) {
+        Errors.GeomNull
+      } else if (!wholeWorldEnvelope.contains(g.getEnvelopeInternal)) {
+        Errors.GeomBounds
```
jnh5y: As a follow-on, how about an XZ2 validator which only requires that the envelope/geometry intersects with the wholeWorldEnvelope?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
   }
 
-  object NoneValidator extends SimpleFeatureValidator {
-    override def name: String = "none"
-    override def validate(sf: SimpleFeature): Boolean = true
-    override def init(sft: SimpleFeatureType): Unit = {}
-    override def lastError: String = null
+  private object Errors {
+    val GeomNull   = "geometry is null"
+    val DateNull   = "date is null"
```
jnh5y: Idly, I wonder if it is worth naming the field which is null (or out of bounds).

elahrvivaz: the values will be printed out by the converter library, so it should be somewhat apparent... i don't have a strong opinion either way

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/SimpleFeatureValidator.scala
```diff
@@ -8,168 +8,256 @@
...
+    * @param minDate min z3 date
+    * @param maxDate max z3 date
+    */
+  private class Z3Validator(geom: Int, dtg: Int, minDate: Date, maxDate: Date) extends Validator {
+    private val dateBefore = s"date is before minimum indexable date ($minDate)"
+    private val dateAfter = s"date is after maximum indexable date ($maxDate)"
+
+    override def validate(sf: SimpleFeature): String = {
+      val d = sf.getAttribute(dtg).asInstanceOf[Date]
+      val g = sf.getAttribute(geom).asInstanceOf[Geometry]
```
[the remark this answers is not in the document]
elahrvivaz: it could definitely be more compact - i was going for minimal processing. i was trying to figure out a way to avoid evaluating anything unnecessarily and and this was all i could come up with

PR #2038, opened 2018-08-09
GEOMESA-2369 - Various logging improvements

jnh5y:

### DIFF
--- a/geomesa-convert/geomesa-convert-shp/src/main/scala/org/locationtech/geomesa/convert/shp/ShapefileConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-shp/src/main/scala/org/locationtech/geomesa/convert/shp/ShapefileConverterFactory.scala
@@ -88,9 +88,6 @@ object ShapefileConverterFactory extends LazyLogging {
     } catch {
       case NonFatal(e) =>
         logger.debug(s"Could not infer Shapefile converter from path '$path':", e)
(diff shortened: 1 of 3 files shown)


geomesa-convert/geomesa-convert-shp/src/main/scala/org/locationtech/geomesa/convert/shp/ShapefileConverterFactory.scala
```diff
@@ -88,6 +88,9 @@ object ShapefileConverterFactory extends LazyLogging {
     } catch {
       case NonFatal(e) =>
         logger.debug(s"Could not infer Shapefile converter from path '$path':", e)
+        if (logger.underlying.isTraceEnabled()) {
+          e.printStackTrace()
```
elahrvivaz: we shouldn't print to stdout - but isn't the stack trace already being logged on the line above?

jnh5y: I think those logger calls drop out the first line of the stack trace rather than the whole thing...  I can live without this if need be.

jnh5y: Cool, I'll 86 that line.

PR #2040, opened 2018-08-15
GEOMESA-2371 Use cached filters in HBase region server scans

elahrvivaz:
* Replaced JSimpleFeatureFilter with scala CqlTransformFilter


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/filters/CqlTransformFilter.scala
```diff
@@ -0,0 +1,346 @@
...
+  * serialized bytes, to avoid re-processing setup code whenever possible
+  *
+  * @param delegate delegate filter
+  * @param serialized serialized delegate filter
+  */
+class CqlTransformFilter(delegate: DelegateFilter, serialized: Array[Byte])
+    extends org.apache.hadoop.hbase.filter.Filter {
+
+  /**
+    * A filter can expect the following call sequence:
```
jnh5y: Is this from the HBase JavaDoc?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/filters/CqlTransformFilter.scala
```diff
@@ -0,0 +1,346 @@
...
+        var offset = 0
+        ByteArrays.writeInt(sftBytes.length, array, offset)
+        offset += 4
+        System.arraycopy(sftBytes, 0, array, offset, sftBytes.length)
+        offset += sftBytes.length
+        ByteArrays.writeInt(cqlBytes.length, array, offset)
+        offset += 4
+        System.arraycopy(cqlBytes, 0, array, offset, cqlBytes.length)
+        offset += cqlBytes.length
+        ByteArrays.writeInt(0, array, offset)
```
jnh5y: Is this to indicate that the length of the transforms is 0?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/filters/CqlTransformFilter.scala
```diff
@@ -0,0 +1,346 @@
...
+  /**
+    * Override of static method from org.apache.hadoop.hbase.filter.Filter
+    *
+    * @param pbBytes serialized bytes
+    * @throws org.apache.hadoop.hbase.exceptions.DeserializationException serialization exception
+    * @return
+    */
+  @throws(classOf[DeserializationException])
+  def parseFrom(pbBytes: Array[Byte]): org.apache.hadoop.hbase.filter.Filter =
+    // note: we copy KryoBufferSimpleFeatures as they seem to cause problems if re-used, even if thread-local
```
jnh5y: We are copying something other than the KryoBufferSF here, right?  

Also, if we have to copy, is the cache worth it?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/filters/CqlTransformFilter.scala
elahrvivaz: right, I'll add a comment

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/filters/CqlTransformFilter.scala
elahrvivaz: yeah, we're instantiating a new instance of the delegate filter, but it should be very lightweight. the KryoBufferSimpleFeature copy is slightly more expensive, as it will re-run some code to generate the transform function [here](https://github.com/locationtech/geomesa/blob/master/geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala#L120-L185) (if there are transforms set). But it saves creating a new serializer, and some cache lookups for getting the readers/writers... I think it's probably worthwhile, and it would make it easy to optimize further if we can figure out a way to fully re-use the features (possibly some kind of object pool, but that seemed slower when we looked at other pieces of the code that were using them).
Originally I was re-using the same KryoBufferSimpleFeature, but for some reason that was causing occasional thread conflicts, so it seems like each filter instance needs to have a separate copy

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/filters/CqlTransformFilter.scala
elahrvivaz: yeah, i'll add a comment to clarify

PR #2043, opened 2018-08-16
GEOMESA-2376 - Change converter field dependency calc to be non-recursive

elahrvivaz:
* Recursive function is slow when there are 100s of dependent fields


elahrvivaz:
haven't been able to reproduce the problem yet, but this solution doesn't seem to fix it

PR #2047, opened 2018-08-20
GEOMESA-2366 Partition data store tables based on time intervals

elahrvivaz:
* Create schema with 'geomesa.table.partition=time'
* 'manage-partitions' tools command
* Supported in HBase, Accumulo and Cassandra


elahrvivaz:
- [ ] need to update docs for adopting tables

docs/user/datastores/index_config.rst
```diff
@@ -384,6 +384,52 @@ To set the cardinality of an attribute, use the key ``cardinality`` on the attri
...
+    .. code-tab:: scala SchemaBuilder
+
+        import org.locationtech.geomesa.utils.geotools.SchemaBuilder
+
+        val sft = SchemaBuilder.builder()
+            .addString("name")
+            .addDate("dtg")
+            .addPoint("geom", default = true)
+            .userData
+            .partitioned()
```
jnh5y: To be 'future-proof' do we need to have the partitioned bit take an argument?

Also, I suppose this is a bad idea, but can I partition by a different dtg field than the indexed one?

elahrvivaz: what would the argument be?
the partitioning is SPI loaded, so you could plug in whatever you want. the only one i've provided here is based on the default dtg. the partitioning API passes in the sft, so you could set other flags in the user data to configure a specific partition impl. but i haven't defined that as a common key or anything.

elahrvivaz: the `partitioned` call here is just a convenience, you can always set arbitrary keys using `.userData` instead

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloQueryPlan.scala
```diff
@@ -124,23 +135,29 @@ case class BatchScanPlan(filter: AccumuloFilterStrategyType,
...
-        scanner.setRanges(ranges)
-        configure(scanner)
-        SelfClosingIterator(scanner.iterator, scanner.close())
-      }
+    if (PartitionParallelScan.toBoolean.contains(true)) {
+      // kick off all the scans at once
+      tables.map(scanner(ds, auths, _)).foldLeft(CloseableIterator.empty[Entry[Key, Value]])(_ ++ _)
+    } else {
+      // kick off the scans sequentially as they finish
+      SelfClosingIterator(tables.iterator).flatMap(scanner(ds, auths, _))
```
jnh5y: In either case, does the number of query threads/resources get applied?

And in the sequential case, does one see a 'gating' effect?  I.e. 10 query threads happen against one table until they all finish and then the threads move onto the next table?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/z3/Z3WritableIndex.scala
```diff
@@ -37,14 +37,15 @@ trait Z3WritableIndex extends AccumuloFeatureIndex {
...
-  override def delete(sft: SimpleFeatureType, ds: AccumuloDataStore, shared: Boolean): Unit = {
-    val table = getTableName(sft.getTypeName, ds)
-    if (ds.tableOps.exists(table)) {
-      // we need to synchronize deleting of tables in mock accumulo as it's not thread safe
-      if (ds.connector.isInstanceOf[MockConnector]) {
-        ds.connector.synchronized(ds.tableOps.delete(table))
-      } else {
-        ds.tableOps.delete(table)
+  override def delete(sft: SimpleFeatureType, ds: AccumuloDataStore, partition: Option[String]): Unit = {
+    getTableNames(sft, ds, partition).par.foreach { table =>
```
jnh5y: .par has some parallelism settings implicitly set.  Do we want to surface those in any configurable way?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloQueryPlan.scala
elahrvivaz: each 'query' (against at different table) will use the normal number of threads, so running in parallel will multiply the total threads. in the sequential case, one 'query' won't start until the previous one finishes.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/legacy/z3/Z3WritableIndex.scala
elahrvivaz: i don't think it's worth it... from my understanding you can just use a different thread pool than the default?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/data/TableConfCommand.scala
```diff
@@ -61,15 +59,15 @@ class TableConfUpdateCommand extends AccumuloDataStoreCommand {
...
 
-      if (params.newValue != value) {
+      if (values != Seq(params.newValue)) {
         Command.user.info(s"Updating configuration parameter to '${params.newValue}'...")
-        setValue(ds, table, params.key, params.newValue)
-        val updated = getProp(ds, table, params.key)
-        Command.user.info(s"  updated value: ${params.key}=$updated")
+        tables.foreach(setValue(ds, _, params.key, params.newValue))
+        val updated = tables.map(getProp(ds, _, params.key)).distinct
+        updated.foreach(v => Command.user.info(s"  updated value: ${params.key}=$v"))
```
jnh5y: Hmm... do we want to update all the table confs uniformly?!  Seems like there may be occasional benefits to configuring different partitions differently.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/GeoMesaFeatureIndex.scala
```diff
@@ -27,13 +29,14 @@ import org.opengis.filter.Filter
...
-trait GeoMesaFeatureIndex[DS <: GeoMesaDataStore[DS, F, WriteResult], F <: WrappedFeature, WriteResult] {
+trait GeoMesaFeatureIndex[DS <: GeoMesaDataStore[DS, F, WriteResult], F <: WrappedFeature, WriteResult]
+    extends LazyLogging {
 
   type TypedFilterStrategy = FilterStrategy[DS, F, WriteResult]
 
   lazy val identifier: String = s"$name:$version"
 
-  lazy val tableNameKey: String = s"table.$name.v$version"
+  lazy private val tableNameKey: String = s"table.$name.v$version"
```
jnh5y: Hmmm... I wonder if this should be private or public.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/package.scala
```diff
@@ -8,10 +8,18 @@
...
 import org.locationtech.geomesa.utils.conf.GeoMesaSystemProperties.SystemProperty
 
 package object index {
 
+  type FlushableFeatureWriter = SimpleFeatureWriter with Flushable
+
   val FilterCacheSize = SystemProperty("geomesa.cache.filters.size", "1000")
   val ZFilterCacheSize = SystemProperty("geomesa.cache.z-filters.size", "1000")
+
+  val PartitionParallelScan = SystemProperty("geomesa.partition.scan.parallel", "false")
```
jnh5y: I'd tempted to argue that parallel scans will be fine (as long as they are capped by a scans-thread setting).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloQueryPlan.scala
jnh5y: Ok, so pertaining to another question... flattening the list of queries out to use one common pool would be another execution option and would require additional development.

Sounds like a new ticket.:)

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/data/TableConfCommand.scala
elahrvivaz: hmm, like what?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/GeoMesaFeatureIndex.scala
elahrvivaz: i made it private as the partition logic means that there won't always be a value under this exact key, so you should use the getTableNames method instead

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
```diff
@@ -87,173 +153,222 @@ object GeoMesaFeatureWriter extends LazyLogging {
...
-    mutators.foreach {
-      case m: Flushable => FlushQuietly(m).foreach(exceptions.+=)
-      case _ => // no-op
+    private def mutators(feature: SimpleFeature): (Seq[(T, F => Seq[W])], Seq[(T, F => Seq[W])]) = {
+      val p = partition.partition(feature)
+      cache.getOrElseUpdate(p, {
+        // reconfigure the partition each time - this should be idempotent, and block
+        // until it is fully created (which may happen in some other thread)
+        val mutators = indices.par.map(i => createMutator(i.configure(sft, ds, Some(p)))).seq
+        (mutators.zip(writeConverters), mutators.zip(removeConverters))
```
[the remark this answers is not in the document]
elahrvivaz: they're functions that contain the logic to go from a simple feature to e.g. a mutation. We could separate it out but I don't think there's a lot of overhead in creating them.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/package.scala
elahrvivaz: that's why there's a sys prop :)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloQueryPlan.scala
elahrvivaz: Well the thread pool is currently managed by the batch scanner, so I'm not actually sure that it would be possible to share a thread pool. You could divide the number of threads among the various scans, but that doesn't seem like a great solution either.

jnh5y: Sounds great.  I thought we may have been managing the thread.  Since we are not, this is an excellent way to handle this.

elahrvivaz: we do handle the threading in hbase, so if you think it's worthwhile we could do something there. It may be possible to hook into the accumulo API too, but not that I know of.

jnh5y: Eh, how about we toss a ticket in for managing one thread pool in HBase?  (I.e., let's think about it for a few days...)

The Accumulo case is subtle.  @jahhulbert-ccri do you know if there's way to share a thread pool among BatchScanners?!

PR #2049, opened 2018-08-24
GEOMESA-2381 Test and fix OR

khobbs-ccri:
1. Add an OR query to a Kafka unit test that.
2. Children of OrHashEquality and OrSequentialEquality filters should not be
   lazy. The FastFilterFactory.sfts must be available when children are
   evaluated.

Signed-off-by: Kevin Hobbs <khobbs@ccri.com>


jnh5y:
@khobbs-ccri have you filed out an ECA?

PR #2052, opened 2018-09-05
GEOMESA-2386 Adding WPS module to FileSystem gs-plugin

elahrvivaz:


elahrvivaz:
I haven't been able to re-create the original issue, but bundles the jars needed for WPS in with the FSDS gs-plugin (where previously you had to also install the e.g. accumulo gs-plugin, or manually copy the correct jars)

PR #2059, opened 2018-09-19
GEOMESA-2404 Fixes bugs with FSDS Orc Mapreduce Formats

jnh5y:
* Creates Spark+FSDS test framework and applies it to ORC format.
* Handles count(*) projection case.
* Fixes various ORC MR format reading bugs.
* Allows SimpleFeatureSpecs to have -'s in userData (without attributes).

### DIFF
--- a/geomesa-fs/geomesa-fs-spark-runtime/pom.xml
+++ b/geomesa-fs/geomesa-fs-spark-runtime/pom.xml
@@ -49,6 +49,7 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>14.0.1</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.specs2</groupId>
@@ -79,10 +80,12 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-mapreduce-client-common</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-minicluster</artifactId>
+            <scope>test</scope>
         </dependency>
 
     </dependencies>
--- a/geomesa-fs/geomesa-fs-spark-runtime/pom.xml
+++ b/geomesa-fs/geomesa-fs-spark-runtime/pom.xml
@@ -46,6 +46,11 @@
 
         <!-- test dependencies -->
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>14.0.1</version>
+        </dependency>
+        <dependency>
             <groupId>org.specs2</groupId>
             <artifactId>specs2-core_${scala.binary.version}</artifactId>
         </dependency>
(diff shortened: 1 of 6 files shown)


geomesa-fs/geomesa-fs-spark-runtime/pom.xml
```diff
@@ -45,6 +45,11 @@
         </dependency>
 
         <!-- test dependencies -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>14.0.1</version>
+        </dependency>
```
elahrvivaz: do you want this to be 'test' scope? i think it's compile, if you don't specify.

geomesa-fs/geomesa-fs-spark-runtime/pom.xml
```diff
@@ -60,6 +65,26 @@
...
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <version>${spark.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-mapreduce-client-common</artifactId>
+        </dependency>
```
elahrvivaz: this looks like it's coming in as provided - just for sanity, could you move it up to the 'provided' section of the pom? (or alternatively, manage it to test scope)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-orc/src/main/scala/org/locationtech/geomesa/fs/storage/orc/utils/OrcInputFormatReader.scala
```diff
@@ -34,24 +34,24 @@ object OrcInputFormatReader {
     var i = 0
     var col = 0
     while (i < sft.getAttributeCount) {
+      val bindings = ObjectType.selectType(sft.getDescriptor(i))
+      val reader = bindings.head match {
+        case ObjectType.GEOMETRY => col += 1; createGeometryReader(bindings(1), col - 1, col, i)
```
elahrvivaz: i guess this needs to be outside the if to increment the col index for geom fields? maybe we could just add an else statement instead of instantiating an unneeded reader

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-orc/src/main/scala/org/locationtech/geomesa/fs/storage/orc/jobs/OrcSimpleFeatureInputFormat.scala
```diff
@@ -89,7 +89,9 @@ object OrcSimpleFeatureInputFormat {
   }
 
   def getReadColumns(conf: Configuration): Option[Set[Int]] =
-    Option(conf.get(ReadColumnsConfig)).map(_.split(",").map(_.toInt).toSet)
+    Option(conf.get(ReadColumnsConfig))
+      .filter(_.nonEmpty)
```
elahrvivaz: do you know if this is going to make it select all the cols? we should always at least be returning the feature id, right? it might be worthwhile to check out if we are just scanning that col or all cols and discarding them.

jnh5y: In the case of SparkSQL, we might return none of the columns (like in a count(*) situation).  Maybe those should be handled better/differently?

PR #2060, opened 2018-09-19
GEOMESA-2400 Remove netty from spark-runtime jars

jnh5y:
* Removes io.netty from all 5 Spark runtimes.


jnh5y:
I've tested with the FSDS.  Spark 2.2 and 2.3 do ship with netty jars.

jahhulbert-ccri:
if they ship with the right jars we should be OK...?

jnh5y:
@jahhulbert-ccri (I believe) Spark relies on Netty to do their web service layer.  If that's the case, then a working Spark has Netty.

PR #2064, opened 2018-09-24
GEOMESA-2388 FSDS - Support distributed metadata operations

elahrvivaz:
* Metadata is written as a series of changelogs
* New 'manage-metadata' command replaces 'update-metadata'
* Restores back-compatibility with 2.0.x metadata files
* Breaks storage API-level compatibility
* Breaks MethodProfiling API compatibility


docs/user/filesystem/commandline.rst
```diff
@@ -133,8 +131,50 @@ Note that in our example January 1st and 2nd both do not have all four quadrants
 dataset for that day didn't have any data in that region of the world. If additional data were ingested, the directory
 and a corresponding file would be created.
 
-``update-metadata``
+``manage-metadata``
```
jnh5y: I wonder if this should be shortened to 'metadata'.  

Also, is there any reason that one might want to get a portion of the metadata?  I suppose I'm wondering what other metadata commands might be useful.

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemFeatureStore.scala
```diff
@@ -89,14 +88,27 @@ class FileSystemFeatureStore(val storage: FileSystemStorage,
...
+    }
+  }
+
   override def getCountInternal(query: Query): Int = {
-    if (query == Query.ALL || query.getFilter.isInstanceOf[IncludeFilter]) {
-      storage.getMetadata.getFeatureCount.toInt
-    } else {
-      -1L
-    }.toInt
+    val partitions = storage.getPartitions(query.getFilter).iterator
```
jnh5y: Are there any sane options for memoizing this?

docs/user/filesystem/commandline.rst
elahrvivaz: for consistency, we've been trying to keep command names as 'verb-noun' (at least I have...). Do the existing commands for listing partitions, files, etc cover your question? I didn't modify those.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/MetadataFileSystemStorage.scala
```diff
@@ -97,24 +102,29 @@ abstract class MetadataFileSystemStorage(conf: Configuration,
...
+        val baseDir = if (metadata.getPartitionScheme.isLeafStorage) { partitionPath.getParent } else { partitionPath }
+        val paths = new java.util.ArrayList[Path](files.size())
+
+        val iter = files.iterator
+        while (iter.hasNext) {
+          val path = new Path(baseDir, iter.next)
+          if (PathCache.exists(metadata.getFileContext, path)) {
+            paths.add(path)
+          } else {
+            logger.warn(s"Inconsistent metadata for ${metadata.getSchema.getTypeName}: $path")
```
jnh5y: Is there something that the user/admin should do in this case?  (I.e., should one manage-metadata to rectify?)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/MetadataFileSystemStorage.scala
```diff
@@ -153,18 +164,28 @@ abstract class MetadataFileSystemStorage(conf: Configuration,
...
 
-  override def updateMetadata(): Unit = {
-    profile {
-      val fc = metadata.getFileContext
-      val scheme = metadata.getPartitionScheme
-      metadata.setFiles(StorageUtils.partitionsAndFiles(fc, metadata.getRoot, scheme, extension, cache = false))
-    } ((_, time) => logger.debug(s"Metadata update took ${time}ms."))
+  class AddCallback(partition: String, file: Path) extends WriterCallback {
+    override def onClose(count: Long, bounds: Envelope): Unit =
+      metadata.addPartition(new PartitionMetadata(partition, Collections.singletonList(file.getName), count, bounds))
```
jnh5y: As a sanity check, we are assuming that the metadata can always be compacted into one file, right?

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemFeatureStore.scala
elahrvivaz: i don't think that this should be a particularly expensive call

elahrvivaz: the partitions are already cached in memory, so it's just iterating through them to match the filter

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/MetadataFileSystemStorage.scala
elahrvivaz: yeah, you could use `manage-metadata deregister` to remove the reference, or figure out why the file is missing and replace it

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/MetadataFileSystemStorage.scala
elahrvivaz: no, there will be 1-n files per partition. Metadata compaction will reduce that down to one per partition

elahrvivaz: as a follow on, it might be good to put metadata files into separate folders so we don't have too many files in one dir. I opened a ticket for that: https://geomesa.atlassian.net/browse/GEOMESA-2415

PR #2067, opened 2018-09-26
GEOMESA-2417 - FSDS - Classpath issue with orc+hive in spark-runtime jar

jnh5y:


elahrvivaz:
nice find. looks good, do we need to run the cq script?

jnh5y:
Good call.  Doing that now.....

elahrvivaz:
did the build pass the first time? I'll merge if so, if we're not sure I'll wait for travis. @jnh5y I'll add a sign-off by you, if that's ok

jnh5y:
@elahrvivaz it looks like it did build both times.  Feel free to add the sign-off and shippit!

PR #2074, opened 2018-09-27
GEOMESA-2419 Fix version of json4s-ext

elahrvivaz:


elahrvivaz:
the bug fix version was off, our CQ is actually for 3.5.4 but it slipped through transitively as 3.5.2

PR #2081, opened 2018-10-10
GEOMESA-2429 Converters - converter-level caches in evaluation contex…

elahrvivaz:
…ts can leak resources

* SimpleFeatureConverter changed to extend Closeable
* Caches configured at the converter level are shared between evaluation contexts

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/CompositeConverter.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/CompositeConverter.scala
@@ -40,7 +40,6 @@ class CompositeConverterFactory[I] extends SimpleFeatureConverterFactory[I] {
 class CompositeConverter[I](val targetSFT: SimpleFeatureType, converters: Seq[(Predicate, SimpleFeatureConverter[I])])
     extends SimpleFeatureConverter[I] {
 
-
   override val caches: Map[String, EnrichmentCache] = Map.empty
 
   val predsWithIndex = converters.map(_._1).zipWithIndex.toIndexedSeq
@@ -101,6 +100,8 @@ class CompositeConverter[I](val targetSFT: SimpleFeatureType, converters: Seq[(P
   override def processSingleInput(i: I, ec: EvaluationContext): Iterator[SimpleFeature] = ???
 
   override def process(is: InputStream, ec: EvaluationContext): Iterator[SimpleFeature] = ???
+
+  override def close(): Unit = indexedConverters.foreach(_.close())
 }
 
 case class CompositeEvaluationContext(contexts: IndexedSeq[EvaluationContext]) extends EvaluationContext {
--- a/geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
+++ b/geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
@@ -9,6 +9,7 @@
 
 package org.locationtech.geomesa.convert.redis
 
+import java.io.Closeable
 import java.util.concurrent.TimeUnit
 
 import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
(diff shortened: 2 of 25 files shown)


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/CompositeConverter.scala
```diff
@@ -101,6 +100,8 @@ class CompositeConverter[I](val targetSFT: SimpleFeatureType, converters: Seq[(P
   override def processSingleInput(i: I, ec: EvaluationContext): Iterator[SimpleFeature] = ???
 
   override def process(is: InputStream, ec: EvaluationContext): Iterator[SimpleFeature] = ???
+
+  override def close(): Unit = indexedConverters.foreach(_.close())
```
jahhulbert-ccri: if one converter throws an exception closing will the others close properly? Question I guess is really around resource leaking.

elahrvivaz: yeah, i was lazy b/c this was the old deprecated one, but I'll fix it

elahrvivaz: fixed

geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
```diff
@@ -64,14 +66,13 @@ class RedisEnrichmentCacheFactory extends EnrichmentCacheFactory {
   override def build(conf: Config): EnrichmentCache = {
     val Array(host, port) = parseRedisURL(conf.getString("redis-url"))
     val timeout = if (conf.hasPath("expiration")) conf.getLong("expiration") else -1
-    val connBuilder = new RedisConnectionBuilder {
-      val pool = new JedisPool(host, port.toInt)
-      override def getConn: Jedis = {
-        pool.getResource
-      }
+    val connBuilder: RedisConnectionBuilder = new RedisConnectionBuilder {
```
jnh5y: Since this is a pool are connections returned?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-convert/src/main/scala/org/locationtech/geomesa/fs/storage/converter/ConverterStorage.scala
```diff
@@ -27,6 +27,8 @@ class ConverterStorage(fc: FileContext,
                        converter: SimpleFeatureConverter,
                        partitionScheme: PartitionScheme) extends FileSystemStorage {
 
+  // TODO close converter...
```
jnh5y: Do we need to address this TODO?  If it isn't stray, can we enter a ticket describing the work a bit more?

geomesa-convert/geomesa-convert-redis-cache/src/main/scala/org/locationtech/geomesa/convert/redis/RedisEnrichmentCache.scala
elahrvivaz: no, they are closed up on line 52, which is how you 'return' them

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-convert/src/main/scala/org/locationtech/geomesa/fs/storage/converter/ConverterStorage.scala
elahrvivaz: the problem is that we aggressively cache storage instances, so even if we wired a 'close' method through the entire storage api, we'd also have to implement a 'borrow/return' paradigm and expire idle convictions. Since this only affects redis caches, and since it will only open a single connection now (instead of one per evaluation context), there shouldn't really be an impact to not closing it. i'll update the TODO but am hesitant to add a ticket as i think the solution would be worse than the problem

PR #2086, opened 2018-10-19
GEOMESA-2426 Fixing Spark version errors with logical relations

elahrvivaz:


jnh5y:
Which versions of Spark does this work with?  Can we document that?  And does it cover 2.1, 2.2, and 2.3 by chance?:)

elahrvivaz:
I doubt that it works with 2.1... I think that it has a fighting chance with 2.2 and should definitely work with 2.3 since that's what we compile with.

elahrvivaz:
looks like we've already documented it: https://github.com/locationtech/geomesa/blob/master/docs/common.py#L129

PR #2087, opened 2018-10-22
GEOMESA-2443 CLI - Follow symlinks when looking up zookeeper JAR

amcdonaldccri:
* GEOMESA-2443 CLI - Follow symlinks when looking up zookeeper JAR


elahrvivaz:
merged as 0722a05ce56c75e1262f6a9fa60bc469c47def58

PR #2091, opened 2018-10-24
GEOMESA-2444 Never calculate exact counts unless configured to do so

elahrvivaz:


jnh5y:
Is this something worth documenting?

elahrvivaz:
I think this actually pulls the functionality into alignment with the documentation.

khobbs-ccri:
I have tested this on an optix test system, and found it to allow wfs 1.1 and json output.

PR #2092, opened 2018-10-26
GEOMESA-2447 Refactor the CartesianProductIterator to remove costly m…

cne1x:
…odulus calls

Reworked the implementation so that it no longer uses the integer
division, but does a more direct iteration over array indexes.
The JIRA ticket contains a chart that shows, for the few test
cases run, the new implementation takes approximately half as
long to generate and exhaust the iterators over the product.

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/CartesianProductIterable.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/CartesianProductIterable.scala
@@ -14,32 +14,51 @@ package org.locationtech.geomesa.utils.iterators
  * way to query the list size that is independent of the iterator itself.
  * (That is, asking for the size does not exhaust any iterator.)
  *
+ * NB:  The first sequence is the least significant; that is, it will
+ * increment fast while the last sequence is the most significant (will
+ * increment slowly).
+ *
  * @param seqs the list-of-lists whose items are to be recombined
  * @tparam T the type of items
  */
-case class CartesianProductIterable[T](seqs:Seq[Seq[T]]) extends Iterable[Seq[T]] {
-  // how many (total) combinations there will be in an iterator
-  lazy val expectedSize : Long = seqs.map(seq => seq.size.toLong).product
 
-  // create an iterator that will visit all possible combinations
-  // (every time you call this, you get a NEW iterator)
-  def iterator : Iterator[Seq[T]] = new Iterator[Seq[T]] {
-    // variable index that counts off the combinations
-    var index : Long = 0L
-    // current value (combination)
-    private def value : Seq[T] =
-      seqs.foldLeft(Seq[T](),index)((t,src) => t match { case (seqSoFar,idx) => {
-        val modulus : Int = src.size
-        val itemIndex : Int = (idx % modulus).toInt
-        val item : T = src(itemIndex)
-        (seqSoFar++Seq(item), idx / modulus)
-      }})._1
-    def next() : Seq[T] = {
-      val result = if (hasNext) value else null
-      index = index + 1L
+case class CartesianProductIterable(seqs: Seq[Seq[_]]) extends Iterable[Seq[_]] {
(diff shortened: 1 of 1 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/CartesianProductIterable.scala
```diff
@@ -14,32 +14,51 @@ package org.locationtech.geomesa.utils.iterators
...
-        val modulus : Int = src.size
-        val itemIndex : Int = (idx % modulus).toInt
-        val item : T = src(itemIndex)
-        (seqSoFar++Seq(item), idx / modulus)
-      }})._1
-    def next() : Seq[T] = {
-      val result = if (hasNext) value else null
-      index = index + 1L
+case class CartesianProductIterable(seqs: Seq[Seq[_]]) extends Iterable[Seq[_]] {
+  lazy val expectedSize: Int = seqs.map(_.size).product
```
aheyne: Should this still be long?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/CartesianProductIterable.scala
```diff
@@ -14,32 +14,51 @@ package org.locationtech.geomesa.utils.iterators
...
+
+  def iterator: Iterator[Seq[_]] = new Iterator[Seq[_]] {
+    val n = seqs.size
+    val maxes = seqs.map(seq => seq.size)
+    val indexes = scala.collection.mutable.MutableList.fill(seqs.size)(0)
+    var nextItem: Option[Seq[_]] = if (isValid) Option(realize) else None
```
jnh5y: If we are using vars for optimization, we can do null checks rather than instantiating Options.  @cne1x can you try without Options and see if it is any faster/slower?

Also, I wonder if using a Vector would be better than/different from using a MutableList....

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/CartesianProductIterable.scala
cne1x: Prolly; it was in the old version.  Good eye!

elahrvivaz: I hope we don't every actually exceed Int.MaxValue here :P

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/CartesianProductIterable.scala
cne1x: For fun, removing `Option` reduced the new runtime by an additional ~5%; introducing `ArraySeq` and `Vector` reduced the new runtime by an additional ~65%.  I had forgotten we talked about `Vector` yesterday at our Scala for Cats session; thanks for the reminder!

elahrvivaz: any thoughts on comparisons between `Vector` and `Array` or `ArrayBuffer`? Those are usually my go tos, but I haven't compared them.

cne1x: image

cne1x: Yeah, `Array` with `ArrayBuffer` have indistinguishable performance (at least in this case) from `Vector` and `ArraySeq`.

cne1x: Pushed a commit with the refined version.

cne1x:
@elahrvivaz, @jnh5y, @aheyne:  I think all concerns have been addressed.

PR #2095, opened 2018-11-01
GEOMESA-2451 HBase - empty transform causes filter deserialization error

elahrvivaz:


jnh5y:
@elahrvivaz would it be possible to add a quick Spark SQL test for this?  That's where this came up initially, right?

elahrvivaz:
@jnh5y we don't currently have any spark sql hbase tests... I believe you added an integration-style test that gets skipped by default, do you know if there's a reason we don't have any other ones?

jnh5y:
@elahrvivaz doesn't https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-spark-runtime/src/test/scala/org/locationtech/geomesa/hbase/spark/HBaseSparkProviderIntegrationTest.scala run during the build?

Maybe I'm making things up...

elahrvivaz:
@jnh5y it does run but all the tests have `skipped` in them

elahrvivaz:
This fixes the one problem but the spark sql count operation still isn't working - I think we could merge this one and then open a new ticket for spark

elahrvivaz:
Ok, finally tracked down the bug `Filter.isFamilyEssential` needs to be true not false. This doesn't seem to matter in the embedded hbase, but does in an actual hbase instance.

PR #2097, opened 2018-11-07
GEOMESA-2458 Add distributed ingest command parameters to silence out…

aheyne:
…put and close on job submission.

### DIFF
--- a/geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FileSystemConverterJob.scala
+++ b/geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FileSystemConverterJob.scala
@@ -44,7 +44,7 @@ trait FileSystemConverterJob extends StorageConfiguration with JobWithLibJars wi
           reducers: Int,
           libjarsFile: String,
           libjarsPaths: Iterator[() => Seq[File]],
-          statusCallback: Option[StatusCallback],
+          statusCallback: StatusCallback,
           waitForCompletion: Boolean = true): Option[(Long, Long)] = {
 
     import scala.collection.JavaConversions._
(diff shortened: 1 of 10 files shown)


geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FileSystemConverterJob.scala
```diff
@@ -113,38 +114,45 @@ trait FileSystemConverterJob extends StorageConfiguration with JobWithLibJars wi
...
+            qualifiedTempPath.foreach { tp =>
+              StorageJobUtils.distCopy(tp, metadata.getRoot, callback, 3, stageCount)
+            }
+          }
 
-      counterResult
+          Some(counterResult)
+        case None =>
+          Command.user.info("Job Submitted")
+          if (waitForCompletion) job.waitForCompletion(false)
```
cne1x: I suspect I'm just missing the point of the ticket, but the terminology inclines me to believe that if `waitForCompletion` is true, then shouldn't `waitForCompletion(true)` be set?  In which case, can't we just use `job.waitForCompletion(waitForCompletion)`?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -200,4 +200,10 @@ trait IngestParams extends OptionalTypeNameParam with OptionalFeatureSpecParam w
 
   @Parameter(names = Array("--src-list"), description = "Input files are text files with lists of files, one per line, to ingest.")
   var srcList: Boolean = false
+
+  @Parameter(names = Array("--no-job-status"), description = "Progress of the ingest job is not printed to the console, however, the application will close when job completes.")
+  var noJobStatusTracking: Boolean = false
```
cne1x: I understand why the options are written as "don't do ___", but the (inconsistent) double negatives throughout the code make it just that much more complicated to read.  Any chance we could rename these variables to be consistently `var jobStatusTracking: Boolean = true` and `var waitForCompletion: Boolean = true`?  Or, if that complicates the command-line annotation, create a second variable that is the positive assertion?  Or is it just me that's stuck on this?  (That happens.)

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FileSystemConverterJob.scala
```diff
@@ -44,7 +44,8 @@ trait FileSystemConverterJob extends StorageConfiguration with JobWithLibJars wi
           reducers: Int,
           libjarsFile: String,
           libjarsPaths: Iterator[() => Seq[File]],
-          statusCallback: StatusCallback): (Long, Long) = {
+          statusCallback: Option[StatusCallback],
```
elahrvivaz: I think we've got two conflicting things going on here - as it is, if you set `waitForCompletion` to false but provide a `statusCallback`, it will still wait. I'd revert back to `statusCallback` not being an `Option`

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FileSystemConverterJob.scala
elahrvivaz: I think the logic is correct - `job.waitForCompletion` will always wait before returning, the Boolean passed into it is actually a flag for 'verbose' output.

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FsIngestCommand.scala
```diff
@@ -80,10 +82,13 @@ object FsIngestCommand {
                                   libjarsPaths: Iterator[() => Seq[File]],
                                   numLocalThreads: Int,
                                   tempPath: Option[Path],
-                                  reducers: Option[java.lang.Integer])
-      extends ConverterIngest(sft, dsParams, converterConfig, inputs, mode, libjarsFile, libjarsPaths, numLocalThreads) {
+                                  reducers: Option[java.lang.Integer],
+                                  jobStatusTracking: Boolean,
```
elahrvivaz: I don't think we need this flag though - it's easy enough to accomplish the same thing with shell redirects, so we don't need to complicate the code

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
```diff
@@ -212,13 +214,12 @@ abstract class AbstractIngest(val dsParams: Map[String, String],
   }
 
   protected def runDistributed(): Unit = {
-    beforeRunTasks()
-    Command.user.info("Running ingestion in distributed mode")
```
elahrvivaz: was there a reason to move these two lines out of this method? I'd rather keep them here for consistency with `runLocal()`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: I think we should keep the negative, as that allows you on the command line to do `--no-job-status` instead of `--job-status false`, which seems nicer IMO. But we could easily add a `def jobStatusTracking = !noJobStatusTracking` and use that in our code

elahrvivaz: Although I think we should get rid of this particular parameter, but we could apply the theory to `--no-wait`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -200,4 +200,10 @@ trait IngestParams extends OptionalTypeNameParam with OptionalFeatureSpecParam w
...
+
+  @Parameter(names = Array("--no-wait"), description = "This application closes when ingest job is submitted. Useful for launching jobs with a script.")
+  var noWaitForCompletion: Boolean = false
```
elahrvivaz: can we rename this to `--no-tracking` and update the description to "Return once the job is submitted, instead of printing job status updates to the console"

elahrvivaz: We need to update the docs for these commands as well

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FileSystemConverterJob.scala
aheyne: Yeah, we should probably be better about using named params when passing in a boolean.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/AbstractIngest.scala
aheyne: Yes, because runDistributed gets overwritten in ConverterIngest.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
aheyne: Added

PR #2103, opened 2018-11-12
GEOMESA-2464 FSDS - fixing spark runtime jar with ORC input format

elahrvivaz:

### DIFF
--- a/geomesa-fs/geomesa-fs-spark-runtime/pom.xml
+++ b/geomesa-fs/geomesa-fs-spark-runtime/pom.xml
@@ -146,6 +146,16 @@
                                     <pattern>org.apache.parquet</pattern>
                                     <shadedPattern>org.locationtech.geomesa.fs.shaded.org.apache.parquet</shadedPattern>
                                 </relocation>
+                                <!-- spark also includes orc/hive, relocate to avoid version conflicts -->
+                                <relocation>
+                                    <pattern>org.apache.orc</pattern>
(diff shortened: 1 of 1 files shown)


geomesa-fs/geomesa-fs-spark-runtime/pom.xml
```diff
@@ -146,6 +146,16 @@
                                     <pattern>org.apache.parquet</pattern>
                                     <shadedPattern>org.locationtech.geomesa.fs.shaded.org.apache.parquet</shadedPattern>
                                 </relocation>
+                                <!-- spark also includes orc/hive, relocate to avoid version conflicts -->
+                                <relocation>
+                                    <pattern>org.apache.orc</pattern>
+                                    <shadedPattern>org.locationtech.geomesa.fs.shaded.org.apache.orc</shadedPattern>
```
jnh5y: Hmm.... the orc-no-hive stuff should be taking care of this...

elahrvivaz: We might technically not need this after I added the kryo relocation, but it might save us some grief in the future :)

elahrvivaz: Would you prefer removing it?

elahrvivaz: Actually, I think it's better to keep it in, as the kryo code is getting invoked from the orc code, so we want to ensure it's coming from our jar and not anywhere else.

PR #2104, opened 2018-11-16
GEOMESA-2454 Introduce st_castToGeometry

mojodna:
This allows Geometry subclasses to be treated as Geometries in Dataset domain-specific objects.

### DIFF
--- a/geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricCastFunctionsTest.scala
+++ b/geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricCastFunctionsTest.scala
@@ -12,6 +12,7 @@ import com.vividsolutions.jts.geom._
 import org.apache.spark.sql.functions.lit
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.spark.jts._
+import org.locationtech.geomesa.spark.jts.util.util.{GeometryContainer, LineStringContainer, PointContainer, PolygonContainer}
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
 
@@ -31,14 +32,20 @@ class GeometricCastFunctionsTest extends Specification with TestEnvironment {
       "null" >> {
         sc.sql("select st_castToPoint(null)").collect.head(0) must beNull
         dfBlank.select(st_castToPoint(lit(null))).first must beNull
+
+        import spark.implicits._
+        dfBlank.select(st_castToPoint(lit(null)) as 'geom).as[PointContainer].head must haveClass[PointContainer]
       }
 
       "point" >> {
         val pointTxt = "POINT(1 1)"
         val point = s"st_geomFromWKT('$pointTxt')"
-        val df = sc.sql(s"select st_castToPoint($point)")
+        val df = sc.sql(s"select st_castToPoint($point) geom")
         df.collect.head(0).asInstanceOf[AnyRef] must haveClass[Point]
         dfBlank.select(st_castToPoint(st_geomFromWKT(pointTxt))).first must haveClass[Point]
+
+        import spark.implicits._
+        df.as[PointContainer].head must haveClass[PointContainer]
       }
     }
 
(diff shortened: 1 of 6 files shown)


geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/util/util.scala
```diff
@@ -0,0 +1,10 @@
+package org.locationtech.geomesa.spark.jts.util
```
elahrvivaz: this class is causing the license check to fail - @mojodna do you want copyright on these files? We can add a new header template with your copyright, otherwise we can add the standard CCRi one:
```
 * Copyright (c) 2013-2018 Commonwealth Computer Research, Inc.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Apache License, Version 2.0
 * which accompanies this distribution and is available at
```

mojodna: CCRi is fine.

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/util/util.scala
```diff
@@ -0,0 +1,10 @@
...
+
+import com.vividsolutions.jts.geom.{Geometry, LineString, Point, Polygon}
+
+package object util {
+  case class PointContainer(geom: Point)
```
elahrvivaz: I'm slightly confused (probably due to my unfamiliarity with spark), but why are these containers needed?

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/util/util.scala
elahrvivaz: thanks, I'll go ahead and add it

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/util/util.scala
mojodna: The original cast error surfaced when DataFrames (`type DataFrame = Dataset[Row]`) were converted to `Dataset[Container]` (using `df.as[Container]`) in order to use rows as strongly typed objects.

More here:
https://databricks.com/blog/2016/07/14/a-tale-of-three-apache-spark-apis-rdds-dataframes-and-datasets.html

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricCastFunctionsTest.scala
```diff
@@ -31,44 +32,105 @@ class GeometricCastFunctionsTest extends Specification with TestEnvironment {
...
+
+        import spark.implicits._
+        dfBlank.select(st_castToPolygon(lit(null)) as 'geom).as[PolygonContainer].head must haveClass[PolygonContainer]
       }
 
       "polygon" >> {
         val polygonTxt = "POLYGON((1 1, 1 2, 2 2, 2 1, 1 1))"
         val polygon = s"st_geomFromWKT('$polygonTxt')"
-        val df = sc.sql(s"select st_castToPolygon($polygon)")
+        val df = sc.sql(s"select st_castToPolygon($polygon) geom")
```
jnh5y: @mojodna, this is maybe a dumb question on my part...  What does adding 'geom' to the end of this SQL statement do?

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/util/util.scala
jnh5y: Cool.  Am I understanding this correctly that this adds better DataSet support to the JTS+Spark work here?

geomesa-spark/geomesa-spark-jts/src/test/scala/org/locationtech/geomesa/spark/jts/udf/GeometricCastFunctionsTest.scala
mojodna: It creates a better alias for the column created. With this, the column is `geom` within the DataFrame and will automatically map to the container's `geom` property. Previously, the column name was `st_castToPolygon(<something>)`.

PR #2106, opened 2018-11-21
GEOMESA-2474 Tools - considering file type for converter inference

elahrvivaz:


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/DataFormats.scala
```diff
@@ -18,7 +18,7 @@ import org.apache.commons.io.FilenameUtils
 object DataFormats extends Enumeration {
 
   type DataFormat = Value
-  val Arrow, Avro, Bin, Csv, GeoJson, Gml, Json, Leaflet , Null, Shp, Tsv, Xml = Value
```
jnh5y: Why is GeoJson removed here?

PR #2109, opened 2018-11-27
GEOMESA-2472 Log warning if simple feature attribute can't be convert…

elahrvivaz:
…ed to the correct type

* New FastConveter class to cache converters and log a warning

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/converters/FastConverter.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/converters/FastConverter.scala
@@ -1,0 +1,84 @@
+/***********************************************************************
+ * Copyright (c) 2013-2019 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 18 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/converters/FastConverter.scala
```diff
@@ -0,0 +1,84 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+import com.typesafe.scalalogging.StrictLogging
+import org.geotools.factory.GeoTools
+import org.geotools.util.{Converter, Converters}
+
+import scala.util.control.NonFatal
+
+/**
+  * Replacement for geotools `Converters`, which caches the converters for each pair of (from -> to) instead of
```
jnh5y: The GeoTools converters recreates the class each time?

elahrvivaz: yeah... https://github.com/geotools/geotools/blob/master/modules/library/metadata/src/main/java/org/geotools/util/Converters.java#L148

PR #2112, opened 2018-11-29
GEOMESA-2483 Docs - Fix build-and-mirror to delete old pages

elahrvivaz:

### DIFF
--- a/docs/bin/build-and-mirror.sh
+++ b/docs/bin/build-and-mirror.sh
@@ -20,6 +20,7 @@ function build_and_mirror {
 	set +x
 	if [ "$release" == "root" ] ; then
 		set -x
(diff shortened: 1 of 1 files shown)


docs/bin/build-and-mirror.sh
```diff
@@ -20,6 +20,7 @@ function build_and_mirror {
 	set +x
 	if [ "$release" == "root" ] ; then
 		set -x
+		for file in $(ls $GEOMESA/docs/target/html/); do rm -r $GEOMESA_GITHUB_IO/documentation/$file; done
```
jnh5y: Just to be clear, this doesn't do anything with the versioned documentation?

elahrvivaz: right - they're all in the same folder, so it's a bit tricky to not modify them

elahrvivaz: I played around with rsync --include and --exclude but couldn't really get anything to work right

PR #2114, opened 2018-12-05
GEOMESA-2485 Fix SimpleFeatureSimpleFeatureConverter field transforms

elahrvivaz:


jnh5y:
@elahrvivaz thanks for verifying that!.

jnh5y:
Ok.... I get 20 files when I run that command.  Do the ones where individual tests are annotated work?

elahrvivaz:
Did you run it on this branch? There should be fewer now - all the other tests were running as expected, but I figured it would be good to standardize on specs2. There was one test that I didn't update from straight junit, as it was fairly complex. And there are several in hbase and lambda that get invoked differently to avoid spinning up the embedded clusters multiple times.

PR #2115, opened 2018-12-05
GEOMESA-2484 Add converting of parsed values in AbstractConverter

elahrvivaz:
* Add `convert` method to AbstractConverter
* Add type bound to AbstractConverter to reflect intermediate parsing
* Add FeatureToFeatureConverter for v2 simple feature transforms
* Add SafeClose methods to close/flush multiple objects at once
* Add `single` and `fill` CloseableIterators
  * Moved CloseableIterator implementations into the companion object
  * Moved SelfClosingIterator into separate class file
* Update CopyingInputStream to avoid allocating an extra array on copy

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2116, opened 2018-12-06
GEOMESA-2456 Composite JSON converter

elahrvivaz:
* Allow for converter predicates to use `==`, `<` etc


elahrvivaz:
stacks on top of #2115

PR #2118, opened 2018-12-06
GEOMESA-2487 Update StreamDataStore to use converters2

elahrvivaz:


jnh5y:
Is this a 'new feature'?  Seems like we could do this in 2.2.0, no?

elahrvivaz:
It's not new, but it changes the API some - some downstream projects were manually invoking this class, so I didn't want to force a change on them quite yet.

PR #2119, opened 2018-12-10
GEOMESA-2488 Add try/catch and cql copying in CqlTransformFilter

elahrvivaz:


jnh5y:
I don't suppose there's an easy unit test for this?

elahrvivaz:
I can't think of one... I'm not sure that this will actually fix the issue tbh, as I couldn't reproduce it. But that was the only difference I could see between the old JSimpleFeatureFilter and the new scala impl. I think that copying the filters is probably not even needed, but out of an abundance of caution seemed wise

PR #2128, opened 2018-12-12
GEOMESA-2495 Fix empty column binding in type inference

elahrvivaz:


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/TypeInference.scala
```diff
@@ -282,7 +282,7 @@ object TypeInference {
     * @return
     */
   private def binding(typed: ObjectType): String = typed match {
-    case STRING              => "String"
+    case STRING | null       => "String"
```
jnh5y: String is the most 'general' type, right?

elahrvivaz: yeah everything can be converted to a string

PR #2136, opened 2018-12-14
GEOMESA-2414 Support for specifying attributes used in each index

elahrvivaz:
* Supports customization of index attributes through 'geomesa.indices.enabled' user data
  * For example, to create two Z3 indices on different dates specify `geomesa.indices.enabled='z3:geom:dtg,z3:geom:dtg2'`
* Index implementations are now in geomesa-index-api
  * All back-end specific code is contained in `IndexAdapter` implementations
* New index implementations can be SPI loaded through `GeoMesaFeatureIndexFactory`
* New date-only option for AttributeIndex secondary index tiering
  * Enable with `geomesa.indices.enabled='attr:foo:dtg'`
* The following features have been removed:
  * Accumulo table-sharing for new schemas (existing schemas still supported)
  * Accumulo pre-computed BIN values
  * Accumulo map-aggregating analytic scans
  * Z2/Z3 non-point support from geomesa-1.2.2 to geomesa-1.2.4 (use XZ instead)


elahrvivaz:
This change also opens the door to having redundant index implementations bundled with geomesa, e.g. multiple spatial implementations using h3, hilbert, etc.

docs/user/upgrade.rst
```diff
@@ -90,6 +90,16 @@ Compatibility Matrix
...
 
+Version 2.3.0 Upgrade Guide
++++++++++++++++++++++++++++
+
+Internal Metadata Format Change
+-------------------------------
+
+The first time a schema is accessed with a GeoMesa 2.3.x data store, the metadata will be transparently rewritten.
+After this, the schema will no longer be readable by older versions of GeoMesa. The old metadata values are
+backed up in the catalog table, in case a rollback is required.
```
jnh5y: Interesting.  Is this necessary?

jnh5y: Ok, I think we'll have situations where folks may want/need to read a table with multiple versions of GeoMesa.  

Either I'm misunderstanding or this would completely prevent someone from doing A/B testing like that.  How hard would it be to use a different metadata namespace or something similar?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -194,39 +240,24 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
...
-    }
+    super.onSchemaUpdated(sft, previous)
     // configure the stats combining iterator on the table for this sft
     stats.configureStatCombiner(connector, sft)
   }
 
-  override def getQueryPlan(query: Query,
-                            index: Option[AccumuloFeatureIndexType],
-                            explainer: Explainer): Seq[AccumuloQueryPlan] =
+  override def getQueryPlan(query: Query, index: Option[String], explainer: Explainer): Seq[AccumuloQueryPlan] =
```
jnh5y: Since this is just calling to super, can the call be removed?  

Maybe this is more clear to read?

elahrvivaz: the super method returns the generic `QueryPlan[AccumuloDataStore]`, this specifies that it is an `AccumuloQueryPlan`, which exposes all the bits of the plan like ranges, etc.

docs/user/upgrade.rst
elahrvivaz: reading with different versions of geomesa is generally not going to work perfectly, due to the distributed runtime jar, and we discourage it (although certain queries *may* work). did you have something concrete in mind, or are you just speculating? If someone really wanted to test the same data with two different versions, you could always clone the metadata table and point it at the same set of index tables, but I don't think that's necessarily a standard use case we need to be supporting. And going forward, any schema created with 2.3+ is not going to be readable with 2.2 or earlier anyway.

jnh5y: @jahhulbert-ccri has noted that one can configure different Accumulo contexts (with separate classpaths) at scan time.  It is with that use case in mind that I'm nervous having tables which once read by a newer version cannot be read with a previous version.

elahrvivaz: that seems like a terrible idea haha. but as I said, you could do that if you cloned the metadata table before upgrading.

elahrvivaz: scanning separate classpaths would have to be functionality that we added, right? so seems like a moot point anyway, as it's not in any older versions (or newer ones yet)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
```diff
@@ -0,0 +1,452 @@
...
+      val ranges = byteRanges.map {
+        case BoundedByteRange(start, end) =>
+            val startKey = if (start.length == 0) { null } else { new Key(new Text(start)) }
+            val endKey = if (end.length == 0) { null } else { new Key(new Text(end)) }
+            new Range(startKey, true, endKey, false)
+
+        case SingleRowByteRange(row) =>
+          new Range(new Text(row))
+      }
+      val numThreads = if (index.name == IdIndex.name) { ds.config.recordThreads } else { ds.config.queryThreads }
```
jnh5y: Interesting...  The number of query threads for the recordThreads can be different?!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
```diff
@@ -0,0 +1,452 @@
...
+      val tables = index.getTablesForQuery(filter.filter)
+      val (colFamily, schema) = {
+        val (cf, s) = groups.group(index.sft, hints.getTransformDefinition, ecql)
+        (Some(new Text(AccumuloIndexAdapter.mapColumnFamily(index)(cf))), s)
+      }
+
+      index match {
+        case i: AccumuloJoinIndex =>
+          i.createQueryPlan(filter, tables, ranges, colFamily, schema, ecql, hints, numThreads)
```
jnh5y: Does the Join Index support Arrow/Bin/Heatmaps/Stats?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
```diff
@@ -0,0 +1,452 @@
...
+    * @param returnSft return simple feature type (transform, etc)
+    * @return
+    */
+  def entriesToFeatures(index: GeoMesaFeatureIndex[_, _],
+                        returnSft: SimpleFeatureType): Entry[Key, Value] => SimpleFeature = {
+    // Perform a projecting decode of the simple feature
+    if (index.serializedWithId) {
+      entriesToFeaturesWithId(KryoFeatureSerializer(returnSft))
+    } else {
+      entriesToFeatures(index, KryoFeatureSerializer(returnSft, SerializationOptions.withoutId))
```
jnh5y: Just to remind myself... we cache the KryoFeatureSerializers somehow, right?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
```diff
@@ -63,27 +61,6 @@ class AccumuloGeoMesaStats(val ds: AccumuloDataStore, statsTable: String, val ge
 
   compactor.run() // schedule initial compaction
 
-  override def getCount(sft: SimpleFeatureType, filter: Filter, exact: Boolean): Option[Long] = {
```
jnh5y: Out of curiosity, where did this go?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloJoinIndex.scala
```diff
@@ -0,0 +1,395 @@
...
+  override val name: String = JoinIndex.name
+  override val identifier: String = GeoMesaFeatureIndex.identifier(name, version, attributes)
+
+  abstract override def getFilterStrategy(filter: Filter,
+                                          transform: Option[SimpleFeatureType],
+                                          stats: Option[GeoMesaStats]): Option[FilterStrategy] = {
+    super.getFilterStrategy(filter, transform, stats).flatMap { strategy =>
+      val join = requiresJoin(strategy.secondary, transform)
+      // verify that it's ok to return join plans, and filter them out if not
+      if (join && !JoinIndex.AllowJoinPlans.get) { None } else {
```
jnh5y: The naked call to .get on an Option makes me nervous:).

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
```diff
@@ -88,43 +79,3 @@ abstract class BaseAggregatingIterator[T <: AnyRef { def isEmpty: Boolean; def c
   override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] =
     throw new NotImplementedError()
 }
-
-trait DeduplicatingScan[T <: AnyRef { def isEmpty: Boolean; def clear(): Unit }] extends AggregatingScan[T] {
```
jnh5y: Have we removed the need for this?  Or did it move?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/AccumuloSchemaBuilder.scala
```diff
@@ -110,8 +112,8 @@ object AccumuloSchemaBuilder {
       * @param sharing table sharing
       * @return user data builder for call chaining
       */
-    def tableSharing(sharing: Boolean): AccumuloUserDataBuilder =
-      userData(SimpleFeatureTypes.Configs.TABLE_SHARING_KEY, sharing.toString)
+    @deprecated("table sharing is no longer supported")
+    def tableSharing(sharing: Boolean): AccumuloUserDataBuilder = this
```
jnh5y: Seems like this should log an error / warning, etc.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithDataStore.scala
```diff
@@ -36,9 +36,6 @@ trait TestWithDataStore extends Specification {
   def spec: String
   def dtgField: Option[String] = Some("dtg")
 
-  // TODO GEOMESA-1146 refactor to allow running of tests with table sharing on and off...
```
jnh5y: Can we closed GM-1146 as OBE?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
```diff
@@ -376,25 +371,6 @@ class AccumuloDataStoreQueryTest extends Specification with TestWithMultipleSfts
       (System.currentTimeMillis() - start) must beLessThan(30000L)
     }
 
-    "avoid deduplication when possible" in {
```
jnh5y: Is deduplication all done?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloPartitioningTest.scala
```diff
@@ -33,6 +32,8 @@ import scala.collection.JavaConversions._
 @RunWith(classOf[JUnitRunner])
 class AccumuloPartitioningTest extends Specification with TestWithDataStore {
 
+  // note: using `Seq.foreach; ok` instead of `foreach(Seq)` shaves several seconds off the time to run this test
```
jnh5y: Interesting... any idea why?!

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -0,0 +1,558 @@
...
+    tables.par.foreach { name =>
+      val tableName = TableName.valueOf(name)
+      WithClose(ds.connection.getTable(tableName)) { table =>
+        val scan = new Scan().setFilter(new KeyOnlyFilter)
+        prefix.foreach(scan.setRowPrefixFilter)
+        ds.applySecurity(scan)
+        val mutateParams = new BufferedMutatorParams(tableName)
+        WithClose(table.getScanner(scan), ds.connection.getBufferedMutator(mutateParams)) { case (scanner, mutator) =>
+          scanner.iterator.asScala.grouped(10000).foreach { result =>
+            // TODO set delete visibilities
```
jnh5y: Can you enter a ticket for this?

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseArrowTest.scala
```diff
@@ -18,6 +18,7 @@ import org.geotools.filter.text.ecql.ECQL
 import org.locationtech.geomesa.arrow.io.SimpleFeatureArrowFileReader
 import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.locationtech.geomesa.hbase.data.HBaseDataStoreParams.{ConnectionParam, HBaseCatalogParam}
+import org.locationtech.geomesa.hbase.data.HBaseQueryPlan.CoprocessorPlan
```
jnh5y: Used or not?

geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
```diff
@@ -17,6 +17,7 @@ import org.apache.hadoop.io.Text
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 import org.geotools.data.{Query, Transaction}
+import org.locationtech.geomesa.hbase.data.HBaseQueryPlan.{EmptyPlan, ScanPlan}
```
jnh5y: Used?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z2Filter.scala
```diff
@@ -10,19 +10,15 @@ package org.locationtech.geomesa.index.filters
...
 
 class Z2Filter(val xy: Array[Array[Int]]) {
 
   def inBounds(buf: Array[Byte], offset: Int): Boolean = {
-    val keyZ = Z2Filter.rowToZ(buf, offset)
-    pointInBounds(keyZ)
-  }
-
-  def pointInBounds(z: Long): Boolean = {
+    val z = ByteArrays.readLong(buf, offset)
```
jnh5y: Nice refactoring:).

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
```diff
@@ -110,188 +104,115 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
...
-  }
-}
-
-object GeoMesaFeatureCollection {
-  private val oneUp = new AtomicLong(0)
-  def nextId: String = s"GeoMesaFeatureCollection-${oneUp.getAndIncrement()}"
-}
-
-trait CachingFeatureSource extends GeoMesaFeatureSource {
+  /*_*/
```
jnh5y: What do these comment do?

jnh5y: I mean, for some reason, they make me think of Marvin the robot from Hitchhiker's Guide ...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
```diff
@@ -95,7 +104,7 @@ abstract class InMemoryQueryRunner(stats: GeoMesaStats, authProvider: Option[Aut
   }
 }
 
-object InMemoryQueryRunner {
+object LocalQueryRunner {
```
jnh5y: Did you combine the foo for the 'InMemory' CQ QueryRunner with the stuff necessary for C* and BigTable?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryRunner.scala
```diff
@@ -90,7 +92,17 @@ trait QueryRunner {
     */
   protected [geomesa] def getReturnSft(sft: SimpleFeatureType, hints: Hints): SimpleFeatureType = {
     import org.locationtech.geomesa.index.conf.QueryHints.RichHints
-    hints.getTransformSchema.getOrElse(sft)
+    if (hints.isBinQuery) {
```
jnh5y: Does this mean that *all* the Bin/Arrow/Stats/Density processing is handled in the GeoMesa DataStore QueryRunner?

If so, does this change any of the GeoMesa WPS Processes?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
elahrvivaz: yeah, we've had this for a long time... it's used for id lookups and join queries - since each scan to the id index is a single row, the thread to performance ratio is different than against say the z3 index. I forget now if it's better to have it higher or lower, but I believe lower works better.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
elahrvivaz: yep!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
elahrvivaz: we cache parts of them - the functions to read/write out of the kryo stream based on the simple feature type, the kryo input/output buffers... but we generally don't cache serializer instances. I'm not sure how much benefit we would get, as they're generally instantiated once per scan, and they're not super heavy-weight.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
elahrvivaz: The implementation was in the super class already, but this had some special handling for duplicate rows. Since I remove support for the non-point Z2/Z3 indices, we don't really need the duplicate handling anymore, since none of our indices have duplicates.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/index/AccumuloJoinIndex.scala
elahrvivaz: that's actually a thread-local, not an option :)

elahrvivaz: with a default value!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
elahrvivaz: Yeah, I removed the de-duplication logic, as it was meant for the non-point z2/z3 indices that I dropped support for.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/AccumuloSchemaBuilder.scala
elahrvivaz: added

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/TestWithDataStore.scala
elahrvivaz: done

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
elahrvivaz: yeah, it was for the non-point z3/z2 indices

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloPartitioningTest.scala
elahrvivaz: maybe because it builds up complex matchers and tracks the results?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/TableSharingTest.scala
```diff
@@ -1,176 +0,0 @@
...
-import org.locationtech.geomesa.accumulo.iterators.TestData._
-import org.locationtech.geomesa.utils.collection.SelfClosingIterator
-import org.opengis.filter._
-import org.specs2.mutable.Specification
-import org.specs2.runner.JUnitRunner
-
-import scala.collection.JavaConversions._
-
-@RunWith(classOf[JUnitRunner])
-class TableSharingTest extends Specification with LazyLogging {
```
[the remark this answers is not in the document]
elahrvivaz: yeah, most of the BackCompatibilityTest versions have table sharing enabled (as it was the default for a while)

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexValuesTest.scala
```diff
@@ -15,6 +15,7 @@ import org.geotools.data.{Query, Transaction}
 import org.geotools.filter.text.ecql.ECQL
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.accumulo.TestWithDataStore
+import org.locationtech.geomesa.accumulo.data.AccumuloQueryPlan.BatchScanPlan
```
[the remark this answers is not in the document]
elahrvivaz: No, I moved the query plan implementing classes out of the top-level package (where they used to be in scope without an import) and into the companion object

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/index/AttributeIndexJob.scala
```diff
@@ -152,51 +155,72 @@ class AttributeMapper extends Mapper[Text, SimpleFeature, Text, Mutation] {
 
   private var counter: Counter = _
 
-  private var writer: AccumuloFeature => Seq[Mutation] = _
-  private var toWritable: SimpleFeature => AccumuloFeature = _
-  private var table: SimpleFeature => Text = _
+  // TODO have a standardized writer that returns mutations instead of using a batch writer
```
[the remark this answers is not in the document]
elahrvivaz: added

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
elahrvivaz: added

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseArrowTest.scala
elahrvivaz: used, as noted above

geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
elahrvivaz: used, as noted above

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
elahrvivaz: haha, nice. they disable intelij's syntax highlighting - this snippet for some reason always shows up as an error if you don't disable it.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
elahrvivaz: Yeah, this should be an easy way to get all our query handling/filtering/transforming against any random iterator of features. And since it's all in one place, we save a lot of redundancy.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryRunner.scala
elahrvivaz: previously not all our data stores supported all the different query types we do, so this was overridden per data store. but now they all support all our query types, so the logic was consolidated here.
It shouldn't affect any of our wps processes - possibly you will be able to do some wps processes against data stores that you weren't able to before, although most of our processes check to see if the query was handled and if not they do it locally.

PR #2138, opened 2018-12-17
GEOMESA-2500 Add support to the KDS for reading confluent avro-serialized topics

dseapy:
* Infers an SFT based on the avro schema stored in the Confluent Avro Schema Registry
* Reads the simple features into the KDS


build/cqs.tsv
```diff
@@ -248,6 +248,8 @@ tec.uom:uom-se	1.0.8	compile
 com.101tec:zkclient	0.10	provided
 com.datastax.cassandra:cassandra-driver-core	3.0.0	provided
 com.datastax.cassandra:cassandra-driver-mapping	3.0.0	provided
+io.confluent:kafka-avro-serializer	5.1.0	provided
```
dseapy: I had this at 5.0.0, but just bumped it to 5.1.0.  License was apache 2.0, but 5.1.0+ is not.  Do I need to leave this at 5.0.0 and allow people to provide 5.1.0?  Or does this dependency need to be removed since we might want to bump kafka in the future without an old confluent potentially breaking the build?

https://twitter.com/nehanarkhede/status/1073594002945929218

dseapy: Actually this one is apache 2:
https://github.com/confluentinc/schema-registry/blob/master/avro-serializer/LICENSE

The 2 below are not:
https://github.com/confluentinc/schema-registry/blob/master/core/LICENSE

As they just broke out the client, I will want to change to that and then we should be good to roll with 5.1.0:
https://github.com/confluentinc/schema-registry/blob/master/client/LICENSE

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/ConfluentMetadata.scala
```diff
@@ -0,0 +1,129 @@
...
+object ConfluentMetadata extends LazyLogging {
+  val SUBJECT_POSTFIX = "-value"
+
+  def schemaToSft(schema: Schema, sftName: String): SimpleFeatureType = {
+    val builder = new SimpleFeatureTypeBuilder
+    builder.setName(sftName)
+    builder.setDefaultGeometry(ConfluentFeatureSerializer.geomAttributeName)
+    builder.add(ConfluentFeatureSerializer.geomAttributeName, classOf[Geometry])
+    builder.add(ConfluentFeatureSerializer.dateAttributeName, classOf[Date])
+    schema.getFields.asScala.foreach(addSchemaToBuilder(builder, _))
```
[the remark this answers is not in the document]
dseapy: On second thought any sft overrides should likely go somewhere other than the registry or the topic.  Registry schema should probably only care about people being able to read the avro data and not be involved with how geomesa ultimately stores/indexes that data (which could be different for different readers).

build/cqs.tsv
elahrvivaz: It seems like the confluent license is fairly permissive. @jnh5y do you think it would pass muster with eclipse?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
```diff
@@ -53,7 +57,11 @@ class KafkaDataStore(val config: KafkaDataStoreConfig)
   override protected def catalog: String = config.catalog
 
   override val metadata: GeoMesaMetadata[String] =
-    new ZookeeperMetadata(s"${config.catalog}/$MetadataPath", config.zookeepers, MetadataStringSerializer)
+    if (config.serialization == SerializationType.CONFLUENT && config.schemaRegistryUrl.isDefined) {
```
elahrvivaz: is there any case where we would want schema registry, but not confluent serialization, or vice versa? maybe we could simplify the config by assuming that a schema registry url implies confluent serialization.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/ConfluentMetadata.scala
```diff
@@ -0,0 +1,129 @@
...
+import org.locationtech.geomesa.kafka.utils.ConfluentMetadata.{schemaToSft, SUBJECT_POSTFIX}
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
+import org.opengis.feature.simple.SimpleFeatureType
+
+import scala.collection.JavaConverters._
+import scala.util.control.NonFatal
+
+class ConfluentMetadata(val schemaRegistry: SchemaRegistryClient) extends GeoMesaMetadata[String] with LazyLogging {
+
+  val topicSftCache: LoadingCache[String, String] =
```
elahrvivaz: could we make this private? generally it makes things simpler if you don't expose things that aren't needed externally.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/ConfluentMetadata.scala
```diff
@@ -0,0 +1,129 @@
...
+    builder.add(ConfluentFeatureSerializer.dateAttributeName, classOf[Date])
+    schema.getFields.asScala.foreach(addSchemaToBuilder(builder, _))
+    builder.buildFeatureType()
+  }
+
+  def addSchemaToBuilder(builder: SimpleFeatureTypeBuilder,
+                         field: Schema.Field,
+                         typeOverride: Option[Schema.Type] = None): Unit = {
+    typeOverride.getOrElse(field.schema().getType) match {
+      case STRING => builder.add(field.name(), classOf[java.lang.String])
```
elahrvivaz: [formatting] align the case statements

geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/RecordVersions.scala
```diff
@@ -54,6 +56,15 @@ object RecordVersions extends LazyLogging {
     }
   }
 
+  private val _getTimestamp: ConsumerRecord[_, _] => Option[Long] = {
+    consumerMethods.find(m => m.getName == "timestamp" && m.getParameterCount == 0).map{ method =>
+      record: ConsumerRecord[_,_] => Option(method.invoke(record).asInstanceOf[java.lang.Long]).map(Long2long)
```
elahrvivaz: I think the timestamp returns a primitive Long (at least in 2.0), so you don't need to cast to java.lang.Long and unbox

geomesa-features/geomesa-feature-confluent/src/main/scala/org/locationtech/geomesa/features/confluent/ConfluentFeatureSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+class ConfluentFeatureSerializer(sft: SimpleFeatureType,
+                                 schemaRegistryClient: SchemaRegistryClient,
+                                 val options: Set[SerializationOption] = Set.empty)
+    extends SimpleFeatureSerializer with LazyLogging {
+
+  private val kafkaAvroDeserializer = new KafkaAvroDeserializer(schemaRegistryClient)
+  private val wktReader = new WKTReader()
+  var geomSrcAttributeName: Option[String] = None
+
+  override def deserialize(id: String, bytes: Array[Byte], timestamp: Option[Long]): SimpleFeature = {
```
elahrvivaz: I don't really like adding the timestamp to the serializer API just for this one use case - would it be possible to handle it in GeoMessageSerializer instead? So this would deserialize all the actual avro fields, and then GeoMessageSerializer could set the date based on the timestamp.

geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/RecordVersions.scala
elahrvivaz: Actually, could we make this just return `Long` and not `Option[Long]`, and return `System.currentTimeMillis` for unsupported versions? That would match the previous behavior, and we don't *really* care about old kafka versions enough to handle missing timestamps I'd say

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
```diff
@@ -128,7 +132,10 @@ object KafkaCacheLoader {
     }
 
     override protected [KafkaCacheLoader] def consume(record: ConsumerRecord[Array[Byte], Array[Byte]]): Unit = {
-      val message = serializer.deserialize(record.key(), record.value(), RecordVersions.getHeaders(record))
+      val message = serializer.deserialize(record.key(),
+                                           record.value(),
+                                           RecordVersions.getHeaders(record),
+                                           RecordVersions.getTimestamp(record))
       val timestamp = try { record.timestamp() } catch { case _: NoSuchMethodError => System.currentTimeMillis() }
```
elahrvivaz: timestamp comment (below) applies here too

geomesa-features/geomesa-feature-confluent/src/main/scala/org/locationtech/geomesa/features/confluent/ConfluentFeatureSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+  val geomAttributeName = "_geom"
+  val dateAttributeName = "_date"
+}
+
+class ConfluentFeatureSerializer(sft: SimpleFeatureType,
+                                 schemaRegistryClient: SchemaRegistryClient,
+                                 val options: Set[SerializationOption] = Set.empty)
+    extends SimpleFeatureSerializer with LazyLogging {
+
+  private val kafkaAvroDeserializer = new KafkaAvroDeserializer(schemaRegistryClient)
```
elahrvivaz: do you know if `KafkaAvroDeserializer` is thread safe? generally we try to make our serializers thread safe, so we might want to wrap this in a ThreadLocal or object pool

geomesa-features/geomesa-feature-confluent/src/main/scala/org/locationtech/geomesa/features/confluent/ConfluentFeatureSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+  private val wktReader = new WKTReader()
+  var geomSrcAttributeName: Option[String] = None
```
elahrvivaz: this should probably be private

elahrvivaz:
looks good, thanks for putting it together!

elahrvivaz:
I'll wait to open CQs until you make the change to confluent-client

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
dseapy: Are you suggesting omit SerializationType.CONFLUENT?

elahrvivaz: no, just to ignore the serialization type that is configured if a confluent url is specified, and always use CONFLUENT as the type in that case

docs/user/kafka/usage.rst
```diff
@@ -65,6 +66,8 @@ Parameter                            Type    Description
                                              See :ref:`kafka_ssi`
 ``kafka.serialization.lazy``         Boolean Use lazy deserialization of features. This may improve processing load at
                                              the expense of slightly slower query times
+``kafka.schema.registry.url``        String  URL to a confluent schema registry server, used to read Confluent schemas for kafka topics
+                                             serialized with the Confluent avro serializer. Experimental, read-only, record timestamps required.
```
elahrvivaz: I believe timestamps were included in 0.10, right? Could we update this to be 'requires Kafka 0.10 or later' instead of 'record timestamps required'? Otherwise it's not clear what is actually needed. Same thing above with the serialization type.

elahrvivaz: Actually I guess we are defaulting to system time now if there is no timestamp, so we can probably remove the bit about timestamps entirely

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreFactory.scala
```diff
@@ -112,7 +114,17 @@ object KafkaDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogging {
       KafkaDataStore.ProducerConfig(props)
     }
 
-    val serialization = org.locationtech.geomesa.features.SerializationType.withName(SerializationType.lookup(params))
+    val schemaRegistryUrl = SchemaRegistryUrl.lookupOpt(params)
+
+    val serialization = {
+      val ser = org.locationtech.geomesa.features.SerializationType.withName(SerializationType.lookup(params))
```
elahrvivaz: We could tweak this a little so that it doesn't use the default value (`kryo`) if nothing is specified. Like:
```
val ser = if (schemaRegistryUrl.isDefined && !params.containsKey(SerializationType.key)) {
  org.locationtech.geomesa.features.SerializationType.CONFLUENT
} else {
  org.locationtech.geomesa.features.SerializationType.withName(SerializationType.lookup(params))
}
```
(doing this from memory so code may not be exactly right)

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/ConfluentMetadata.scala
```diff
@@ -0,0 +1,129 @@
...
+    throw new NotImplementedError(s"ConfluentMetadata only supports ATTRIBUTES_KEY")
+
+  override def insert(typeName: String, key: String, value: String): Unit = {}
+  override def insert(typeName: String, kvPairs: Map[String, String]): Unit = {}
+  override def remove(typeName: String, key: String): Unit = {}
+  override def delete(typeName: String): Unit = {}
+}
+
+object ConfluentMetadata extends LazyLogging {
+  val SUBJECT_POSTFIX = "-value"
```
elahrvivaz: Could you add a comment here on why we are looking for just topics ending in '-value'? Is that a confluent standard?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/GeoMessageSerializer.scala
```diff
@@ -66,6 +69,7 @@ object GeoMessageSerializer {
 
   val KryoVersion: Byte = 2
   val AvroVersion: Byte = 3
+  val ConfluentVersion: Byte = 3
```
elahrvivaz: This isn't used is it? could be removed

elahrvivaz:
@ddseapy is the dependency list finalized? I'll go ahead and open up CQs.

dseapy:
Thanks for the comments, i'll work on those now.  The dependency list should be done (assuming test-scope for 5.1.0 is ok).  Still haven't retested though (other than the unit test).

Actually maybe i should change avro version to .1: https://github.com/confluentinc/common/blob/master/pom.xml#L48

elahrvivaz:
Looks like it breaks recursive schemas: https://issues.apache.org/jira/browse/AVRO-2122
I'll leave it up to you if that outweighs whatever other fixes are in .2.

elahrvivaz:
there's 108 tickets against 1.8.1 and 139 against 1.8.2

dseapy:
I think it matters more for the rest of geomesa, than for my confluent pr.  So maybe you should decide :)

elahrvivaz:
haha, well I'd say go with the latest I guess. @jahhulbert-ccri any opinions on avro version?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/ConfluentMetadata.scala
dseapy: Yeah, i'll definitely add a comment.  Before I do, do you know if there is a need to support more?  It's default behavior of the serializer/deserializer, also how ksql registers all schemas it generates.
For my use-cases it's sufficient.

dseapy: I'll probably just add in the comment that we can support more in the future if needed

elahrvivaz: I don't know - I think if it supports your use case then that is good enough for now

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/GeoMessageSerializer.scala
```diff
@@ -89,6 +94,13 @@ object GeoMessageSerializer {
     val (serializer, version) = serialization match {
       case SerializationType.KRYO => (kryoSerializer, KryoVersion)
       case SerializationType.AVRO => (avroSerializer, AvroVersion)
+      case SerializationType.CONFLUENT =>
+        val confluentBuilder = schemaRegistryUrl match {
+          case Some(srUrl) => ConfluentFeatureSerializer.builder(sft, srUrl).withoutId.withUserData.immutable
+          case None =>
+            throw new IllegalArgumentException("Confluent serialization used, however no schema-registry url provided.")
+        }
+        (confluentBuilder.build(), ConfluentVersion)
```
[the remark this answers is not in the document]
elahrvivaz: How about we just make it -1 to indicate that it is not used? The deserializer matches on the header number to determine the serialization type, so we would want it to be unique (i.e. not 1, 2 or 3). Since confluent is read-only for now, it doesn't really matter, but would help avoid confusion.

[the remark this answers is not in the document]
dseapy: Sounds good, thanks.

build/cqs.tsv
```diff
@@ -126,7 +126,7 @@ org.apache.accumulo:accumulo-start	1.8.1	compile
 org.apache.arrow:arrow-format	0.10.0	compile
 org.apache.arrow:arrow-memory	0.10.0	compile
 org.apache.arrow:arrow-vector	0.10.0	compile
-org.apache.avro:avro	1.7.5	compile
+org.apache.avro:avro	1.8.2	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18582

build/cqs.tsv
```diff
@@ -248,6 +248,8 @@ tec.uom:uom-se	1.0.8	compile
 com.101tec:zkclient	0.10	provided
 com.datastax.cassandra:cassandra-driver-core	3.0.0	provided
 com.datastax.cassandra:cassandra-driver-mapping	3.0.0	provided
+io.confluent:kafka-avro-serializer	5.1.0	provided
+io.confluent:kafka-schema-registry-client	5.1.0	provided
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18584

build/cqs.tsv
```diff
@@ -259,8 +261,8 @@ org.apache.hadoop:hadoop-mapreduce-client-common	2.6.4	provided
 org.apache.hadoop:hadoop-mapreduce-client-core	2.6.4	provided
 org.apache.hadoop:hadoop-yarn-api	2.6.4	provided
 org.apache.hadoop:hadoop-yarn-common	2.6.4	provided
-org.apache.kafka:kafka_2.11	2.0.0	provided
-org.apache.kafka:kafka-clients	2.0.0	provided
+org.apache.kafka:kafka_2.11	2.1.0	provided
+org.apache.kafka:kafka-clients	2.1.0	provided
```
elahrvivaz: covered by our 'works with' cq: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=9082

build/cqs.tsv
```diff
@@ -270,6 +272,8 @@ org.springframework.security:spring-security-core	4.2.7.RELEASE	provided
 org.springframework.security:spring-security-web	4.2.7.RELEASE	provided
 org.springframework:spring-web	4.3.18.RELEASE	provided
 
+io.confluent:kafka-schema-registry	5.1.0	test
+io.confluent:kafka-schema-registry	tests:5.1.0	test
```
elahrvivaz: @jnh5y we don't need to do anything for these test scope dependencies, right?

jnh5y: Yeah, test scope we should be able to skip on doing anything.

jnh5y:
@jahhulbert-ccri thoughts on upgrading to Avro 1.8?

elahrvivaz:
@ddseapy is avro 1.8 required for compile? Or could we make it an install step to swap the jar in geoserver?
@jahhulbert-ccri is concerned with upgrading, since hadoop 2.x bundles avro 1.7.

dseapy:
Not required for compile, but needed for runtime and unit test (java.lang.ClassNotFoundException: org.apache.avro.SchemaValidationException).

Is there a good way to have the unit test be against 1.8?
https://stackoverflow.com/questions/6575742/maven-2-different-dependency-versions-in-test-and-compile#comment30375891_6575863

jnh5y:
@elahrvivaz are you in a position to get this across the finish line?

elahrvivaz:
it's on my list but low-priority at the moment

jnh5y:
Roger wilco.

elahrvivaz:
replacing with #2240

PR #2155, opened 2019-01-07
GEOMESA-2494 Query re-writing

elahrvivaz:
* QueryInterceptors can be defined in the sft user data using
  `geomesa.query.interceptors`, and must be available on the classpath


elahrvivaz:
note: stacks on top of #2136

jnh5y:
Needs to be rebased.

elahrvivaz:
rebased

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
```diff
@@ -0,0 +1,122 @@
...
+import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.Configs
+import org.locationtech.geomesa.utils.io.CloseWithLogging
+import org.opengis.feature.simple.SimpleFeatureType
+
+import scala.util.control.NonFatal
+
+/**
+  * Provides a hook to modify a query before executing it
+  */
+trait QueryInterceptor extends Closeable {
```
jnh5y: Do we expected folks to implement this from Java or Scala?  (I didn't know if we had a strong preference for Java interop or not.)

elahrvivaz: i think this class actually would be easily extensible in java. In general I would say we can use scala for most things... has anyone actually ever extended any of our java apis?

docs/user/datastores/index_config.rst
```diff
@@ -559,6 +559,50 @@ Full Example
...
+Configuring Query Interceptors
+------------------------------
+
+GeoMesa provides a chance for custom logic to be applied to a query before executing it. Query interceptors must
+be specified through user data in the simple feature type, and may be set before calling ``createSchema``, or
+updated by calling ``updateSchema``. To indicate query interceptors, use the key ``geomesa.query.interceptors``:
+
+.. code-block:: java
+
+    sft.getUserData().put("geomesa.query.interceptors", "com.example.MyQueryInterceptor");
```
jnh5y: Since this key is plural... is it possible to set multiple interceptors?!

elahrvivaz: yep! comma-separated

elahrvivaz: see next line haha :)

jnh5y: Dang!  I was trying to maintain the illusion that I read these PRs...

jnh5y: Ok, Is the order of operations clear?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
```diff
@@ -68,5 +71,32 @@ class GeoMesaDataStoreTest extends Specification {
...
+      results must beEmpty
+
+      // other queries should go through as normal
+      results = SelfClosingIterator(ds.getFeatureReader(new Query(sft.getTypeName, ECQL.toFilter("bbox(geom,39,54,51,56)")), Transaction.AUTO_COMMIT)).toSeq
+      results must haveLength(10)
+    }
+  }
+}
+
+object GeoMesaDataStoreTest {
```
jnh5y: How hard would it be to add a unit test demonstrating multiple query interceptors which clearly shows the order of operations?

docs/user/datastores/index_config.rst
elahrvivaz: ah, possibly not in the docs - should it be clarified? they are executed in the order they are declared.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryRunner.scala
```diff
@@ -40,6 +56,9 @@ trait QueryRunner {
 
     val query = new Query(original) // note: this ends up sharing a hints object between the two queries
 
+    // query rewriting
+    interceptors(sft).foreach(_.rewrite(query))
```
jnh5y: Do we indicate if a query has been re-written?  I could imagine it being VERY helpful to list the re-writers used.  Potentially logging the query after each rewrite...  (Not sold on that, just spitballing.)

docs/user/datastores/index_config.rst
elahrvivaz: added a note

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryRunner.scala
elahrvivaz: the original query gets logged, and the query that gets executed gets logged, so you can see the delta by comparing them. There isn't an explicit message on whether it changed or not. The query interceptors are defined in the sft user data, so i don't think we need to log them each time...

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
elahrvivaz: probably not too hard, but it's a `.foreach` so dunno if we need to check it so rigorously :)

elahrvivaz: https://github.com/locationtech/geomesa/pull/2155/files#diff-70d7b065fbfde37dcee7f8ab3966d3cdR102

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryRunner.scala
elahrvivaz: oh you mean log the query, i misunderstood

elahrvivaz: could leave that logging up to the interceptors... i could see it both ways, but i feel like it might be kind of verbose to log the query each time, especially if it doesn't actually change - so we'd want to check for that too?

jnh5y: Yeah, I know I'm thinking of something verbose...

PR #2157, opened 2019-01-07
GEOMESA-2513 Catch errors in deserialization of corrupt data

elahrvivaz:

### DIFF
--- a/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/TwkbSerialization.scala
+++ b/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/TwkbSerialization.scala
@@ -8,8 +8,11 @@
 
 package org.locationtech.geomesa.features.serialization
 
-import org.locationtech.jts.geom._
+import com.typesafe.scalalogging.LazyLogging
 import org.locationtech.geomesa.utils.geometry.GeometryPrecision.TwkbPrecision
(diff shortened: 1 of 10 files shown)


geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/TwkbSerialization.scala
```diff
@@ -48,7 +51,7 @@ trait TwkbSerialization[T <: NumericWriter, V <: NumericReader]
         val coord = geometry.getCoordinate
         // check for dimensions - use NaN != NaN to verify z coordinate
```
jnh5y: Update comment?

jnh5y:
@elahrvivaz have you checked to see if the try-catchs are slowing things done?

elahrvivaz:
trys that don't throw should be essentially free: https://stackoverflow.com/questions/16451777/is-it-expensive-to-use-try-catch-blocks-even-if-an-exception-is-never-thrown

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/TwkbSerialization.scala
elahrvivaz: oops, sorry didn't see this. I think it's still somewhat valid as that's what `Double.isNaN` does anyway

PR #2162, opened 2019-01-08
GEOMESA-2515 Add -p to log dir creation

jahhulbert-ccri:

### DIFF
--- a/geomesa-tools/bin/common-functions.sh
+++ b/geomesa-tools/bin/common-functions.sh
@@ -17,7 +17,7 @@ function setGeoLog() {
     export GEOMESA_OPTS="-Dgeomesa.log.dir=${GEOMESA_LOG_DIR} ${GEOMESA_OPTS}"
   fi
   if [[ ! -d "${GEOMESA_LOG_DIR}" ]]; then
-    mkdir "${GEOMESA_LOG_DIR}"
+    mkdir -p "${GEOMESA_LOG_DIR}"
   fi
 }
 


elahrvivaz:
@jahhulbert-ccri do you want this back-ported to 2.1/2.2?

jahhulbert-ccri:
@elahrvivaz i will probably want to backport this to 2.1.x as well. I think that @amcdonaldccri had some other things that I might add before merging.

jahhulbert-ccri:
oh yeah to 2.2 as well. So the use case here is supporting multiple users with access to the tools which are installed as an RPM. The first user to write to the log wins and the other users can't delete it. So we are going to make it based on the $USER var in linux by setting GEOMESA_LOG_DIR in the env. Perhaps we should do that by default in the tools somehow as well.

elahrvivaz:
I think putting the log into the user's home dir by default might be confusing... I know in general I'm often struggling to find where (non-geomesa) log files end up. Seems like we have a pretty good ability to specify it per-user though, so I like the current approach.

geomesa-tools/bin/common-functions.sh
```diff
@@ -17,7 +17,7 @@ function setGeoLog() {
     export GEOMESA_OPTS="-Dgeomesa.log.dir=${GEOMESA_LOG_DIR} ${GEOMESA_OPTS}"
```
amcdonaldccri: The export GEOMESA_OPTS needs to be set

elahrvivaz: is there another change that needs to be made? this pr was already merged

PR #2163, opened 2019-01-09
GEOMESA-2517 HBase - fix spark-sql projections with remote-filtering disabled

misslzyan:
…nstance to deserialize the encoded byte data

fix bug :
  when use spark-sql specified the fields   for querying data from hbase data ,the wrong deserializer can't deserialize the encoded byte data .
  change it with ProjectingKryoFeatureDeserializer . fix it .

PR #2164, opened 2019-01-09
GEOMESA-2517 hbase-datastore: spark-sql with the wrong deserializer i…

misslzyan:
…nstance to deserialize the encoded byte data

fix bug :
  when use spark-sql specified the fields   for querying data from hbase data ,the wrong deserializer can't deserialize the encoded byte data .
  change it with ProjectingKryoFeatureDeserializer . fix it .


elahrvivaz:
closing in favor of #2163

PR #2165, opened 2019-01-10
GEOMESA-2518 Add fast temporal filter implementations

elahrvivaz:
* FastDuring, FastAfter and FastBefore
* GeoTools temporal filters try to convert every value into
  a Period, which fails every time for our common use cases


elahrvivaz:
After adding `FastConverter` I noticed a bunch of warnings in the build log:
```
WARN FastConverter$: Could not convert Sun Jan 05 00:00:01 UTC 2014 (of type java.util.Date) to org.opengis.temporal.Period
```
Traced it down to: https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/filter/temporal/BinaryTemporalOperatorImpl.java#L71

jnh5y:
@elahrvivaz I'm confused by the FastConverter warning.  Is there a change that should be made to GeoMesa or GeoTools relative to this case?

Generally, should we propose changes to the GeoTools converters?

elahrvivaz:
I assume there are some reasons that they have it the way it is. In our case, we know the sft and are optimizing for the 99% case for filtering on a Date.

PR #2171, opened 2019-01-16
GEOMESA-2521 Handle null TypeName in Query from GS querylayer extension

tkunicki:
simple fix, not sure on use of mutation on Query instance, but consistent w/ other mutations in rest of implementation

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
@@ -86,12 +86,15 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
     getFeatures(new Query(sft.getTypeName, filter))
 
   override def getFeatures(query: Query): SimpleFeatureCollection = {
-    // GeoServer querylayer extension generates query instances w/o type name set
-    if(query.getTypeName == null) {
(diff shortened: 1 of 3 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
```diff
@@ -85,7 +85,14 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
   override def getFeatures(filter: Filter): SimpleFeatureCollection =
     getFeatures(new Query(sft.getTypeName, filter))
 
-  override def getFeatures(query: Query): SimpleFeatureCollection = collection(query, this)
+  override def getFeatures(query: Query): SimpleFeatureCollection = {
+    // GeoServer querylayer extension generates query instances w/o type name set
+    if(query.getTypeName == null) {
+      logger.warn(s"Received Query with null TypeName, setting to ${sft.getTypeName}")
```
elahrvivaz: do we want to log this as a warning, or is it expected behavior?

tkunicki: it's expected after looking at [this](https://github.com/geoserver/geoserver/blob/master/src/extension/querylayer/src/main/java/org/geoserver/filter/function/QueryFunction.java#L118) but i don't know if it's typically expected.  no strong opinion on whether to keep or not...

PR #2189, opened 2019-01-23
GEOMESA-2530 Add fast comparison filters

elahrvivaz:
* Default geotools CompareFilterImpl performs lots of instanceof checks
  that make it extremely slow

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyIsEqualTo.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyIsEqualTo.scala
@@ -23,6 +23,18 @@ abstract class FastPropertyIsEqualTo(exp1: Expression, exp2: Literal) extends Pr
   override def getMatchAction: MatchAction = MatchAction.ANY
 
   override def toString: String = s"[ $exp1 = $exp2 ]"
+
+  def canEqual(other: Any): Boolean = other.isInstanceOf[FastPropertyIsEqualTo]
+
+  override def equals(other: Any): Boolean = other match {
+    case that: FastPropertyIsEqualTo =>
+      (that canEqual this) && exp1 == that.getExpression1 && exp2 == that.getExpression2 &&
+          isMatchingCase == that.isMatchingCase
+    case _ => false
+  }
+
+  override def hashCode(): Int =
+    Seq(exp1, exp2, isMatchingCase).map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
 }
 
 object FastPropertyIsEqualTo {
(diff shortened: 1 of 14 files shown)


elahrvivaz:
note: stacks on top of #2165

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyIsEqualTo.scala
```diff
@@ -23,6 +23,18 @@ abstract class FastPropertyIsEqualTo(exp1: Expression, exp2: Literal) extends Pr
   override def getMatchAction: MatchAction = MatchAction.ANY
 
   override def toString: String = s"[ $exp1 = $exp2 ]"
+
+  def canEqual(other: Any): Boolean = other.isInstanceOf[FastPropertyIsEqualTo]
+
+  override def equals(other: Any): Boolean = other match {
```
jnh5y: Errm... why do we have to implement equals and hashCode?

Is it because we've got our own type?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyIsEqualTo.scala
```diff
@@ -23,6 +23,18 @@ abstract class FastPropertyIsEqualTo(exp1: Expression, exp2: Literal) extends Pr
...
+    case that: FastPropertyIsEqualTo =>
+      (that canEqual this) && exp1 == that.getExpression1 && exp2 == that.getExpression2 &&
```
jnh5y: Should we call exp1 == ... or exp1.equals(...)?  (Me just asking silly questions that I cannot remember the answer to.)

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyIsEqualTo.scala
elahrvivaz: makes unit tests pass... may be other benefits i guess

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastPropertyIsEqualTo.scala
elahrvivaz: `==` in scala is equivalent to java `.equals`, although it's also null safe

elahrvivaz:
link to the geotools filter silliness we're replacing: https://github.com/geotools/geotools/blob/20.2/modules/library/main/src/main/java/org/geotools/filter/CompareFilterImpl.java#L84

PR #2194, opened 2019-01-25
GEOMESA-2532 fix bug: user-data not support Chinese character

misslzyan:
Signed-off-by: 段卫东 <misslzyan108@gmail.com>

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/StorageMetadata.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/StorageMetadata.scala
@@ -349,7 +349,7 @@ object StorageMetadata extends MethodProfiling with LazyLogging {
     }
     profile("Persisted storage configuration") {
       WithClose(fc.create(file, java.util.EnumSet.of(CreateFlag.CREATE), CreateOpts.createParent)) { out =>
-        out.writeBytes(data)
+        out.write(data.getBytes("UTF-8"))
         out.hflush()
         out.hsync()
       }
@@ -368,7 +368,7 @@ object StorageMetadata extends MethodProfiling with LazyLogging {
     val file = new Path(root, MetadataPath)
     if (!PathCache.exists(fc, file)) { None } else {
       val config = profile("Loaded storage configuration") {
-        WithClose(new InputStreamReader(fc.open(file))) { in =>
+        WithClose(new InputStreamReader(fc.open(file), "UTF-8")) { in =>
           ConfigFactory.parseReader(in, ConfigParseOptions.defaults().setSyntax(ConfigSyntax.JSON))
         }
       }
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/FileMetadataTest.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/FileMetadataTest.scala
@@ -30,7 +30,7 @@ class FileMetadataTest extends Specification with AllExpectations {
   import scala.collection.JavaConverters._
 
   lazy val fc = FileContext.getFileContext(new Configuration())
-  val sft = SimpleFeatureTypes.createType("metadata", "name:String,dtg:Date,*geom:Point:srid=4326")
+  val sft = SimpleFeatureTypes.createType("metadata", "name:String,dtg:Date,*geom:Point:srid=4326;geomesa.user-data.prefix=desc,desc.name=姓名,desc.dtg=ひづけ,desc.geom=좌표")
   val encoding = "parquet"
   val scheme = PartitionScheme(sft, "hourly,z2-2bit", Collections.emptyMap())
 
@@ -57,6 +57,7 @@ class FileMetadataTest extends Specification with AllExpectations {
         foreach(Seq(created, loaded)) { metadata =>
           metadata.getEncoding mustEqual encoding
           metadata.getSchema mustEqual sft
+          metadata.getSchema.getUserData mustEqual sft.getUserData
           metadata.getPartitionScheme mustEqual scheme
           metadata.getPartitions.asScala.map(_.name) must containTheSameElementsAs(Seq("1", "2"))
           metadata.getPartition("1").files.asScala must containTheSameElementsAs((1 to 3).map(i => s"file$i"))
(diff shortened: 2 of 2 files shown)


misslzyan:
@elahrvivaz  should UTF-8  cover all   demands   ?  UTF-8 is suitable for me , but I am afraid of  other's 
demands and scalability , so I make a configuration for charset .

misslzyan:
another problem .  
/home/travis/build/locationtech/geomesa/geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemDataStoreFactory.scala:41: missing parameter type for expanded function ((x$3) => CharsetParam.key.$minus$greater(x$3))
[error]       .map(_ => (CharsetParam.key -> _))
[error]                                      ^
[error] /home/travis/build/locationtech/geomesa/geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemDataStoreFactory.scala:42: type mismatch;
[error]  found   : Any
[error]  required: String
[error]       .foreach { case (k, v) => conf.set(k, v) }
[error]                                          ^
[error] /home/travis/build/locationtech/geomesa/geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemDataStoreFactory.scala:42: type mismatch;
[error]  found   : Any
[error]  required: String
[error]       .foreach { case (k, v) => conf.set(k, v) }
[error]                                             ^
[error] three errors found

I compiled this code on my mac  and successfully . Why  it  comes  wrong  on the Travis CI build ?

misslzyan:
@elahrvivaz  Thank you for your reply! I will change the default charset to utf-8 ,and can you accept that ?

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemDataStoreFactory.scala
```diff
@@ -37,7 +37,8 @@ class FileSystemDataStoreFactory extends DataStoreFactorySpi {
       props.asScala.foreach { case (k, v) => conf.set(k, v) }
       conf
     }.getOrElse(configuration)
-
+    CharsetParam.lookupOpt(params)
+      .map(v => conf.set(CharsetParam.key,v))
```
elahrvivaz: the `conf` here could be the cached singleton instance from the `FileSystemDataStoreFactory` object, so we don't want to modify it directly.

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemDataStoreFactory.scala
```diff
@@ -110,5 +111,6 @@ object FileSystemDataStoreFactory extends GeoMesaDataStoreInfo {
     val ConfParam         = new GeoMesaParam[Properties]("fs.config", "Values to set in the root Configuration, in Java properties format", largeText = true)
     val ReadThreadsParam  = new GeoMesaParam[Integer]("fs.read-threads", "Read Threads", default = 4)
     val WriteTimeoutParam = new GeoMesaParam[Duration]("fs.writer.partition.timeout", "Timeout for closing a partition file after write, e.g. '60 seconds'", default = Duration("60s"), systemProperty = Some(SystemPropertyDurationParam(WriterFileTimeout)))
+    val CharsetParam      = new GeoMesaParam[String]("fs.metadata.charset","Charset for transforming string to/from bytes array")
```
elahrvivaz: we should add this to the docs under https://github.com/locationtech/geomesa/blob/master/docs/user/filesystem/usage.rst#filesystem-data-store-parameters

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/StorageMetadata.scala
```diff
@@ -342,14 +343,16 @@ object StorageMetadata extends MethodProfiling with LazyLogging {
...
     val file = new Path(root, MetadataPath)
     val data = profile("Serialized storage configuration") {
       ConfigWriter[StorageConfig].to(config).render(options)
     }
     profile("Persisted storage configuration") {
       WithClose(fc.create(file, java.util.EnumSet.of(CreateFlag.CREATE), CreateOpts.createParent)) { out =>
-        out.writeBytes(data)
+        charset.map(c => {
+          out.write(data.getBytes(c))
+        }).getOrElse(out.writeBytes(data))
```
elahrvivaz: it would be better to use a `match` here, since the return value from `.map` is not used:
```
charset match {
  case None => out.writeBytes(data)
  case Some(c) => out.writeBytes(data.getBytes(c))
}
```
I noticed that there is also a `writeUTF` method, though - I wonder if there is some difference in just writing the raw bytes, as the `writeUTF` method also writes the byte length

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/FileMetadataTest.scala
```diff
@@ -30,7 +30,7 @@ class FileMetadataTest extends Specification with AllExpectations {
   import scala.collection.JavaConverters._
 
   lazy val fc = FileContext.getFileContext(new Configuration())
-  val sft = SimpleFeatureTypes.createType("metadata", "name:String,dtg:Date,*geom:Point:srid=4326")
+  val sft = SimpleFeatureTypes.createType("metadata", "name:String,dtg:Date,*geom:Point:srid=4326;geomesa.user-data.prefix=desc,desc.name=姓名,desc.dtg=ひづけ,desc.geom=좌표")
```
elahrvivaz: it would be good to add an explicit test that verifies that the user data is readable as expected.

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/FileSystemDataStoreFactory.scala
misslzyan: Thank you for your review. I should learn more about the coding style which accepted by the majority . About the charset configuration ,I agree with you ,though I change  some lines to make it look better .  I will improve myself so that I can talk with you at the same level . Thank you for your time ! And I wish to communicate with you  more !

elahrvivaz: no problem, thanks for contributing!

PR #2195, opened 2019-01-29
GEOMESA-2537 fix:FileMetadataTest will not test correctly

misslzyan:
add reload for clear the cache when do test cases . 
Signed-off-by: 段卫东 <misslzyan108@gmail.com>

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/StorageMetadata.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/StorageMetadata.scala
@@ -227,7 +227,18 @@ object StorageMetadata extends MethodProfiling with LazyLogging {
     * @param root path to the persisted metadata
     * @return
     */
-  def load(fc: FileContext, root: Path): Option[StorageMetadata] = Option(cache.get((fc, root)))
+  def load(fc: FileContext, root: Path, reload: Boolean = false): Option[StorageMetadata] = {
+    if (reload) {
+      val metadata = loadStorage(fc, root)
+      metadata match {
+        case Some(v) => cache.put((fc, root), v)
+        case None    => cache.put((fc, root), null)
+      }
+      metadata
+    } else {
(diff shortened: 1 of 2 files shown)


elahrvivaz:
@misslzyan what is this trying to fix? it seems like the tests pass ok without the change. Thanks,

misslzyan:
@elahrvivaz   The tests with persist and load , the load will use cache but not the metadata which was persisted .  This case will cause the tests pass but not checks the code correctly.
eg.
data.writeUTF(str)   
the test case 'create and persist an empty metadata file' will pass through , but this line code is error .

elahrvivaz:
ah, I see. makes sense, thanks!

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/StorageMetadata.scala
```diff
@@ -227,7 +227,16 @@ object StorageMetadata extends MethodProfiling with LazyLogging {
...
-  def load(fc: FileContext, root: Path): Option[StorageMetadata] = Option(cache.get((fc, root)))
+  def load(fc: FileContext, root: Path , reload: Boolean = false): Option[StorageMetadata] = {
+    if(reload) {
+      val metadata = loadStorage(fc, root)
+      metadata match {
+        case Some(v) => cache.put((fc, root), v)
+        case None    => cache.put((fc, root), null)
+      }
+      metadata
+    }else Option(cache.get((fc, root)))
```
elahrvivaz: could you clean up the formatting a bit here? there's an extra space before the comma on line 230, we try to put a space after an if, and we always try to use {} for an if/else. e.g.
```
if (reload) {
...
} else {
  ...
}
```

elahrvivaz:
@misslzyan thanks for the fix! it looks good, if you don't mind just cleaning up the code formatting I will merge it.

misslzyan:
@elahrvivaz   I will clean up the code soon

misslzyan:
@elahrvivaz   Done ! I should go to sleep . You can leave words here ,and I will read them tomorrow.Thanks.

PR #2200, opened 2019-01-31
GEOMESA-2539 FSDS - don't apply filters that are covered by the partition

elahrvivaz:
* Partitions are grouped based on the filter that needs to be applied
* Instead of searching all potential partitions, partitions from
  the partition scheme are matched against the partition metadata


elahrvivaz:
rebased

PR #2206, opened 2019-02-05
GEOMESA-2493 Configurable query filtering for merged data store view

elahrvivaz:
* Use `geomesa.merged.store.filter` to apply a filter to each store

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreView.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreView.scala
@@ -19,6 +19,7 @@ import org.locationtech.geomesa.index.view.MergedQueryRunner.MergedStats
 import org.locationtech.geomesa.utils.geotools.{SchemaBuilder, SimpleFeatureTypes}
 import org.opengis.feature.`type`.{GeometryDescriptor, Name}
 import org.opengis.feature.simple.SimpleFeatureType
+import org.opengis.filter.Filter
 
 /**
   * Merged querying against multiple data stores
@@ -26,7 +27,7 @@ import org.opengis.feature.simple.SimpleFeatureType
   * @param stores delegate stores
   * @param namespace namespace
   */
-class MergedDataStoreView(val stores: Seq[DataStore], namespace: Option[String] = None)
+class MergedDataStoreView(val stores: Seq[(DataStore, Option[Filter])], namespace: Option[String] = None)
     extends HasGeoMesaStats with ReadOnlyDataStore {
 
   import scala.collection.JavaConverters._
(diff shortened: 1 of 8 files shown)


jnh5y:
Is this ready to be rebased?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreView.scala
```diff
@@ -95,13 +96,15 @@ class MergedDataStoreView(val stores: Seq[DataStore], namespace: Option[String]
 
   override def getFeatureSource(name: Name): SimpleFeatureSource = getFeatureSource(name.getLocalPart)
 
-  override def getFeatureSource(typeName: String): SimpleFeatureSource =
-    new MergedFeatureSourceView(this, stores.map(_.getFeatureSource(typeName)), getSchema(typeName))
+  override def getFeatureSource(typeName: String): SimpleFeatureSource = {
+    val sources = stores.map { case (store, filter) => (store.getFeatureSource(typeName), filter) }
```
jnh5y: What happens if the desired featureType is only available in some datastores?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreView.scala
```diff
@@ -19,14 +19,15 @@ import org.locationtech.geomesa.index.view.MergedQueryRunner.MergedStats
...
+import org.opengis.filter.Filter
 
 /**
   * Merged querying against multiple data stores
   *
   * @param stores delegate stores
   * @param namespace namespace
   */
-class MergedDataStoreView(val stores: Seq[DataStore], namespace: Option[String] = None)
+class MergedDataStoreView(val stores: Seq[(DataStore, Option[Filter])], namespace: Option[String] = None)
```
jnh5y: Hmm... looks a little rough here.  Seems like one could have a filter per featureSource in the DataStore.  Since it covers the use, Imma ship it!

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreView.scala
elahrvivaz: in getSchema and getTypeNames, it should only return the feature types common in each store. So if you request one that doesn't exist in one of the stores, you'll usually get a schema doesn't exist exception from that store.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreView.scala
elahrvivaz: you want to specify a filter per feature source? what would be the use case?

jnh5y: I'm imagining a situation where several types are in the same DataStore.

PR #2208, opened 2019-02-06
GEOMESA-2544 Composite XML converter

elahrvivaz:
* Adds 'xpath' converter function for use in predicates
* Literal expressions are now pre-evaluated during expression parsing

### DIFF
--- a/geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XmlConverter.scala
+++ b/geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XmlConverter.scala
@@ -24,7 +24,7 @@ import org.apache.commons.io.IOUtils
 import org.apache.commons.io.input.BOMInputStream
 import org.locationtech.geomesa.convert.Modes.{ErrorMode, LineMode, ParseMode}
 import org.locationtech.geomesa.convert._
-import org.locationtech.geomesa.convert.xml.XmlConverter.{XmlConfig, XmlField, XmlOptions}
+import org.locationtech.geomesa.convert.xml.XmlConverter.{DocParser, XmlConfig, XmlField, XmlOptions}
 import org.locationtech.geomesa.convert2.transforms.Expression
 import org.locationtech.geomesa.convert2.{AbstractConverter, ConverterConfig, ConverterOptions, Field}
 import org.locationtech.geomesa.utils.collection.CloseableIterator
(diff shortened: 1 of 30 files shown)


elahrvivaz:
note: stacks on top of #2116

jnh5y:
Is this ready to be rebased?

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XmlConverter.scala
```diff
@@ -39,69 +39,23 @@ import scala.util.control.NonFatal
...
-    }
-    xp
-  }
+  private val xpath = XmlConverter.createXPath(config.xpathFactory, config.xmlNamespaces)
 
   private val rootPath = config.featurePath.map(xpath.compile)
 
   fields.foreach(_.compile(xpath))
 
   // TODO GEOMESA-1039 more efficient InputStream processing for multi mode
```
jnh5y: Does this comment need to move?

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XmlCompositeConverterTest.scala
```diff
@@ -0,0 +1,100 @@
...
+        |  ]
+        |}
+      """.stripMargin)
+  )
+
+  "XmlCompositeConverter" should {
+
+    "select a parser based on a predicate" >> {
+      val xmlStr =
+        """<doc><type>a</type><id>1</id><number>123</number><color>red</color><physical><weight>127.5</weight><height>5'11"</height></physical><lat>0</lat><lon>0</lon></doc>
```
jnh5y: Does XML composite converters work by applying different converters to each file?  Or can one process sub-documents differently?

geomesa-convert/geomesa-convert-xml/src/main/scala/org/locationtech/geomesa/convert/xml/XmlConverter.scala
elahrvivaz: ah, yeah it should probably go in the 'iterator' method, but i guess it's still relevant here where it's invoked...

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XmlCompositeConverterTest.scala
elahrvivaz: you can only apply a single converter to each xml doc, so you couldn't use a feature-path and handle each sub-document separately, if that's what you're asking.

elahrvivaz: if that's a use case, we can open a ticket to support it.

PR #2209, opened 2019-02-06
GEOMESA-2542 Kafka - provide option to clear topic on startup

elahrvivaz:
* Use `kafka.producer.clear` set to `true`

### DIFF
--- a/docs/user/kafka/usage.rst
+++ b/docs/user/kafka/usage.rst
@@ -50,6 +50,8 @@ Parameter                            Type    Description
                                              all existing messages are processed. However, feature listeners will still be invoked as normal.
                                              See :ref:`kafka_initial_load`
 ``kafka.consumer.count``             Integer Number of kafka consumers used per feature type. Set to 0 to disable consuming (i.e. producer only)
+``kafka.producer.clear``             Boolean Send a 'clear' message on startup. This will cause clients to ignore any data that was in the
+                                             topic prior to startup
 ``kafka.topic.partitions``           Integer Number of partitions to use in new kafka topics
 ``kafka.topic.replication``          Integer Replication factor to use in new kafka topics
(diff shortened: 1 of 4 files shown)


docs/user/kafka/usage.rst
```diff
@@ -50,6 +50,8 @@ Parameter                            Type    Description
                                              all existing messages are processed. However, feature listeners will still be invoked as normal.
                                              See :ref:`kafka_initial_load`
 ``kafka.consumer.count``             Integer Number of kafka consumers used per feature type. Set to 0 to disable consuming (i.e. producer only)
+``kafka.producer.clear``             Boolean Send a 'clear' message on startup. This will cause clients to ignore any data that was in the
```
jnh5y: This seems like it introduces a really dumb possibility.  What happens with any of the readback options and clear-on-start?  

Seems like it may be worth a warning.

elahrvivaz: yeah, good point. generally the producer and consumer are separate datastores though, so it's not something we can really check effectively... what do you think?

jnh5y: That's true that they are usually separate.  Is that something that we can reflect in the documentation?

In the code, I could imagine labeling a config as 'producer' or 'consumer'.  If some producer-only and consumer-only things are set, then there could be a warning.

elahrvivaz: I think we do cover it in the docs, [here](https://www.geomesa.org/documentation/user/kafka/producers.html) and [here](https://www.geomesa.org/documentation/user/kafka/consumers.html). In producer scenarios (e.g. nifi, CLI), we set the consumers to 0, and in geoserver we only expose the consumer properties. Technically you could both consume and produce, so it's not really an error to set both types of config. But in practice, checking for conflicting configs would only cover that particular case.

PR #2218, opened 2019-02-19
GEOMESA-2559 Add system property to disable AM wrap handling

elahrvivaz:
* Use `geomesa.geometry.processing` set to one of `jts` or `none`


docs/user/datastores/runtime_config.rst
```diff
@@ -97,6 +97,33 @@ front, so estimates will cause problems. To force GeoMesa to calculate the exact
 set, you may set this property to ``true``. You may also override this behavior on a per-query basis
 by using the query hint ``org.locationtech.geomesa.accumulo.index.QueryHints.EXACT_COUNT``.
 
+geomesa.geometry.processing
++++++++++++++++++++++++++++
+
+This property controls how query geometries will be handled with respect to the anti-meridian. Acceptable values are
+one of ``jts`` or ``none``. ``jts`` (the default) will use the JTS library, which will interpret a geometry with a
```
jnh5y: Is this 'JTS' handling or Spatial4J handling?  I'd argue it is the latter.

JTS wouldn't wrap...

elahrvivaz: oh, you're right - I'll change it

elahrvivaz: updated - when we merge, we should update the commit message to say 'spatial4j' instead of 'jts'

PR #2221, opened 2019-02-20
GEOMESA-2564 Lazily evaluate feature ID in remote filtering

elahrvivaz:


elahrvivaz:
note: stacks on top of #2214

PR #2223, opened 2019-02-20
GEOMESA-2561 CLI option to override converter error mode

elahrvivaz:


elahrvivaz:
This adds an option to CLI commands - I will also open a PR against geomesa-nifi to override in the processor config.

PR #2224, opened 2019-02-22
can 2.2.x call s2-geometry to implement spatial index

wyfSunflower:
I know geomesa using geohash to implement spatial index.
can I use s2-geometry[https://github.com/google/s2geometry](url) to implement spatial index?
what should I do,thanks!


elahrvivaz:
@wyfSunflower can you post this question to our user list or gitter channel? The links are in the main README. Pull requests aren't a good forum for this type of thing.
Thanks,
Emilio

PR #2228, opened 2019-03-08
Switch to gt-epsg-wkt

malinsinbigler:
All *-tools now use wkt instead of hsql. Tested geomesa-fs cli in both local and distributed run modes with this change.


elahrvivaz:
I think your git user doesn't match your signed-off footer user, and your signed-off footer user doesn't have a valid CLA...

PR #2245, opened 2019-03-22
GEOMESA-2587 Catch errors in distributed aggregating scans

elahrvivaz:

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
@@ -8,6 +8,7 @@
 
 package org.locationtech.geomesa.index.iterators
 
+import com.typesafe.scalalogging.LazyLogging
 import org.geotools.filter.text.ecql.ECQL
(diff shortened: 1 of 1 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
```diff
@@ -78,14 +81,18 @@ trait AggregatingScan[T <: AnyRef { def isEmpty: Boolean; def clear(): Unit }]
...
     result.clear()
     while (hasNextData && notFull(result)) {
-      nextData(setValues)
-      if (validateFeature(reusableSf)) {
-        // write the record to our aggregated results
-        if (hasTransform) {
-          aggregateResult(reusableTransformSf, result)
-        } else {
-          aggregateResult(reusableSf, result)
+      try {
```
jnh5y: Do we believe this would cause any performance impact?  (I think it is the right thing to do, etc. just asking...)

elahrvivaz: trys that don't throw anything are generally considered to be "free"...

PR #2258, opened 2019-04-02
GEOMESA-2598 Update validators for new converter API

elahrvivaz:
* Replace `org.locationtech.geomesa.convert.SimpleFeatureValidator$ValidatorFactory`
  with `org.locationtech.geomesa.convert2.validators.SimpleFeatureValidatorFactory`
* AbstractConverter API changed to store validator config as `Seq[String]`


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/validators/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,267 @@
...
+    ServiceLoader.load[org.locationtech.geomesa.convert.SimpleFeatureValidator.ValidatorFactory]()
+
+  val DefaultValidators = SystemProperty("geomesa.converter.validators", IndexValidatorFactory.Name)
+
+  /**
+    * Default validator names
+    *
+    * @return
+    */
+  def default: Seq[String] = DefaultValidators.get.split(",")
```
jnh5y: How often does this get called?  Should this be something that isn't recomputed?

elahrvivaz: I think it's only called when using type inference to create a converter conf, so should not be frequently. I made it a def so that it could be modified via the thread-local value.

jnh5y: Hmm...  from reading the code, I kinda expect DefaultValidators to be static and unchangeable. 

Not a problem; just my expectation:).

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/validators/SimpleFeatureValidator.scala
```diff
@@ -0,0 +1,267 @@
...
+        if (e != null) {
+          error = if (error == null) { e } else { s"$error, $e" }
+        }
+      }
+      error
+    }
+  }
+
+  /**
+    * Validator based on the indices used by the feature type. Currently only the x/z2 and x/z3 indices have
```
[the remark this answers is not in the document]
jnh5y: Maybe the NullValidator needs a better name.  Or I need read better.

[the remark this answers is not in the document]
elahrvivaz: attributes are allowed to be null, so we don't actually have a validator for them... the null validator is just common logic used by the 'required dtg/geom' validators

PR #2261, opened 2019-04-24
GEOMESA-2606 Fix client-side stats for transform schemas

elahrvivaz:


elahrvivaz:
We should cherry-pick back to 2.3.x

PR #2262, opened 2019-04-26
GEOMESA-1936,GEOMESA-2427 Support non-point geometries and Booleans in Parquet

elahrvivaz:
* Support linestrings, multipoints, polygons, multilinestrings, multipolygons
* Move ObjectTypes.JSON to secondary typing (primary now STRING)
* Make parquet FilterConverter an object, run in a single pass

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2259, opened 2019-04-02
GEOMESA-2600 Java API for converters

elahrvivaz:
* `org.locationtech.geomesa.convert2.interop.SimpleFeatureConverterLoader#load`
* `org.locationtech.geomesa.convert2.SimpleFeatureConverter#createJavaEvaluationContext`

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2263, opened 2019-04-30
GEOMESA-2609 Parquet converter implementation

elahrvivaz:
* Add file path to converter inference


elahrvivaz:
note: stacks on top of #2262

jnh5y:
Needs a rebase.:)

I'm happy to merge #2259 first if it'd help.

elahrvivaz:
rebased

docs/user/convert/parquet.rst
```diff
@@ -0,0 +1,171 @@
...
+
+.. _parquet_converter_functions:
+
+Parquet Transform Functions
+---------------------------
+
+GeoMesa defines several Parquet-specific transform functions, in addition to the ones defined under
+:ref:`avro_converter_functions`.
+
+parquetPoint
```
jnh5y: When are these needed?  The example below uses paths and existing functions.  Is this syntactical sugar?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/SimpleFeatureConverterFactory.scala
```diff
@@ -31,5 +32,26 @@ trait SimpleFeatureConverterFactory {
     * @param sft simple feature type, if known ahead of time
     * @return
     */
-  def infer(is: InputStream, sft: Option[SimpleFeatureType] = None): Option[(SimpleFeatureType, Config)] = None
+  @deprecated("Use infer(InputStream, Option[String], Option[SimpleFeatureType]")
```
jnh5y: Do you mean:
  @deprecated("Use infer(InputStream, Option[SimpleFeatureType],  Option[String]")
?

docs/user/convert/parquet.rst
elahrvivaz: These are for geomesa-encoded geometry columns, as written by geomesa-fs-storage-parquet. The examples below are just pulling lat/lon out of generic parquet data, so they're not needed. Our geometry columns use more complicated nested structures, that wouldn't be easy (possible?) to read using regular paths and functions.

elahrvivaz: generally these will probably only show up in type-inferenced ingestion of geomesa-parquet files.

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/SimpleFeatureConverterFactory.scala
elahrvivaz: yes, thanks

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/parquet/io/SimpleFeatureReadSupport.scala
```diff
@@ -181,12 +216,12 @@ object SimpleFeatureReadSupport {
     override def addBinary(value: Binary): Unit = callback.set(index, value.getBytes)
   }
 
-  class ListConverter(index: Int, binding: Seq[ObjectType], callback: Settable) extends GroupConverter {
+  class ListConverter(binding: ObjectType, index: Int, callback: Settable) extends GroupConverter {
```
jnh5y: Any particular reason for the argument order change?

elahrvivaz: i changed it from a Seq to a single type (same with the map converter below), to make it a little clearer what's required. Once the signature had changed, seemed like the binding was 'higher importance' so i put it first... no real reason.

PR #2260, opened 2019-04-05
GEOMESA-2455 Converters - support dropwizard metrics

elahrvivaz:
* Metrics are exposed through the EvaluationContext or configurable reporters
* Validators publish success/failed metrics
* API change - AbstractConverter.ConverterOptions requires a `reporters` field

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2264, opened 2019-05-02
GEOMESA-2615 Converters - provide lifecycle for validators

elahrvivaz:
* SimpleFeatureConverter extends Closeable


elahrvivaz:
note: stacks on top of #2260

jnh5y:
Ok, why do we want a close for a validator?

elahrvivaz:
presumably so we can lookup valid values in some kind of database

PR #2266, opened 2019-05-07
GEOMESA-2617 Generic data store command line tools

elahrvivaz:
* New release artifact `geomesa-gt-dist_2.11`
* geomesa-spark-geotools module deprecated and replaced with geomesa-gt-spark
* `WithStore` utility method to load/dispose of data stores
* Fixed dispose of stores in GeoMesaSparkSql


build/cqs.tsv
```diff
@@ -36,6 +36,7 @@ commons-cli:commons-cli	1.2	compile
 commons-codec:commons-codec	1.9	compile
 commons-collections:commons-collections	3.2.2	compile
 commons-configuration:commons-configuration	1.6	compile
+commons-dbcp:commons-dbcp	1.4	compile
```
elahrvivaz: approved https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19645

build/cqs.tsv
```diff
@@ -227,6 +230,7 @@ org.mortbay.jetty:jetty-util	6.1.26	compile
 org.objenesis:objenesis	1.0	compile
 org.parboiled:parboiled-core	1.1.7	compile
 org.parboiled:parboiled-scala_2.11	1.1.7	compile
+org.postgresql:postgresql	42.2.2	compile
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19646

docs/user/cli/geotools.rst
```diff
@@ -0,0 +1,102 @@
...
+
+    You can configure environment variables and classpath settings in ``geomesa-gt_2.11-$VERSION/conf/geomesa-env.sh``.
+
+.. note::
+
+    ``geomesa-gt`` will load JARs from the ``$GEOMESA_EXTRA_CLASSPATHS`` environment variable
+    into the class path. Use the ``geomesa-gt classpath`` command in order to see what JARs are being used.
+
+
+Due to licensing restrictions, dependencies for shape file support must be separately installed. Do this with the
```
jnh5y: Can you change it to read 'shapefile'?

docs/user/cli/geotools.rst
```diff
@@ -0,0 +1,102 @@
...
+Run ``geomesa-gt`` without arguments to confirm that the tools work.
+
+.. code::
+
+    $ bin/geomesa-gt
+    INFO  Usage: geomesa-gt [command] [command options]
+      Commands:
+      ...
+
+Setting up Distributed Processing
```
jnh5y: Does MR ingest, etc work with the various GeoTools datastores?  Seems like this might deserve a note/warning;).

docs/user/cli/geotools.rst
```diff
@@ -0,0 +1,102 @@
...
+argument can be used to specify a Java properties file containing the parameters. This may be useful for
+simplifying the command invocation, or to hide sensitive parameters from bash history and process lists.
+
+For example, to connect to a Postgis data store, you may use the following command:
+
+.. code::
+
+    $ bin/geomesa-gt export --param dbtype=postgis --param host=localhost \
+      --param user=postgres --param passwd=postgres --param port=5432 \
+      --param database=example --feature-name gdelt
```
jnh5y: Hmm...  with this --param option...  I wonder if this would be a good way to allow for extra datastore params to be passed into our other command line tools.

docs/user/spark/providers.rst
```diff
@@ -135,31 +152,36 @@ details.
...
+``GeoToolsSpatialRDDProvider`` is provided by the ``geomesa-gt-spark`` module.
 
 ``GeoToolsSpatialRDDProvider`` generates and saves ``RDD``\ s of features stored in
 a generic GeoTools ``DataStore``. The configuration parameters passed are the same as
 those passed to ``DataStoreFinder.getDataStore()`` to create the data store of interest,
 plus a required boolean parameter called "geotools" to indicate to the SPI to load
-``GeoToolsSpatialRDDProvider``. For example, the `CSVDataStore`_ described in the
-`GeoTools ContentDataStore tutorial`_ takes a single parameter called "file". To use
-this data store with GeoMesa Spark, do the following:
+``GeoToolsSpatialRDDProvider``. For example, to use the `Postgis DataStore`_ with
```
jnh5y: Did you test this with Spark?

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
```diff
@@ -37,10 +37,6 @@
             <groupId>org.locationtech.geomesa</groupId>
             <artifactId>geomesa-spark-converter_${scala.binary.version}</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-spark-geotools_${scala.binary.version}</artifactId>
```
jnh5y: Was there any implicit dependencies inside 'geomesa-spark-geotools' that GM+Accumulo+Spark needed?

geomesa-gt/geomesa-gt-dist/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>pom</packaging>
+
+    <artifactId>geomesa-gt-dist_2.11</artifactId>
+    <name>GeoMesa GeoTools Distribution</name>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <gmtools.dist.name>GEOMESA_GEOTOOLS</gmtools.dist.name> <!-- used in tools filtering -->
```
jnh5y: Ok... that's interesting.  Does the param need to be uppercase?

geomesa-gt/geomesa-gt-dist/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+        </dependency>
+    </dependencies>
+
+    <profiles>
+        <profile>
+            <id>python</id>
+            <dependencies>
+                <dependency>
+                    <groupId>org.locationtech.geomesa</groupId>
+                    <artifactId>geomesa_pyspark</artifactId>
```
jnh5y: Ok, so if we build the release with and without the Python profile, we'd get a different dist?

geomesa-gt/geomesa-gt-spark/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-spark-core_${scala.binary.version}</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools.jdbc</groupId>
+            <artifactId>gt-jdbc-postgis</artifactId>
+        </dependency>
```
jnh5y: Are the other GeoTools dependencies coming in transitively?  (I'm thinking of gt-main, gt-api, etc.)

geomesa-gt/geomesa-gt-spark/src/main/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+  * implementation (e.g. HBaseSpatialRDDProvider), as it will be better optimized.
+  *
+  * To avoid being picked up inadvertently, the data store params must include "geotools" -> "true"
+  */
+class GeoToolsSpatialRDDProvider extends SpatialRDDProvider with LazyLogging {
+
+  import scala.collection.JavaConverters._
+
+  override def canProcess(params: java.util.Map[String, java.io.Serializable]): Boolean = {
+    Option(params.get("geotools")).exists(FastConverter.convert(_, classOf[java.lang.Boolean])) &&
```
jnh5y: Is this a change for how one gets an RDD from a GT datastore presently?  (If so, the change might be worth mentioning.)

geomesa-gt/geomesa-gt-spark/src/test/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProviderTest.scala
```diff
@@ -6,19 +6,16 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.spark.geotools
+package org.locationtech.geomesa.geotools.spark
```
jnh5y: Hmm...  Have we settled on a scheme here?

I see
package org.locationtech.geomesa.accumulo.spark
package org.locationtech.geomesa.bigtable.spark
package org.locationtech.geomesa.bigtable.spark;
...
package org.locationtech.geomesa.fs.spark
package org.locationtech.geomesa.hbase.spark
package org.locationtech.geomesa.kudu.spark
package org.locationtech.geomesa.spark
package org.locationtech.geomesa.spark.accumulo
package org.locationtech.geomesa.spark.api.java
package org.locationtech.geomesa.spark.api.python
package org.locationtech.geomesa.spark.converter
package org.locationtech.geomesa.spark.geotools
package org.locationtech.geomesa.spark.hbase

Is this a preference for o.l.g.THING.spark?

docs/user/cli/geotools.rst
elahrvivaz: done

docs/user/cli/geotools.rst
elahrvivaz: Yeah it does, what note/warning were you thinking of? I thought this section was basically the note/warning.

geomesa-gt/geomesa-gt-tools/examples/ingest/avro/ingest_avro.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# type and converter example-xml is registered in $GEOMESA_HOME/conf/application.conf
+bin/geomesa-gt ingest --param dbtype=postgis --param host=localhost --param port=5432 --param schema=public --param database=database --param user=postgres --param passwd=postgres -s example-avro -C example-avro-no-header example_no_header.avro
```
jnh5y: Do we have a script that ties together these examples and tries 'em out?

geomesa-gt/geomesa-gt-tools/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-utils_${scala.binary.version}</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-convert-all_${scala.binary.version}</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-csv</artifactId>
```
jnh5y: Hmm... what pulls this in?  (Idle question)

docs/user/cli/geotools.rst
elahrvivaz: we could, yeah - were there any params in particular you were thinking of? generally it's nicer to have named parameters so that they can be validated, enforced as required/not, etc.

docs/user/spark/providers.rst
elahrvivaz: I tested postgis, yeah.

geomesa-gt/geomesa-gt-tools/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+                             also adding the project build directory -->
+                        <classpath />
+                        <argument>org.locationtech.geomesa.geotools.tools.GeoToolsRunner</argument>
+                        <argument>help</argument>
+                        <argument>--autocomplete-function</argument>
+                        <argument>${project.build.directory}/autocomplete.sh,geomesa</argument>
+                    </arguments>
+                </configuration>
+            </plugin>
+            <plugin>
```
jnh5y: Hmm...  Is there a good way to pull this config up to a common place?

geomesa-gt/geomesa-gt-dist/pom.xml
elahrvivaz: it's an environment variable I believe... we created this pattern a while ago to inject things into common scripts.

geomesa-gt/geomesa-gt-dist/pom.xml
elahrvivaz: Ah, I'm not sure. I copied this from geomesa-accumulo. @tkunicki ?

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
```diff
@@ -0,0 +1,63 @@
...
+      val error = file.collect {
+        case f if !f.exists     => s"${f.getAbsolutePath} does not exist"
+        case f if !f.canRead    => s"${f.getAbsolutePath} can't be read"
+        case f if f.isDirectory => s"${f.getAbsolutePath} is a directory"
+      }
+      error.foreach(e => throw new ParameterException(s"Invalid parameter file: $e"))
+    }
+    val props = new Properties()
+    file.foreach(f => props.load(new FileInputStream(f)))
+    params.params.asScala.foreach { case (k, v) => props.put(k, v) }
```
jnh5y: Ok, it looks like one could give param(s) and a file of params.  Is the fact that the params would override the file documented?  (And should it be?)

geomesa-gt/geomesa-gt-spark/pom.xml
elahrvivaz: Yeah, they're coming from the other geomesa-* jars i believe, although also this one probably.

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
```diff
@@ -0,0 +1,63 @@
...
+    props.asScala.toMap
```
jnh5y: Further, is there a logging level which would echo the params so that they can be verified/debugged?

geomesa-gt/geomesa-gt-spark/src/main/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProvider.scala
elahrvivaz: It's not really a change, I think before it just accepted strings which seemed potentially confusing, so I made it also accept Booleans.

geomesa-gt/geomesa-gt-spark/src/test/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProviderTest.scala
elahrvivaz: I was mainly thinking about package isolation for java 11. If we want to be able to put anything into `org.locationtech.geomesa.spark` that means that jars like this would have to be `org.locationtech.geomesa.<THING>.spark` right?

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseColumnGroupsTest.scala
```diff
@@ -125,10 +125,7 @@ class HBaseColumnGroupsTest extends HBaseTest with LazyLogging  {
     ds = DataStoreFinder.getDataStore(params.asJava).asInstanceOf[HBaseDataStore]
     ds.createSchema(sft)
     val writer = ds.getFeatureWriterAppend(sft.getTypeName, Transaction.AUTO_COMMIT)
-    features.foreach { f =>
-      FeatureUtils.copyToWriter(writer, f, useProvidedFid = true)
-      writer.write()
-    }
+    features.foreach(FeatureUtils.write(writer, _, useProvidedFid = true))
```
jnh5y: Hmm... I wonder if we have a page of documentation to show 'best practices' like this.

geomesa-gt/geomesa-gt-dist/pom.xml
tkunicki: yeah, this was to avoid issues on build systems w/o even a minimal python install (python + setuptools module).  If -Ppython is forced in the release plugin profile this shouldn't be an issue...

geomesa-gt/geomesa-gt-tools/examples/ingest/avro/ingest_avro.sh
elahrvivaz: we have, e.g. [this](https://github.com/locationtech/geomesa/blob/geomesa_2.11-2.3.0/geomesa-accumulo/geomesa-accumulo-tools/examples/test/geomesa-test-accumulo.sh), but I don't think that those scripts are used and/or maintained...

geomesa-gt/geomesa-gt-tools/pom.xml
elahrvivaz: Ah, I no longer recall... it might be that it's pulled in by one of the provided deps, and this is to re-scope it to compile.

geomesa-gt/geomesa-gt-spark/src/main/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProvider.scala
tkunicki: might want to match with with [JavaGeoMesaSpark](https://github.com/locationtech/geomesa/blob/master/geomesa-spark/geomesa-spark-core/src/main/scala/org/locationtech/geomesa/spark/api/java/JavaGeoMesaSpark.scala#L27), tho, i don't think it will be an issue w/ type erasure

geomesa-gt/geomesa-gt-tools/pom.xml
elahrvivaz: I think due to maven inheritance, this is the best we can do... we could create a 'tools archetype' like we have to the geoserver plugin modules, but then it wouldn't inherit the parent module, so we'd just be reversing what gets duplicated...
maybe we could use the maven tiles plugin??

geomesa-gt/geomesa-gt-spark/src/test/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProviderTest.scala
tkunicki: the scheme i had settled on was to match module homing on file system...  base spark stuff was `org.locationtech.geomesa.spark.<extension>` then datastore extensions were `org.locationtech.geomesa.<extention>.spark`  this was a little munged w/ geotools and converter extentions.  I think this is Java11+ safe

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
elahrvivaz: I clarified it in the docs.

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
elahrvivaz: there is not - I assume you'd like one? is there a concern with logging e.g. password?

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseColumnGroupsTest.scala
elahrvivaz: we don't, but I think if we wanted something like that we should maybe use a wiki page or just try to check during code reviews; I don't want to make them part of the 'public API' to where we have to worry about semver. That said, I'm sure a lot of downstream projects using geomesa could be a bit simpler using things like this.

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
elahrvivaz: I'm fairly sure not - I don't remember if I tested accumulo or not, but we cover testing spark during our release process, so we should catch any problems then.

docs/user/cli/geotools.rst
jnh5y: It is nicer.  I'm just wondering if this is useful as an escape hatch in case we miss one...  Eh, maybe not?

geomesa-gt/geomesa-gt-spark/src/test/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProviderTest.scala
jnh5y: On the Java 11 front, I think jars can add things further down in a hierarchy.  I'm 95% sure about that:).

geomesa-gt/geomesa-gt-tools/pom.xml
jnh5y: @tkunicki thoughts on Maven tiles here?

geomesa-gt/geomesa-gt-spark/src/main/scala/org/locationtech/geomesa/geotools/spark/GeoToolsSpatialRDDProvider.scala
elahrvivaz: hmm, right now SpatialRDDPRovider is `def canProcess(params: java.util.Map[String, java.io.Serializable]): Boolean` and the java API is `def apply(params: java.util.Map[String, _ <: java.io.Serializable])` - probably the java API signature is the one we should go with.

Although I'm not sure where we settled on Serializable, as DataStoreFinder uses `getDataStore(Map params)` (I guess for java 1.4 compat?) and Param uses `lookUp(Map<String, ?> map)`. We should probably make our signatures (java.util.)`Map[String, _]`.

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
jnh5y: Yes, I'd like something like that.  It wouldn't say that it is a requirement, but a nice-to-have.  If it ought to be a separate ticket, that's fine.

For passwords, I think GeoTools has a Parameter keyword to know if a field is sensitive.  We could use that to avoid rendering passwords, etc.

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseColumnGroupsTest.scala
jnh5y: +1 to wiki or something like that.  Not a high priority, but yeah, good idea!

PR #2268, opened 2019-05-22
GEOMESA-2628 Table partitioning indicates if no tables can match a filter

elahrvivaz:
* TablePartition.partitions changed to return an Option[Seq[String]] to
  indicate disjoint filters

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/partition/TimePartition.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/partition/TimePartition.scala
@@ -17,8 +17,8 @@ import org.locationtech.geomesa.curve.TimePeriod.TimePeriod
 import org.locationtech.geomesa.filter.{Bounds, FilterHelper, FilterValues}
 import org.locationtech.geomesa.index.conf.partition.TimePartition.CustomPartitionCache
 import org.locationtech.geomesa.index.metadata.{CachedLazyMetadata, GeoMesaMetadata, HasGeoMesaMetadata}
-import org.locationtech.geomesa.utils.text.DateParsing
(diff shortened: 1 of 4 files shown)


khobbs-ccri:
I tried this out and it behaves as advertised. I am now able to quickly query my merged and partitioned data store with a temporal query that does not overlap any of the partitions in one of the stores.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/partition/TimePartition.scala
```diff
@@ -60,13 +60,18 @@ class TimePartition(metadata: GeoMesaMetadata[String], typeName: String, dtg: St
     f"${toBin(ZonedDateTime.ofInstant(toInstant(date), ZoneOffset.UTC))}%05d" // a short should fit into 5 digits
   }
 
-  override def partitions(filter: Filter): Seq[String] = {
+  override def partitions(filter: Filter): Option[Seq[String]] = {
```
jnh5y: Ok, do you have a really good reason for using Option[Seq]?  Seems like this could be done with a Seq.  

That said, tell me you needed it, and I'll ship the PR:).

elahrvivaz: yeah, I tried to show cover it in the scaladocs: https://github.com/locationtech/geomesa/blob/05d69812e866566b46880ecd3facd587d0017253/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/partition/TablePartition.scala#L33
Basically we need to differentiate between "there are no matching partitions" and "could not extract any partitions from the predicate"

PR #2275, opened 2019-06-10
GEOMESA-2649 Queries don't consistently hit all partitioned tables

elahrvivaz:


elahrvivaz:
I think this will cherry-pick cleanly back to 2.3.1

PR #2272, opened 2019-06-03
GEOMESA-2640 SparkSQL - Fix schema comparison on save

elahrvivaz:
Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2283, opened 2019-06-13
GEOMESA-2630,GEOMESA-2631 Fix JSON output for Spark dataframes

elahrvivaz:
* New conversion class in spark-sql module
* Deprecated spark-jts conversion class
* Separated out classes in GeoMesaSparkSQL file


elahrvivaz:
stacks on top of #2272

docs/user/spark/core.rst
```diff
@@ -139,3 +139,20 @@ To save features, use the ``save()`` method:
 Note that some providers may be read-only.
 
 See :doc:`./providers` for details on specific provider implementations.
+
+
+GeoJSON Output
```
jnh5y: Does 'core' depend on JTS?  Or is it the other way?

I do like having GeoJSON support in the most basic module.  If the docs are indicating a move of functionality from the Spark+JTS module, I'd like to consider alternatives for a minute.

elahrvivaz: I think core depends on JTS. The old method is still there, but we don't have geotools on the classpath in the JTS module so we'd have to re-implement geojson encoding to keep it there...

elahrvivaz: possibly we could use some combination of a generic json writer and the geojson geometry writer from jts... but I dunno seems like people would be using the core module anyway.

jnh5y: Yeah, I wonder how hard it'd be to do that? 

I'd hate for us to pull in GeoTools for something that'd be really, really quick.

jnh5y: Am I asking for a different feature/capability?  Aka, should this idea be a separate ticket?

elahrvivaz: Is this an actual use case though? Who is going to be using just the JTS module and nothing else? All our spark-runtime jars bundle core up too. I'd opine it's not worth trying to implement something just in JTS, when we can easily use geotools in the other module.

jnh5y: RasterFrames and other projects are pointed at GeoMesa's JTS Spark module.

jnh5y: That said, it is a different issue, so I'm shipping this PR!

PR #2282, opened 2019-06-13
GEOMESA-2627 Update-schema CLI command

elahrvivaz:
* Add support for renaming attributes
* Deprecate `keywords` command, as the functionality is covered by
  `update-schema`/`describe-schema`
* Standardize config key vals to CamelCase, deprecate old key vals
* Rename CachedLazyMetadata to TableBasedMetadata
* Rename CachedLazyBinaryMetadata to KeyValueStoreMetadata

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2299, opened 2019-06-18
GEOMESA-2619 Fix sort by an attribute that is not returned

elahrvivaz:

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
@@ -15,7 +15,6 @@ import org.geotools.feature.simple.SimpleFeatureTypeBuilder
 import org.geotools.filter.{FunctionExpressionImpl, MathExpressionImpl}
 import org.geotools.process.vector.TransformProcess
 import org.geotools.process.vector.TransformProcess.Definition
-import org.locationtech.geomesa.filter.FilterHelper
 import org.locationtech.geomesa.index.api.QueryPlan
 import org.locationtech.geomesa.index.conf.QueryHints
 import org.locationtech.geomesa.index.conf.QueryHints.RichHints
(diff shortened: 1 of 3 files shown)


elahrvivaz:
stacks on top of #2282

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
```diff
@@ -179,32 +178,32 @@ object QueryPlanner extends LazyLogging {
...
+        None
       }
     }
 
-    Option(query.getPropertyNames).map(_.toSeq)
-        .filter(_ != sft.getAttributeDescriptors.asScala.map(_.getLocalName))
-        .orElse(Some(transformsFromQueryType).filter(_.nonEmpty))
-        .foreach { props =>
+    // since we do sorting on the client, just add any sort-by attributes to the transform
+    // TODO GEOMESA-2655 we should sort and then transform back to the requested props, but it's complicated...
```
jnh5y: Given that returning more columns/attributes than requested may not be expected, should we document this somehow?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureOrdering.scala
```diff
@@ -71,6 +124,11 @@ object SimpleFeatureOrdering {
       nullCompare(property.evaluate(x).asInstanceOf[Comparable[Any]], property.evaluate(y))
   }
 
+  private class UserDataOrdering(key: String) extends Ordering[SimpleFeature] {
```
jnh5y: Is this used?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
elahrvivaz: it's documented in the ticket :D
maybe we should have a known bugs section? but seems somewhat redundant with jira

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureOrdering.scala
elahrvivaz: no - I was initially playing around with stuffing the sort field into the user data so that we wouldn't have to return it, so I figured i'd leave it here in case we want to pursue that later. I can remove it if you want.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
elahrvivaz: maybe we could have a link to JIRA 'affects version' searches for each version, like:
```
Known Bugs: 2.0.0, 2.0.1, 2.0.2, ...
```

jnh5y: *shrugs*  Not sure what's a best practice here.

PR #2304, opened 2019-06-24
GEOMESA-2656 Cache writers during local ingest

elahrvivaz:
* Flush each writer every 20k features
* Configurable via `geomesa.ingest.local.batch.size`
* New scala-friendly object pool impl:
  `org.locationtech.geomesa.utils.io.CloseablePool`


crazyxxl:
Geomesa is version 2.3.0 and hbase is version 1.1.2. By default, the time split table is used. When data is inserted, a new table can be created dynamically, but the table information cannot be stored in the metadata,why?the version?or somethins else?

jnh5y:
@crazyxxl Can you ask your question on Gitter or the GeoMesa-user list?

PR #2307, opened 2019-07-09
GEOMESA-2660 Log warning for unused/undefined converter fields

elahrvivaz:

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Modes.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/Modes.scala
@@ -8,26 +8,26 @@
 
 package org.locationtech.geomesa.convert
 
-import org.locationtech.geomesa.convert.Modes.ParseMode.Incremental
 import org.locationtech.geomesa.utils.conf.GeoMesaSystemProperties.SystemProperty
 
 trait Modes {
+
   this: Enumeration =>
 
   type Mode
+
   protected def defaultValue: Mode
+
   def systemProperty: SystemProperty
 
   def apply(): Mode = {
     val string = systemProperty.get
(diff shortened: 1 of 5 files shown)


elahrvivaz:
in the CLI tools, the warnings get logged to geomesa.log by default - this may not be extremely high visibility but I'm wary of spamming the console with warnings. We could define some special logger just for this that we could configure to log to the console by default...

elahrvivaz:
@jrs53 @lrobb any thoughts?

jrs53:
I quite often find that when I am having problems, I need to look at geomesa.log (vs default console output). Should the default log4j.properties echo WARN to the console? If it ends up too spammy, then maybe those warnings should't be warnings after all

jnh5y:
@elahrvivaz thoughts on James's comment on the command line tools going to the console?  

Maybe that's a separate discussion from this ticket, and it ought to be merged as-is?

elahrvivaz:
Yeah, I think the logging should probably be a separate ticket. It's a complicated problem, as most of our logging is meant to go to (duh) log files. I don't think that we want to propagate the idea of separating out log messages for users throughout our whole codebase. And, I would prefer more 'chatty' warnings than logging at a lower level or not logging at all, because it can be very useful when you go back to look for a problem. But it also seems somewhat infeasible to configure the CLI output just via log4j.properties - although we might be able to mitigate the issue somewhat that way, it would be fairly brittle to keep in sync with new logging or classes.
Maybe we could add a generic 'verbose' argument to each command that switches the log4j.properties to print everything to console.

jrs53:
My view is that the message should be at the WARN level (i.e. processing can continue, but you might not get what you expected). Agree there should be a separate ticket to look at the correct log level for the file and stderr logs for the command line tools (I would guess WARN?)

jnh5y:
All, shout if I closed this too soon!  I'm happy to encourage debate/conversations!

PR #2313, opened 2019-07-16
GEOMESA-2647 Converters - publish default metrics

elahrvivaz:
* Rate, parse duration, convert duration, feature date, feature latency

### DIFF
--- a/docs/user/convert/parsing_and_validation.rst
+++ b/docs/user/convert/parsing_and_validation.rst
@@ -171,15 +171,15 @@ Metrics
 ~~~~~~~
 
 Converters use the `Dropwizard Metrics <https://metrics.dropwizard.io/>`__ library to register metrics on
-successful conversions, failed conversions, and validation errors. Metrics can be accessed through the converter
-evaluation context, or can be exposed through reporters configured in the converter options:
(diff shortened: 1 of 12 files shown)


docs/user/convert/parsing_and_validation.rst
```diff
@@ -201,16 +201,16 @@ by default. The reporters can be configured as follows:
...
+          {
             type           = "graphite"
             url            = "localhost:9000"
             prefix         = "example"
             rate-units     = "SECONDS"
             duration-units = "MILLISECONDS"
             interval       = "10 seconds"
-          }
-          ganglia = {
+          },
```
jnh5y: Do the docs line up with how to configure the different reporters?

elahrvivaz: yeah, I believe so.

elahrvivaz: the configs themselves didn't change, but the outer container changed from a named map to an array (since the names were not used/irrelevant)

elahrvivaz: the named map will still work for backwards compatibility, but we don't document it.

PR #2314, opened 2019-07-18
GEOMESA-2667 Redis - per-feature time-to-live

elahrvivaz:
* Add FeatureExpiration sft config
* Add type parameter to FeatureWrapper

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

PR #2317, opened 2019-07-19
GEOMESA-2670 Standardize feature expiration config

elahrvivaz:
* Accumulo expiration configurable/modifiable via sft hints


elahrvivaz:
stacks on top of #2314

docs/user/accumulo/internals.rst
```diff
@@ -1,24 +0,0 @@
-Internals
-=========
-
-Indexing Strategies
```
jnh5y: Did this discussion go away?

elahrvivaz: Yeah... it was out of date, and seemed deeper in the weeds than I'd guess most people would go. Seems like it would be more appropriate in the developers guide, if we ever wrote one.

PR #2318, opened 2019-07-24
GEOMESA-2673 - Performance - don't return from scala iterable methods

elahrvivaz:


elahrvivaz:
I grepped through the code for 'return' statements, and updated all the occurrences I found. I verified that javap no longer shows any references to `NonLocalReturnControl` in the updated classes.

PR #2329, opened 2019-08-06
GEOMESA-2684 Bypass decorating feature collections to optimize WPS calls

elahrvivaz:

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureCollection.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureCollection.scala
@@ -11,34 +11,57 @@ package org.locationtech.geomesa.index.geotools
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicLong}
 
 import com.typesafe.scalalogging.LazyLogging
+import org.geotools.data.simple.{SimpleFeatureCollection, SimpleFeatureSource}
 import org.geotools.data.store.DataFeatureCollection
+import org.geotools.data.util.NullProgressListener
 import org.geotools.data.{FeatureReader, Query, Transaction}
+import org.geotools.feature.FeatureCollection
+import org.geotools.feature.collection.{DecoratingFeatureCollection, DecoratingSimpleFeatureCollection}
 import org.geotools.feature.visitor.{BoundsVisitor, MaxVisitor, MinVisitor}
 import org.geotools.geometry.jts.ReferencedEnvelope
+import org.geotools.util.factory.Hints
+import org.locationtech.geomesa.filter.FilterHelper
+import org.locationtech.geomesa.index.geotools.GeoMesaFeatureCollection.GeoMesaFeatureVisitingCollection
 import org.locationtech.geomesa.index.process.GeoMesaProcessVisitor
+import org.locationtech.geomesa.index.stats.GeoMesaStats
+import org.locationtech.geomesa.utils.collection.CloseableIterator
+import org.locationtech.geomesa.utils.io.WithClose
 import org.opengis.feature.FeatureVisitor
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+import org.opengis.filter.Filter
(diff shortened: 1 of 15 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureCollection.scala
```diff
@@ -11,96 +11,248 @@ package org.locationtech.geomesa.index.geotools
...
+  @tailrec
+  private def unwrap(
+      collection: FeatureCollection[SimpleFeatureType, SimpleFeature],
+      level: Int = 1): FeatureCollection[SimpleFeatureType, SimpleFeature] = {
+
+    // noinspection TypeCheckCanBeMatch
+    val unwrapped = if (collection.isInstanceOf[DecoratingSimpleFeatureCollection]) {
+      getDelegate(collection, classOf[DecoratingSimpleFeatureCollection])
+    } else if (collection.isInstanceOf[DecoratingFeatureCollection[SimpleFeatureType, SimpleFeature]]) {
+      getDelegate(collection, classOf[DecoratingFeatureCollection[SimpleFeatureType, SimpleFeature]])
```
jnh5y: Does GeoServer have a separate decorating interface that they could be using?  Aka, do we need to do anything extreme to cover another corner case?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedFeatureSourceView.scala
```diff
@@ -117,35 +112,6 @@ class MergedFeatureSourceView(
...
-          }
-
-        case v: GeoMesaProcessVisitor =>
-          v.execute(MergedFeatureSourceView.this, query)
-
-        case _ =>
-          super.accepts(visitor, progress)
-      }
-
-    private def minMax(attribute: String, exact: Boolean): Option[(Any, Any)] =
```
jnh5y: Is this function supposed to go away?

elahrvivaz: it's only called in the visit method, so it's in the new trait now.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureCollection.scala
elahrvivaz: I didn't find anything... I looked through all the classes that I could find that extend SimpleFeatureCollection or FeatureCollection. They all either delegate appropriately, or extend one of these two.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureCollection.scala
```diff
@@ -11,96 +11,248 @@ package org.locationtech.geomesa.index.geotools
...
+    * @param collection feature collection
+    * @param visitor visitor
+    * @param progress progress monitor
+    */
+  def visit(
+      collection: FeatureCollection[SimpleFeatureType, SimpleFeature],
+      visitor: FeatureVisitor,
+      progress: ProgressListener = new NullProgressListener): Unit = {
+    val unwrapped = if (collection.isInstanceOf[GeoMesaFeatureVisitingCollection]) { collection } else {
+      try { unwrap(collection) } catch {
```
jnh5y: Is it worth adding another logging statement or two here?  I suppose I'm suggesting one for the case where we see a GeoMesaFeatureVisitingCollection and do nothing.

elahrvivaz: I believe all our processes log the feature collection class, so it would be somewhat redundant, what do you think?

jnh5y: That makes sense.  I'm being overly defensive;).

PR #2332, opened 2019-08-09
GEOMESA-2689 Reduce memory footprint of deserialized simple features

elahrvivaz:
* New V3 serialization format
* Doesn't require tracking offsets into serialized bytes
* De-references backing array when all attributes are read
* Stores nulls in a bitmask to save space

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
@@ -44,7 +44,7 @@ import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleF
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.AttributeOptions
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.Configs.OverrideDtgJoin
-import org.locationtech.geomesa.utils.index.{GeoMesaSchemaValidator, IndexMode}
+import org.locationtech.geomesa.utils.index.{GeoMesaSchemaValidator, IndexMode, VisibilityLevel}
 import org.locationtech.geomesa.utils.io.WithClose
 import org.locationtech.geomesa.utils.stats.{IndexCoverage, Stat}
 import org.opengis.feature.simple.SimpleFeatureType
@@ -198,6 +198,10 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
       }
     }
 
+    if (sft.getVisibilityLevel == VisibilityLevel.Attribute && sft.getAttributeCount > 255) {
+      throw new IllegalArgumentException("Attribute level visibility only supports up to 255 attributes")
+    }
+
     super.preSchemaCreate(sft)
 
     // note: dtg should be set appropriately before calling this method
@@ -230,6 +234,10 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
 
   @throws(classOf[IllegalArgumentException])
   override protected def preSchemaUpdate(sft: SimpleFeatureType, previous: SimpleFeatureType): Unit = {
+    if (sft.getVisibilityLevel == VisibilityLevel.Attribute && sft.getAttributeCount > 255) {
+      throw new IllegalArgumentException("Attribute level visibility only supports up to 255 attributes")
+    }
+
     // check for attributes flagged 'index=join' and convert them to sft-level user data
     sft.getAttributeDescriptors.asScala.foreach { d =>
       val index = d.getUserData.get(AttributeOptions.OptIndex).asInstanceOf[String]
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
@@ -14,9 +14,10 @@ import org.apache.accumulo.core.data.{Key, Value}
 import org.apache.accumulo.core.iterators.user.RowEncodingIterator
 import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
 import org.locationtech.geomesa.features.kryo.KryoFeatureSerializer
-import org.locationtech.geomesa.features.kryo.impl.{KryoFeatureDeserialization, KryoFeatureSerialization}
+import org.locationtech.geomesa.features.kryo.impl.KryoFeatureDeserialization
+import org.locationtech.geomesa.features.serialization.ObjectType
 import org.locationtech.geomesa.index.iterators.IteratorCache
-import org.locationtech.geomesa.utils.cache.CacheKeyGenerator
+import org.locationtech.geomesa.utils.collection.IntBitSet
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.opengis.feature.simple.SimpleFeatureType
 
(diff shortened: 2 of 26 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStore.scala
```diff
@@ -198,6 +198,10 @@ class AccumuloDataStore(val connector: Connector, override val config: AccumuloD
       }
     }
 
+    if (sft.getVisibilityLevel == VisibilityLevel.Attribute && sft.getAttributeCount > 255) {
```
jnh5y: *blinks*  Ok, I'd love to know where this came from:)

elahrvivaz: When I was reviewing the changes, I realized that our attribute level vis would not work correctly if you had more than 127 (Byte.MaxValue) attributes, because the attributes are parsed out of the col family without any delimiters [here](https://github.com/locationtech/geomesa/blob/geomesa_2.11-2.3.1/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala#L62).
I switched the logic there to use unsigned bytes, which will be the same for 0-128 but lets us also go up to 255, then I added the explicit check here so that we'd fail-fast up front instead of getting bad behavior at runtime.
We could fix it by delimiting the col family values, but that would introduce back-compatibility issues, and 255 seems reasonable enough...

elahrvivaz:
Some background:

The old serialization scheme using variable length int encoding to store the byte offsets of each attribute. Because the offset block length wasn't known up front (due to the variable length encoding), we put the offset values after all the serialized attributes (we actually could have computed the variable length up front, as we know the number of attributes and which ints take more than one byte, but we didn't do that). During serialization, we left a 4 byte space at a known offset that we back-filled with the location of the offsets block, using a fixed-length 4-byte int encoding. When decoding, we read the offset block location, then skipped ahead to the offsets and read them into an `Array[Int]`. For lazy deserialization, we could reference the offset to a particular feature by looking at the offset array. This scheme worked pretty well, however it required the allocation of the offset array for each feature.

This PR updates the serialization scheme to use a fixed-length 2-byte encoding for the offsets. Because the length is known up front, we can put the offsets at the start of the serialized value, and know that the offset for attribute `i` will be at a particular spot. This lets us avoid allocating an array for the offests - instead, when reading, we can skip to the offset value, read it, then skip to the serialized attribute value. Because we use 2-byte shorts, we can't handle more than 32k attributes (!?) or 32MB total size of the serialized feature.

An additional change is to use a bitset to store the null values - this bitset is loaded up once for each feature, so represents a small allocation, however it is optimized to only require 1-2 Ints (and no array) for schemas with less than 64 attributes. For larger schemas, it allocates an `Array[Int]`, but only requires one Int per 32 attributes. (this could be optimized further by adding additional IntBitSet implementations if needed). Because all the null checks are stored in the bitset, we save 1 byte per attribute in the serialized array (minus the 4 bytes per 32 attributes required for the bitset).

elahrvivaz:
@jahhulbert-ccri do you think adding additional BitSet impls [here](https://github.com/locationtech/geomesa/blob/61defb088f88c61a286238cfd6e60009370d745a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/IntBitSet.scala#L106-L246) would be worthwhile to avoid allocating an array for schemas with > 64 attributes?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -26,52 +27,117 @@ import org.opengis.feature.simple.SimpleFeatureType
...
-    val cacheKey = CacheKeyGenerator.cacheKey(sft)
-    nullBytes = KryoFeatureSerialization.getWriters(cacheKey, sft).map { writer =>
-      output.clear()
-      writer(output, null)
-      output.toBytes
+    count = sft.getAttributeCount
+    if (attributes == null || attributes.length < count) {
+      attributes = Array.ofDim[(Array[Byte], Int, Int)](count)
     }
+    nullBytesV2 = null // lazily initialized when we hit v2 encoded data
```
jnh5y: Doesn't line 33 set this to null already?  Or is the _ an empty array?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/IntBitSet.scala
```diff
@@ -0,0 +1,508 @@
...
+
+  /**
+    * Serialize to a kryo output stream
+    *
+    * @param output output
+    */
+  def serialize(output: Output): Unit
+}
+
+object IntBitSet {
```
jnh5y: This is an improvement over other BitSets in that it can be very small?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -26,52 +27,117 @@ import org.opengis.feature.simple.SimpleFeatureType
...
       return values.get(0)
     }
+    // TODO if we don't have a geometry, skip the record?
+    values.get(0).get.head match {
+      case KryoFeatureSerializer.Version  => encodeV3(keys, values)
+      case KryoFeatureSerializer.Version2 => encodeV2(keys, values)
+    }
+  }
+
+  private def encodeV3(keys: java.util.List[Key], values: java.util.List[Value]): Value = {
```
jnh5y: Could you add a comment to describe the v3 and v2 formats?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -146,21 +148,34 @@ object ScalaSimpleFeature {
...
+      private var readUserData: () => java.util.Map[AnyRef, AnyRef]
+    ) extends AbstractImmutableSimpleFeature(sft, id) {
+
+    private val bits = IntBitSet(sft.getAttributeCount)
+    private lazy val attributes = Array.ofDim[AnyRef](sft.getAttributeCount)
+    private var count = 0
+
+    override lazy val getUserData: java.util.Map[AnyRef, AnyRef] = {
+      val read = synchronized { readUserData() }
+      readUserData = null // dereference any backing resources
```
jnh5y: Ok, I'll bite.  Can you explain this?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -146,21 +148,34 @@ object ScalaSimpleFeature {
...
+      Collections.unmodifiableMap(read)
+    }
 
     override def getAttribute(index: Int): AnyRef = {
-      if (bits.add(index)) {
-        attributes(index) = synchronized { readAttribute(index) }
+      if (readAttribute != null) {
+        bits.synchronized {
+          if (bits.add(index)) {
+            attributes(index) = synchronized { readAttribute(index) }
```
jnh5y: Do we need to worry about all the synchronized calls in here?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
```diff
@@ -91,27 +118,14 @@ class KryoBufferSimpleFeature(sft: SimpleFeatureType,
...
-      i += 1
-    }
-    if (i < offsets.length) {
-      // attributes have been added to the sft since this feature was serialized
-      missingAttributes = true
-      do { offsets(i) = -1; i += 1 } while (i < offsets.length)
-    } else {
-      missingAttributes = false
+    delegate = input.readByte() match {
+      case KryoFeatureSerializer.Version  => delegateV3
```
jnh5y: Should Version 3 be named Version3?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/ActiveDeserialization.scala
```diff
@@ -6,38 +6,44 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.features.kryo.impl
+package org.locationtech.geomesa.features.kryo
+package impl
```
jnh5y: What cool thing does this do again?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureSerialization.scala
```diff
@@ -47,155 +49,191 @@ trait KryoFeatureSerialization extends SimpleFeatureSerializer {
...
+      output.writeInt(list.size(), true)
+      val iter = list.iterator()
+      while (iter.hasNext) {
+        elements(output, iter.next())
+      }
+    }
+  }
+
+  case class KryoMapWriter(keys: KryoAttributeWriter, values: KryoAttributeWriter) extends KryoAttributeWriter {
+    // TODO handle null keys/values?
```
jnh5y: Do we have unit tests which cover this?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureSerialization.scala
```diff
@@ -47,155 +49,191 @@ trait KryoFeatureSerialization extends SimpleFeatureSerializer {
...
+      KryoGeometrySerialization.serializeWkb(output, value.asInstanceOf[Geometry])
+  }
+
+  case class KryoGeometryTwkbWriter(precision: GeometryPrecision.TwkbPrecision) extends KryoAttributeWriter {
+    override def apply(output: Output, value: AnyRef): Unit =
+      KryoGeometrySerialization.serialize(output, value.asInstanceOf[Geometry], precision)
+  }
+
+  case class KryoListWriter(elements: KryoAttributeWriter) extends KryoAttributeWriter {
+    // TODO handle null elements?
```
jnh5y: Do we have unit tests which cover this?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
elahrvivaz: it would be null from the `_`, but can't an iterator instance be re-used? the init method gets called before every scan, so if e.g. the sft changes the null bytes would need to be re-calculated

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/ActiveDeserialization.scala
elahrvivaz: this brings the `kryo` package into scope, so you can use the package object and classes without importing them. if you're interested, more details here: https://www.artima.com/scalazine/articles/chained_package_clauses_in_scala.html

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
elahrvivaz: good call, thanks. I had comments in-line in the code, but capturing the overall layout is useful. I added it in the scaladocs for `KryoFeatureSerializer` and `KryoFeatureDeserializationV2`

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: once the user data has been deserialized, we set the read function to null, that way the garbage collector can reclaim the serialized byte array and any other intermediate constructs used for deserialization. otherwise we're hanging on to both the serialized bytes and the deserialized objects, so 2x the memory footprint.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: do you mean the nested synchronized blocks, or just the synchronized blocks in general? I am fairly certain the nested blocks can't cause a deadlock, as the inner locks are not dependent on anything else. we need to synchronize on reading attributes/user data, as it uses the same single-threaded kryo reader under the hood. previously we were using an AtomicBitSet to avoid synchronizing on the bitset, but I think that wasn't entirely thread-safe in that it could potentially return null attributes while an attribute was being read. the synchronized blocks here are guarded by a null check, so once all the attributes have been read it won't synchronize on access anymore. I am fairly sure that is safe to do, since we are only setting `readAttribute` null inside a synchronized block, after all attributes are read.

elahrvivaz: so, in short, yes we should worry about it but it's necessary and I think it's implemented correctly :)

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureSerialization.scala
elahrvivaz: no, I don't think that we had any tests for this, and I didn't add any. do you think we should support null values in a list or map? I can add it, but it would be some extra bytes per list element.

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureSerialization.scala
elahrvivaz: no, same as above, do you think we should support it?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/collection/IntBitSet.scala
elahrvivaz: it just minimizes the serialization footprint - normally bitsets store values in a Long, so 64 values and 8 bytes per 'word'. using ints instead means 32 values and 4 bytes per word, so we can save some bytes if we have between 0-32 values, or 64-96, etc.

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoBufferSimpleFeature.scala
elahrvivaz: we had kept `Version` pointing to the most recent version (it used to be 2), but yeah I can see it either way. what do you think?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoFeatureSerializer.scala
```diff
@@ -15,6 +15,27 @@ import org.locationtech.geomesa.features.kryo.impl.LazyDeserialization.{Immutabl
...
+/**
+  * Kryo feature serialization and deserialization.
+  *
+  * The current serialization scheme (version 3):
+  *
+  * <ol>
+  *   <li>one byte containing the serialization version (3)</li>
+  *   <li>two bytes for a short containing the number of serialized attributes. this supports
+  *       appending attributes to the schema, as we know how many attributes were written when deserializing</li>
+  *   <li>a known number of bytes for metadata, which consists of:
```
jnh5y: Do we have unit tests which verify that one can add attributes and read 'old' data?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoFeatureSerializer.scala
```diff
@@ -15,6 +15,27 @@ import org.locationtech.geomesa.features.kryo.impl.LazyDeserialization.{Immutabl
...
+  *     <ol>
+  *       <li>2 byte short per attribute for the offset to the start of the attribute in the serialized bytes</li>
```
jnh5y: This puts a roughly 65kb upper limit on the size of a serialized SimpleFeature, right?  v2 had a varInt, right?  (Making sure I'm understanding; no judgment here:))

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureDeserializationV2.scala
```diff
@@ -24,6 +24,18 @@ import org.opengis.feature.simple.SimpleFeatureType
 
 import scala.util.control.NonFatal
 
+/**
+  * Version 2 serialization scheme consists of:
```
jnh5y: Can new SFTs select the v2 Kryo format?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoFeatureSerializer.scala
elahrvivaz: ah, i guess it would actually be 32kb since we're using signed shorts. for some reason in my head it was 32mb...  maybe this limit is too small

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureDeserializationV2.scala
elahrvivaz: no... we could make that an option i guess. ideally there'd be no reason to

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/KryoFeatureSerializer.scala
elahrvivaz: yeah, we have the 'back compatibility tests' and there's also an accumulo test here: https://github.com/locationtech/geomesa/blob/master/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreAlterSchemaTest.scala

PR #2333, opened 2019-08-09
GEOMESA-2690 Provide callback for handling kafka consumer errors

elahrvivaz:

### DIFF
--- a/geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
+++ b/geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
@@ -17,6 +17,7 @@ import org.apache.kafka.clients.consumer.{Consumer, ConsumerRecord, OffsetAndMet
 import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.errors.{InterruptException, WakeupException}
 import org.locationtech.geomesa.kafka.KafkaConsumerVersions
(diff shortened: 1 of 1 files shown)


geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
```diff
@@ -110,3 +114,21 @@ abstract class ThreadedConsumer(
...
+  trait ConsumerErrorHandler {
+
+    /**
+      * Invoked on a fatal error
+      *
+      * @param consumer consumer identifier
+      * @param e exception
+      * @return true to continue processing, false to terminate the consumer
+      */
+    def handle(consumer: String, e: Throwable): Boolean
```
jnh5y: Ok, this is cool.  How does one configure one?:)

elahrvivaz: you pass it off to the `startConsumers` method here: https://github.com/locationtech/geomesa/pull/2333/files#diff-7f9d3345259eca082a59f009c3f485a5R52

jnh5y: Ok, so one would have to do this quite manually?

elahrvivaz: Yeah, the idea is to give access to the extending class, as currently everything is private and inaccessible, but without exposing too much inner logic.

PR #2335, opened 2019-08-12
GEOMESA-2688 Fix heatmap rendering outside [-180,180] longitude

elahrvivaz:
* Geometries are translated to overlap the rendering envelope
* Filters extracted from the envelope are split for AntiMeridian handling
* Translation logic consolidated in org.locationtech.geomesa.utils.geotools.RenderingGrid

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
@@ -7,57 +7,44 @@
  ***********************************************************************/
 
 package org.locationtech.geomesa.index.iterators
-import java.awt.image.BufferedImage
-
 import com.typesafe.scalalogging.LazyLogging
+import org.geotools.filter.text.ecql.ECQL
 import org.geotools.util.factory.Hints
 import org.geotools.util.factory.Hints.ClassKey
-import org.geotools.filter.text.ecql.ECQL
 import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.locationtech.geomesa.features.kryo.impl.{KryoFeatureDeserialization, KryoFeatureSerialization}
 import org.locationtech.geomesa.filter.FilterHelper
 import org.locationtech.geomesa.index.api.GeoMesaFeatureIndex
 import org.locationtech.geomesa.index.api.QueryPlan.ResultsToFeatures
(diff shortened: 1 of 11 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
```diff
@@ -203,235 +198,110 @@ object DensityScan extends LazyLogging {
...
 
   /**
     * Evaluates an arbitrary expression against the simple feature to return a weight
     */
-  private def getWeightFromExpression(e: Expression)(sf: SimpleFeature): Double = {
-    val d = e.evaluate(sf, classOf[java.lang.Double])
-    if (d == null) 0.0 else d
+  class WeightByExpression(e: Expression) extends Weigher {
+    override def weight(sf: SimpleFeature): Double = {
+      val d = e.evaluate(sf, classOf[java.lang.Double])
```
jnh5y: Does this have any issues if the type of the attribute is an int / long / not-a-double?

geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/geotools/GeometryUtilsTest.scala
```diff
@@ -37,5 +38,36 @@ class GeometryUtilsTest extends Specification {
         min must beLessThanOrEqualTo(max)
       }
     }
+
+    "split bounding box envelopes" >> {
+      val unchanged = Seq(
+        new ReferencedEnvelope(-180, 180, -90, 90, CRS_EPSG_4326),
```
jnh5y: What happens if a bbox covers and contains this one?  E.g., if the bbox is larger than the world?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/DensityScan.scala
elahrvivaz: when you pass in classOf double, it tries to convert the result to a double, using the geotools Converters class.

geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/geotools/GeometryUtilsTest.scala
elahrvivaz: thanks, added a check and test for that.

PR #2339, opened 2019-08-13
GEOMESA-2672 Stats - skip null GroupBy values

elahrvivaz:


jnh5y:
Is this a unit test that sets a null key?

elahrvivaz:
ha, actually there was not... I added one. The previous changes were because I changed the json output of group by to be a map instead of a list of pairs, which conceptually makes more sense.

PR #2342, opened 2019-08-14
GEOMESA-2692 Kudu - Fix metadata scans for index tables

elahrvivaz:


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/GeoMesaFeatureIndex.scala
```diff
@@ -349,7 +349,7 @@ abstract class GeoMesaFeatureIndex[T, U](val ds: GeoMesaDataStore[_],
     import StringSerialization.alphaNumericSafeString
 
     val prefix = (ds.config.catalog +: Seq(sft.getTypeName, name).map(alphaNumericSafeString)).mkString("_")
-    val suffix = s"_v$version${partition.map(p => s"_$p").getOrElse("")}"
+    val suffix = s"v$version${partition.map(p => s"_$p").getOrElse("")}"
```
jnh5y: Does this break things?!

elahrvivaz: no, once the table name is stored in the metadata it will use that. This only gets called for new tables - I noticed that an extra underscore had crept in, so I went ahead and fixed it.

PR #2348, opened 2019-08-15
GEOMESA-1614 CLI - use feature name for spec lookup

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
@@ -305,7 +305,7 @@ object SimpleFeatureTypes {
   def encodeUserData(sft: SimpleFeatureType): String = {
     val prefixes = sft.getUserDataPrefixes
     val result = new StringBuilder(";")
-    sft.getUserData.asScala.foreach { case (k, v) =>
+    sft.getUserData.asScala.foreach { case (k: AnyRef, v: AnyRef) =>
       if (v != null && prefixes.exists(k.toString.startsWith)) {
         result.append(encodeUserData(k, v)).append(",")
       }
@@ -313,7 +313,7 @@ object SimpleFeatureTypes {
     if (result.lengthCompare(1) > 0) { result.substring(0, result.length - 1) } else { "" }
   }
 
-  def encodeUserData(data: java.util.Map[AnyRef, AnyRef]): String = {
+  def encodeUserData(data: java.util.Map[_ <: AnyRef, _ <: AnyRef]): String = {
     if (data.isEmpty) { "" } else {
       val result = new StringBuilder(";")
       data.asScala.foreach { case (k, v) =>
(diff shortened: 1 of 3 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -447,6 +447,59 @@ object SimpleFeatureTypes {
...
+    *
+    * <ul>
+    *   <li>a positive int if type A is a sub type or reorder of type B</li>
+    *   <li>zero if type A has the same schema as type B</li>
+    *   <li>a negative int if none of the above</li>
+    * </ul>
+    *
+    * Note that this shouldn't be used for sorting.
+    *
+    * Note: GeoTools DataUtilities has a similar method that does not work correctly
```
jnh5y: How hard would it be to submit a PR for a fix to GeoTools?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -447,6 +447,59 @@ object SimpleFeatureTypes {
...
+    *
+    * @param a schema A
+    * @param b schema B
+    * @return comparison
+    */
+  def compare(a: SimpleFeatureType, b: SimpleFeatureType): Int = {
```
jnh5y: Oh man, does this compare function satisfy the contract for Comparator: https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: I think the main issue is this [line](https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java#L472) should be `a.getType.getBinding.equals(b.getType.getBinding)` - I'm somewhat hesitant to change the behavior though... What do you think?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/sft/package.scala
```diff
@@ -26,41 +26,44 @@ package object sft {
...
-                                     userData: java.util.Map[AnyRef, AnyRef])
-      extends AttributeDescriptorImpl(`type`, name, minOccurs, maxOccurs, isNillable, defaultValue) {
+  class ImmutableAttributeDescriptor(
+      `type`: AttributeType,
+      name: Name,
+      minOccurs: Int,
+      maxOccurs: Int,
+      isNillable: Boolean,
+      defaultValue: AnyRef,
+      ud: java.util.Map[_ <: AnyRef, _ <: AnyRef]
```
jnh5y: Meh, any reason to abbreviate userData here?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: no, that's why it says don't use for sorting :)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/sft/package.scala
elahrvivaz: there's a protected/package level variable called 'userData' in one of the super-classes that makes intellij complain when we reference this one...

elahrvivaz: Although maven seems to handle it ok either way

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
jnh5y: Hmmm....  if it is 'broken' as-is, then there ought to be a unit test which would show it.   

If it isn't a major pain to gin up a test and patch, then we could discuss it with the community in a PR. 

That said, DataUtilities is considered 'sample' code; despite that, (I think) it is used widely.  Those two considerations are at odds with each other, so GeoTools may or may not merge the fix.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
jnh5y: Just like a compiler, I don't read comments.

Unlike a compiler, I only read some of the actual code.

I wonder if there's a better name for this function (so that folks don't blow up in a sort)....

elahrvivaz: I'm not even sure what sorting simple feature types would mean... I copied the name from data utilities for consistency, but I can change it if you want. We're also not using the 'subtype' comparison part, but it seemed like we might as well have it available if needed.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
elahrvivaz: I'll put up a pr and we can see.

elahrvivaz: https://github.com/geotools/geotools/pull/2525

PR #2351, opened 2019-08-19
GEOMESA-2624 Improve wildcard handling

elahrvivaz:
* Handle '_' and '%' in attribute indices
* Handle non-trailing wildcards

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
@@ -898,10 +898,16 @@ class AttributeIndexStrategyTest extends Specification with TestWithDataStore {
 
   "AttributeIndexLikeStrategy" should {
 
-    "correctly query on strings" in {
+    "correctly query on strings with multi-char match" in {
       val features = execute("name LIKE 'b%'")
       features must haveLength(2)
       features must contain("bill", "bob")
+    }
+
+    "correctly query on strings with single-char match" in {
+      val features = execute("name LIKE 'b_b'")
+      features must haveLength(1)
+      features must contain("bob")
     }
 
     "correctly query on non-strings" in {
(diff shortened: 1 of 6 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -337,15 +337,18 @@ object FilterHelper {
...
-            // Remove the trailing wildcard and create a range prefix
+            // find the first wildcard and create a range prefix
             val literal = f.getLiteral
-            val lower = if (literal.endsWith(MULTICHAR_WILDCARD)) {
-              literal.substring(0, literal.length - MULTICHAR_WILDCARD.length)
-            } else {
-              literal
+            var i = literal.indexWhere(Wildcards.contains)
+            // check for escaped wildcards
+            while (i > 1 && literal.charAt(i - 1) == '\\' && literal.charAt(i - 2) == '\\') {
```
elahrvivaz: @jnh5y does this seem like the right escape behavior?

elahrvivaz: seems to be... https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/filter/LikeFilterImpl.java#L47

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/index/AttributeIndexStrategyTest.scala
```diff
@@ -898,12 +898,18 @@ class AttributeIndexStrategyTest extends Specification with TestWithDataStore {
...
       val features = execute("name LIKE 'b%'")
       features must haveLength(2)
       features must contain("bill", "bob")
     }
 
+    "correctly query on strings with single-char match" in {
+      val features = execute("name LIKE 'b_b'")
+      features must haveLength(1)
+      features must contain("bob")
+    }
```
jnh5y: Should we add a word like 'bobbed' to the test to show that an underscore doesn't match multi-chars?

elahrvivaz: sure, I'll add it to the one in index-api, so I don't have to update all the other tests in this class.

PR #2360, opened 2019-09-09
GEOMESA-2707 Converters - check for nulls in date parsing methods

elahrvivaz:
* Date parsing methods pass through nulls instead of throwing exceptions
  * Fix the anti-pattern of `try(date($0), null)`, which is expensive
* `require` function added to preserve old behavior if desired

### DIFF
--- a/docs/user/convert/function_overview.rst
+++ b/docs/user/convert/function_overview.rst
@@ -104,6 +104,7 @@ Control Functions
 
 -  ``try``
 -  ``withDefault``
+-  ``require``
 
 State Functions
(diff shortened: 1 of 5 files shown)


jnh5y:
I like this.  Two questions:

1.  With the pattern "try(date($0), null)", the try covers the case where the input is misformatted, right?

jnh5y:
2.  Would this make sense to backport to 2.3.x?

elahrvivaz:
yeah, `try` is still useful for misformatted dates. I would err on not back porting since it changes the behavior in a somewhat subtle way...

PR #2361, opened 2019-09-10
GEOMESA-2708 Strict index to table for adopt

khobbs-ccri:
Make the mapping of indexes to user provided tables strict by
  1. Reversing the direction of the match from table -> index to
     index -> table.
  2. Matching the index with the longest attribute string first.
  3. Match only if the table has not already been matched.
  4. Match only if all of the index name, version and attributes
     are included in the table name.

Signed-off-by: Kevin Hobbs <khobbs@ccri.com>

### DIFF
--- a/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/data/ManagePartitionsCommand.scala
+++ b/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/data/ManagePartitionsCommand.scala
@@ -12,6 +12,7 @@ import java.time.{ZoneOffset, ZonedDateTime}
 
 import com.beust.jcommander.{JCommander, Parameter, ParameterException, Parameters}
 import org.locationtech.geomesa.features.ScalaSimpleFeature
+import org.locationtech.geomesa.index.api.GeoMesaFeatureIndex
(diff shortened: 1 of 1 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/data/ManagePartitionsCommand.scala
```diff
@@ -110,22 +111,25 @@ object ManagePartitionsCommand {
...
-        if (matchedByName.lengthCompare(1) == 0) {
-          (matchedByName.head._1, table)
-        } else {
-          val matchedByAttributes = matchedByName.filter(m => table.contains(m._3))
-          if (matchedByAttributes.lengthCompare(1) == 0) {
-            (matchedByAttributes.head._1, table)
+      val tableToIndexName = indexNames
+        .toList
+        .sortWith(_._3.size > _._3.size) // Sorted so we get best match first
+        .foldLeft(Map[String,(GeoMesaFeatureIndex[_,_],String,String,String)]()){
```
elahrvivaz: nit: this works, but probably would be simpler code to build up a mutable map in a foreach loop. don't need to change it though

PR #2364, opened 2019-09-17
GEOMESA-2710 Converters - resolve substitutions in config parsing

elahrvivaz:
* Allow specifying sft if there are multiple present

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypeLoader.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypeLoader.scala
@@ -9,16 +9,11 @@
 package org.locationtech.geomesa.utils.geotools
 
 import java.net.URL
-import java.util.{ServiceLoader, List => JList}
+import java.util.Collections
 
-import com.typesafe.config.{Config, ConfigFactory, ConfigRenderOptions}
-import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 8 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypeLoader.scala
```diff
@@ -28,91 +23,65 @@ import scala.collection.JavaConverters._
...
-  }
-  
-  // Will also pick things up from the SystemProperties
-  def configURLs: Seq[URL] = {
-    val config = ConfigFactory.load()
-    if (config.hasPath(SftConfigURLs)) {
-      config.getAnyRef(SftConfigURLs) match {
-        case s: String          => s.split(',').map(s => s.trim).toList.map(new URL(_))
-        case lst: JList[String] => lst.map(new URL(_))
+    // Will also pick things up from the SystemProperties
```
jnh5y: Are all the ways of applying substitutions documented?  Would it be easy to add that to the docs?

elahrvivaz: do you mean HOCON substitutions? we don't explicitly document it, but it's in the HOCON docs themselves.

elahrvivaz: which we link to, in some places at least

PR #2371, opened 2019-09-25
GEOMESA-2714 ADSBx Converter

malinsinbigler:
Closes https://geomesa.atlassian.net/browse/GEOMESA-2714


geomesa-tools/conf/sfts/adsbx/reference.conf
```diff
@@ -0,0 +1,248 @@
...
+      user-data {
+        "geomesa.indices.enabled"           = "attr,z3"
+        "geomesa.table.compression.enabled" = "true"
+        "geomesa.table.compression.type"    = "snappy"
+        "parquet.summary.metadata.level"    = "ALL"
+        "parquet.compression"               = "SNAPPY"
+        "geomesa.table.sharing"             = "false"
+        "geomesa.attr.splits"               = "4"
+        "geomesa.index.tiering"             = "date"
+        "geomesa.z.splits"                  = "4"
```
elahrvivaz: we probably want to get rid of most of these - they seem deployment specific and not really related to the data type per say.

geomesa-tools/conf/sfts/adsbx/reference.conf
```diff
@@ -0,0 +1,248 @@
...
+        { name = "Sat"                , type = "Boolean"                  }
+        { name = "PosStale"           , type = "Boolean"                  }
+        { name = "dtg"                , type = "Date"    , default = "true"                 }
+        { name = "geom"               , type = "Point"   , srid = "4326" , default = "true" }
+      ]
+    }
+  } // end sfts
+
+  converters = {
+    adsbx-Base = {
```
elahrvivaz: [minor] maybe make this lower case to match the rest

geomesa-tools/conf/sfts/adsbx/reference.conf
```diff
@@ -0,0 +1,248 @@
...
+      type = json
+      options = {
+        validators = ["z-index", "has-geo", "has-dtg"]
```
elahrvivaz: you can just make this `validators = [ "index" ]` - that will work with whatever indices get created, and the has-geo and has-dtg are redundant if there's an index validator

geomesa-tools/conf/sfts/adsbx/reference.conf
```diff
@@ -0,0 +1,248 @@
...
+        { name = "Sat"                , json-type = "bool"       , path = "$.Sat"          }
+        { name = "PosStale"           , json-type = "bool"       , path = "$.PosStale"     }
+        { name = "dtg"                , transform = "withDefault($PosTime, now())"         }
+        { name = "geom"               , transform = "point($Long::double, $Lat::double)"   }
+      ]
+    }
+
+    adsbx = ${geomesa.converters.adsbx-Base} {
+      id-field = "md5(stringToBytes(concatenate($Icao::string, $dtg::string, $Lat::string, $Long::string)))"
+      //feature-path = "$.acList[*]"
```
elahrvivaz: this one and adsbx-single are the same, right? maybe we should make this the default, then add one for adsbx-multi that has the feature path uncommented.

geomesa-tools/conf/sfts/adsbx/reference.conf
```diff
@@ -0,0 +1,248 @@
...
+    }
+  }
+  vsit_codes = {
+    type = "simple"
+    data = {
+      "0" = { code = "Barometric" },
+      "1" = { code = "Geometric" }
+    }
+  }
+  adsbx_static = {
```
elahrvivaz: this cache would need to be populated with something, right? i'm hesitant to include it in the default, given that it won't really work out of the box (or it will just throw a lot of errors/timeouts?). maybe we should have separate converters that include the enrichment, e.g. `adsbx-enriched`, `adsbx-multi-enriched`, etc. Per Austin's comments, some more mention of this in the docs would be helpful too.

elahrvivaz:
I have a few comments but looks pretty good, thanks for adding it!

geomesa-tools/conf/sfts/adsbx/reference.conf
malinsinbigler: Can caches be defined within a converter?

malinsinbigler: Or are you talking about making references to a cache within a different converter?

elahrvivaz: yeah, have one set of converters that don't invoke the redis cache, and then a second set that do (but require some additional work, since you'd have to set up redis and populate it)

elahrvivaz: i think caches have to be defined outside converters

geomesa-tools/conf/sfts/adsbx/reference.conf
malinsinbigler: Are there any that should be retained?

docs/user/convert/premade/adsbx.rst
```diff
@@ -0,0 +1,37 @@
...
+The converter definitions provided include:
+- adsbx : For historical storage. Uncomment ``feature-path`` to ingest a json array.
+- adsbx-single : For historical storage.
+- adsbx-live-single : For live storage Ex. Kafka, Redis
+
+These extend from a common converter named ``adsbx-base`` and provide additional configuration.
+For historical data storage, ``id-field`` is a unique concatenation formed from Icao, dtg, Lat, and Long.
+For live storage, only Icao is used as the ``id-field`` which ensures that the most recently processed observation is correctly presented for that aircraft.
+
+While not required for ingest, the user can optionally set up a redis instance as referenced by the ``adsbx_static`` cache in the converter.
```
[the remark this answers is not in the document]
malinsinbigler: All references to enrichment removed

PR #2372, opened 2019-10-02
GEOMESA-2716 Support embedded hadoop configs in HBase and FSDS

elahrvivaz:
* New data store parameters "hbase.config.xml" and "fs.config.xml"

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
@@ -8,6 +8,7 @@
 
 package org.locationtech.geomesa.hbase.data
 
+import java.io.File
 import java.util.Collections
 
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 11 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
```diff
@@ -334,6 +335,22 @@ class HBaseDataStoreTest extends HBaseTest with LazyLogging {
...
+          |    <name>foo.embedded</name>
+          |    <value>bar</value>
+          |  </property>
+          |</configuration>""".stripMargin
+      val resource = new File(getClass.getClassLoader.getResource("custom-site.xml").toURI).getAbsolutePath
+      val params = Map(ConfigsParam.getName -> embedded, ConfigPathsParam.getName -> resource)
+      val conf = HBaseConnectionPool.getConfiguration(params.asJava)
+
+      conf.get("foo.embedded") mustEqual "bar"
+      conf.get("foo.resource") mustEqual "baz"
```
jnh5y: How does this interact with any *-site.xml files which may be in the environment?

jnh5y:
Hmm... is there a good place to document using these options?

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseDataStoreTest.scala
elahrvivaz: it will load the default resources as per usual. I *think* these additional ones will take precedence if there is a conflict, since they're loaded second, but I'm not sure about that and wouldn't guarantee it.

elahrvivaz:
I added it in the data store docs for each store. We can probably call it out in the release notes as well. Were you thinking of something else?

jnh5y:
> I added it in the data store docs for each store. We can probably call it out in the release notes as well. Were you thinking of something else?

Nope, sounds good.

PR #2384, opened 2019-10-21
GEOMESA-2718 HBase - Add server-side coprocessor timeouts

elahrvivaz:
* Make HBaseConnectionPool lazily initialize

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/GeoMesaCoprocessor.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/GeoMesaCoprocessor.scala
@@ -13,7 +13,6 @@ import java.util.Collections
 import java.util.concurrent._
 import java.util.concurrent.atomic.AtomicBoolean
 
-import com.google.common.util.concurrent.MoreExecutors
 import com.google.protobuf.{ByteString, RpcCallback, RpcController, Service}
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.hadoop.hbase.client.coprocessor.Batch.Call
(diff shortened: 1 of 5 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/GeoMesaCoprocessor.scala
```diff
@@ -94,14 +94,13 @@ class GeoMesaCoprocessor extends GeoMesaCoprocessorService with Coprocessor with
           scanner.close()
         }
       }
-
-      GeoMesaCoprocessorResponse.newBuilder.addAllPayload(results.map(ByteString.copyFrom).asJava).build
     } catch {
-      case e: IOException => ResponseConverter.setControllerException(controller, e); null
-      case NonFatal(e) => ResponseConverter.setControllerException(controller, new IOException(e)); null
+      case _: InterruptedException | _ : InterruptedIOException => // stop processing, but don't return an error to prevent retries
```
jnh5y: Are there situations where we ought to retry?!

elahrvivaz: I think if we get an interrupted exception then we don't want to retry... if we get a non-interrupted exception then we return that and it causes hbase to retry the call. At least, that's the whole basis for this PR :)

@malinsinbigler do you have time to test out the interrupt with this branch?

malinsinbigler: Sure thing, I'll test the 2.4.0 co-processor

malinsinbigler: @elahrvivaz running `./geomesa-hbase stats-count` does not time out as expected.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/coprocessor/GeoMesaCoprocessor.scala
```diff
@@ -60,18 +58,17 @@ class GeoMesaCoprocessor extends GeoMesaCoprocessorService with Coprocessor with
...
-      if (!notifier.canceled.get) {
+    try {
+      if (!controller.isCanceled) {
         val options = GeoMesaCoprocessor.deserializeOptions(request.getOptions.toByteArray)
+        val timeout = options.get(GeoMesaCoprocessor.TimeoutOpt).map(_.toLong + System.currentTimeMillis())
         val aggregator = {
           val classname = options(GeoMesaCoprocessor.AggregatorClass)
           Class.forName(classname).newInstance().asInstanceOf[HBaseAggregator[_]]
         }
+        logger.debug(s"Initializing aggregator $aggregator with options ${options.mkString(", ")}")
```
jnh5y: Just as a sanity check, the timeout config is logged, right?

elahrvivaz: yeah it would be in the options map

PR #2400, opened 2019-11-09
GEOMESA-2737 GeoCQEngine STRtree and Quadtree indexes support

rtotaro:
I provided the new features just via API, is not possible to configure them in the sft schema. This  avoids using them as indexes for the GeoCQEngineDataStore

Signed-off-by: Rodolfo Totaro <rodolfo_totaro@yahoo.it>

### DIFF
--- a/geomesa-memory/geomesa-cqengine/src/main/java/org/locationtech/geomesa/memory/cqengine/index/AbstractGeoIndex.java
+++ b/geomesa-memory/geomesa-cqengine/src/main/java/org/locationtech/geomesa/memory/cqengine/index/AbstractGeoIndex.java
@@ -1,0 +1,195 @@
+/***********************************************************************
+ * Copyright (c) 2013-2019 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.memory.cqengine.index;
+
+import com.googlecode.cqengine.attribute.Attribute;
+import com.googlecode.cqengine.index.Index;
+import com.googlecode.cqengine.index.support.AbstractAttributeIndex;
+import com.googlecode.cqengine.index.support.indextype.OnHeapTypeIndex;
+import com.googlecode.cqengine.persistence.support.ObjectSet;
+import com.googlecode.cqengine.persistence.support.ObjectStore;
+import com.googlecode.cqengine.query.Query;
+import com.googlecode.cqengine.query.option.QueryOptions;
+import com.googlecode.cqengine.resultset.ResultSet;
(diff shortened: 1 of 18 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/WrapperIndex.scala
```diff
@@ -0,0 +1,47 @@
...
+package org.locationtech.geomesa.utils.index
+
+import org.locationtech.jts.geom.{Envelope, Geometry}
+import org.locationtech.jts.index.{SpatialIndex => JSTSpatialIndex}
+
+import scala.collection.JavaConverters._
+
+class WrapperIndex[T, Index <: JSTSpatialIndex](val indexBuider : () => Index) extends SpatialIndex[T] with Serializable {
+
+  var index = indexBuider()
```
elahrvivaz: this should probably be private so we don't expose mutable state

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/WrappedSTRtree.scala
```diff
@@ -0,0 +1,24 @@
...
+
+package org.locationtech.geomesa.utils.index
+
+import org.locationtech.jts.index.strtree.STRtree
+
+class WrappedSTRtree[T](nodeCapacity:Int) extends WrapperIndex[T,STRtree](
+  indexBuider = () => new STRtree(nodeCapacity)
+) with Serializable {
+
+  def this()
```
elahrvivaz: [minor] instead of an alternate constructor, scala convention is to use a default value, e.g. `class WrappedSTRtree[T](nodeCapacity: Int = 10)`

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/WrapperIndex.scala
rtotaro: Changed visibility to protected: must be accessible to subclasses

geomesa-memory/geomesa-cqengine/src/main/java/org/locationtech/geomesa/memory/cqengine/index/AbstractGeoIndex.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Iterator;
+import java.util.Set;
+
+public abstract class AbstractGeoIndex<A extends Geometry, O extends SimpleFeature> extends AbstractAttributeIndex<A, O> implements OnHeapTypeIndex {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractGeoIndex.class);
+
+    private static final int INDEX_RETRIEVAL_COST = 40;
+
+    protected SpatialIndex<O> index;
+    protected int geomAttributeIndex;
```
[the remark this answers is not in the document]
rtotaro: ok for geomAttributeIndex, but Bucket index instantiate a different index in accord with geometry instance type so i cannot make index final

elahrvivaz:
thanks for the contribution!

PR #2406, opened 2019-11-19
GEOMESA-2743 FSDS - Custom file callbacks

elahrvivaz:
* SimpleFeatureTypes can define file writer callbacks
  * User data key is `geomesa.fs.observers`
  * Trait is `org.locationtech.geomesa.fs.storage.common.observer.FileSystemObserverFactory`

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/observer/FileSystemObserverFactory.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/observer/FileSystemObserverFactory.scala
@@ -12,7 +12,7 @@ package observer
 import java.io.Closeable
 
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.{FileContext, Path}
+import org.apache.hadoop.fs.Path
 import org.locationtech.geomesa.utils.io.{CloseQuietly, FlushQuietly}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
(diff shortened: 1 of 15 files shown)


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/observer/FileSystemObserverFactory.scala
```diff
@@ -0,0 +1,60 @@
...
+    override def close(): Unit = {}
+  }
+
+  /**
+   * Composite observer
+   *
+   * @param observers observers
+   */
+  class CompositeObserver(observers: Seq[FileSystemObserver]) extends FileSystemObserver {
+    override def write(feature: SimpleFeature): Unit = observers.foreach(_.write(feature))
```
jnh5y: How should this work in light two different types of issues/failures?

1.  How should missing Observers be handled?  (E.g., what if an SFT has a custom Observer which is not available on the writer's classpath?)
2.  How should exceptions during an Observer be handled?

elahrvivaz: 1. The observers are loaded lazily, so if you aren't writing anything then you don't need them on the classpath. If you are writing, then it will fail with a ClassNotFoundException - we could make it just log a warning, but I think if you set an observer you would want it to fail hard.
2. Right now observers would throw an exception and it would go back to the caller. Any subsequent observers wouldn't be invoked. On close, I made sure they would all get closed and then the exception would get propagated. I could add the same logic here, if you want. I think that failures in an observer should get propagated and not just logged though, same as if it can't be loaded up front. Individual observers can catch and log exceptions if they're not fatal.

PR #2409, opened 2019-11-25
Wschultz 2745 add int to boolean

w-schultz:
I've added a new transform function with tests. I also tested the other misc. transform function so the class was consistent.

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/transforms/MiscFunctionFactoryTest.scala
+++ b/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/transforms/MiscFunctionFactoryTest.scala
@@ -10,39 +10,39 @@ class MiscFunctionFactoryTest extends Specification {
   "MiscFunctionFactory" >> {
     "intToBoolean" should {
       "convert a 0 to false" >> {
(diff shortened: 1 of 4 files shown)


geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/transforms/MiscFunctionFactoryTest.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class MiscFunctionFactoryTest extends Specification {
+
+  "MiscFunctionFactory" >> {
+    "intToBoolean" should {
+      "convert a 0 to false" >> {
+        MiscFunctionFactory.intToBoolean(Array(0)) mustEqual false
```
elahrvivaz: FYI, you should be able to test functions through the Expression interface, as in https://github.com/locationtech/geomesa/blob/master/geomesa-convert/geomesa-convert-common/src/test/scala/org/locationtech/geomesa/convert2/transforms/ExpressionTest.scala#L42-L43. But the tests here are fine, too.

elahrvivaz:
Closing, replaced with #2411

PR #2412, opened 2019-11-26
GEOMESA-2746 Stats - Include bounds on Histogram json

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Histogram.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Histogram.scala
@@ -157,18 +157,14 @@ class Histogram[T](
 
   override def toJsonObject: Map[String, Any] = {
     val binSeq = Seq.tabulate(bins.length) { bin =>
-      val builder = ListMap.newBuilder[String, Any]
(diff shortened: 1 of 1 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Histogram.scala
```diff
@@ -157,18 +157,14 @@ class Histogram[T](
 
   override def toJsonObject: Map[String, Any] = {
     val binSeq = Seq.tabulate(bins.length) { bin =>
-      val builder = ListMap.newBuilder[String, Any]
-      builder.sizeHint(if (bin == 0) { 4 } else { 3 })
```
jnh5y: Is the size hint used after this change?

elahrvivaz: it's a fixed size map so i got rid of the builder construct and just created it directly. i believe that will size it appropriately...

elahrvivaz: tracing through the source code, it looks like the apply method being used now does not use a size hint - but the ListMap builder also ignores the size hint, so in actuality this call was a no-op.

PR #2418, opened 2019-12-03
GEOMESA-2748  Add support for dsParams which contains hbase Connection when creatin…

fsdzz:
Sometimes, our dsParams contains hbase connection, as shown below.
 val dsParams = Map(
      "hbase.connection"  -> connection,
      "hbase.catalog"     -> "xxx"
    )
For instance, we have to create a special hbase connection for kerberos . We use  "hbase.connection" rather than  "hbase.zookeepers", so that we can set related configuration files. 
However，we cannot create a rdd through "spatialRDDProvider.rdd(conf, sc, dsParams, query)" , for there is only Map[String, String] is allowed as a dsParams. 
Since we have this type of dsParams which contains hbase connection，we'd better add support for these dsParams to create rdd. Aimed at this case, I add a function createRdd, so that we can use "createRdd(conf, sc, dsParams, query) " to create rdd.


elahrvivaz:
Is there a something specific that requires you to pass in a connection instead of using hbase-site.xml to configure it? I don't think this approach is feasible, because spark is a distributed system - the parameter map has to be serialized and sent out to each worker. Since the HBase connection object is not serializable, it can't be passed in directly.

elahrvivaz:
As of 2.4.0, you can set [additional configuration files ](https://www.geomesa.org/documentation/user/hbase/usage.html#hbase-data-store-parameters)as parameters in the datastore. This may solve your issue with configuring the connection.

PR #2423, opened 2019-12-06
GEOMESA-2752 Fix before/after date queries against Z3 index for 2.2 or before

elahrvivaz:


jahhulbert:
Is this the same as an issue with dtg > abc? Wondering if we need to backport.

elahrvivaz:
I think the bug was only introduced in 2.3, so we should probably backport to 2.3.x. @jahhulbert Are you seeing an issue with `dtg > abc`?

PR #2424, opened 2019-12-09
GEOMESA-2754 Support intersection and difference of geometries in spark 

fsdzz:
We've have many methods for geometry in spark.  We can judge if geometies intersects or disjoint. However, getting the intersection or difference part is unsupported in spark now. It is expected to add support for getting intersection and difference geometries are often required.
I add support for getting intersection and difference of geometries in spark. The intersection and difference can be accessed using SQL queries of spark SQL, as follows.
image
image


fsdzz:
@elahrvivaz OK. I've added the new functions to the docs and signed the eclipse CLA.

PR #2429, opened 2019-12-13
GEOMESA-2757 Fix ilike attribute scans

elahrvivaz:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,7 +18,7 @@ Alternatively use ``sudo`` with the ``pip`` command to install the packages in t
 Optional:  if you want to build the PDF version of the manual, install LaTeX:
 
     # on Ubuntu
-    $ sudo apt-get install texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
+    $ sudo apt-get install texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended latexmk
 
 The LaTeX distribution is pretty big, so you can skip it if you're just interested in the HTML docs.
 You will also need ``make``.
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
@@ -9,15 +9,15 @@
 package org.locationtech.geomesa.filter
 
 import java.time.{ZoneOffset, ZonedDateTime}
-import java.util.Date
+import java.util.{Date, Locale}
 
 import com.typesafe.scalalogging.LazyLogging
 import org.geotools.data.DataUtilities
 import org.locationtech.geomesa.filter.Bounds.Bound
 import org.locationtech.geomesa.filter.expression.AttributeExpression.{FunctionLiteral, PropertyLiteral}
 import org.locationtech.geomesa.filter.visitor.IdDetectingFilterVisitor
-import org.locationtech.geomesa.utils.geotools.GeometryUtils
 import org.locationtech.geomesa.utils.date.DateUtils.toInstant
+import org.locationtech.geomesa.utils.geotools.GeometryUtils
 import org.locationtech.geomesa.utils.geotools.converters.FastConverter
 import org.locationtech.jts.geom._
 import org.opengis.feature.simple.SimpleFeatureType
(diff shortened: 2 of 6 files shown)


elahrvivaz:
I'll cherry-pick back to 2.4

docs/README.md
```diff
@@ -18,7 +18,7 @@ Alternatively use ``sudo`` with the ``pip`` command to install the packages in t
 Optional:  if you want to build the PDF version of the manual, install LaTeX:
 
     # on Ubuntu
-    $ sudo apt-get install texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
+    $ sudo apt-get install texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended latexmk
```
jnh5y: *blinks*  Why do we need latexmk now?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -418,6 +430,54 @@ object FilterHelper {
...
+    val upper = string.toUpperCase(Locale.US)
+    // account for chars without upper/lower cases, which we don't need to permute
+    val count = (0 until lower.length).count(i => lower(i) != upper(i))
+
+    if (count > max) {
+      FilterHelperLogger.log.warn(s"Not expanding case-insensitive prefix due to length: $string")
+      Seq.empty
+    } else {
+      // there will be 2^n different permutations, accounting for chars that don't have an upper/lower case
+      val permutations = Array.fill(math.pow(2, count).toInt)(Array(lower: _*))
```
jnh5y: I wonder if the call to 'fill' is 'expensive' here?  E.g., do we end up setting and resetting the values?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -418,6 +430,54 @@ object FilterHelper {
...
 
+  /**
+   * Calculates all the different case permutations of a string.
+   *
+   * For example, "foo" -> Seq("foo", "Foo", "fOo", "foO", "fOO", "FoO", "FOo", "FOO")
+   *
+   * @param string input string
+   * @return
+   */
+  private def casePermutations(string: String): Seq[String] = {
```
jnh5y: Oh man... this is nerd sniping at its best...  It feels like this ought to be something that can be done with the Scala collections library in one line.:)

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/package.scala
```diff
@@ -30,7 +30,9 @@ package object filter {
   implicit val ff: FilterFactory2 = CommonFactoryFinder.getFilterFactory2
 
   object FilterProperties {
-    val GeometryProcessing = SystemProperty("geomesa.geometry.processing", "spatial4j")
+    val GeometryProcessing: SystemProperty = SystemProperty("geomesa.geometry.processing", "spatial4j")
```
jnh5y: Did this line change?

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
```diff
@@ -195,6 +196,25 @@ class FilterHelperTest extends Specification {
...
+        matcher(f, expected, o => o.map(_ + WildcardSuffix))
+
+      literal("name ilike 'foo'", Seq("foo", "Foo", "fOo", "foO", "fOO", "FoO", "FOo", "FOO"))
+      wildcard("name ilike 'foo%'", Seq("foo", "Foo", "fOo", "foO", "fOO", "FoO", "FOo", "FOO"))
+      literal("name ilike 'f2o'", Seq("f2o", "F2o", "f2O", "F2O"))
+      wildcard("name ilike 'f2o%'", Seq("f2o", "F2o", "f2O", "F2O"))
+      // test max expansion limit of 10 chars
+      literal("name ilike 'fooooooooooo'", Seq.empty)
+      wildcard("name ilike 'fooooooooooo%'", Seq.empty)
+    }
```
jnh5y: As an *incredibly picky* ask, should we add a unit test where the System property controlling the length of wildcard expansion is overridden?

docs/README.md
elahrvivaz: eh, I tried building the pdf and it didn't work... found [this](https://github.com/sphinx-doc/sphinx/issues/3543) which indicates it requires latexmk now. installed that, and it worked

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
elahrvivaz: we set them originally to the lowercase bytes (copying the array for each permutation). then we overwrite the uppercase ones in our loop. so the array values are being modified later but the arrays are just created here up front. alternatively we could instantiate empty arrays and then overwrite both upper and lower case chars in the loop, but i don't think that would be any better...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
elahrvivaz: yeah, i had a `foldLeft` that did it, but it built up the arrays one char at a time so i switched to this...

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/package.scala
elahrvivaz: i just added the type since intellij was complaining about it

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
elahrvivaz: sure

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
elahrvivaz: i couldn't come up with a more elegant way... but definitely open to suggestions if you've got some. i'm not crazy about this piece of code. and i would have thought there would be a common library somewhere that does this but i didn't find anything with some light googling.

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
elahrvivaz: added

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
jnh5y: Yeah, the FP part of me wants to see a cool foldLeft or some such.  The 'fast' part of me is really, really happy with this.:)

PR #2446, opened 2020-01-15
GEOMESA-2769 Add expiry for cached filters in distributed runtimes

elahrvivaz:
* `geomesa.filter.remote.cache.expiry` for setting filter expiration
* `geomesa.serializer.cache.expiry` for setting serializer expiration

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
@@ -1,0 +1,146 @@
+/***********************************************************************
+ * Copyright (c) 2013-2020 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.utils.cache
+
+import java.io.Closeable
+import java.lang.ref.WeakReference
+import java.util.concurrent._
+
+import com.github.benmanes.caffeine.cache.{Cache, Caffeine}
+import com.google.common.util.concurrent.MoreExecutors
+
+import scala.concurrent.duration.Duration
+
+/**
+ * Creates a per-thread cache of values with a timed expiration.
+ *
+ * Map operations will only affect/reflect the state of the current thread. Additional methods `globalIterator`
+ * and `estimatedGlobalSize` are provided for global views across all threads, which generally should only be used
+ * for debugging.
+ *
+ * Since caches are only cleaned up when accessed, uses an asynchronous thread to
+ * actively clean up any orphaned thread local values
+ *
+ * @tparam K key type
+ * @tparam V value type
+ */
+class ThreadLocalCache[K <: AnyRef, V <: AnyRef](
+    expiry: Duration,
+    executor: ScheduledExecutorService = ThreadLocalCache.executor
+  ) extends scala.collection.mutable.Map[K, V]
+    with scala.collection.mutable.MapLike[K, V, ThreadLocalCache[K, V]]
+    with Runnable
+    with Closeable {
+
+  import scala.collection.JavaConverters._
+
+  // weak references to our current caches, to allow cleanup + GC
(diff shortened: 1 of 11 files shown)


elahrvivaz:
I'll cherry-pick back to 2.4

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
```diff
@@ -0,0 +1,146 @@
...
+ * @tparam K key type
+ * @tparam V value type
+ */
+class ThreadLocalCache[K <: AnyRef, V <: AnyRef](
+    expiry: Duration,
+    executor: ScheduledExecutorService = ThreadLocalCache.executor
+  ) extends scala.collection.mutable.Map[K, V]
+    with scala.collection.mutable.MapLike[K, V, ThreadLocalCache[K, V]]
+    with Runnable
+    with Closeable {
```
jnh5y: This is a lovely collection of interfaces to implement.  Maybe it should be serializable too just for kicks.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  // weak references to our current caches, to allow cleanup + GC
+  private val references = new ConcurrentLinkedQueue[(Long, WeakReference[Cache[K, V]])]()
+
+  private val caches = new ThreadLocal[Cache[K, V]]() {
+    override def initialValue(): Cache[K, V] = {
+      val cache = Caffeine.newBuilder().expireAfterAccess(expiry.toMillis, TimeUnit.MILLISECONDS).build[K, V]()
+      // this will always succeed as our queue is unbounded
+      references.offer((Thread.currentThread().getId, new WeakReference(cache)))
+      cache
```
jnh5y: Ok, how does having a WeakReference to the 'cache' help?  

It seems like each thread will have a non-Weak reference to the same object, right?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
```diff
@@ -0,0 +1,146 @@
...
+      }
+      iter.remove()
+    }
+  }
+}
+
+object ThreadLocalCache {
+  // use a 2 thread executor service for all the caches - we only use a handful across the code base
+  private val executor = MoreExecutors.getExitingScheduledExecutorService {
+    Executors.newScheduledThreadPool(2).asInstanceOf[ScheduledThreadPoolExecutor]
```
jnh5y: Any reason to use 2 instead of 1?

elahrvivaz: 1 seemed too few, 3 seemed too many?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
elahrvivaz: yes for long-running threads, but if the calling thread dies then the cache could be gc'd and we don't want to prevent that.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
jnh5y: I can be onboard with 2 being greater than 1, but not much greater.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
elahrvivaz: yeah... i could make it private and just expose the Map interface to end users

jnh5y:
Maybe delete SoftThreadLocalCache?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/cache/ThreadLocalCache.scala
jnh5y: Nah, that's fine.  Just noting the interfaces; fun class!

PR #2451, opened 2020-01-21
GEOMESA-2713 Unit test for Timestamp attribute stat support

elahrvivaz:


elahrvivaz:
The bug was fixed in 2.4.0 due to a refactoring, but this adds a unit test to prevent regression.

PR #2453, opened 2020-01-22
GEOMESA-2735 FSDS - support S3 object tags

elahrvivaz:

### DIFF
--- a/build/cqs.tsv
+++ b/build/cqs.tsv
@@ -262,6 +262,7 @@ tec.uom.lib:uom-lib-common	1.0.2	compile
 tec.uom:uom-se	1.0.8	compile
 
 com.101tec:zkclient	0.10	provided
+com.amazonaws:aws-java-sdk-s3	1.11.412	provided
 com.datastax.cassandra:cassandra-driver-core	3.0.0	provided
 com.datastax.cassandra:cassandra-driver-mapping	3.0.0	provided
 io.confluent:kafka-avro-serializer	5.1.0	provided
--- a/build/cqs.tsv
+++ b/build/cqs.tsv
@@ -262,7 +262,7 @@ tec.uom.lib:uom-lib-common	1.0.2	compile
 tec.uom:uom-se	1.0.8	compile
 
 com.101tec:zkclient	0.10	provided
-com.amazonaws:aws-java-sdk-s3	1.11.412	provided
+com.amazonaws:aws-java-sdk-s3	1.11.421	provided
 com.datastax.cassandra:cassandra-driver-core	3.0.0	provided
 com.datastax.cassandra:cassandra-driver-mapping	3.0.0	provided
 io.confluent:kafka-avro-serializer	5.1.0	provided
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/s3/S3VisibilityObserverFactory.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/s3/S3VisibilityObserverFactory.scala
@@ -1,0 +1,60 @@
+/***********************************************************************
+ * Copyright (c) 2013-2020 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.fs.storage.common.s3
+
+import java.io.IOException
+
+import com.amazonaws.services.s3.AmazonS3
+import org.apache.hadoop.conf.Configuration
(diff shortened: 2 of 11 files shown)


jnh5y:
Is there any chance that we could gin up a unit test?  Or is there not a good S3 mock library?

elahrvivaz:
I did add a unit test here: https://github.com/locationtech/geomesa/pull/2453/files#diff-a6e977537b2b29aaee2c0fae9fa8c384

elahrvivaz:
I guess there are some more realistic mock frameworks like this we could use, but it's probably overkill in this instance: https://github.com/adobe/S3Mock

elahrvivaz:
Another option: https://github.com/findify/s3mock

build/cqs.tsv
```diff
@@ -262,6 +262,7 @@ tec.uom.lib:uom-lib-common	1.0.2	compile
 tec.uom:uom-se	1.0.8	compile
 
 com.101tec:zkclient	0.10	provided
+com.amazonaws:aws-java-sdk-s3	1.11.421	provided
```
elahrvivaz: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21682

elahrvivaz: approved

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/s3/S3VisibilityObserverFactory.scala
```diff
@@ -0,0 +1,60 @@
...
+      tag = conf.get(S3VisibilityObserverFactory.TagNameConfig, S3VisibilityObserverFactory.DefaultTag)
+    } catch {
+      case e: Exception => throw new RuntimeException("Unable to get s3 client", e)
+    }
+  }
+
+  override def apply(path: Path): FileSystemObserver = new S3VisibilityObserver(s3, path, tag)
+
+  override def close(): Unit = {
+    s3 = null
```
jnh5y: Is there any locking that needs to happen so that something weird doesn't happen with init and close being called simultaneously?  

Or is that just an insane situation?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/s3/S3VisibilityObserverTest.scala
```diff
@@ -0,0 +1,148 @@
...
+        val encodedFront = Base64.getEncoder.encodeToString("user&admin".getBytes(StandardCharsets.UTF_8))
+        val encodedBack = Base64.getEncoder.encodeToString("admin&user".getBytes(StandardCharsets.UTF_8))
+        request.getTagging.getTagSet.asScala must beOneOf(
+          Seq(new Tag(S3VisibilityObserverFactory.DefaultTag, encodedFront)),
+          Seq(new Tag(S3VisibilityObserverFactory.DefaultTag, encodedBack))
+        )
+      }
+    }
+
+    "simplify tag expressions" >> {
```
jnh5y: What's does the simplification?  

Does it handle complicated conjunctions and disjunctions?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/s3/S3VisibilityObserverFactory.scala
elahrvivaz: these observers have their lifecycle tied to the datastore, and due to them being lazy vals i don't think it's possible for them to be closed asynchronously before or while they're initialized: https://github.com/locationtech/geomesa/blob/master/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala#L50-L58
but all this threading stuff we've been doing definitely made me go back and double check... i think for now it's ok as is, with the hope that if we change anything that would affect that we'll realize it and update this to be synchronized as needed.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/s3/S3VisibilityObserverTest.scala
elahrvivaz: the simplification is here: https://github.com/locationtech/geomesa/pull/2453/files/9b4cf3563a078f7aeb687586b1d9d92a2a52ca0b#diff-25df5de0778453f8b01569cccd7efd95R40-R46
It's not that complicated, it just flattens AND clauses and deduplicates. so e.g. `foo&bar` plus `foo` ends up as `foo&bar` and not as `(foo&bar)&foo`

PR #2463, opened 2020-02-19
GEOMESA-2778 Check for timeouts in highly selective aggregating scans

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
@@ -10,8 +10,9 @@ package org.locationtech.geomesa.accumulo.iterators
 
 import org.apache.accumulo.core.data.{Range => aRange, _}
 import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
+import org.locationtech.geomesa.accumulo.iterators.BaseAggregatingIterator.BatchScanCallback
 import org.locationtech.geomesa.index.iterators.AggregatingScan
-import org.locationtech.geomesa.index.iterators.AggregatingScan.RowValue
+import org.locationtech.geomesa.index.iterators.AggregatingScan.{RowValue, AggregateCallback}
 import org.locationtech.geomesa.utils.io.CloseWithLogging
 
 /**
@@ -50,7 +51,7 @@ abstract class BaseAggregatingIterator[T <: AggregatingScan.Result]
   override def next(): Unit = findTop()
 
   private def findTop(): Unit = {
-    val result = aggregate()
+    val result = aggregate(new BatchScanCallback()).result
     if (result == null) {
       CloseWithLogging(this)
       topKey = null // hasTop will be false
@@ -81,4 +82,22 @@ abstract class BaseAggregatingIterator[T <: AggregatingScan.Result]
 
   override def deepCopy(env: IteratorEnvironment): SortedKeyValueIterator[Key, Value] =
     throw new NotImplementedError()
+}
+
+object BaseAggregatingIterator {
+
+  private class BatchScanCallback extends AggregateCallback {
+
+    private var bytes: Array[Byte] = _
+
+    override def batch(bytes: Array[Byte]): Boolean = {
+      this.bytes = bytes
+      false // we want to stop scanning and return the batch
+    }
+
+    // we always keep scanning and rely on client connections to stop the scan
+    override def partial(bytes: => Array[Byte]): Boolean = true
+
+    def result: Array[Byte] = bytes
+  }
 }
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
@@ -11,16 +11,18 @@ package org.locationtech.geomesa.index.iterators
 import java.io.Closeable
 
 import com.typesafe.scalalogging.LazyLogging
+import org.geotools.data.DataUtilities
 import org.geotools.filter.text.ecql.ECQL
 import org.locationtech.geomesa.features.SerializationOption.SerializationOptions
 import org.locationtech.geomesa.features.TransformSimpleFeature
 import org.locationtech.geomesa.features.kryo.KryoBufferSimpleFeature
 import org.locationtech.geomesa.index.api.GeoMesaFeatureIndex
-import org.locationtech.geomesa.index.iterators.AggregatingScan.RowValue
+import org.locationtech.geomesa.index.iterators.AggregatingScan.{AggregateCallback, CqlSampleValidator, CqlValidator, RowValidator, RowValue, SampleValidator, ValidateAll}
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import org.opengis.filter.Filter
 
+import scala.util.Try
 import scala.util.control.NonFatal
 
 trait AggregatingScan[T <: AggregatingScan.Result]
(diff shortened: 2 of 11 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
```diff
@@ -11,41 +11,40 @@ package org.locationtech.geomesa.index.iterators
...
-  private var reusableTransformSf: TransformSimpleFeature = _
-  private var hasTransform: Boolean = _
+  private var aggregateSf: SimpleFeature = _
 
   override def init(options: Map[String, String]): Unit = {
     val spec = options(SftOpt)
-    sft = IteratorCache.sft(spec)
-    index = options.get(IndexSftOpt) match {
+    val sft = IteratorCache.sft(spec)
+    // note: index won't have a hook to the data store, so some operations aren't available
```
jnh5y: Any particular operations worth noting?

elahrvivaz: it doesn't affect the code here, just a warning for the future. it would be things like `getTablesNames` that need to read the catalog metadata

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
```diff
@@ -54,106 +53,164 @@ trait AggregatingScan[T <: AggregatingScan.Result]
...
         reusableSf.setBuffer(rowValue.value, rowValue.valueOffset, rowValue.valueLength)
-        if (validateFeature(reusableSf)) {
+        if (validate(reusableSf)) {
           // write the record to our aggregated results
-          if (hasTransform) {
-            aggregateResult(reusableTransformSf, result)
-          } else {
-            aggregateResult(reusableSf, result)
-          }
+          status.aggregated(aggregateResult(aggregateSf, result))
```
jnh5y: How does 'aggregateSf' get updated?

elahrvivaz: aggregateSf is going to be either reusableSf (on line 56) or transformSf (on line 65). either way, it gets updated through reusableSf, as transformSf is a view on top of reusableSf. clear as mud?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
```diff
@@ -54,106 +53,164 @@ trait AggregatingScan[T <: AggregatingScan.Result]
...
+        } else {
+          status.skipped()
         }
       } catch {
-        case NonFatal(e) => logger.error("Error aggregating value:", e)
+        case NonFatal(e) =>
+          logger.error(s"Error aggregating value for ${debugSf()}:", e)
```
jnh5y: The best error message folks will not look at in their logs...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/AggregatingScan.scala
elahrvivaz: it's slightly convoluted b/c we run the filter against the non-transformed feature, then apply the transform in the result

jnh5y: Yeah, got it the second or third time I read through the class.:)  I'm good now.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/BinAggregatingScan.scala
```diff
@@ -15,96 +15,83 @@ import org.geotools.util.factory.Hints
...
+  override protected def defaultBatchSize: Int =
+    throw new IllegalArgumentException("Batch scan is specified per scan")
 
-  override protected def notFull(result: ByteBufferResult): Boolean =
-    result.buffer.position < result.buffer.limit
+  // add the feature to the current aggregated result
+  override protected def aggregateResult(sf: SimpleFeature, result: ResultCallback): Int = {
+    val pos = result.position
+    encoder.encode(sf, result)
+    (result.position - pos) / binSize
```
jnh5y: Do we ever encode more than one result at a time here?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
```diff
@@ -82,3 +83,21 @@ abstract class BaseAggregatingIterator[T <: AggregatingScan.Result]
...
+
+    private var bytes: Array[Byte] = _
+
+    override def batch(bytes: Array[Byte]): Boolean = {
+      this.bytes = bytes
+      false // we want to stop scanning and return the batch
+    }
+
+    // we always keep scanning and rely on client connections to stop the scan
+    override def partial(bytes: => Array[Byte]): Boolean = true
```
jnh5y: Ok, for this class.... if partial is called, what happens to 'bytes'?  When would it actually be executed to continue the scan?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/BinAggregatingScan.scala
```diff
@@ -155,9 +141,43 @@ object BinAggregatingScan {
     (Seq(hints.getBinTrackIdField) ++ geom ++ dtg ++ hints.getBinLabelField).distinct.filter(_ != "id")
   }
 
-  // wrapper for java's byte buffer that adds scala methods for the aggregating iterator
-  class ByteBufferResult(val buffer: ByteBuffer, var overflow: ByteBuffer) {
-    def ensureCapacity(size: Int): ByteBuffer = {
+  class ResultCallback(buffer: ByteBuffer, private var overflow: ByteBuffer) extends BinaryOutputCallback {
```
jnh5y: I'm tempted to suggest that this class name should include the word "Binary" for 'clarity'.  That said, anyone who can read the rest of this can probably figure it out.:)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/RenderingGrid.scala
```diff
@@ -218,6 +218,7 @@ class RenderingGrid(env: Envelope, xSize: Int, ySize: Int) extends LazyLogging {
         i += 1
       }
     }
+    count += 1
```
jnh5y: How did you reach the conclusion that we needed to update this line?

The counts for the RenderingGrid seem to be very loose/complex.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/BinAggregatingScan.scala
elahrvivaz: nope, this is called per feature

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
elahrvivaz: bytes is lazily evaluated, so in this case it will never be calculated, and then we return true so the scan continues on.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/BinAggregatingScan.scala
jnh5y: If that's the case, will this calculation always be 1?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
jnh5y: Ok....  I thought calling 'bytes' would be what continues the scan...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/RenderingGrid.scala
elahrvivaz: yeah... right now they're not actually used though. i realized it was wrong b/c i had an initial implementation that did rely on the counts to track the features and it was failing due to the missing +1 here. i figured i should fix it regardless.
The original thought was to count the number of pixels that got rendered, but i think that is kind of irrelevant anyway.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/BinAggregatingScan.scala
elahrvivaz: ah, sorry, no. if it's a line string then it will encode to multiple 'bin' records so we still have to track it.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
elahrvivaz: i tried to explain it in the scala docs, but possibly not very well: https://github.com/locationtech/geomesa/pull/2463/files#diff-2feb8b7c8aaa8331044930a8332e7291R282
returning true/false is what tells it to continue the scan, the bytes are provided in case the caller wants to stop the scan and return any partial results, otherwise they will be bundled in the next batch callback.

PR #2471, opened 2020-03-02
GEOMESA-2791 Arrow - Accumulo scans close prematurely

elahrvivaz:
* Move Arrow cleanup into aggregate lifecycle


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/BaseAggregatingIterator.scala
```diff
@@ -53,7 +52,6 @@ abstract class BaseAggregatingIterator[T <: AggregatingScan.Result]
   private def findTop(): Unit = {
     val result = aggregate(new BatchScanCallback()).result
     if (result == null) {
-      CloseWithLogging(this)
```
jnh5y: OK, why did this close go?

elahrvivaz: this was the underlying cause... the result was getting closed, but then the iterator was getting re-used and the result was not in a usable state anymore. so i had to move the open/close lifecycle into each call to aggregate, which is not ideal as it is less re-use, but at least should clean things up correctly.

jnh5y: Good job finding this!

PR #2480, opened 2020-03-13
GEOMESA-2787,GEOMESA-2637 Support Accumulo 2.0, Hadoop 3.1

elahrvivaz:
* Compile time Hadoop dependency is still 2.8.5
* Remove most dependencies on guava
* Remove reflection code related to Accumulo 1.5/1.6
* Remove accumulo-jobs assembly jar
* Remove NormalizedDimensions for SFC traits
* Update Accumulo install docs


docs/user/accumulo/install.rst
```diff
@@ -44,8 +44,8 @@ Installing the Accumulo Distributed Runtime Library
 ---------------------------------------------------
 
 The ``geomesa-accumulo_2.11-$VERSION/dist/accumulo/`` directory contains the distributed
-runtime JARs that contains server-side code for Accumulo that must be made
-available on each of the Accumulo tablet servers in the cluster. These JARs
+runtime JAR that contains server-side code for Accumulo that must be made
+available on each of the Accumulo tablet servers in the cluster. This JAR
 contain GeoMesa code and the Accumulo iterators required for querying GeoMesa data.
```
jnh5y: This JAR *contains*...

(Just proving that I read:))

docs/user/accumulo/install.rst
```diff
@@ -57,25 +57,6 @@ The version of the distributed runtime JAR must match the version of the GeoMesa
 data store client JAR (usually installed in GeoServer; see below). If not,
 queries might not work correctly or at all.
 
-.. _install_accumulo_runtime_manual:
```
jnh5y: Deleting this is probably the right call.  That said, lib/ext based installs are still available/possible on 1.9.x, right?

docs/user/accumulo/install.rst
```diff
@@ -84,49 +65,57 @@ Namespace Install
...
-    the ``tableName`` parameter with the namespace. For example, refer to the
+    the ``accumulo.catalog`` parameter with the namespace. For example, refer to the
     ``my_catalog`` table as ``myNamespace.my_catalog``.
 
+.. _install_accumulo_runtime_manual:
+
+System Install
+^^^^^^^^^^^^^^
+
+Alternatively, the GeoMesa runtime JAR can be copied into the ``$ACCUMULO_HOME/lib/ext`` folder on
```
jnh5y: Ah, nevermind the above comment.  Is lib/ext still a thing in Accumulo 2.0?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/audit/AccumuloEventWriter.scala
```diff
@@ -79,8 +78,7 @@ class AccumuloEventWriter(connector: Connector, table: String) extends Runnable
 
 object AccumuloEventWriter {
 
-  val WriteInterval = SystemProperty("geomesa.accumulo.audit.interval", "5 seconds")
+  val WriteInterval: SystemProperty = SystemProperty("geomesa.accumulo.audit.interval", "5 seconds")
 
-  private val executor = MoreExecutors.getExitingScheduledExecutorService(new ScheduledThreadPoolExecutor(5))
-  sys.addShutdownHook(executor.shutdownNow())
+  private val executor = ExitingExecutor(new ScheduledThreadPoolExecutor(5), force = true)
```
jnh5y: Nice refactoring.  Is there elsewhere this should be done?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreParams.scala
```diff
@@ -9,23 +9,13 @@
...
-import org.geotools.data.DataAccessFactory.Param
-import org.geotools.data.Parameter
 import org.locationtech.geomesa.index.geotools.GeoMesaDataStoreFactory.GeoMesaDataStoreParams
 import org.locationtech.geomesa.security.SecurityParams
 import org.locationtech.geomesa.utils.geotools.GeoMesaParam
 
 // keep params in a separate object so we don't require accumulo classes on the build path to access it
 object AccumuloDataStoreParams extends GeoMesaDataStoreParams with SecurityParams {
 
-  val ConnectorParam =
```
jnh5y: Is this us dropping support for passing in an existing connector instance?  (That's fine if it is, just checking.)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
```diff
@@ -79,9 +78,8 @@ class AccumuloGeoMesaStats(ds: AccumuloDataStore, val metadata: AccumuloBackedMe
     StatsCombiner.configure(sft, connector, metadata.table, metadata.typeNameSeparator.toString)
 
     val keys = Seq(CountKey, BoundsKeyPrefix, TopKKeyPrefix, FrequencyKeyPrefix, HistogramKeyPrefix)
-    val splits = keys.map(k => new Text(metadata.encodeRow(sft.getTypeName, k)))
-    // noinspection RedundantCollectionConversion
-    connector.tableOperations().addSplits(metadata.table, ImmutableSortedSet.copyOf(splits.toIterable))
+    val splits = new java.util.TreeSet[Text](keys.map(k => new Text(metadata.encodeRow(sft.getTypeName, k))).asJava)
```
jnh5y: The new TreeSet is immutable and/or the previous code that made an immutable copy is unnecessary?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/resources/log4j.xml
```diff
@@ -22,9 +22,6 @@
     <category name="org.apache.curator">
         <priority value="error"/>
     </category>
-    <category name="org.apache.curator">
```
jnh5y: Seeing the above block, one of them is unnecessary.  Are any of the 'warn' level log messages worth anything?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/MiniCluster.scala
```diff
@@ -52,15 +60,19 @@ case object MiniCluster extends LazyLogging {
       }
       connector.securityOperations().changeUserAuthorizations(name, auths)
     }
+
+    AccumuloVersion.close(connector)
+
+    logger.info("Started Accmulo minicluster")
```
jnh5y: Shouldn't this be right after the cluster.start?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/BackCompatibilityTest.scala
```diff
@@ -49,7 +49,7 @@ class BackCompatibilityTest extends Specification with LazyLogging {
 
   sequential
 
-  lazy val connector = MiniCluster.connector
+  lazy val connector = MiniCluster.cluster.getConnector(Users.root.name, Users.root.password)
```
jnh5y: Would it be better to use the ds.connector for this test?  (I don't know if that's a good idea or not.)

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
```diff
@@ -141,6 +141,11 @@
             <artifactId>scala-compiler</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+            <scope>provided</scope>
```
jnh5y: Is this added to remove commons-text from the distribution?

geomesa-accumulo/geomesa-accumulo-jobs/pom.xml
```diff
@@ -61,77 +70,9 @@
...
-            <groupId>org.specs2</groupId>
-            <artifactId>specs2-junit_${scala.binary.version}</artifactId>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
         </dependency>
     </dependencies>
 
-    <profiles>
-        <profile>
-            <id>assemble</id>
```
jnh5y: *blinks*  Was this being used?!  Good eye to remove it!

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
```diff
@@ -55,36 +44,12 @@ class AccumuloSpatialRDDProvider extends SpatialRDDProvider with LazyLogging {
 
     lazy val transform = query.getHints.getTransformSchema
 
-    def queryPlanToRDD(sft: SimpleFeatureType, qp: AccumuloQueryPlan, conf: Configuration): RDD[SimpleFeature] = {
+    def queryPlanToRDD(sft: SimpleFeatureType, qp: AccumuloQueryPlan): RDD[SimpleFeature] = {
       if (ds == null || sft == null || qp.isInstanceOf[EmptyPlan]) {
         sc.emptyRDD[SimpleFeature]
       } else {
-        // note: we've ensured there is only one table per query plan, below
-        InputConfigurator.setInputTableName(classOf[AccumuloInputFormat], conf, qp.tables.head)
```
jnh5y: All this is now in the GeoMesaAccumuloInputFormat, right?

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
```diff
@@ -94,34 +59,9 @@ class AccumuloSpatialRDDProvider extends SpatialRDDProvider with LazyLogging {
 
         // From sc.newAPIHadoopRDD
         // Add necessary security credentials to the JobConf. Required to access secure HDFS.
-        val jconf = new JobConf(conf)
+        val jconf = job.getConfiguration.asInstanceOf[JobConf]
         SparkHadoopUtil.get.addCredentials(jconf)
 
-        // Get username from params
-        val username = AccumuloDataStoreParams.UserParam.lookup(paramsAsJava)
```
jnh5y: Where does/did this logic / configuration / etc end up?

geomesa-accumulo/geomesa-accumulo-tools/bin/install-hadoop-accumulo.sh
```diff
@@ -59,6 +57,14 @@ declare -a urls=(
   "${base_url}org/apache/htrace/htrace-core4/${htrace4_core_version}/htrace-core4-${htrace4_core_version}.jar"
 )
 
+accumulo_maj_ver="$(expr match "$accumulo_version" '\([0-9][0-9]*\)\.')"
+if [[ "$accumulo_maj_ver" -lt 2 ]]; then
+  urls+=(
+    "${base_url}org/apache/accumulo/accumulo-fate/${accumulo_version}/accumulo-fate-${accumulo_version}.jar"
```
jnh5y: Accumulo-{fate,trace} are gone in 2.0?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
```diff
@@ -101,48 +98,6 @@ trait RunnerWithAccumuloEnvironment extends Runner {
           }
         }
       }
-
-      if (p.instance == null) {
```
jnh5y: I'm fine with this code going.  Do we want any of its logic?

geomesa-blobstore/geomesa-blobstore-api/src/main/java/org/locationtech/geomesa/blob/api/Blob.java
```diff
@@ -37,9 +36,10 @@ public String getLocalName() {
 
     @Override
     public String toString() {
-        return Objects.toStringHelper(this)
-                .add("id", id)
-                .add("localName", localName)
-                .toString();
+        return "Blob{" +
```
jnh5y: Hmm... would it be better to use a StringBuilder?  Does it matter?

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
```diff
@@ -18,7 +18,7 @@ htrace_core_version="4.1.0-incubating"
 
 # for hadoop 2.5 and 2.6 to work we need these
```
jnh5y: Do the comments in the install-hadoop.sh scripts need to be updated?  (Hate to ask, but this may be a good time to grep for 2.5/2.6.)

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
```diff
@@ -18,7 +18,7 @@ htrace_core_version="4.1.0-incubating"
...
 # These should match up to what the hadoop version desires
-guava_version="11.0.2"
+guava_version="%%guava.version%%"
```
jnh5y: If we don't have a dependency on Guava, does the install script need to download it still?

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/StorageJobUtils.scala
```diff
@@ -49,4 +47,29 @@ object StorageJobUtils extends LazyLogging {
     }
     success
   }
+
+  private def distCpOptions(src: Path, dest: Path): DistCpOptions =
+    try { distCpOptions3(src, dest) } catch { case _: ClassNotFoundException => distCpOptions2(src, dest) }
```
jnh5y: Should we have debug/trace logging around this try and fallback approach?

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/MiniCluster.scala
```diff
@@ -20,7 +20,7 @@ case object MiniCluster extends LazyLogging {
     logger.info("Starting embedded hbase")
     cluster.getConfiguration.set("hbase.superuser", "admin")
     cluster.getConfiguration.set(CoprocessorHost.USER_REGION_COPROCESSOR_CONF_KEY, HBaseIndexAdapter.CoprocessorClass)
-    cluster.startMiniCluster(sys.props.get("geomesa.hbase.test.servers").map(_.toInt).getOrElse(3))
+    cluster.startMiniCluster(sys.props.get("geomesa.hbase.test.region.servers").map(_.toInt).getOrElse(2))
```
jnh5y: Is this system property documented / used anywhere?  Also, any particular reason for the change from 3 to 2?

geomesa-hbase/geomesa-hbase-spark-runtime-hbase1/pom.xml
```diff
@@ -59,6 +59,12 @@
             <version>${hbase.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
jnh5y: Are we including it in order to relocate Guava?

geomesa-kafka/pom.xml
```diff
@@ -21,17 +21,11 @@
...
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
     <dependencyManagement>
         <dependencies>
-            <dependency>
-                <groupId>com.google.guava</groupId>
-                <artifactId>guava</artifactId>
-                <version>${kafka.guava.version}</version>
-            </dependency>
```
jnh5y: How did we remove the Kafka module's dependency on Guava?

geomesa-lambda/geomesa-lambda-datastore/src/test/scala/org/locationtech/geomesa/lambda/LambdaTestRunnerTest.scala
```diff
@@ -64,16 +64,18 @@ object LambdaTestRunnerTest {
     var zookeepers: String = _
 
     lazy val dsParams = Map(
-      "lambda.accumulo.connector"  -> MiniCluster.connector,
+      "lambda.accumulo.instance.id" -> MiniCluster.cluster.getInstanceName,
+      "lambda.accumulo.zookeepers"  -> MiniCluster.cluster.getZooKeepers,
+      "lambda.accumulo.user"        -> MiniCluster.Users.root.name,
+      "lambda.accumulo.password"    -> MiniCluster.Users.root.password,
       // note the table needs to be different to prevent testing errors
```
jnh5y: Does the test table still need to be different? 

Does this show that a Lambda DS catalog and an Accumulo DS catalog may not be shared?  If so, is that documented?

geomesa-spark/geomesa-spark-sql/src/test/scala/org/locationtech/geomesa/spark/SparkSQLTestUtils.scala
```diff
@@ -28,6 +28,7 @@ object SparkSQLTestUtils {
       .config("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
       .config("spark.kryo.registrator", classOf[GeoMesaSparkKryoRegistrator].getName)
       .config("spark.sql.crossJoin.enabled", "true")
+      .config("spark.ui.enabled", value = false)
```
jnh5y: Is the default 'true'?

docs/user/accumulo/install.rst
elahrvivaz: yeah, i think it actually works on 2.0 too

docs/user/accumulo/install.rst
elahrvivaz: i believe so... i didn't actually verify it though, and it doesn't exist by default.

elahrvivaz: actually I think @khobbs-ccri has been using lib/ext with accumulo 2

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/audit/AccumuloEventWriter.scala
elahrvivaz: yeah, i replaced all the usages of guava's MoreExecutors with this

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreParams.scala
elahrvivaz: right - we only used it in tests, and technically it's not serializable so it breaks the params contract... if someone feels strongly i don't mind putting it back in though.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
elahrvivaz: the new one is *not* immutable, but i don't think it's necessary. guava offers an immutable one which i assume was used just out of preference for using immutable collections, but it doesn't seem worth requiring guava here just for that.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/resources/log4j.xml
elahrvivaz: generally i'd say yes, but the mini cluster tends to spam a lot of warnings so we moved everything to error in that pr. this was just to fix the oversight of leaving in the duplicate entry.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/MiniCluster.scala
elahrvivaz: i thought it made sense to add the user permissions stuff in the startup time, since the mini cluster isn't available until that returns, but we could log that separately if you want.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/BackCompatibilityTest.scala
elahrvivaz: yeah that probably makes sense. I'll update it

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
elahrvivaz: yeah, it ships with accumulo 2 now, so we don't need to bundle it into the lib/ folder. I added it to the install-hadoop-accumulo.sh script for the use case of not having an ACCUMULO_HOME set. I don't think we're using it directly, as things seemed to still work ok with accumulo 1.9 which doesn't ship it. we should definitely do some thorough testing of different installs/versions once everything has stabilized.

geomesa-accumulo/geomesa-accumulo-jobs/pom.xml
elahrvivaz: eh, we had docs on it... but it's one of those things that we never test, so it likely doesn't even work anymore.

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
elahrvivaz: right, i created a Job instance to temporarily pass through to those methods and then retrieve the config from it. seemed to work ok...

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala
```diff
@@ -23,9 +23,9 @@ class Z3SFC(period: TimePeriod, precision: Int = 21) extends SpaceTimeFillingCur
 
   require(precision > 0 && precision < 22, "Precision (bits) per dimension must be in [1,21]")
 
-  override val lon: NormalizedDimension  = NormalizedLon(precision)
-  override val lat: NormalizedDimension  = NormalizedLat(precision)
-  override val time: NormalizedDimension = NormalizedTime(precision, BinnedTime.maxOffset(period).toDouble)
+  val lon: NormalizedDimension  = NormalizedLon(precision)
+  val lat: NormalizedDimension  = NormalizedLat(precision)
+  val time: NormalizedDimension = NormalizedTime(precision, BinnedTime.maxOffset(period).toDouble)
```
jnh5y: As a sanity check, we are tossing lon/lat/time out of the common interface, but Z2/Z3 still have them in use?

pom.xml
```diff
@@ -83,15 +83,16 @@
         <metrics.version>3.2.6</metrics.version>
         <aws.sdk.version>1.11.421</aws.sdk.version>
 
-        <specs2.version>4.3.2</specs2.version>
-        <junit.version>4.12</junit.version>
+        <specs2.version>4.9.2</specs2.version>
+        <junit.version>4.13</junit.version>
```
jnh5y: @jeffklassen you may be interested to see that @elahrvivaz is bumping up the Specs 2 and Junit versions here.

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
elahrvivaz: this is all handled by `GeoMesaAccumuloInputFormat.configure` now. we had a lot of duplicate code between these two classes, i think b/c the main input format requires a Job instance and we didn't have one here.

pom.xml
```diff
@@ -1658,6 +1668,17 @@
...
+            <!-- standardize httpcomponents version, used by various libraries including hadoop and thrift -->
+            <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpclient</artifactId>
+                <version>4.5.3</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpcore</artifactId>
+                <version>4.4.6</version>
```
jnh5y: The versions of httpclient and httpcore in the Hadoop/Apache ecosystem don't match?

geomesa-accumulo/geomesa-accumulo-tools/bin/install-hadoop-accumulo.sh
elahrvivaz: yeah. although i notice that there is an `accumulo-tracer-2.0.0.jar` now... possibly only needed if you run remote traces? i'm not entirely sure what the old trace jar was used for either.

pom.xml
```diff
@@ -1770,22 +1791,55 @@
...
                 <scope>provided</scope>
+                <exclusions>
+                    <!-- exclude hadoop 3+ only jars, since we're still compiling against haddop 2.8 -->
+                    <exclusion>
+                        <groupId>org.apache.hadoop</groupId>
+                        <artifactId>hadoop-client-api</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.apache.hadoop</groupId>
+                        <artifactId>hadoop-client-runtime</artifactId>
```
jnh5y: Do we need to add them back in bundles and/or install scripts?

Do we need install-hadoop2 / install-hadoop3 scripts?

pom.xml
```diff
@@ -1922,6 +1976,38 @@
...
+                        <artifactId>j2objc-annotations</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>com.google.errorprone</groupId>
+                        <artifactId>error_prone_annotations</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>animal-sniffer-annotations</artifactId>
+                    </exclusion>
```
jnh5y: Ok, so I don't mind the exclusions, but let's talk about 'em for a second.

Can you say a little bit about why some of these exclusions are here?

pom.xml
```diff
@@ -2324,13 +2414,13 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <!-- 3.0.0-M1 was released 2018-11-07. -->
-                    <version>3.0.0-M1</version>
+                    <!-- 3.0.0-M4 was released 2019-11-13 -->
+                    <version>3.0.0-M4</version>
```
jnh5y: @jeffklassen NB: Upgrade to the Apache Maven surefire plugin.

pom.xml
```diff
@@ -2349,8 +2439,11 @@
                             <user.language>en</user.language>
                             <user.region>US</user.region>
                             <geomesa.scan.ranges.target>500</geomesa.scan.ranges.target>
-                            <geomesa.hbase.test.servers>2</geomesa.hbase.test.servers>
+                            <geomesa.accumulo.test.tablet.servers>1</geomesa.accumulo.test.tablet.servers>
+                            <geomesa.hbase.test.region.servers>1</geomesa.hbase.test.region.servers>
```
jnh5y: Ok, somewhere else one of the defaults is set to 2.  Thoughts?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
elahrvivaz: the underlying accumulo classes don't exist anymore, so i'm not sure we can do the same thing. Accumulo 2 uses a `accumulo-client.properties` file that you can specify connection params in - we should probably open a ticket to support that, but it seemed a little more complicated than i wanted to do at the moment (we'd have to make all the ds params optional for one thing).

elahrvivaz: opened https://geomesa.atlassian.net/browse/GEOMESA-2800

geomesa-blobstore/geomesa-blobstore-api/src/main/java/org/locationtech/geomesa/blob/api/Blob.java
elahrvivaz: possibly... i just used intellij generate->toString to make this :D

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
elahrvivaz: i think the comments are still correct? i'm not 100% sure though so i didn't want to just remove them.

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
elahrvivaz: i think so b/c hadoop still needs it. we also do still have a dependency on guava byte comparators - they have some native code performance things that seemed useful to keep, and afaik we haven't ever hit a version issue with that particular class.

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/StorageJobUtils.scala
elahrvivaz: mmm, we could i guess... would anyone ever care though unless it failed? it's only called once on a m/r job configure, so i think the performance penalty from throwing the exception should be negligible.

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/MiniCluster.scala
elahrvivaz: it's in the main pom.xml properties section. i actually dialed it down to 1 there, to reduce the load/complexity of the mini cluster. it seemed to improve the stability of the accumulo mini cluster immensely. 2 is the hbase default i think...

geomesa-hbase/geomesa-hbase-spark-runtime-hbase1/pom.xml
elahrvivaz: right

geomesa-kafka/pom.xml
elahrvivaz: this may be one of those things we need to test haha. but i removed the use of guava loading caches and executors, so i don't think there are any direct dependencies on guava anymore (i don't think kafka uses the byte comparators i mentioned earlier).

geomesa-lambda/geomesa-lambda-datastore/src/test/scala/org/locationtech/geomesa/lambda/LambdaTestRunnerTest.scala
elahrvivaz: you can share an accumulo and lambda ds table - this is just meant for the unit tests, so that they don't interfere with each other.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreParams.scala
jnh5y: Right!  Not serializable; it is outta here!

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/MiniCluster.scala
jnh5y: Cool.  I didn't read closely enough.

geomesa-spark/geomesa-spark-sql/src/test/scala/org/locationtech/geomesa/spark/SparkSQLTestUtils.scala
elahrvivaz: yeah, this just prevents it from spinning up the spark UI jetty servlet for no reason

elahrvivaz: the flag is mainly used by spark for unit testing i believe

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
jnh5y: Wow, that's a lot of considerations in a short span.  Can I put you on the hook for indicating a few test environments / use cases that'd cover our bases?

geomesa-accumulo/geomesa-accumulo-jobs/pom.xml
jnh5y: Are the docs gone?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala
elahrvivaz: right. does that seem reasonable? we do use them directly in the Z2/Z3 feature index code. but the s2 index doesn't have them so it seemed more an implementation detail.

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
jnh5y: Sweet.  I assume you tested on all 8 million combinations of MapReduce, HBase, Accumulo, Linux, public cloud provider, and day of the week.  Works for me!

pom.xml
elahrvivaz: i think that i bumped this hoping to get some better stability in the tests - i didn't see any appreciable difference but seems like a good idea to use latest anyway.

pom.xml
elahrvivaz: it's not hadoop, for some reason the two artifacts are on different versioning schemes: https://search.maven.org/search?q=g:org.apache.httpcomponents

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
jnh5y: Great!  Is there a similar refactoring that should happen on the HBase side?  Or is that code happy?

geomesa-accumulo/geomesa-accumulo-tools/bin/install-hadoop-accumulo.sh
jnh5y: @jahhulbert @jahhulbert-ccri any thoughts or info on the use of Accumulo Tracer?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/AccumuloRunner.scala
jnh5y: Yeah, that's great separate ticket.  Thanks for opening it.

pom.xml
elahrvivaz: i just left the install scripts as installing hadoop 2.8, and that seemed to work with a hadoop 3 environment... but you're right, it would probably be a good idea to modify the install scripts to add these if hadoop version > 3, as likely some user will update the version number. I'll make that change

pom.xml
elahrvivaz: i didn't want to open more cqs and everything still passed ;)

pom.xml
elahrvivaz: this was another attempt at fixing test stability, that didn't seem to actually affect anything. What finally worked was reducing the mini cluster tablet servers from 2 to 1.

pom.xml
elahrvivaz: I think 2 is the default used by hbase/accumulo, which is why it's 2 in the code but 1 here. if that doesn't make sense i don't mind updating it

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
elahrvivaz: for the CLI tools i think the matrix looks like:
accumulo version: (1.7, 1.8)?, 1.9, 2.0
accumulo jars: ACCUMULO_HOME + HADOOP_HOME, install-hadoop-accumulo.sh
I ran through those 4 combos (for accumulo >= 1.9) with ingest, distributed ingest, export, distributed export and spark (although i didn't test distributed processing without HADOOP_HOME, as you need hadoop installed for that...)
for geoserver i used install-hadoop-accumulo.

geomesa-accumulo/geomesa-accumulo-jobs/pom.xml
elahrvivaz: yeah i deleted them

geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
elahrvivaz: hbase is good already: https://github.com/locationtech/geomesa/blob/master/geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala#L53

docs/user/accumulo/install.rst
elahrvivaz: fixed, thanks

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/BackCompatibilityTest.scala
elahrvivaz: fixed

pom.xml
elahrvivaz: fixed

build/cqs.tsv
```diff
@@ -28,24 +28,16 @@ com.googlecode.juniversalchardet:juniversalchardet	1.0.3	compile
 com.google.errorprone:error_prone_annotations	2.3.3	compile
 com.google.flatbuffers:flatbuffers-java	1.9.0	compile
 com.google.guava:guava	11.0.2	compile
-com.google.guava:guava	12.0.1	compile
-com.google.guava:guava	14.0.1	compile
-com.google.guava:guava	17.0	compile
-com.google.guava:guava	18.0	compile
+com.google.guava:guava	27.0-jre	compile
```
jnh5y: Already approved here: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21084.  

@elahrvivaz just to confirm, your understanding is that we needn't enter piggyback CQs anymore, right?

elahrvivaz: that's my understanding based on the cq page verbiage, but maybe we should confirm with Wayne or someone at eclipse.

build/cqs.tsv
```diff
@@ -126,22 +118,22 @@ net.sf.geographiclib:GeographicLib-Java	1.49	compile
 net.sf.saxon:Saxon-HE	9.7.0-20	compile
 net.sf.trove4j:trove4j	3.0.3	compile
 org.antlr:antlr4-runtime	4.7.1	compile
-org.apache.accumulo:accumulo-core	1.8.1	compile
-org.apache.accumulo:accumulo-fate	1.8.1	compile
+org.apache.accumulo:accumulo-core	2.0.0	compile
```
jnh5y: To be entered.

build/cqs.tsv
```diff
@@ -126,22 +118,22 @@ net.sf.geographiclib:GeographicLib-Java	1.49	compile
...
 org.apache.arrow:arrow-format	0.16.0	compile
 org.apache.arrow:arrow-memory	0.16.0	compile
 org.apache.arrow:arrow-vector	0.16.0	compile
 org.apache.avro:avro	1.7.5	compile
 org.apache.camel:camel-core	2.24.1	compile
 org.apache.camel:camel-scala	2.24.1	compile
+org.apache.commons:commons-collections4	4.3	compile
```
jnh5y: Hmm... 4.4 is approved: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21262.

I wonder if we should manage it up:).

build/cqs.tsv
```diff
@@ -126,22 +118,22 @@ net.sf.geographiclib:GeographicLib-Java	1.49	compile
...
 org.apache.commons:commons-compress	1.18	compile
+org.apache.commons:commons-configuration2	2.5	compile
```
jnh5y: To be entered.

build/cqs.tsv
```diff
@@ -126,22 +118,22 @@ net.sf.geographiclib:GeographicLib-Java	1.49	compile
...
 org.apache.commons:commons-crypto	1.0.0	compile
 org.apache.commons:commons-csv	1.5	compile
 org.apache.commons:commons-dbcp2	2.6.0	compile
 org.apache.commons:commons-lang3	3.7	compile
 org.apache.commons:commons-pool2	2.6.1	compile
-org.apache.commons:commons-text	1.4	compile
-org.apache.commons:commons-vfs2	2.1	compile
+org.apache.commons:commons-text	1.6	compile
```
jnh5y: We can PB on https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19028

build/cqs.tsv
```diff
@@ -163,6 +155,7 @@ org.apache.hbase.thirdparty:hbase-shaded-netty	2.2.1	compile
 org.apache.hbase.thirdparty:hbase-shaded-protobuf	2.2.1	compile
 org.apache.hive:hive-storage-api	2.6.0	compile
 org.apache.htrace:htrace-core	3.1.0-incubating	compile
+org.apache.htrace:htrace-core	3.2.0-incubating	compile
```
jnh5y: Need to enter.

build/cqs.tsv
```diff
@@ -177,11 +170,11 @@ org.apache.parquet:parquet-encoding	1.9.0	compile
 org.apache.parquet:parquet-format	2.3.1	compile
 org.apache.parquet:parquet-hadoop	1.9.0	compile
 org.apache.parquet:parquet-jackson	1.9.0	compile
-org.apache.thrift:libthrift	0.9.3	compile
+org.apache.thrift:libthrift	0.12.0	compile
```
jnh5y: PB on https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20905

build/cqs.tsv
```diff
@@ -177,11 +170,11 @@ org.apache.parquet:parquet-encoding	1.9.0	compile
...
 org.apache.xml:xml-commons-resolver	1.2	compile
 org.apache.yetus:audience-annotations	0.5.0	compile
 org.apache.yetus:audience-annotations	0.7.0	compile
-org.apache.zookeeper:zookeeper	3.4.10	compile
+org.apache.zookeeper:zookeeper	3.4.14	compile
```
jnh5y: Need to enter or manage to 3.4.11

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -103,51 +103,50 @@ object AccumuloDataStoreFactory extends GeoMesaDataStoreInfo {
...
-    def hasKeytab = !PasswordParam.exists(params) && KeytabPathParam.exists(params) && isKerberosAvailable
-    hasConnector || (hasConnection && (hasPassword || hasKeytab))
+    val hasConnection = InstanceIdParam.exists(params) && ZookeepersParam.exists(params) && UserParam.exists(params)
+    // exactly one of password or keytab
+    hasConnection && (PasswordParam.exists(params) != KeytabPathParam.exists(params))
   }
 
   def buildAccumuloConnector(params: java.util.Map[String, _ <: Serializable]): Connector = {
     if (MockParam.lookup(params)) {
       throw new IllegalArgumentException("Mock Accumulo connections are not supported")
```
jnh5y: Should we remove this? 

Did you git grep -l mock:)?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/stats/AccumuloGeoMesaStats.scala
jnh5y: Cool.  Thanks.

geomesa-blobstore/geomesa-blobstore-api/src/main/java/org/locationtech/geomesa/blob/api/Blob.java
jnh5y: Hehehe, yeah, let's leave it.

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/StorageJobUtils.scala
jnh5y: Meh, works for me.

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala
jnh5y: Works for me.

pom.xml
jnh5y: Good to know.

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
jnh5y: I just wonder if we want to drop any mention of unsupported versions of Hadoop?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: i did git grep mock. and removed most references. I think keeping this here is useful in case someone is still trying to use the old param for whatever reason.

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
elahrvivaz: ah, i forgot we had dropped support for hadoop before 2.8. let me take a look...

build/cqs.tsv
elahrvivaz: i'll give it a quick try and see if it builds ok

geomesa-fs/geomesa-fs-tools/bin/install-hadoop.sh
elahrvivaz: hadoop 2.8 still uses the same version, so i just updated the comment

build/cqs.tsv
elahrvivaz: i think it's worthwhile to put in a new cq...

build/cqs.tsv
elahrvivaz: builds ok so i went with it

build/cqs.tsv
jnh5y: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21802 (Waiting)

build/cqs.tsv
jnh5y: Entered: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21803

build/cqs.tsv
jnh5y: Entered: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21804

build/cqs.tsv
jnh5y: New CQ for you: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21805

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
jnh5y: @khobbs-ccri can you capture this information in your V&V test plan?

elahrvivaz:
thanks for putting in the cqs @jnh5y !

jnh5y:
> thanks for putting in the cqs @jnh5y !

Of course!  I think it is my 'turn' for the next 3 or 4 hundred, right?

khobbs-ccri:
What information?

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
khobbs-ccri: Wow, the e-mail summary was cryptic!

docs/user/accumulo/install.rst
khobbs-ccri: No, I've been using the namespace install. I think lib ext went away and, i think, creating it didn't work.

jnh5y: Ok, given that, should this documentation be updated?  (If so, can we do it in this PR?  Either way, who's the right person to spill some words on that?)

elahrvivaz: accumulo still mentions `lib/ext` in their docs: https://accumulo.apache.org/docs/2.x/getting-started/table_configuration#iterators

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
jnh5y: @khobbs-ccri lemme know when you've got these pearls recorded!

docs/user/accumulo/install.rst
khobbs-ccri: The namespace install worked perfectly. but I really didn't explore my one failed attempt at using lib/ext further.

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
khobbs-ccri: done

jnh5y:
Actually, it is  3d10058b105cd578f6e1c1337c5de73c67886f70 on master after correcting the author.

PR #2483, opened 2020-03-19
WIP GeoTools Upgrade

jnh5y:
This branch is to show the WIP for upgrading to GeoTools 23.


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/GeoJsonExporter.scala
```diff
@@ -8,54 +8,29 @@
...
-      count += 1L
-    }
     while (features.hasNext) {
-      writer.write(",\n")
-      json.writeFeature(features.next, writer)
+      writer.write(features.next)
       count += 1L
     }
-    writer.flush()
+    // TODO writer.flush()
```
jnh5y: Is the flush commented out since the implementation happens to be AutoCloseable?

elahrvivaz: ah, that's another thing i forgot - it would be nice if the geojsondatastore writer implemented Flushable instead of flushing on every feature. currently there is no way to flush it

PR #2490, opened 2020-03-23
GEOMESA-2805 - create first integration test

jeffklassen:
Signed-off-by: Jeffrey Klassen <jeffrey.s.klassen@gmail.com>

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/pom.xml
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/pom.xml
@@ -180,21 +180,21 @@
                 </executions>
             </plugin>
             <plugin>
-                    <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-failsafe-plugin</artifactId>
-                    <version>3.0.0-M3</version>
-                    <executions>
-                        <execution>
-                        <goals>
-                            <goal>integration-test</goal>
-                            <goal>verify</goal>
-                        </goals>
-                        </execution>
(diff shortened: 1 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/pom.xml
```diff
@@ -179,6 +179,22 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-failsafe-plugin</artifactId>
```
elahrvivaz: we should just need the groupId and artifactId here, right? seems like we could delete the rest since it's the same as the pluginManagement definition

elahrvivaz: also, could you fix the indentation here? :D

jeffklassen: @elahrvivaz actually, we need this configuration. The one in the root pom does not work.

jeffklassen: Indentation fixed

jeffklassen: @elahrvivaz  - fixed

jnh5y:
merged as 29fd728ec7c5fcc9787e9c73c2ad1394a2fd11ef

PR #2492, opened 2020-03-23
If expiry is zero, just pass features through without storing them in…

elahrvivaz:
… memory


elahrvivaz:
merged as 5e56df4bfd93b4ecf3c8c24b9cfce80f822c2d2c

PR #2503, opened 2020-03-31
GEOMESA-2814 Configure GitHub Actions

jeffklassen:
Signed-off-by: Jeffrey Klassen <jeffrey.s.klassen@gmail.com>

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,30 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+    
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+                ${{ runner.os }}-maven-
+      - name: Build with Maven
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,30 @@
+name: CI
+
+on: [push]
```
elahrvivaz: we should probably make this push or pr

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          java-version: 1.8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+                ${{ runner.os }}-maven-
+      - name: Build with Maven
+        run: mvn clean install -DskipTests -T4 2>&1| grep -e 'Building GeoMesa' -e '\(maven-surefire-plugin\|maven-jar-plugin\|scala-maven-plugin.*:compile\)'
```
elahrvivaz: we could use build/mvn and possibly cache the downloads of zinc if that's easy

elahrvivaz:
this looks good - the only concern i have is that we won't be able to see the errors if it fails, due to the grepping of the output. the travis build `tee`d to a log file and then dumped the last 500 lines of the log file on a failure - we could do that, or just remove the grep.

.github/workflows/build-and-test.yml
jeffklassen: done

.github/workflows/build-and-test.yml
jeffklassen: @elahrvivaz implemented. I've seen that caching is not mega fast, while the zinc download is. I'm going to investigate this w/o the cached zinc download

elahrvivaz:
merged as fb633e31d98e68eac358404a843dea3fde086240

PR #2504, opened 2020-03-31
GEOMESA-2817 Fix Arrow memory cleanup for delta sorting queries

elahrvivaz:

### DIFF
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
@@ -423,7 +423,7 @@ object DeltaWriter extends StrictLogging {
     logger.trace(s"merging sorted deltas - read schema: ${result.underlying.getField}")
 
     // builder for our merge array - (vector, reader for sort values, transfers to result, dictionary mappings)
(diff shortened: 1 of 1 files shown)


geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -471,7 +471,7 @@ object DeltaWriter extends StrictLogging {
         val mapVector = toLoad.underlying
         val dict = dictionaries.get(sortBy)
         val sortReader = ArrowAttributeReader(sft.getDescriptor(sortBy), mapVector.getChild(sortBy), dict, encoding)
-        mergeBuilder += ((mapVector, sortReader, transfers, mappings.get(sortBy).orNull))
+        mergeBuilder += ((toLoad, sortReader, transfers, mappings.get(sortBy).orNull))
```
jnh5y: Ok what's the difference between mapVector and toLoad?

elahrvivaz: `mapVector` is the underlying arrow vector, `toLoad` is the SimpleFeatureVector that wraps it. the issue was that the SimpleFeatureVector holds a reference to an allocator, and it wasn't being passed through so it wasn't being closed.

jnh5y:
merged as 98aebafdba5143f6167f5041383e8f1a61bab4cd

PR #2507, opened 2020-03-31
GEOMESA-2795 Use wait/notify for awaitTermination in CachedThreadPool

elahrvivaz:

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/concurrent/CachedThreadPool.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/concurrent/CachedThreadPool.scala
@@ -34,6 +34,7 @@ class CachedThreadPool(maxThreads: Int) extends AbstractExecutorService {
   private val tasks = new java.util.HashSet[Future[_]]()
   private val stopped = new AtomicBoolean(false)
   private val lock = new ReentrantLock()
+  private val done = lock.newCondition()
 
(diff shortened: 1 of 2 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/concurrent/CachedThreadPool.scala
```diff
@@ -59,12 +60,15 @@ class CachedThreadPool(maxThreads: Int) extends AbstractExecutorService {
...
 
   override def awaitTermination(timeout: Long, unit: TimeUnit): Boolean = {
-    // TODO GEOMESA-2795 could probably make a wait instead of a poll
-    val end = System.nanoTime() + unit.toNanos(timeout)
-    while (!isTerminated && System.nanoTime() < end) {
-      Thread.sleep(100)
+    lock.lock()
+    try {
+      if (isTerminated) { true } else {
+        done.await(timeout, unit)
```
jnh5y: Could the 'done' here be a def/lazy val?  (Probably doesn't matter)

elahrvivaz: i don't think so - it gets called by the last thread that runs so it would always be created. can't be a def b/c it's shared between the waiter and notifier thread.

jnh5y:
merged as f841898c555e111abad70f46108e28f84b63e5e4

PR #2509, opened 2020-04-01
GEOMESA-2820 Update HBase install scripts and docs

elahrvivaz:


docs/user/hbase/install.rst
```diff
@@ -105,47 +105,52 @@ registered manually. See :ref:`coprocessor_alternate` for details.
...
 .. tabs::
 
+    .. group-tab:: Standard
+
+        Configure GeoMesa to use pre-installed HBase and Hadoop distributions:
+
+        .. code-block:: bash
+
+            export HADOOP_HOME=/path/to/hadoop
+            export HBASE_HOME=/path/to/hbase
```
khobbs-ccri, line 150: I always find HADOOP_HOME, and HBASE_HOME somewhat frustrating on a managed cluster. We think of these paths as "the place where the tarball was extracted," or the path to the hadoop executable, but what do we actually need from them? I'm looking at an executable at /usr/bin/hadoop, and a core-site.xml in  /etc/hadoop/conf/core-site.xml. How do I know if these will work or if I need to use more advanced variables?

elahrvivaz: there's a separate tab for EMR here, does that suit your use case? https://github.com/locationtech/geomesa/pull/2509/files/57736e5315c87316e2a4655306080fbe27bfa70f#diff-9afc2c38411542d4ba4bb4656a2b3500R154
I'm not sure if there's a good way to know when you need different configuration setups... the most control is through using the explicit classpaths as described here:
https://github.com/locationtech/geomesa/pull/2509/files/57736e5315c87316e2a4655306080fbe27bfa70f#diff-9afc2c38411542d4ba4bb4656a2b3500R136
we could probably update our scripts to be better about pulling out hadoop/hbase paths from standard managed install locations... I think generally we look for the 'conf' folder and any jar files and add them all to the classpath.

jnh5y:
merged as 611b5be7279810698a3edb484469ff08278f4816

PR #2511, opened 2020-04-02
GEOMESA-2823 Group HBase ranges by region instead of region server 

jnh5y:


elahrvivaz:
the github actions build was stuck on building maven for 16 hours, so i killed it, but i don't see a way to restart it... it also seems to not show the console output for tasks that don't complete, so hard to say what went wrong.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -451,20 +451,19 @@ class HBaseIndexAdapter(ds: HBaseDataStore) extends IndexAdapter[HBaseDataStore]
   private def groupRange(
       locator: RegionLocator,
       range: RowRange,
-      result: scala.collection.mutable.Map[ServerName, java.util.List[RowRange]]): Unit = {
-    var regionServer: ServerName = null
+      result: scala.collection.mutable.Map[String, java.util.List[RowRange]]): Unit = {
+    var regionInfo: HRegionInfo = null
```
elahrvivaz, line 455: i'd change this to be the encoded name, so that if there's an exception getting the location it won't throw an NPE down on line 466/7

elahrvivaz:
merged as f14604ac1f370091ba82c0dab6c1e99b26cd5471

PR #2520, opened 2020-04-13
GEOMESA-2800 Accumulo - load connection from `accumulo-client.properties`

elahrvivaz:
* Or `client.conf` from Accumulo 1.x
* Remove default visibilities configuration from data store (use feature user data)
* Update `canProcess` to look just for catalog param, resulting in better
  errors if parameters are incorrect
* Add `accumulo.zookeepers.timeout` data store param


docs/user/hbase/visibilities.rst
```diff
@@ -38,21 +38,6 @@ When connecting to your datastore you'll need to enable visibilities with the fo
     parameters.put("hbase.security.enabled", "true");
     DataStore ds = DataStoreFinder.getDataStore(parameters);
 
-Data Store Level Visibilities
```
elahrvivaz, line 41: these docs were actually incorrect, we haven't ever supported data-store-level vis for HBase

elahrvivaz:
tutorials PR: https://github.com/geomesa/geomesa-tutorials/pull/85

docs/tutorials/geomesa-examples-authorizations.rst
```diff
@@ -78,21 +78,7 @@ to show how authorizations work.
...
 
-GeoMesa supports applying a single set of visibilities to all data in a
-``DataStore``. When configuring a ``DataStore``, the visibilities can be
-set with the ``geomesa.security.visibilities`` parameter:
-
-.. code-block:: java
-
-    // create a map containing initialization data for the GeoMesa data store
-    Map<String, String> configuration = new HashMap<>();
-    configuration.put("geomesa.security.visibilities", "user&admin");
```
jnh5y, line 89: When did this option / capability go away?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
```diff
@@ -0,0 +1,130 @@
...
+    override protected val instanceKey: String = "instance.name"
+    override protected val zookeeperKey: String = "instance.zookeepers"
+    override protected val zookeeperTimeoutKey: String = "instance.zookeepers.timeout"
+    override protected val authTypeKey: Option[String] = Some("auth.type")
+    override protected val principalKey: Option[String] = Some("auth.principal")
+    override protected val tokenKey: Option[String] = Some("auth.token")
+  }
+
+  /**
+   * This class mimics the Accumulo 1.x logic for reading 'client.conf'
```
jnh5y, line 81: Mimics or implements?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -104,45 +103,60 @@ object AccumuloDataStoreFactory extends GeoMesaDataStoreInfo {
       null,
       Map(Parameter.DEPRECATED -> true, Parameter.IS_PASSWORD -> true).asJava)
 
-  override def canProcess(params: java.util.Map[String, _ <: Serializable]): Boolean = {
-    val hasConnection = InstanceIdParam.exists(params) && ZookeepersParam.exists(params) && UserParam.exists(params)
-    // exactly one of password or keytab
-    hasConnection && (PasswordParam.exists(params) != KeytabPathParam.exists(params))
-  }
+  override def canProcess(params: java.util.Map[String, _ <: Serializable]): Boolean =
+    CatalogParam.exists(params)
```
jnh5y, line 107: Would anything depend on the particular behavior of the canProcess method?  (I'm wondering how big a deal this change would be.)

I think the answer should be no, but just wanted to discuss it.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactoryTest.scala
```diff
@@ -76,8 +78,8 @@ class AccumuloDataStoreFactoryTest extends Specification {
         AccumuloDataStoreParams.PasswordParam.key   -> MiniCluster.Users.root.password,
         AccumuloDataStoreParams.CatalogParam.key    -> catalog
       ).asJava
-      AccumuloDataStoreFactory.canProcess(params) must beFalse
-      DataStoreFinder.getDataStore(params) must beNull
+      AccumuloDataStoreFactory.canProcess(params) must beTrue
+      DataStoreFinder.getDataStore(params) must throwAn[IOException]
```
jnh5y, line 82: Ok, this is an interesting change.  If we know that we'd thrown an exception, does it make sense that canProcess should be true?

docs/tutorials/geomesa-examples-authorizations.rst
elahrvivaz: i got rid of it in this PR - i think there's not really a use-case for it anymore.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
elahrvivaz: well there's not really an API or anything, just some conventions they follow...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: i don't think so... i think that having `canProcess` be more lenient and then having `createDataStore` more strict results in better error messages. E.g. before:
```
scala> DataStoreFinder.getDataStore(Map("foo" -> "bar"))
res1: org.geotools.data.DataStore = null
scala> DataStoreFinder.getDataStore(Map("accumulo.catalog" -> "bar"))
res2: org.geotools.data.DataStore = null
```
and after:
```
scala> DataStoreFinder.getDataStore(Map("accumulo.catalog" -> "bar"))
Apr 13, 2020 5:54:22 PM org.geotools.data.DataAccessFinder getDataStore
WARNING: Accumulo (GeoMesa) should be used, but could not connect
java.io.IOException: Parameter accumulo.instance.id is required: Accumulo Instance ID
	at org.locationtech.geomesa.accumulo.data.AccumuloDataStoreFactory$$anonfun$lookup$1$1.apply(AccumuloDataStoreFactory.scala:116)
<more stack trace>
java.io.IOException: Parameter accumulo.instance.id is required: Accumulo Instance ID
  at org.locationtech.geomesa.accumulo.data.AccumuloDataStoreFactory$$anonfun$lookup$1$1.apply(AccumuloDataStoreFactory.scala:116)
<more stack trace>
```

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactoryTest.scala
elahrvivaz: see above

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: i think technically this breaks the data store API, but seems better really. What do you think?

docs/user/accumulo/commandline.rst
```diff
@@ -21,22 +21,21 @@ here are Accumulo-specific.
...
+and ``--zookeepers``, and the username and password with ``--user`` and ``--password``. The password argument may be
+omitted in order to avoid plaintext credentials in the bash history and process list - in this case it will be
+prompted case for later. To use Kerberos authentication instead of a password, use ``--keytab`` with a path to a
+Kerberos keytab file containing an entry for the specified user. Since a keytab file allows authentication
+without any further constraints, it should be protected appropriately.
+
+Instead of specifying the cluster connection explicitly, an appropriate ``accumulo-client.properties`` (for Accumulo
+2) or ``client.conf`` (for Accumulo 1) may be added to the classpath. See the
+`Accumulo documentation <https://accumulo.apache.org/docs/2.x/getting-started/clients#creating-an-accumulo-client>`_
+for information on the necessary configuration keys.
```
jnh5y, line 35: Can we be explicit that once the accumulo-client.properties is on the classpath that the user does not need to specify all those parameters above?

Also, what happens if they mix and match?

Separately, is there a good way to use one set of command line tools to connect to multiple clusters?  (For that, I could imagine having a --accumulo-config option being helpful.

Just random thoughts...

docs/user/upgrade.rst
```diff
@@ -141,6 +141,12 @@ The signature for ``org.locationtech.geomesa.security.AuthorizationsProvider#con
 interfaces will need to update their method signature. Any classes invoking these methods should not need to updated,
 as the new signature is compatible with the old one.
 
+Accumulo Default Visibilities Removed
+-------------------------------------
+
+The Accumulo data store parameter ``geomesa.security.visibilities`` have been removed. Visibilities should be set
+per-feature, as per :ref:`accumulo_visibilities`.
```
jnh5y, line 148: Are per-feature visibilities only described in the Accumulo section (as this seems to imply)?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
jnh5y: I don't know what the exact API is.  I think this is sensible.  

A 'true' result from canProcess doesn't promise that the connection will work.  Throwing IllegalArgumentExceptions seems rather reasonable here.  (I don't know when IOExceptions or IllegalArgumentExceptions make more sense in a situation like this.)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -104,45 +103,60 @@ object AccumuloDataStoreFactory extends GeoMesaDataStoreInfo {
...
-      new PasswordToken(password.getBytes(StandardCharsets.UTF_8))
-    } else if (password == null && keytabPath != null) {
+    val auth: AuthenticationToken = if (authType == AccumuloClientConfig.PasswordAuthType) {
+      new PasswordToken(lookup(PasswordParam, config.token).getBytes(StandardCharsets.UTF_8))
+    } else if (authType == AccumuloClientConfig.KerberosAuthType) {
       // explicitly enable SASL for kerberos connections
       // this shouldn't be required if Accumulo client.conf is set appropriately, but it doesn't seem to work
       conf.withSasl(true)
-      val file = new java.io.File(keytabPath)
+      val file = new java.io.File(lookup(KeytabPathParam, config.token))
```
jnh5y, line 153: (Random question) Have we seen Accumulo 2.0 in a Kerberized environment?  

If not, is it worth noting that in the documentation somehow?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreParams.scala
```diff
@@ -20,23 +22,27 @@ object AccumuloDataStoreParams extends GeoMesaDataStoreParams with SecurityParam
...
 
   val ZookeepersParam =
     new GeoMesaParam[String](
       "accumulo.zookeepers",
       "Zookeepers",
-      optional = false,
       deprecatedKeys = Seq("zookeepers"),
       supportsNiFiExpressions = true)
 
+  val ZookeeperTimeoutParam =
```
jnh5y, line 35: This is only used for Accumulo 2.0, right?  

If so, is it worth adding something about that in the description so that Accumulo 1.9 NiFi/GeoServer users don't get excited about tuning this parameter?

docs/user/upgrade.rst
elahrvivaz: Currently we have a section in each data store and no generic section. That might be worth revisiting since it is largely the same across data stores. I opened a ticket: https://geomesa.atlassian.net/browse/GEOMESA-2830

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
```diff
@@ -61,12 +61,7 @@ trait KerberosParams {
   var keytab: String = _
 }
 
-trait RequiredCredentialsParams extends PasswordParams {
-  @Parameter(names = Array("-u", "--user"), description = "Connection user name", required = true)
-  var user: String = _
-}
-
-trait OptionalCredentialsParams extends PasswordParams {
+trait CredentialsParams extends PasswordParams {
```
jnh5y, line 64: Ok, why did the required username parameter go away?  Don't we have at least one DS that requires a username somewhere:)?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: we have not... we have some caveats on how it was tested already here: https://www.geomesa.org/documentation/user/accumulo/kerberos.html#development-testing

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreParams.scala
elahrvivaz: this will also be used for accumulo 1.9, we just didn't expose it as a ds param before, it was only as a system prop.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/CommonParams.scala
elahrvivaz: nope :)

docs/user/accumulo/commandline.rst
elahrvivaz: I thought it was pretty explicit, especially if you read the linked accumulo docs... i can make it more so.
data store params are first precedence and the props file is second.
i originally had logic to load the props from various places that accumulo checks, but decided that was a bit opaque... instead i modeled it after the hbase-site.xml logic. if you want to connect to multiple clusters, you can specify the connection directly, so i don't think we need to support an accumulo-config option.

jnh5y:
Merged as 2d4d550dc0079af6dff9a7033b7a397e47020bec.

PR #2522, opened 2020-04-14
GEOMESA-2783 Heatmaps on non-default geometries

elahrvivaz:
* New WPS parameter and query hint DENSITY_GEOM


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
```diff
@@ -237,7 +237,11 @@ object LocalQueryRunner {
     } else if (hints.isDensityQuery) {
       val Some(envelope) = hints.getDensityEnvelope
       val Some((width, height)) = hints.getDensityBounds
-      densityTransform(sampled, sft, envelope, width, height, hints.getDensityWeight)
+      val geom = hints.getDensityGeometry.getOrElse(sft.getGeometryDescriptor.getLocalName)
+      require(
```
jnh5y, line 241: Meh, since this require is in a few places, it is worth having a method that gets the DensityGeometry and throws an exception if it cannot?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/DensityProcess.scala
```diff
@@ -161,6 +167,9 @@ class DensityProcess extends GeoMesaProcess {
     invertedQuery.getHints.put(QueryHints.DENSITY_BBOX, envelope)
     invertedQuery.getHints.put(QueryHints.DENSITY_WIDTH, outputWidth)
     invertedQuery.getHints.put(QueryHints.DENSITY_HEIGHT, outputHeight)
+    if (argGeomAttr != null) {
```
jnh5y, line 170: As the query is inverted...  is the bbox applied to the default geometry or the 'desnity' geometry?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
elahrvivaz: yeah, i didn't want to make the RichHints too complicated but i can add a method in DensityScan.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/DensityProcess.scala
elahrvivaz: The envelope is pulled out and applied against the 'density' geometry in `QueryRunner.configureQuery`

jnh5y: Should we add a logger.warn in case folks query a non-indexed geometry?  

I'm wondering if we are letting people do something they'll regret...

elahrvivaz: haha, this was your idea...
if you query against a non-indexed geom we already log a warning about full table scans.

elahrvivaz: i think if they're calling it through geoserver there will be a bbox filter on the primary geometry, and then a secondary filter on the second geometry. that may result in some 'missing' data if the primary geom isn't in view but the second one is, but _shrugs_

jnh5y: For the literal 'edge' case, I wonder if we need to provide some buffering mechanism?  Hmm...

elahrvivaz: we do buffer the envelope here: https://github.com/locationtech/geomesa/pull/2522/files#diff-37e08ab303976058cacaadf59c39c4c7R157

jnh5y:
Merged as 4e1ce4a0c2b6bd2b33f95eac146d1dff8046b8ea.

PR #2526, opened 2020-04-17
GEOMESA-2698 SimpleFeatureType config to prioritize time-based indices

elahrvivaz:


docs/user/datastores/index_config.rst
```diff
@@ -649,6 +649,19 @@ For example:
 
 See :ref:`cli_analytic` and :ref:`stats_api` for information on reading cached stats.
 
+Configuring Temporal Priority
+-----------------------------
+
+For some large time-based datasets, an index that leverages a temporal predicate will almost always be faster
```
jnh5y, line 655: If we believe this, should we make the default 'true'?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/FilterSplitter.scala
```diff
@@ -285,7 +286,9 @@ object FilterSplitter {
     }
 
     if (canMergePrimary && toMerge.secondary == mergeTo.secondary) {
-      FilterStrategy(mergeTo.index, orOption(toMerge.primary.toSeq ++ mergeTo.primary), mergeTo.secondary, mergeTo.cost)
+      val primary = orOption(toMerge.primary.toSeq ++ mergeTo.primary)
+      val temporal = mergeTo.temporal && toMerge.temporal // should be the same...
```
jnh5y, line 290: One can only hope.:)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/StrategyDecider.scala
```diff
@@ -131,23 +131,39 @@ object StrategyDecider extends MethodProfiling with LazyLogging {
...
+
+    private def cost(option: FilterPlan): FilterPlanCost = {
+      profile((fpc: FilterPlanCost, time: Long) => fpc.copy(time = time)) {
+        FilterPlanCost(option, option.strategies.map(_.cost).sum, 0L)
+      }
+    }
+
+    private case class FilterPlanCost(plan: FilterPlan, cost: Long, time: Long) extends Comparable[FilterPlanCost] {
+      override def compareTo(o: FilterPlanCost): Int = java.lang.Long.compare(cost, o.cost)
+      override def toString: String = s"$plan (Cost $cost in ${time}ms)"
```
jnh5y, line 166: Interesting... are we seeing cost calculations taking a noticeable amount of time?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/IdFilterStrategy.scala
```diff
@@ -17,20 +17,22 @@ import org.opengis.filter.{And, Filter, Id, Or}
...
+      Some(FilterStrategy(this, None, None, temporal = false, Long.MaxValue))
     } else if (filter == Filter.EXCLUDE) {
       None
     } else {
       val (ids, notIds) = IdExtractingVisitor(filter)
       if (ids.isDefined) {
         // top-priority index - always 1 if there are actually ID filters
-        Some(FilterStrategy(this, ids, notIds, IdFilterStrategy.StaticCost))
+        // note: although there's no temporal predicate, there's an implied exact date for the given feature
+        Some(FilterStrategy(this, ids, notIds, temporal = true, IdFilterStrategy.StaticCost))
```
jnh5y, line 33: Nice corner case.  

This does seem to suggest that we may want to think through all the cases our query planning handles at some point and think through it holistically.  

It seems good and bad to associate an id query (which has the ids specified) as 'temporal'.  Locally, right now, it is the Right Thing though.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/SpatialFilterStrategy.scala
```diff
@@ -19,23 +19,24 @@ trait SpatialFilterStrategy[T, U] extends GeoMesaFeatureIndex[T, U] {
 
   import SpatialFilterStrategy.{StaticCost, spatialCheck}
 
-  def geom: String
+  // attributes are assumed to be a single geometry field
+  lazy private val Seq(geom) = attributes
```
jnh5y, line 23: Is there any way that this is not true?

jnh5y: I'm just wondering if there's any chance for an exception, etc.  (It seems believable that everything is ok here; just asking.)

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
```diff
@@ -235,6 +237,24 @@ class GeoMesaDataStoreTest extends Specification {
...
+      val spec = "name:String:index=true,age:Int,dtg:Date,*geom:Point:srid=4326"
+
+      val ds = new TestGeoMesaDataStore(true)
+      ds.createSchema(SimpleFeatureTypes.createType("default", spec))
+      ds.createSchema(SimpleFeatureTypes.createType("temporal", s"$spec;geomesa.temporal.priority=true"))
+
+      val filters = Seq(
+        ("name like 'a%' and dtg DURING 2020-01-01T00:00:00.00Z/2020-01-02T00:00:00.00Z", AttributeIndex, Z3Index),
+        ("name > 'a' AND name < 'b' and dtg DURING 2020-01-01T00:00:00.00Z/2020-01-02T00:00:00.00Z", AttributeIndex, Z3Index),
+        ("name = 'a' and dtg DURING 2020-01-01T00:00:00.00Z/2020-01-02T00:00:00.00Z", AttributeIndex, AttributeIndex)
```
jnh5y, line 250: Would it be 'easy' to add tests for the id cases above?

docs/user/datastores/index_config.rst
elahrvivaz: well, there's the `For some large time-based datasets` part...

elahrvivaz: in practice, i think the only thing this affects is prefix attribute scans against high-cardinality attributes get de-prioritized below z3.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/StrategyDecider.scala
elahrvivaz: not with the current heuristic impl - using stats could take a bit longer. we've always had this logging in here, i just cleaned it up a little.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/IdFilterStrategy.scala
elahrvivaz: yeah, the current approach of using costs and delegating the cost to the index means that we don't have a single place to rank everything... but the tradeoff is that it's extensible.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/strategies/SpatialFilterStrategy.scala
elahrvivaz: only if there's a new class that extends this trait and doesn't follow the current conventions... i thought about adding some validation here but seemed like overkill.

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
elahrvivaz: easy enough - but id filters are always #1 priority so it wouldn't show any change. i guess that would be a meaningful result in itself. i'll add one

elahrvivaz: added

jnh5y:
Merged as c11aa3c279e1c7106dbaf8864c5b1dd17a658951

PR #2527, opened 2020-04-17
GEOMESA-2831 Single configuration for table compression

elahrvivaz:


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -53,7 +53,6 @@ object SimpleFeatureTypes {
     val OverrideReservedWords = "override.reserved.words"
     val QueryInterceptors     = "geomesa.query.interceptors"
     val StatsEnabled          = "geomesa.stats.enable"
-    val TableCompression      = "geomesa.table.compression.enabled"
```
jnh5y, line 56: Should we note in the upgrade guide that this configuration is gone?

elahrvivaz: we still support the old functionality, so i think it's ok to not mention it. if we ever drop that support we can mention it.

jnh5y:
Merged as 9736b9ff3ed9697f6232c8b7f5815d51cb981b20.

PR #2530, opened 2020-04-28
GEOMESA-2799 Option to sort results on disk

elahrvivaz:
* System property `geomesa.sort.memory.threshold` controls memory threshold
* New AccumuloDataStore parameters for toggling Accumulo remote processing
* Reduce in-memory footprint of simple feature implementations
* Replace LambdaQueryRunner with MergedQueryRunner
* Distributed Arrow encoding will *not* use disk for sorting


build/cqs.tsv
```diff
@@ -182,6 +182,7 @@ org.eclipse.emf:org.eclipse.emf.common	2.15.0	compile
 org.eclipse.emf:org.eclipse.emf.ecore	2.15.0	compile
 org.eclipse.emf:org.eclipse.emf.ecore.xmi	2.15.0	compile
 org.eclipse.xsd:org.eclipse.xsd	2.12.0	compile
+org.ehcache:sizeof	0.3.0	compile
```
elahrvivaz, line 185: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22019

elahrvivaz: approved

docs/user/accumulo/configuration.rst
```diff
@@ -40,6 +40,39 @@ The timeout is defined as a duration, e.g. ``60 seconds`` or ``100 millis``. See
 
 __ https://accumulo.apache.org/1.9/apidocs/org/apache/accumulo/core/client/BatchWriterConfig.html#setTimeout(long,%20java.util.concurrent.TimeUnit)
 
+Remote Processing Properties
+----------------------------
+
+The following properties control the push-down processing for certain queries to the Accumulo tablet servers, vs
```
jnh5y, line 46: Hmm...  I wonder if this would read better as 

"The following properties control the push-down processing for certain queries to the Accumulo tablet servers versus processing in the client."

I'm trying to figure out if the comma here is good or bad...

docs/user/datastores/runtime_config.rst
```diff
@@ -240,6 +240,19 @@ This property controls how long simple feature serializers will be cached in mem
...
+
+This property can be used to constrain the memory used to sort result sets. GeoMesa sorts results in the client
+process memory, since the supported back-end databases don't offer native ordering. To avoid having large
+result sets exceed the client memory capacity, a memory threshold can be set. Once the size of a result set
+exceeds this threshold, additional results will be written to disk and sorted there. Note that the actual memory
+used may exceed the threshold, as the memory footprint calculation is an estimation. The threshold is specified
+as a number of bytes, e.g. ``10MB``. The default is to always sort in memory.
+
+Note that distributed Arrow queries will never use disk to sort, due to the nature of Arrow result batches. For
+supported back-ends, sorting on disk for Arrow queries can be achieved by disabling remote Arrow processing.
```
jnh5y, line 254: Later, I'm going to be looking for a warning if a user has set this setting and remote arrow queries are still enabled...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
```diff
@@ -255,4 +275,5 @@ object AccumuloDataStoreFactory extends GeoMesaDataStoreInfo {
       caching: Boolean
     ) extends DataStoreQueryConfig
 
+  case class RemoteScansEnabled(arrow: Boolean, bin: Boolean, density: Boolean, stats: Boolean)
```
jnh5y, line 278: Hmm...  Should this factored out?  Meh, probably not?  Maybe later?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
```diff
@@ -155,29 +156,71 @@ class AccumuloIndexAdapter(ds: AccumuloDataStore) extends IndexAdapter[AccumuloD
...
+          if (ds.config.remote.bin) {
+            val iter = BinAggregatingIterator.configure(schema, index, ecql, hints)
+            (Seq(iter), new AccumuloBinResultsToFeatures(), None)
+          } else {
+            if (hints.isSkipReduce) {
+              // override the return sft to reflect what we're actually returning,
+              // since the bin sft is only created in the local reduce step
+              hints.hints.put(QueryHints.Internal.RETURN_SFT, returnSchema)
+            }
+            (fti, resultsToFeatures, localReducer)
```
jnh5y, line 183: Above it looks like the localReducer gets an 'arrowHook'.  Does that get used / matter here?  

Or is it just expedient to express things that way?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
```diff
@@ -155,29 +156,71 @@ class AccumuloIndexAdapter(ds: AccumuloDataStore) extends IndexAdapter[AccumuloD
...
         } else if (strategy.hints.isStatsQuery) {
-          val iter = StatsIterator.configure(schema, index, ecql, hints)
-          val reduce = Some(StatsScan.StatsReducer(schema, hints))
-          (Seq(iter), new AccumuloStatsResultsToFeatures(), reduce)
+          if (ds.config.remote.stats) {
+            val iter = StatsIterator.configure(schema, index, ecql, hints)
+            val reduce = Some(StatsScan.StatsReducer(schema, hints))
+            (Seq(iter), new AccumuloStatsResultsToFeatures(), reduce)
+          } else {
+            if (hints.isSkipReduce) {
```
jnh5y, line 215: Ok, I'll admit that I'm only half getting this. 

For a given 'function' (e.g. Density), are there situations where isSkipReduce is both true and false (when remote is false)?  If so, can you give an example of each?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -224,42 +224,46 @@ object DeltaWriter extends StrictLogging {
...
+   * @param batchSize record batch size
+   * @param threadedBatches record batches, grouped by threading key
+   * @return
+   */
   private def reduceNoSort(
       sft: SimpleFeatureType,
       dictionaryFields: Seq[String],
       encoding: SimpleFeatureEncoding,
       mergedDictionaries: MergedDictionaries,
+      sort: Option[(String, Boolean)],
```
jnh5y, line 266: This is indicating the field to sort and whether or not it is ascending?  

Makes me think of the GeoTools Sort/SortOrder classes.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/AbstractSimpleFeature.scala
```diff
@@ -115,6 +110,24 @@ object AbstractSimpleFeature {
     }
     override def setDefaultGeometryProperty(geoAttr: GeometryAttribute): Unit =
       if (geoAttr == null) { setDefaultGeometry(null) } else { setDefaultGeometry(geoAttr.getValue) }
+
+    /**
+     * Mutable feature ID implementation that references back to the parent class. This allows us to
+     * avoid allocating a new object unless `getIdentifier` is called.
```
jnh5y, line 116: Ok, this is a quick little performance enhancement?  

How did you come to this idea?  I.e., is there a clear need?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -8,32 +8,55 @@
...
   override def getAttribute(index: Int): AnyRef = values(index)
+
+  override def getUserData: java.util.Map[AnyRef, AnyRef] = synchronized {
+    if (userData == null) {
+      userData = new java.util.HashMap[AnyRef, AnyRef]()
+    }
+    userData
+  }
+
+  override def calculateSizeOf(): Long = Sizable.sizeOf(this) + Sizable.deepSizeOf(id, values, userData)
```
jnh5y, line 59: Ah!  I was trying to figure out why these classes were getting TLC!

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -116,142 +139,180 @@ object ScalaSimpleFeature {
...
-          if (word.add(index)) {
-            attributes(index) = reader.read(index)
-            phaser.arriveAndDeregister()
-          }
-        }
-      }
-      attributes(index)
-    }
+    override def calculateSizeOf(): Long =
+      Sizable.sizeOf(this) + Sizable.deepSizeOf(id, bits, attributes, userData, count, reader, userDataReader)
```
jnh5y, line 199: Hmm... is the reader specific to this SimpleFeature?  If so, I suppose counting its size makes sense.

Otherwise, can we share a reader?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -116,142 +139,180 @@ object ScalaSimpleFeature {
...
+              count = null
+            }
           }
         }
       }
       attributes(index) = value
     }
 
+    override def calculateSizeOf(): Long =
+      Sizable.sizeOf(this) + Sizable.deepSizeOf(id, bits, attributes, userData, count, reader, userDataReader)
```
jnh5y, line 248: What order do we expect calculateSizeOf and setAttributeNoConverter to be called in?

I'm wondering if we should be counting the size of things like the reader and count which could null or used?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -116,142 +139,180 @@ object ScalaSimpleFeature {
...
+                attributes = Array.ofDim[AnyRef](getFeatureType.getAttributeCount)
+              }
+            }
             attributes(index) = reader.read(index)
-            phaser.arriveAndDeregister()
+            if (count.decrementAndGet() == 0) {
+              // once all attributes have been read, dereference any backing resources so that they can be gc'd
+              reader = null
+              count = null
+            }
```
jnh5y, line 280: Ok, the first time I read through this...  I thought that after getAttribute was called once, then the next time there'd be an NPE since count is null.  Then I realized that we'd need to zero out 'count', so the 'count+first' time would have an NPE.

But then I realized that's not right.  Since the reader would be null, we'd be ok.

Ok, this next time through, I'm wondering if we only decrement count when we read a new attribute?  If so, then I think I understand it. 

If not, then calling getAttribute(0) 'count' times would remove the reader/count and cause trouble....

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
```diff
@@ -116,142 +139,180 @@ object ScalaSimpleFeature {
...
     }
-  }
 
+    override def getUserData: java.util.Map[AnyRef, AnyRef] = {
+      if (userDataReader != null) {
+        synchronized {
+          if (userData == null) {
+            userData = userDataReader.read()
+            userDataReader == null // dereference any backing resources
+          }
```
jnh5y, line 293: Do we need to do any handling in the userDataReader != null && userData != null case?  

Can someone call 'setUserData' before the userDataReader is used (and hence released)?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureDeserialization.scala
```diff
@@ -48,18 +48,18 @@ trait KryoFeatureDeserialization extends SimpleFeatureSerializer with LazyLoggin
...
-    val in = inputs.getOrElseUpdate(new NonMutatingInput())
+    val in = inputBytes.getOrElseUpdate(new NonMutatingInput())
     in.setBuffer(bytes, offset, count)
     in
   }
 
   def getInput(stream: InputStream): Input = {
-    val in = inputs.getOrElseUpdate(new NonMutatingInput())
-    in.setBuffer(Array.ofDim(1024))
+    val in = inputStreams.getOrElseUpdate(new NonMutatingInput(Array.ofDim(1024)))
```
jnh5y, line 62: Any particular reason for this change?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureDeserialization.scala
```diff
@@ -48,18 +48,18 @@ trait KryoFeatureDeserialization extends SimpleFeatureSerializer with LazyLoggin
 
 object KryoFeatureDeserialization extends LazyLogging {
 
-  private val inputs  = new SoftThreadLocal[Input]()
-  private val readers = new ThreadLocalCache[String, Array[KryoAttributeReader]](SerializerCacheExpiry)
+  private val inputBytes   = new SoftThreadLocal[Input]()
+  private val inputStreams = new SoftThreadLocal[Input]()
```
jnh5y, line 52: Ok, I'm stumped.  Why do we have two 'Input's here?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/LazyDeserialization.scala
```diff
@@ -75,6 +77,11 @@ object LazyDeserialization {
         readers(i).apply(input)
       }
     }
+
+    override def calculateSizeOf(): Long = {
+      // doesn't count shared readers
```
jnh5y, line 82: +1

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
```diff
@@ -282,27 +273,20 @@ object LocalQueryRunner {
...
-                             hook: Option[ArrowDictionaryHook]): CloseableIterator[SimpleFeature] = {
+  private def arrowTransform(
+      original: CloseableIterator[SimpleFeature],
+      sft: SimpleFeatureType,
+      transform: Option[(String, SimpleFeatureType)],
+      hints: Hints,
+      hook: Option[ArrowDictionaryHook]): CloseableIterator[SimpleFeature] = {
 
-    val sort = hints.getArrowSort
+    val sort = hints.getArrowSort.map(Seq.fill(1)(_))
```
jnh5y, line 283: What's this trick?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreFactory.scala
elahrvivaz: maybe when we get to 3 of em ;)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
```diff
@@ -34,15 +36,18 @@ import org.opengis.filter.Filter
   * @param ds merged data store
   * @param stores delegate stores
   */
-class MergedQueryRunner(ds: MergedDataStoreView, stores: Seq[(DataStore, Option[Filter])])
+class MergedQueryRunner(ds: HasGeoMesaStats, stores: Seq[(Queryable, Option[Filter])])
```
jnh5y, line 39: This is an interesting sharpening....

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
elahrvivaz: it's needed in certain query patterns to pre-calculate the dictionaries for an arrow query - it will get used if it's needed but otherwise not. it's basically just a reference back to the full filter for the query and the data store stats for calculating enumerations, which aren't otherwise available down where they're needed.

geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/data/LambdaQueryRunner.scala
```diff
@@ -70,72 +58,25 @@ class LambdaQueryRunner(persistence: DataStore, transients: LoadingCache[String,
           .read(Option(query.getFilter), Option(query.getPropertyNames), Option(query.getHints), explain)
     }
   }
+}
 
-  private def runMergedQuery(sft: SimpleFeatureType, query: Query, explain: Explainer): CloseableIterator[SimpleFeature] = {
```
jnh5y, line 74: Oh!  Did this get factored out between this and the MergedView DS?!  If so, wow!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
elahrvivaz: `skipReduce` will always be either true or false. It's used in the merged data store views, because if we run the reduce step we sometimes get things that can't be merged together easily (i.e. we go from arrow delta batches that we can merge to finalized arrow files that we can't). with remote encoding enabled, the intermediate (pre-reduce) features are already arrow/density/etc encoded, but if we do the encoding locally it's done in the reduce step. in that case, the simple feature type of the iterator doesn't match the simple feature type of the features in the iterator, which causes errors in the merge step when checking to see if the results were already encoded or not.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
```diff
@@ -280,6 +280,11 @@ object ExportCommand extends LazyLogging {
       }
       val order = if (params.sortDescending) { SortOrder.DESCENDING } else { SortOrder.ASCENDING }
       query.setSortBy(fields.map(f => org.locationtech.geomesa.filter.ff.sort(f, order)).toArray)
+    } else if (hints.isArrowQuery) {
```
jnh5y, line 283: This makes sense.  What did we do before?  

How does one call for an Arrow sort from the command line?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: yeah it's equivalent to that - maybe we should extract it out into a case class or trait.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/stats/StatsCountCommand.scala
```diff
@@ -32,12 +32,18 @@ trait StatsCountCommand[DS <: DataStore with HasGeoMesaStats] extends DataStoreC
...
+      val count = ds.stats.getCount(sft, filter, params.exact).map(_.toString).getOrElse("Unknown")
+      Command.output.info(s"Count: $count")
+    } else {
+      ds.stats.getCount(sft, filter, params.exact).map(_.toString) match {
+        case None =>
+          Command.output.info("Estimated count: Unknown")
+          Command.output.info("Re-run with --no-cache to get an exact count")
+
+        case Some(count) =>
+          Command.output.info(s"Estimated count: $count")
```
jnh5y, line 44: I assume this is just clean up from other discussions?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/Sizable.scala
```diff
@@ -0,0 +1,40 @@
...
+   */
+  def calculateSizeOf(): Long
+}
+
+object Sizable extends LazyLogging {
+
+  private val sizeOf = try { SizeOf.newInstance() } catch {
+    case NonFatal(e) =>
+      logger.warn("Could not load SizeOf, memory estimates will be off:", e)
+      new SizeOf(new PassThroughFilter(), true, true) {
```
jnh5y, line 32: Under what conditions would we expect this to happen?!

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/AbstractSimpleFeature.scala
elahrvivaz: I started going through to calculate the size of each feature in memory, and there were some local vals that we probably don't need to keep around - i.e. having a FeatureID instance for each feature, keeping the default geometry index in each feature, etc. This way, the FeatureID and  default geometry index will be re-created each time they get called, but generally I would expect that to only happen at most once? So technically it may be slightly slower for certain operations, but it should reduce the memory footprint and avoid some unnecessary calls (e.g. if we never call getFeatureID and just call getID, if we never call getDefaultGeometry).

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: it is specific to this feature, as it contains the backing data (e.g. byte array) and offset into it for this feature.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: i think they get nulled out after they're no longer needed - the deepSizeOf method will skip over null values, so they won't get counted if they are null.

elahrvivaz: but to answer your more general question, they could be called in any order. in the standard query workflow, `setAttributeNoConvert` won't be called at all, and sizeOf will only be called if we have to sort on disk.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: yeah, it's bounded by the `if(word.add(index))` clause, which is an atomic set addition, so it should only be decremented the first time a given attribute is accessed.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+    extends CloseableIterator[SimpleFeature] {
+
+  import SortingSimpleFeatureIterator.{sortInMemory, sortWithSpillover}
+
+  private val closed = new AtomicBoolean(false)
+
+  private lazy val sorted: CloseableIterator[SimpleFeature] = {
+    if (closed.get || !features.hasNext) { features } else {
+      val head = features.next()
+      if (closed.get || !features.hasNext) { CloseableIterator.single(head, features.close()) } else {
```
jnh5y, line 46: Ok, I'll bite.  Why does the single element iterator get handled separately?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: there is no `setUserData` method, so we don't have to worry about that...
i think the synchronized block should ensure that if the user data is null, then the userDataReader will not be null, since it is only set to null inside that block. it's possible that multiple threads could hit the unsynchronized `if (userDataReader != null)` check, but i don't _think_ that would matter since they'd block on the synchronized block, and userDataReader can't be null until userData has been read and is thus valid to return.
any subtle threading issues i'm missing here?

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureDeserialization.scala
elahrvivaz: lets us re-use a cached byte array instead of always creating a new one for the input stream case.

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/impl/KryoFeatureDeserialization.scala
elahrvivaz: one is for getInput for reading an InputStream, the other for getInput for reading a byte array

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
elahrvivaz: ah, not much of a trick... the arrow sort is defined as a single field (since we only support sorting on one field), while the general geomesa sort can handle a Seq of fields, so this just converts between the two.
i.e. we need to go from `Option[(String, Boolean)]` to `Option[Seq[(String, Boolean)]]`

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
elahrvivaz: yeah, there was a lot of common code between the LambdaQueryRunner and this class - this abstraction lets us use this one for both, so I don't have to fix all the local/remote arrow calls in both places.

geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/data/LambdaQueryRunner.scala
elahrvivaz: yep!

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+      head: SimpleFeature,
+      tail: CloseableIterator[SimpleFeature],
+      ordering: Ordering[SimpleFeature],
+      closed: AtomicBoolean): CloseableIterator[SimpleFeature] = {
+    // use ArrayList for sort-in-place of the underlying array
+    val list = new java.util.ArrayList[SimpleFeature](100)
+    list.add(head)
+    // note: we already ensured tail.hasNext == true before calling this method
+    do { list.add(tail.next()) } while (tail.hasNext && !closed.get)
+    if (closed.get) {
```
jnh5y, line 90: Ok, in what case does the list get sorted *right* it is closed?  

Also, I think a newline above this line may increase readability.  While scanning the code, I was looking for {} associated to the 'while'.  (It took a split second to see the 'do'.)

Related to the do-while, I wonder if it is better as while (even though we have already done the first checks previously).

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
elahrvivaz: this is just to ensure that any arrow sort hints get set as a regular query sort. in geomesa it doesn't matter if it's in one or both places, we will handle it the same. but for other data sources (e.g. postgis or whatever), we want to make sure that the results come back already sorted, so that we don't need complicated sort logic in this ArrowExporter.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/stats/StatsCountCommand.scala
elahrvivaz: yeah, you got me ;)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+    var size = sizable.sizeOf(head)
+
+    // note: we already ensured tail.hasNext == true before calling this method
+    do {
+      if (size >= threshold) {
+        // write out the sorted list to disk
+        list.sort(ordering)
+        files += Files.createTempFile(s"gm-sort-$id-", ".kryo").toFile
+        logger.trace(s"Created temp sort file '${files.last.getAbsolutePath}'")
+        WithClose(new FileOutputStream(files.last)) { os =>
```
jnh5y, line 146: Is files.last 'quick'?  Or would it be better to have a 'val file = Files.createTempFile', operate on it, and then add it to 'files'?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/Sizable.scala
elahrvivaz: I don't think this would necessarily ever happen... you can see the different approaches attempted by ehcache:sizeof here: https://github.com/ehcache/sizeof#the-different-engines
this was more best practice of not killing the whole system if for some reason it can't load a sizeof implementation.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: it's already sorted ;)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: it's an array buffer, so it should be just an array lookup - i thought it was safer to add it directly to the `files` object but i can make it a local val if you think that's better.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+      }
+      CloseableIterator(Iterator.empty, tail.close())
+    } else {
+      if (!list.isEmpty) {
+        list.sort(ordering)
+      }
+      if (files.isEmpty) {
+        CloseableIterator(list.iterator.asScala, tail.close())
+      } else {
+        new MergeSortingIterator(files, serializer, list.iterator.asScala, tail, ordering)
```
jnh5y, line 178: Just as a sanity check, 'tail' here is empty, right?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: > Ok, in what case does the list get sorted right it is closed?
can you clarify what you mean?
I can fix the formatting and get rid of the `do`

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  /**
+   * Does a merge sort on a group of locally sorted files and a left-over in-memory iterator
+   *
+   * @param files files to merge
+   * @param serializer serializer
+   * @param mem left-over in-memory features, already sorted
+   * @param closeable link to the original feature iterator, for cleaning up on close
```
jnh5y, line 189: Can we just close the original iterator after it has been exhausted?  

Isn't it likely already a CloseIterator (and hence closed by exhaustion above)?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: yeah, i clarified in the scaladocs on the iterator class

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: do you mean a SelfClosingIterator? possibly... in general I don't think we use SelfClosingIterators though. I passed it back so that any exceptions on close would be throw to the client.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+  private object SizableFeatureIsSizable extends FeatureIsSizable[SimpleFeature with Sizable] {
+    override def sizeOf(feature: SimpleFeature with Sizable): Long = feature.calculateSizeOf()
+  }
+
+  /**
+   * Use a rough estimate based on attributes and user data
+   */
+  private object UnSizableFeatureIsSizable extends FeatureIsSizable[SimpleFeature] {
+    override def sizeOf(feature: SimpleFeature): Long =
+      Sizable.sizeOf(feature) + Sizable.deepSizeOf(feature.getAttributes, feature.getUserData)
```
jnh5y, line 296: In this case, do we expect Sizable.sizeOf(feature) to return that 64 as a default value?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
```diff
@@ -0,0 +1,289 @@
...
+   * Does a merge sort on a group of locally sorted files and a left-over in-memory iterator
+   *
+   * @param files files to merge
+   * @param serializer serializer
+   * @param mem left-over in-memory features, already sorted
+   * @param closeable link to the original feature iterator, for cleaning up on close
+   * @param ordering feature ordering
+   */
+  private class MergeSortingIterator(
+      files: IndexedSeq[File],
```
jnh5y, line 202: Meh, it might be a 'later' thing, but it seems like we could separate the concern of opening/reading files and merging sorted iterators of SimpleFeatures.

On the other hand, this is the first time we've messed with Merge Sorting features in the project, right?  So maybe it doesn't need to be generalized.

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIteratorTest.scala
```diff
@@ -0,0 +1,157 @@
...
+      test(features, Seq(("name", false), ("", false)), Seq(a, b, c1, c2, d))
+    }
+
+    "be able to sort by an multiple attributes" >> {
+      val features = new TestCloseableIterator(Iterator(a, b, c1, d, c2))
+      test(features, Seq(("age", true), ("name", false)), Seq(b, c2, a, c1, d))
+    }
+
+    "be able to spill to disk" >> {
+      val threshold = Seq(a, b, c1, c2, d).map(_.calculateSizeOf()).min * 2
```
jnh5y, line 116: How many files are you seeing this create on disk?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
elahrvivaz: on the command line you can either use `--sort-by` or `--hints ARROW_SORT_FIELD=foo`

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: only if the actual sizeof agent can't be loaded. deepSizeOf delegates to the sizeof function but has some additional logic, so I'm not sure exactly what it would return

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIteratorTest.scala
elahrvivaz: i think it created 2 files. i couldn't think of a way to check that it was spilling to disk as a unit test assertion without exposing a lot of internal state though...

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
jnh5y: Ok, I'm up on to 75% understanding.  For the last quarter, Does isSkipReduce==true when the features have been processed with their function remotely?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
jnh5y: Do you want to extract that as a class now or later?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
jnh5y: "Maybe".

This seems like a backwards version of the 'lazy initialization' problem.  For that double-locking is either answer or an anti-pattern.

https://en.wikipedia.org/wiki/Double-checked_locking
https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

jnh5y: In short, I don't know.  I think two threads could both get to the synchronized block at the same time.  The second one should throw an error/NPE (line 291), right?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/LocalQueryRunner.scala
jnh5y: Ok, so if/when we refactor the Option[(String, Boolean)] to be Option[SortingClass], this would make more sense....

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
jnh5y: Is the Lambda DS just a MergedView DS with a write path?:)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
jnh5y: I've been in foot races by myself and still come in second...  But yeah, here, that makes sense.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
jnh5y: Ah, I think I meant to ask ....

"In what cases does the list get sorted and then immediately closed?"  Aka, is this check necessary?  (If it is a rare case, we can leave it.)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
jnh5y: Makes sense.   Maybe check that fine detail and maybe add a comment?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIteratorTest.scala
jnh5y: Hmm...  the files are created in 'java.io.tmpdir', but they'd be cleaned up by the sort, right?  If not, then one could check the number of files in that directory, etc.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
elahrvivaz: this isn't quite the same as double checked locking, since the second check is a different variable. I *think* this is ok... taking it line-by-line:
```
if (userDataReader != null) {
  synchronized {
    if (userData == null) {
      userData = userDataReader.read()
      userDataReader == null // dereference any backing resources
    }
  }
}
userData
```

1. It's guaranteed that by the userData will be non-null by the time userDataReader is null. Thus any thread that comes through and sees userDataReader as null will return the non-null userData. Potentially there needs to be a volatile keyword on userDataReader? I assumed setting something null would be a single operation, but maybe nt.
2. If userDataReader is not null, then it will wait on a synchronized block. Once it gains the block, it will check for userData being null and read it if needed. If another thread gained the synchronized block first, then it will just continue on.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloIndexAdapter.scala
elahrvivaz: when they're processed remotely, they get partially encoded in the remote iterator, so the return type is correct and doesn't need to be overridden.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: we can do that later...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
elahrvivaz: kind of... the 'transient' part of it isn't a full-fledged data store, and it also has some logic to persist things that haven't been modified in a while. We could probably abstract it out to where that is the case, which might make it easier to support non-accumulo back-ends.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: this would be to cover the query timeout kill. the while loop is checking the `closed` val, so if the query gets killed before it completes, we'll exit the loop and then this check lets us just return immediately without sorting.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIterator.scala
elahrvivaz: i made it a local val

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/utils/SortingSimpleFeatureIteratorTest.scala
elahrvivaz: they are cleaned up on close - we could check the number of files in /tmp, but there's no guarantee that some other process isn't creating/deleting tmp files, so the counts would not be accurate.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/ScalaSimpleFeature.scala
jnh5y: Ah, ok.  My coffee really isn't working today (or this is slightly subtle).  

I have reached your same level of understanding.  I think this is correct.  Also, we are talking about a somewhat crazy case where multiple threads are reading userData at the same time... 

So yeah, we'd have to be 'wrong' and then we'd have to be 'unlucky'. ;)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
jnh5y: Yeah, that sounds like 1) a lot of work, and 2) a somewhat good idea if the Lambda DS idea 'takes off' and gets used elsewhere.

jnh5y:
Merged as 597b2191c6e07279692c8d96a2a4744de1218598

PR #2531, opened 2020-04-28
GEOMESA-2835 Cleanup test logging

elahrvivaz:
* Standardize log4j config across modules
* Bridge java.util.logging calls
* Upgrade postgresql version for GeoTools requirement


build/cqs.tsv
```diff
@@ -234,7 +234,7 @@ org.mortbay.jetty:jetty-util	6.1.26	compile
 org.objenesis:objenesis	1.0	compile
 org.parboiled:parboiled-core	1.1.7	compile
 org.parboiled:parboiled-scala_2.11	1.1.7	compile
-org.postgresql:postgresql	42.2.2	compile
+org.postgresql:postgresql	42.2.12	compile
```
elahrvivaz, line 237: Actually, we shouldn't need a CQ for a bug-fix bump.

jnh5y:
Heheh, if there any DRY option here?  I mean, standardization is good, but the next time we change a line a logging config file, we shouldn't start getting ~100 files out of sync.

elahrvivaz:
yeah... i could add a script to sync up the log4j files, or we could configure the test classpath to look in a non-standard, shared location for the log4j file. what do you think? could make that a follow-on ticket too.
before this they were each individual per module, but that just lead to lots of bad logging configs, so I'd rather not go back to that.

jnh5y:
Let's see... if we wanted to go the 'non-standard' route, we would find the log4j property and set it in the root pom, right?  And then we could point to one file in root directory?

I think I like that over the script option.  Thoughts?

elahrvivaz:
yes, that would probably work... there might be some trickiness around absolute paths. the other things i was thinking was to add a new 'test/resources' folder to the surefire classpath instead of configuring it via log4j sys prop.

jnh5y:
Merged as 7f35fd428477f80b15f40cf41661659f159b2248.

PR #2532, opened 2020-04-29
GEOMESA-2836 Fix CachedThreadPool jvm shutdown behavior

elahrvivaz:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa-accumulo
+++ b/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa-accumulo
@@ -44,6 +44,8 @@ if [[ -n "${GEOMESA_EXTRA_CLASSPATHS}" ]]; then
 fi
 
 ACCUMULO_CP="$(findJars $ACCUMULO_LIB true true)"
+# if there's a geomesa runtime jar in accumulo, exclude it from the classpath
+ACCUMULO_CP="${ACCUMULO_CP#geomesa-accumulo-distributed-runtime*jar}"
 
 # Get the hadoop jars, ignoring jars with names containing slf4j and test
 # Copied from accumulo classpath
--- a/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa-accumulo
+++ b/geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa-accumulo
@@ -45,7 +45,7 @@ fi
 
 ACCUMULO_CP="$(findJars $ACCUMULO_LIB true true)"
 # if there's a geomesa runtime jar in accumulo, exclude it from the classpath
-ACCUMULO_CP="${ACCUMULO_CP#geomesa-accumulo-distributed-runtime*jar}"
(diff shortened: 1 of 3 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/concurrent/CachedThreadPool.scala
```diff
@@ -86,10 +88,19 @@ class CachedThreadPool(maxThreads: Int) extends AbstractExecutorService {
     lock.lock()
     try {
       if (available > 0) {
+        // note that we could fairly easily create a global thread limit by backing this with a different pool
+        try { CachedThreadPool.pool.execute(task) } catch {
+          case e: RejectedExecutionException =>
+            // we still need to execute the task to fulfill the executor API
```
jnh5y, line 94: Does the ExitingExecutor satisfy this part of the API contract?

jnh5y:
Looks like GitHub can't compile this one quite yet.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/concurrent/CachedThreadPool.scala
elahrvivaz: I think so - if a pool is 'shutdown' then it will reject any incoming execution requests, but it still has to run any tasks that were already submitted. if it's 'shutdownNow' then it can reject already submitted tasks. The issue we hit is that the static backing executor service can be shutdown by the jvm shutdown hook, which then causes us to not execute already submitted tasks.

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa-accumulo
```diff
@@ -44,6 +44,8 @@ if [[ -n "${GEOMESA_EXTRA_CLASSPATHS}" ]]; then
 fi
 
 ACCUMULO_CP="$(findJars $ACCUMULO_LIB true true)"
+# if there's a geomesa runtime jar in accumulo, exclude it from the classpath
+ACCUMULO_CP="${ACCUMULO_CP#geomesa-accumulo-distributed-runtime*jar}"
```
jnh5y, line 48: Wow.  Next, these scripts are gonna start holding people's hands...

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/concurrent/CachedThreadPool.scala
jnh5y: Ok, I think I see how the layers of this are fitting together.  The ExitingExecutor has a System shutdown hook registered.  The thing we build on top of that needs to be able to run one last task...  but the Executor is gonna already due to the System shutdown.  

Nice.

geomesa-accumulo/geomesa-accumulo-tools/bin/geomesa-accumulo
elahrvivaz: oops, this shouldn't be in here... it doesn't work

jnh5y:
Merged as 5663bf1ed8c5011b002fc0c063938dc372fd6a3e

PR #2533, opened 2020-04-29
GEOMESA-2837 Fix HBase yield partial results queries

elahrvivaz:

### DIFF
--- a/geomesa-hbase/geomesa-hbase-rpc/src/main/scala/org/locationtech/geomesa/hbase/rpc/coprocessor/GeoMesaHBaseCallBack.scala
+++ b/geomesa-hbase/geomesa-hbase-rpc/src/main/scala/org/locationtech/geomesa/hbase/rpc/coprocessor/GeoMesaHBaseCallBack.scala
@@ -8,16 +8,65 @@
 
 package org.locationtech.geomesa.hbase.rpc.coprocessor
 
-import java.util.concurrent.LinkedBlockingQueue
+import java.io.InterruptedIOException
+import java.util.Base64
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{CancellationException, LinkedBlockingQueue}
 
-import com.google.protobuf.ByteString
+import com.google.protobuf.{ByteString, RpcCallback}
 import com.typesafe.scalalogging.LazyLogging
-import org.apache.hadoop.hbase.client.coprocessor.Batch.Callback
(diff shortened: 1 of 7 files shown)


geomesa-hbase/geomesa-hbase-rpc/src/main/scala/org/locationtech/geomesa/hbase/rpc/coprocessor/GeoMesaHBaseCallBack.scala
```diff
@@ -8,16 +8,65 @@
...
+          if (controller.failed()) {
+            logger.error(s"Controller failed with error:\n${controller.errorText()}")
+          }
+
+          callback.get()
+        }
+      }
+    }
+
+  override def hasNext: Boolean = !closed.get() && scan.getStartRow != null
```
jnh5y, line 64: *Blinks*  What's up with the 'scan.getStartRow != null' condition?!

elahrvivaz: I set it `null` down on line 89 as an end condition: https://github.com/locationtech/geomesa/pull/2533/files/ae9b5f3182d75fb7192250de8a0ce3a10951e3d6#diff-4e356118c06df279e1a802c46b56d5deR89
I am fairly sure that we don't ever set it null in a normal scan, as we use an empty byte array instead.

elahrvivaz: alternatively, to make it clearer, we could have a separate variable to track the same thing. i didn't go that route b/c the scan class didn't complain about it being null and it seemed redundant to have a separate var.

jnh5y: Ah, ok!  That makes sense.  Would it be worth adding a comment about the fact that we are using 'null' as a sentinel?  And otherwise, we expect an empty byte array?

elahrvivaz: added

jnh5y:
Merged as 1a1609ffc36f7f76f03d2889cb0146dd3e9086fd

PR #2535, opened 2020-05-01
GEOMESA-2841 ILIKE queries with invalid regexp throw tserver errors

jnh5y:

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
@@ -8,9 +8,10 @@
 
 package org.locationtech.geomesa.filter.visitor
 
-import java.util.regex.{Pattern, PatternSyntaxException}
 import java.util.{Collections, Date}
 
+import org.geotools.filter.LikeToRegexConverter
+import org.geotools.filter.text.ecql.ECQL
 import org.geotools.filter.visitor.{DuplicatingFilterVisitor, ExpressionTypeVisitor, IsStaticExpressionVisitor}
 import org.locationtech.geomesa.filter.{FilterHelper, GeometryProcessing}
 import org.locationtech.geomesa.utils.geotools.converters.FastConverter
@@ -318,12 +319,11 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
   }
 
   override def visit(filter: PropertyIsLike, extraData: Any): AnyRef = {
-    val pattern = filter.getLiteral
     try {
-      Pattern.compile(pattern)
+      new LikeToRegexConverter(filter).getPattern
     } catch {
-      case e: PatternSyntaxException =>
-        throw new IllegalArgumentException("The regex for the (i)like filter is invalid.", e)
+      case e: Exception =>
+        throw new IllegalArgumentException(s"The regex filter (${ECQL.toCQL(filter)}) for the (i)like filter is invalid.", e)
     }
     super.visit(filter, extraData)
   }
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
@@ -8,6 +8,7 @@
 
 package org.locationtech.geomesa.filter.visitor
 
+import java.util.regex.Pattern
 import java.util.{Collections, Date}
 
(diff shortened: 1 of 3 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -316,6 +317,17 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
     }
   }
 
+  override def visit(filter: PropertyIsLike, extraData: Any): AnyRef = {
+    val pattern = filter.getLiteral
+    try {
+      Pattern.compile(pattern)
+    } catch {
+      case e: PatternSyntaxException =>
+        throw new IllegalArgumentException("The regex for the (i)like filter is invalid.", e)
```
elahrvivaz, line 326: can we add the pattern into the exception message?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -316,6 +317,17 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
     }
   }
 
+  override def visit(filter: PropertyIsLike, extraData: Any): AnyRef = {
+    val pattern = filter.getLiteral
+    try {
+      Pattern.compile(pattern)
```
elahrvivaz, line 323: i think we might need to use the LikeToRegexConverter to handle the %'s and all that appropriately: https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/filter/LikeFilterImpl.java#L166
https://github.com/geotools/geotools/blob/e8cf256a9e7ad034863d08a2bc680371f2b774b9/modules/library/main/src/main/java/org/geotools/filter/LikeToRegexConverter.java#L30

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -316,6 +318,16 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
     }
   }
 
+  override def visit(filter: PropertyIsLike, extraData: Any): AnyRef = {
+    try {
+      new LikeToRegexConverter(filter).getPattern
+    } catch {
+      case e: Exception =>
+        throw new IllegalArgumentException(s"The regex filter (${ECQL.toCQL(filter)}) for the (i)like filter is invalid.", e)
```
jahhulbert, line 326: is there any possibility that ECQL.toCQL() might throw an exception here as well?

elahrvivaz: you should be able to use `filter.getLiteral` to return the raw string value, which would be safe

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -316,6 +318,16 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
     }
   }
 
+  override def visit(filter: PropertyIsLike, extraData: Any): AnyRef = {
+    try {
+      new LikeToRegexConverter(filter).getPattern
```
elahrvivaz, line 323: do you still need to pass it into `Pattern.compile`? looking at the LikeToRegexConverter I don't think it throws any validation errors, but I could be wrong.

jnh5y: Good question; doing this was sufficient to throw an exception in the provided unit test.  I think that exception may have been in the GeoTools code which was trying to parse / extract the regex?

jnh5y: I'll toss in the call to Pattern.compile.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
jnh5y: Ok, switched to filter.getLiteral.

elahrvivaz:
merged as 68cedbf402f902d9086178d831baa604ab2d7648

PR #2537, opened 2020-05-04
GEOMESA-2842 Update references to OSGeo's 'webdav'

jnh5y:


elahrvivaz:
looks good - should we update those http links in the docs to be https? I assume either would work...

jnh5y:
Yeah, maybe updating to https is better?  Lemme do that real quick.

elahrvivaz:
merged as 8365e82039418876f6b9b3d8ecfed39d3da4b61c

PR #2538, opened 2020-05-04
GEOMESA-2843 - Optimize Arrow sorting in-memory

jnh5y:


geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -218,10 +222,10 @@ object DeltaWriter extends StrictLogging {
   }
 
   // note: ordering is flipped as higher values come off the queue first
-  private val queueOrdering = new Ordering[(AnyRef, Int, Int)] {
-    override def compare(x: (AnyRef, Int, Int), y: (AnyRef, Int, Int)): Int =
-      SimpleFeatureOrdering.nullCompare(y._1.asInstanceOf[Comparable[Any]], x._1)
-  }
+//  private val queueOrdering = new Ordering[(AnyRef, Int, Int)] {
```
jnh5y, line 225: Remove comments.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -503,17 +494,13 @@ object DeltaWriter extends StrictLogging {
...
-          val (_, batch, i) = queue.dequeue()
-          val (vector, sort, transfers, mappings) = toMerge(batch)
-          transfers.foreach(_.apply(i, resultIndex))
+          val next = queue.remove()
+          if (next.transfer(resultIndex)) {
+            queue.add(next)
+          }
           result.underlying.setIndexDefined(resultIndex)
           resultIndex += 1
-          val nextBatchIndex = i + 1
```
jnh5y, line 511: @elahrvivaz can you briefly describe what's happening here with the change in logic?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -526,9 +513,13 @@ object DeltaWriter extends StrictLogging {
     }
 
     val merged: CloseableIterator[Array[Byte]] = new CloseableIterator[Array[Byte]] {
+
+      // note: we synchronize the hasNext and close methods so that query timeouts don't asynchronously
```
jnh5y, line 517: Nice!

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -564,44 +555,54 @@ object DeltaWriter extends StrictLogging {
...
-      builder.result
     }
 
     // final results
     val results = createNewVectors
 
     // re-used queue, gets emptied after each dictionary field
     // [(dictionary value, batch index, index of value in the batch)]
-    val queue = scala.collection.mutable.PriorityQueue.empty[(AnyRef, Int, Int)](queueOrdering)
+    // TODO ordering reverse?
```
jnh5y, line 589: Should we do the TODO or create a ticket?  (Or remove the comment?)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -616,47 +617,59 @@ object DeltaWriter extends StrictLogging {
...
+          offsets(j) += v.getValueCount // note: side-effect in map - update our offsets for the next batch
+          v.vector.makeTransferPair(dictionaries(j).vector)
         }
-        offset
+//        (vectors, transfers)
+
+        DictionaryMerger(vectors, transfers, off, null, -1) // we don't care about the batch number here
       }
 
+      // set the count for each batch so we can offset mappings later
```
jnh5y, line 634: More comment code for one of us to get.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -616,47 +617,59 @@ object DeltaWriter extends StrictLogging {
...
-          transfers(i).copyValueSafe(j, count)
-          mappings(i).put(offsets(batch)(i) + j, count)
-          val jpp = j + 1
-          if (jpp < vectors(i).getValueCount) {
-            queue += ((vectors(i).apply(jpp), batch, jpp))
+        while (!queue.isEmpty) {
+//          val (_, batch, j) = queue.dequeue()
+          val merger = queue.remove()
+//          val (vectors, transfers) = toMerge(batch)
+//          transfers(i).copyValueSafe(j, count)
```
jnh5y, line 666: Comments to remove.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -564,44 +555,54 @@ object DeltaWriter extends StrictLogging {
...
     // [(dictionary value, batch index, index of value in the batch)]
-    val queue = scala.collection.mutable.PriorityQueue.empty[(AnyRef, Int, Int)](queueOrdering)
+    // TODO ordering reverse?
+    val queue = new PriorityQueue[DictionaryMerger](Ordering.ordered[DictionaryMerger])
 
     // merge each threaded delta vector into a single dictionary for that thread
     // Array[(dictionary vector, transfers to result, batch delta mappings)]
-    val allMerges: Array[(Array[ArrowAttributeReader], Array[TransferPair], Array[HashBiMap[Integer, Integer]])] = deltas.map { deltas =>
+//    val allMerges: Array[(Array[ArrowAttributeReader], Array[TransferPair], Array[HashBiMap[Integer, Integer]])] = deltas.map { deltas =>
+    val allMerges: Array[DictionaryMerger] = deltas.mapWithIndex { case (deltas, batch) =>
```
jnh5y, line 595: I know I don't like zipWithIndex since it allocates a bunch of extra tuples, etc.

I imagine there aren't too many batches here, so it isn't an 'inner loop' situation.  That said, would it be better to have a counter and increment?

Also, this anonymous function is huge.  Is there anyway to break it up sensibly?  Or should it be a do-all function in order to optimize things?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: sure - this may not be 100% accurate as I'm going from memory. previously there was an array of attribute readers, one for each batch. the queue contained `(current value, batch number, current reader index)`. We'd pop the top value off the queue, then push the next value read from the array (based on the batch number), and the next reader index. instead, i've moved the state into sortable classes, which will hold the reader, the current reader value, and the reader index, and sort based on the current reader value. we pop one off the queue, then we 'increment' it (read the next value from that batch and increment internal counters), then put it back on the queue with its new value. so we create a lot less classes, and we don't have to track the relationship between the queue values and the batch readers b/c they're the same thing now.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: we can remove it, sorting seems to be correct.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: `mapWithIndex` is syntactic sugar we wrote that increments a var: https://github.com/locationtech/geomesa/blob/master/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/conversions/ScalaImplicits.scala#L33-L40
but yeah, this method is large...

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
```diff
@@ -56,7 +56,12 @@ class SimpleFeatureVector private [arrow] (
     */
   def clear(): Unit = underlying.setValueCount(0)
 
-  override def close(): Unit = CloseWithLogging.raise(Seq(underlying) ++ allocator)
+  override def close(): Unit = {
+    CloseWithLogging.raise(Seq(underlying))
+    this.synchronized {
```
elahrvivaz, line 61: i think we don't need this any more, right? if so, we should try to make sure it gets called even if the first call fails

jnh5y: Yeah, I think it can go.  I'll do that.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
jnh5y: Done.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -411,24 +417,21 @@ object DeltaWriter extends StrictLogging {
...
+    // we do a merge sort of each batch
+    // queue sorted by current value in each batch
+    val queue = {
+      val ordering = if (reverse) { BatchMergerOrdering.reverse } else { BatchMergerOrdering }
+      new PriorityQueue[BatchMerger[Any]](ordering)
+    }
+
+    // track our open vectors to close later
+    val cleanup = ArrayBuffer.empty[SimpleFeatureVector]
+    cleanup.sizeHint(threadedBatches.foldLeft(0)((sum, a) => sum + a.length))
```
jnh5y, line 434: threadedBatches.map(_.length).sum too pedestrian?:)

elahrvivaz: no point in creating an intermediate collection :)

jnh5y:
Merged as f13e64748fa54250ec0d47007855b7e230e336dd

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
jnh5y: Fair enough.  

At the rate we're going, we're gonna start writing this stuff in JVM bytecode directly.;)

PR #2539, opened 2020-05-04
GEOMESA-2845 Performance improvements for evaluating transforms

elahrvivaz:
* Replace unsupported TransformProcess.toDefinitions
* Optimize for simple relational projections


geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/json/JsonPathPropertyAccessor.scala
```diff
@@ -97,66 +131,23 @@ object JsonPathPropertyAccessor {
...
+        target: Class[_],
+        hints: Hints): PropertyAccessor = {
       if (xpath == null || !xpath.startsWith("$.")) {
         null
       } else if (classOf[SimpleFeature].isAssignableFrom(typ)) {
-        JsonPathFeatureAccessor
+        JsonPathPropertyAccessor
       } else if (classOf[SimpleFeatureType].isAssignableFrom(typ)) {
-        JsonPathTypeAccessor
+        JsonPathPropertyAccessor
```
jnh5y, line 144: Ok, just to double-check, both of these should be JsonPathPropertyAccessor, right?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
```diff
@@ -0,0 +1,248 @@
...
+     * @param transform transform expression
+     * @return
+     */
+    private def definition(sft: SimpleFeatureType, transform: String): Transform = {
+      val equals = transform.indexOf('=')
+      if (equals == -1) {
+        val name = transform.trim()
+        val i = sft.indexOf(name)
+        if (i == -1) {
+          attributeExpression(sft, name)
```
jnh5y, line 128: What's an example of a string that hits this branch?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
```diff
@@ -0,0 +1,248 @@
...
+            ExpressionTransform(name, f.getFunctionName.getReturn.getType, f)
+
+          case f: MathExpressionImpl =>
+            ExpressionTransform(name, classOf[java.lang.Double], f) // math ops always return doubles?
+
+          case p: PropertyName =>
+            val orig = p.getPropertyName
+            val i = sft.indexOf(orig)
+            if (i == -1) {
+              attributeExpression(sft, orig).copy(name = name)
```
jnh5y, line 149: Same here, What's an example of a string that hits this branch?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
```diff
@@ -0,0 +1,248 @@
...
+            }
+        }
+
+        schema.add(descriptor)
+      }
+
+      var i = 0
+      while (geom == null && i < schema.size) {
+        schema.get(i) match {
+          case d: ImmutableGeometryDescriptor => geom = d
```
jnh5y, line 237: So, this method makes it clear that currently Transforms cannot have multiple geometries and let the 'user' pick the default geometry. 

There's a chance we'll want to do that one day.;)  Seems like the current mappings (Transforms) could have an extra bit of side car meta-data around.  Some of that information could also explain how to set userData (such as visibilities).

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
```diff
@@ -0,0 +1,248 @@
...
+          case _ => // no-op
+        }
+        i += 1
+      }
+
+      // TODO reconsider default field user data?
```
jnh5y, line 243: Is this a new 'TODO'?  If so, is it worth a ticket?

jnh5y: Nevermind, it is not:).

geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/json/JsonPathPropertyAccessor.scala
elahrvivaz: yeah - i just updated this to consolidate the if checks into one statement to make it clearer. As more background, we originally had 2 different accessors, one for simple feature types and one for simple features. the changes to the transforms caused some changes in how these get called in the iterators - specifically that originally the simple feature type one was never called in the iterators but now it does get called for handling json path attributes. that lead to some exceptions in the geotools code that caches and re-tries the last accessor that was used. the fix for that was to make more robust canHandle/get methods for both accessors, at which point they were basically the same class so i consolidated them.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
elahrvivaz: the main use case is json-path selectors like `$.foo.bar` - other cases would probably be xpath, or anything that uses a non-standard PropertyAccessor.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
elahrvivaz: same thing as before, but in this case the attribute is renamed. previous statement would be `"$.foo.bar"`, this one would be `"baz=$.foo.bar"`

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
elahrvivaz: yeah there's no way to pick the default geometry - however, the default geometry doesn't mean much in this case does it? you could always style on a different geometry if you wanted.

elahrvivaz: this seems like something that should be standardized at the geotools level.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Transform.scala
elahrvivaz: yeah, it seems like maybe we shouldn't be copying over all the user data, but things break hard when i removed it so...

jnh5y:
Merged as 4dfc7001d34d4f5cbf7ebbb9903f2c00e2fc6473

PR #2541, opened 2020-05-07
GEOMESA-2846 Fix ThreadedConsumer topic logging

elahrvivaz:

### DIFF
--- a/geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
+++ b/geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
@@ -31,7 +31,7 @@ abstract class ThreadedConsumer(
 
   protected def consume(record: ConsumerRecord[Array[Byte], Array[Byte]]): Unit
 
-  private lazy val topics = consumers.flatMap(_.subscription().asScala).distinct
+  private val topics = scala.collection.mutable.Set.empty[String]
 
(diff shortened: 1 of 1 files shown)


geomesa-kafka/geomesa-kafka-utils/src/main/scala/org/locationtech/geomesa/kafka/consumer/ThreadedConsumer.scala
```diff
@@ -53,6 +53,7 @@ abstract class ThreadedConsumer(
     val format = if (consumers.lengthCompare(10) > 0) { "%02d" } else { "%d" }
     var i = 0
     consumers.foreach { c =>
+      c.subscription().asScala.foreach(topics.add)
```
jnh5y, line 56: What was happening here before?  I don't have the context to know what's up here:).

elahrvivaz: the topics was a lazy val that was evaluated here, but since consumers are aggressively single threaded, it was throwing ConcurrentAccess exceptions once the consumers were started up in separate threads.

elahrvivaz: so we add the topics before starting the separate threads to ensure that only a single thread accesses the consumer at once

jnh5y:
Merged as 93f2797663582fc0bdfae930ddd87799fe30f6a8

PR #2544, opened 2020-05-14
GEOMESA-2848 Improve query timeout behavior

elahrvivaz:
* Move query termination to low-level scans
  * Prevents errors in 'reduce' logic being interrupted
* Throw all exceptions through 'next' method, to prevent geoserver from swallowing them

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/BatchMultiScanner.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/BatchMultiScanner.scala
@@ -17,6 +17,8 @@ import org.apache.accumulo.core.client.{Connector, ScannerBase}
 import org.apache.accumulo.core.data.{Key, Value}
 import org.apache.accumulo.core.security.Authorizations
 import org.locationtech.geomesa.accumulo.data.AccumuloQueryPlan.{BatchScanPlan, JoinFunction}
+import org.locationtech.geomesa.index.utils.ThreadManagement.Timeout
+import org.locationtech.geomesa.utils.collection.CloseableIterator
 
 import scala.collection.JavaConversions._
 
@@ -31,14 +33,16 @@ import scala.collection.JavaConversions._
   * @param numThreads threads
   * @param batchSize batch size
   */
-class BatchMultiScanner(connector: Connector,
-                        in: ScannerBase,
-                        join: BatchScanPlan,
-                        joinFunction: JoinFunction,
-                        auths: Authorizations,
-                        numThreads: Int = 12,
-                        batchSize: Int = 32768)
-  extends Iterable[java.util.Map.Entry[Key, Value]] with AutoCloseable with LazyLogging {
+class BatchMultiScanner(
+    connector: Connector,
+    in: ScannerBase,
+    join: BatchScanPlan,
+    joinFunction: JoinFunction,
+    auths: Authorizations,
+    timeout: Option[Timeout],
+    numThreads: Int = 12,
+    batchSize: Int = 32768
+  ) extends CloseableIterator[java.util.Map.Entry[Key, Value]] with LazyLogging {
 
   require(batchSize > 0, f"Illegal batchSize ($batchSize%d). Value must be > 0")
   require(numThreads > 0, f"Illegal numThreads ($numThreads%d). Value must be > 0")
@@ -51,6 +55,30 @@ class BatchMultiScanner(connector: Connector,
 
   private val inDone  = new AtomicBoolean(false)
   private val outDone = new AtomicBoolean(false)
+
+  private var prefetch: Entry[Key, Value] = _
+
+  private def prefetchIfNull(): Unit = {
+    // loop while we might have another and we haven't set prefetch
+    while (prefetch == null && (!outDone.get || outQ.size > 0)) {
+      prefetch = outQ.poll(5, TimeUnit.MILLISECONDS)
+    }
+  }
+
+  // must attempt a prefetch since we don't know whether or not the outQ
+  // will actually be filled with an item (filters may not match and the
+  // in scanner may never return a range)
+  override def hasNext(): Boolean = {
+    prefetchIfNull()
+    prefetch != null
+  }
+
+  override def next(): Entry[Key, Value] = {
+    prefetchIfNull()
+    val ret = prefetch
+    prefetch = null
+    ret
+  }
 
   executor.submit(new Runnable {
     override def run(): Unit = {
@@ -73,7 +101,7 @@ class BatchMultiScanner(connector: Connector,
             inQ.drainTo(entries)
             val task = executor.submit(new Runnable {
               override def run(): Unit = {
-                val iterator = join.copy(ranges = entries.map(joinFunction)).scan(connector, auths)
+                val iterator = join.copy(ranges = entries.map(joinFunction)).scan(connector, auths, timeout)
                 try {
                   iterator.foreach(outQ.put)
                 } finally {
@@ -99,32 +127,5 @@ class BatchMultiScanner(connector: Connector,
       executor.shutdownNow()
     }
     in.close()
-  }
-
-  override def iterator: Iterator[Entry[Key, Value]] = new Iterator[Entry[Key, Value]] {
-
-    private var prefetch: Entry[Key, Value] = _
-
-    private def prefetchIfNull(): Unit = {
-      // loop while we might have another and we haven't set prefetch
-      while (prefetch == null && (!outDone.get || outQ.size > 0)) {
-        prefetch = outQ.poll(5, TimeUnit.MILLISECONDS)
-      }
-    }
-
-    // must attempt a prefetch since we don't know whether or not the outQ
-    // will actually be filled with an item (filters may not match and the
-    // in scanner may never return a range)
-    override def hasNext(): Boolean = {
-      prefetchIfNull()
-      prefetch != null
-    }
-
-    override def next(): Entry[Key, Value] = {
-      prefetchIfNull()
-      val ret = prefetch
-      prefetch = null
-      ret
-    }
   }
 }
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
@@ -479,14 +479,14 @@ class AccumuloDataStoreQueryTest extends Specification with TestWithMultipleSfts
     }
 
     "kill queries after a configurable timeout" in {
-      import scala.concurrent.duration._
-
+      skipped("relies on thread.sleep timing")
       val params = dsParams ++ Map(AccumuloDataStoreParams.QueryTimeoutParam.getName -> "1s")
 
       val dsWithTimeout = DataStoreFinder.getDataStore(params).asInstanceOf[AccumuloDataStore]
       val reader = dsWithTimeout.getFeatureReader(new Query(defaultSft.getTypeName, Filter.INCLUDE), Transaction.AUTO_COMMIT)
-      reader.isClosed must beFalse
-      eventually(20, 200.millis)(reader.isClosed must beTrue)
+      reader.hasNext() must beTrue
+      Thread.sleep(5000) // TODO this is error prone...
+      reader.close() must throwAn[RuntimeException]("Scan terminated due to timeout of 1000ms")
     }
 
     "block full table scans" in {
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
@@ -8,8 +8,6 @@
 
 package org.locationtech.geomesa.arrow.io
 
-import java.util.concurrent.atomic.AtomicBoolean
-
 import org.locationtech.geomesa.arrow.io.records.{RecordBatchLoader, RecordBatchUnloader}
 import org.locationtech.geomesa.arrow.vector.SimpleFeatureVector.SimpleFeatureEncoding
 import org.locationtech.geomesa.arrow.vector._
@@ -84,8 +82,6 @@ object BatchWriter {
 
     private var batch: Array[Byte] = _
 
-    private val closed = new AtomicBoolean(false)
-
     // gets the attribute we're sorting by from the i-th feature in the vector
     val getSortAttribute: (SimpleFeatureVector, Int) => AnyRef = {
       val sortByIndex = sft.indexOf(sortBy)
@@ -98,10 +94,9 @@ object BatchWriter {
     }
 
     // this is lazy to allow the query plan to be instantiated without pulling back all the batches first
-    // we also allow for early termination by checking the `closed` flag
     private lazy val inputs: Array[(SimpleFeatureVector, (Int, Int) => Unit)] = {
       val builder = Array.newBuilder[(SimpleFeatureVector, (Int, Int) => Unit)]
-      while (!closed.get && batches.hasNext) {
+      while (batches.hasNext) {
         val vector = SimpleFeatureVector.create(sft, dictionaries, encoding)
         RecordBatchLoader.load(vector.underlying, batches.next)
 
@@ -186,7 +181,6 @@ object BatchWriter {
     }
 
     override def close(): Unit = {
-      closed.set(true)
       CloseWithLogging(result, batches)
       inputs.foreach { case (vector, _) => CloseWithLogging(vector) }
     }
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
@@ -181,8 +181,9 @@ object BatchWriter {
     }
 
     override def close(): Unit = {
-      CloseWithLogging(result, batches)
+      CloseWithLogging(batches)
       inputs.foreach { case (vector, _) => CloseWithLogging(vector) }
+      CloseWithLogging(result)
     }
   }
 }
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
@@ -12,7 +12,6 @@ import java.io.{ByteArrayOutputStream, Closeable, OutputStream}
 import java.nio.channels.Channels
 import java.util.PriorityQueue
 import java.util.concurrent.ThreadLocalRandom
-import java.util.concurrent.atomic.AtomicBoolean
 
 import com.google.common.collect.HashBiMap
 import com.typesafe.scalalogging.StrictLogging
@@ -39,6 +38,7 @@ import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import scala.annotation.tailrec
 import scala.collection.mutable.ArrayBuffer
 import scala.math.Ordering
+import scala.util.control.NonFatal
 
 /**
   * Builds up dictionaries and write record batches. Dictionaries are encoded as deltas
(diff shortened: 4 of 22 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
```diff
@@ -479,14 +479,14 @@ class AccumuloDataStoreQueryTest extends Specification with TestWithMultipleSfts
     }
 
     "kill queries after a configurable timeout" in {
-      import scala.concurrent.duration._
-
+      skipped("relies on thread.sleep timing")
```
jnh5y, line 482: Hmm....  Anything we can do for this test?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/util/BatchMultiScanner.scala
```diff
@@ -31,14 +33,16 @@ import scala.collection.JavaConversions._
...
+class BatchMultiScanner(
+    connector: Connector,
+    in: ScannerBase,
+    join: BatchScanPlan,
+    joinFunction: JoinFunction,
+    auths: Authorizations,
+    timeout: Option[Timeout],
+    numThreads: Int = 12,
+    batchSize: Int = 32768
+  ) extends CloseableIterator[java.util.Map.Entry[Key, Value]] with LazyLogging {
```
jnh5y, line 45: Solid improvement!

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
```diff
@@ -98,10 +94,9 @@ object BatchWriter {
     }
 
     // this is lazy to allow the query plan to be instantiated without pulling back all the batches first
-    // we also allow for early termination by checking the `closed` flag
     private lazy val inputs: Array[(SimpleFeatureVector, (Int, Int) => Unit)] = {
       val builder = Array.newBuilder[(SimpleFeatureVector, (Int, Int) => Unit)]
-      while (!closed.get && batches.hasNext) {
```
jnh5y, line 104: Ok, why can we get away with this at this point?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -320,9 +320,9 @@ object DeltaWriter extends StrictLogging {
       private var mappings: Map[String, java.util.Map[Integer, Integer]] = _
       private var count = 0 // records read in current batch
 
-      override def hasNext: Boolean = synchronized { count < toLoad.reader.getValueCount || loadNextBatch() }
+      override def hasNext: Boolean = count < toLoad.reader.getValueCount || loadNextBatch()
```
jnh5y, line 323: Overall, can you remind me why we think that the removal of the 'synchronized' calls is ok now?

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraQueryPlan.scala
```diff
@@ -75,5 +76,5 @@ case class StatementPlan(
   ) extends CassandraQueryPlan {
 
   override def scan(ds: CassandraDataStore): CloseableIterator[Row] =
-    CassandraBatchScan(ds.session, ranges, numThreads)
+    CassandraBatchScan(this, ds.session, ranges, numThreads, ds.config.queries.timeout.map(Timeout.apply))
```
jnh5y, line 79: Is there any chance that the C* unit tests would let us exercise timeouts?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -255,9 +255,8 @@ class HBaseIndexAdapter(ds: HBaseDataStore) extends IndexAdapter[HBaseDataStore]
         }
         (cqlFilter ++ indexFilter).sortBy(_._1).map(_._2)
       }
-      lazy val coprocessorOptions: Map[String, String] =
-        Map[String, String](GeoMesaCoprocessor.YieldOpt -> ds.config.coprocessors.yieldPartialResults.toString) ++
-            strategy.index.ds.config.queries.timeout.map(GeoMesaCoprocessor.timeout)
```
jnh5y, line 260: Doesn't the timeout still need to go to the RS?  Maybe it goes somewhere else...

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/utils/CoprocessorBatchScan.scala
```diff
@@ -13,31 +13,35 @@ import java.util.concurrent._
...
 private class CoprocessorBatchScan(
     connection: Connection,
     table: TableName,
     ranges: Seq[Scan],
     options: Map[String, String],
-    threads: Int,
     rpcThreads: Int,
     buffer: Int
-  ) extends AbstractBatchScan[Scan, Array[Byte]](ranges, threads, buffer, CoprocessorBatchScan.Sentinel) {
+  ) extends AbstractBatchScan[Scan, Array[Byte]](ranges, rpcThreads * 2, buffer, CoprocessorBatchScan.Sentinel) {
```
jnh5y, line 31: For this commit / PR, should this be 'threads'?

geomesa-hbase/geomesa-hbase-rpc/src/main/scala/org/locationtech/geomesa/hbase/rpc/coprocessor/GeoMesaHBaseCallBack.scala
```diff
@@ -28,42 +28,37 @@ import org.locationtech.geomesa.utils.index.ByteArrays
...
-  ) extends Callback[GeoMesaCoprocessorResponse] with Iterator[Scan] with LazyLogging {
+class GeoMesaHBaseCallBack(scan: Scan, options: Map[String, String], result: LinkedBlockingQueue[ByteString])
+    extends Callback[GeoMesaCoprocessorResponse] with Iterator[Scan] with LazyLogging {
 
   private var updated: Array[Byte] = _ // tracks our status per invocation to detect multiple region updates
 
   val callable: Call[GeoMesaCoprocessorService, GeoMesaCoprocessorResponse] =
     new Call[GeoMesaCoprocessorService, GeoMesaCoprocessorResponse]() {
       override def call(instance: GeoMesaCoprocessorService): GeoMesaCoprocessorResponse = {
-        if (closed.get) { null } else {
```
jnh5y, line 45: *blinks*  I'm trying to remember why we had the 'closed' variable in the first place. 

You've thought through removing it obviously... right;)?  This was the crux of the matter, right?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/AbstractBatchScan.scala
```diff
@@ -156,13 +155,14 @@ abstract class AbstractBatchScan[T, R <: AnyRef](ranges: Seq[T], threads: Int, b
     */
   private class Terminator extends Runnable {
 
-    private val done = new AtomicBoolean(false)
+    @volatile
+    private var done = false
```
jnh5y, line 159: OK, why the switch from AtomicBoolean to volatile boolean?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
```diff
@@ -11,15 +11,23 @@ package org.locationtech.geomesa.index.utils
...
+import scala.concurrent.duration.Duration
 import scala.util.control.NonFatal
 
 /**
  * Singleton for registering and managing running queries.
  */
-object ThreadManagement extends LazyLogging {
+object ThreadManagement {
+
+  private val logger = Logger(LoggerFactory.getLogger(ThreadManagement.getClass.getName.replace("$", "")))
```
jnh5y, line 30: Does this let us use 'org.locationtech.geomesa.index.utils.ThreadManagement' in log4j properties files?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
```diff
@@ -28,28 +36,120 @@ object ThreadManagement extends LazyLogging {
...
+    // used for log messages
+    protected def typeName: String
+    protected def filter: Option[Filter]
+
+    // we can use a volatile var since we only update the value with a single thread
+    @volatile
+    private var terminated = timeout.absolute <= System.currentTimeMillis()
+
+    private val iter = ExceptionalIterator(if (terminated) { Iterator.empty } else { underlying.iterator })
+    private val cancel = if (terminated) { None } else { Some(ThreadManagement.register(this)) }
```
jnh5y, line 79: Ok, just to make sure I'm understanding... this is called as part of the constructor / initialization for the ManagedScan.  So if the Scan is initialized *late*, then it is already terminated, etc, etc, etc...

jnh5y, line 79: Ok, separate thread/comment...

Is there any daylight between System.currentTimeMillis between these two calls?  Is that a corner case even worth thinking through?  (E.g., could one of them see 'terminated' true and the other not?)

It'd look nutty, but would something like....

val (iter, cancel) = { 
    if (terminated) { (Iterator.empty, None) } 
    else { ... }
} 

make sense?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
```diff
@@ -28,28 +36,120 @@ object ThreadManagement extends LazyLogging {
...
       }
     }
   }
+
+  /**
+   * Low level scanner that can be closed to terminate a scan
+   *
+   * @tparam T type
+   */
+  trait LowLevelScanner[T] extends Closeable {
```
jnh5y, line 130: Should it be AutoCloseable?  Meh, I lose track of which one we like.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/iterators/ExceptionalIterator.scala
```diff
@@ -0,0 +1,56 @@
...
+      throw _suppressed
+    } else {
+      delegate.next()
+    }
+  }
+
+  def suppressed: Option[Throwable] = Option(_suppressed)
+}
+
+object ExceptionalIterator {
```
jnh5y, line 45: No unit tests for the ExceptionalIterator:)?  

Don't do it; just messing.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreQueryTest.scala
elahrvivaz: nothing i can think of... we can't use the 'eventually' construct b/c we only throw exceptions on 'next' and that's not a idempotent call. or we could enable the test, but i feel like the CI build will fail randomly on it if we do.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
elahrvivaz: `batches.next` will throw an exception now once it's been terminated, so we don't need to check the closed flag (and closed won't get invoked by the thread management anyway)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: i believe that we only needed the synchronized b/c we were calling next/hasNext/close in different threads. now they will all be called by the same thread, so we shouldn't need to synchronize. the low-level scan will be closed by the thread management, but not this iterator.

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraQueryPlan.scala
elahrvivaz: i think we'd have the same issue in any of the data stores, in that it's requires timing and any unit tests that rely on timing tend to fail occasionally.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
elahrvivaz: it does, but it only gets added in the CoprocessorBatchScan now, since the timeout is not calculated as an absolute timestamp until CoprocessorQueryPlan.scan is invoked.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/utils/CoprocessorBatchScan.scala
elahrvivaz: i think the previous behavior was ranges.length, should i revert it? i didn't bother since i added the other pr to update it.

geomesa-hbase/geomesa-hbase-rpc/src/main/scala/org/locationtech/geomesa/hbase/rpc/coprocessor/GeoMesaHBaseCallBack.scala
elahrvivaz: i think it was from an earlier attempt to get the queries to terminate correctly. i _believe_ it's not needed, but we should verify that it's behaving as expected.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/AbstractBatchScan.scala
elahrvivaz: i remembered that if you just use .get and .set, there's no difference between the 2, other than the overhead of the AtomicBoolean class...
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/atomic/AtomicBoolean.java#L63

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
elahrvivaz: right. spark actually has a little utility that does that automatically for objects :heart:

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
elahrvivaz: correct

elahrvivaz: terminated it set and then compared twice, but it can't change between the 2 comparisons b/c it's a var not a def. the call to `ThreadManagement.register(this)` could potentially change the state of terminated, but it's the last thing we do so it shouldn't affect anything in the initialization. and in the register function, we do a `math.max(1, System.currentTimeMillis - timeout)` to ensure that if there is a gap in the sys time we still execute the termination call.

elahrvivaz: i did initially have it in a single if statement, but it seems easier to grok the code this way...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
elahrvivaz: Closeable exends AutoCloseable... the only difference is the type of exception that gets thrown so it doesn't matter to us. I believe there is some code in geotools that doesn't work right with AutoCloseable, so we always use Closeable.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
jnh5y: What happens if someone gets in there and calls close separately?  

Does that even matter?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
jnh5y: Cool.  So someone would have to go to trouble to pass around GeoMesaReaders to multiple threads in their application and start calling methods like hasNext, next, and close in concurrent manners...  

If someone does that in their application, we cannot help them.  Is that the story?

geomesa-cassandra/geomesa-cassandra-datastore/src/main/scala/org/locationtech/geomesa/cassandra/data/CassandraQueryPlan.scala
jnh5y: Grrr...  "I wonder how it'd be to 'fix' the unit test framework....", he said trailing off in a manner to indicate that this was a bad idea....

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
jnh5y: +1

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
elahrvivaz: hmm, yeah that would not work well since it would block on the lazy val again... maybe i should add the closed check back in

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: right

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/AbstractBatchScan.scala
jnh5y: Good call...  I kinda want about the VarHandle, etc, etc...  I have more questions about how AtomicObjects work now...  but that's for another time.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
jnh5y: Really?!  Neat.  

I mean, could one gin up a ObjectLazyLogging trait?:)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/BatchWriter.scala
elahrvivaz: actually nvm, the close method closes the batches so it wouldn't block. let me go through and double check that in the different places we had synchronization.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/ThreadManagement.scala
jnh5y: I can't read... the var is calculated once.  Mea culpea.  

I'm just searching for any corner case I can make up.  I'll shut up and get back to testing;).

jnh5y:
Merged 746ba8296b30e7a0851715d763a4ac0fcb270055

PR #2546, opened 2020-05-15
GEOMESA-2850 HBase - clean up data store params

elahrvivaz:
* Default arrow coprocessor scans to true
* Move coprocessor scan threads to sys prop
* Rename coprocessor url sys prop to match data store

### DIFF
--- a/docs/user/hbase/configuration.rst
+++ b/docs/user/hbase/configuration.rst
@@ -6,6 +6,12 @@ HBase Configuration
 This section details HBase specific configuration properties. For general properties,
 see :ref:`geomesa_site_xml`.
 
+geomesa.hbase.client.scanner.caching.size
++++++++++++++++++++++++++++++++++++++++++
+
+Set the number of rows that scanners will read ahead. If not set, the default caching will apply as configured in
+``hbase-site.xml``. Higher caching values will enable faster scanners but will use more memory.
+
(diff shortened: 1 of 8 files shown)


docs/user/hbase/configuration.rst
```diff
@@ -51,6 +69,18 @@ Disable remote filtering. Remote filtering and coprocessors speed up queries, ho
 of custom JARs in HBase. Since this is not always possible, they can be disabled by setting this to ``false``.
 This property will be overridden by the data store configuration parameter, if both are specified.
 
+geomesa.hbase.scan.buffer
++++++++++++++++++++++++++
+
+Specify the maximum number of results to pre-buffer in local memory when executing a scan, if the client is not
+consuming the results as fast as they are being returned.
+
+geomesa.hbase.table.availability.timeout
```
jnh5y, line 78: Were these two configs just missing in the docs?  If so, thanks for adding them!

jnh5y:
Merged as 26f01455ccc33db89bb66be08a9bdb60f79d5e1b.

docs/user/hbase/configuration.rst
elahrvivaz: yeah i went through and made sure we had an entry for all the sys props.

PR #2553, opened 2020-05-29
GEOMESA-2860 Ability to read older geotools hints with patched avro writers

elahrvivaz:


jnh5y:
The completionist in me wants to hold this PR until we backport a v5 writer to a version with GT <=20 so that we can attach a sample file.:)

Also, if we decide not to backport, we'll never need this complexity!

elahrvivaz:
yep, I'm fine with that

elahrvivaz:
I'm going to close this, since the consensus seems to be not to back-port any fixes, and since it is superseded by https://github.com/locationtech/geomesa/pull/2560

PR #2563, opened 2020-06-03
Initial work for 'Geometry' class support in Arrow

jnh5y:


geomesa-arrow/geomesa-arrow-gt/src/test/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileTest.scala
```diff
@@ -46,6 +48,16 @@ class SimpleFeatureArrowFileTest extends Specification {
...
   }
+  val points: Seq[Geometry] = Seq("POINT(0 0)", "POINT(1 1)", "POINT(2 2)", "POINT(3 3)", "POINT(4 4)").map(WKTUtils.read)
+  val polys = points.map(_.buffer(1))
+  val geoms = (points ++ polys).toArray
+  val geomFeatures = (0 until 10).map { i =>
+    val name = s"name${i % 2}"
+    val team = s"team$i"
+    val age = i % 5
+    val weight = Option(i % 3).filter(_ != 0).map(Int.box).orNull
+    ScalaSimpleFeature.create(lineSft, s"$i", name, team, age, weight, s"2017-02-03T00:0$i:01.000Z", geoms(i))
```
elahrvivaz, line 59: this should be geomSft right?

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
```diff
@@ -0,0 +1,109 @@
...
+  private VarBinaryVector vector;
+  private WKBWriter writer = new WKBWriter();
+  private WKBReader reader = new WKBReader();
+
+  public static FieldType createFieldType(Map<String, String> metadata) {
+    return new FieldType(true, ArrowType.Binary.INSTANCE, null, metadata);
+  }
+
+  public WKBGeometryVector(String name, BufferAllocator allocator, @Nullable Map<String, String> metadata) {
+    // JNH: How do I use the metadata?
```
elahrvivaz, line 42: i think as long as you pass it through like this then you don't need to do anything else with it

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
```diff
@@ -0,0 +1,109 @@
...
+  @Override
+  public Geometry get(int i) {
+    if (vector.isNull(i)) {
+      return null;
+    } else {
+      Geometry geometry = null;
+      try {
+         geometry = reader.read(vector.get(i));
+      } catch (ParseException exception) {
+        // JNH: Swallow exception?!
```
elahrvivaz, line 73: nooooooooo :scream:

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Catch-all for storing instances of Geometry as WKB
+ */
+public class WKBGeometryVector implements GeometryVector<Geometry, VarBinaryVector> {
+  private VarBinaryVector vector;
+  private WKBWriter writer = new WKBWriter();
+  private WKBReader reader = new WKBReader();
```
elahrvivaz, line 35: i guess we don't have the thread-safe pools on the classpath - but i think that should be ok as vectors aren't really thread-safe anyway.

geomesa-arrow/geomesa-arrow-jts/src/test/java/org/locationtech/geomesa/arrow/vector/GeometryVectorTest.java
```diff
@@ -542,6 +537,107 @@ public void testMultiPolygon() throws Exception {
...
+      Assert.assertEquals(geom1, wktWriter.write(doubles.get(0)));
+      Assert.assertEquals(geom2, wktWriter.write(doubles.get(1)));
+      Assert.assertEquals(geom3, wktWriter.write(doubles.get(3)));
+      Assert.assertNull(doubles.get(2));
+
+      // ensure field was created correctly up front
+      Assert.assertEquals(doubleField, doubles.getVector().getField());
+
+      // JNH:  I don't think there are fields?
+      //Assert.assertEquals(doubleField.getChildren(), PointVector.fields);  // FIX
```
elahrvivaz, line 570: yeah there should just be the one field

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: It seems like the only option here is to log it and move on.  

The API doesn't have this method throwing an exception.

geomesa-arrow/geomesa-arrow-jts/src/test/java/org/locationtech/geomesa/arrow/vector/GeometryVectorTest.java
jnh5y: Cool.  I'll double-check that.  I think I saw an empty array the one thing I ran this in the debugger.

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: Yeah, is there somewhere we could share the Reader/Writers?  (One per vector is kinda gross, right?)

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
elahrvivaz: ah, yeah java typed exceptions. you could throw a runtime exception though, i think that would probably be the best solution, then we can handle it higher up the chain

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
elahrvivaz: it's probably ok to have 1 per vector - usually the issue is you hit thread-safe problems if you do that. maybe we could make them lazy so that we only instantiate them when needed. i think in a lot of use cases we're just copying buffers around so not actually reading/writing through the WKB classes.

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: Ok, so re-throwing as a Runtime exception makes sense?

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: Since we are in Java land, would having a static and/or thread local thing going on here make sense?

elahrvivaz: you could just use 'java lazy', i.e. make it null to start and then check if it's null each time and instantiate it. i'd say we shouldn't use static or thread locals... i think it's fine to just leave it as-is, too.

geomesa-arrow/geomesa-arrow-gt/src/test/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileTest.scala
jnh5y: Yes.  Fixed

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: Done

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: Ok, great.

geomesa-arrow/geomesa-arrow-jts/src/test/java/org/locationtech/geomesa/arrow/vector/GeometryVectorTest.java
jnh5y: Yeah, the field is just one field and no children.  I've updated the test.

geomesa-arrow/geomesa-arrow-gt/src/test/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileTest.scala
```diff
@@ -216,6 +228,38 @@ class SimpleFeatureArrowFileTest extends Specification {
...
+              r.getX must beCloseTo(f.getX, delta=0.001)
+              r.getY must beCloseTo(f.getY, delta=0.001)
+            case (r: Polygon, f: Polygon) =>
+              // because of our limited precision in arrow queries, points don't exactly match up
+              r.getNumPoints mustEqual f.getNumPoints
+              foreach(0 until r.getNumPoints) { n =>
+                r.getExteriorRing.getCoordinateN(n).x must beCloseTo(r.getExteriorRing.getCoordinateN(n).x, delta=0.001)
+                r.getExteriorRing.getCoordinateN(n).y must beCloseTo(r.getExteriorRing.getCoordinateN(n).y, delta=0.001)
+              }
+            case (a, b) => println(s"a: $a b: $b")
```
elahrvivaz, line 255: should this be a `failure` case?

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/vector/WKBGeometryVector.java
jnh5y: Got it!

geomesa-arrow/geomesa-arrow-gt/src/test/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileTest.scala
jnh5y: Updated!

elahrvivaz:
replaced with #2852

PR #2570, opened 2020-06-10
GEOMESA-2678 FSDS - lazy load for metadata

elahrvivaz:
* Fix StringSerialization decoding


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -52,37 +58,61 @@ import scala.util.control.NonFatal
...
-  private val partitions = new ConcurrentHashMap[String, PartitionMetadata]()
+  private val expiry = PathCache.CacheDurationProperty.toDuration.get.toMillis
 
-  override def getPartition(name: String): Option[PartitionMetadata] = Option(partitions.get(name))
+  // cache of files associated with each partition
+  // we use a cache to provide lazy non-blocking refresh, but the cache will only ever have 1 element in it
+  private val partitions: LoadingCache[BoxedUnit, ConcurrentMap[String, PartitionFiles]] =
+    Caffeine.newBuilder().refreshAfterWrite(expiry, TimeUnit.MILLISECONDS).build(
+      new CacheLoader[BoxedUnit, ConcurrentMap[String, PartitionFiles]]() {
+        override def load(key: BoxedUnit): ConcurrentMap[String, PartitionFiles] = readPartitionFiles(8)
```
jnh5y, line 80: Oh man, BoxedUnit as a cache's key.  This is neat and simultaneously, I have to try and think of alternatives.  

I don't have any at the minute.:)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -92,151 +122,83 @@ class FileBasedMetadata(
...
-            partitions.put(m.name, m.toMetadata)
-          }
-          keys.asScala.foreach(partitions.remove)
 
-          if (write) {
-            FileBasedMetadata.writeCompactedConfig(fc, directory, merged)
-            paths.foreach(fc.delete(_, false)) // note: already a parallel collection
-          }
+    // in normal usage, we never pass in a partition to this method
+    partition.foreach(p => logger.warn(s"Ignoring requested partition '$p' and compacting all metadata"))
```
jnh5y, line 139: Admittedly, I'm skimming...  I do like the idea of being able to compact a range/partition/etc.  

Is that a sane addition or should it be a follow-on ticket?

elahrvivaz: we never actually invoked it, so i didn't implement it this time. compacting a single partition means merging the partition-specific files into a compacted file that also contains other partitions, so it's a bit messy... and compacting region by region will end up rewriting the file over and over, so it's inefficient.
i think the reason the method has a partition at all is b/c it inherits from the Compactable trait that we created, otherwise i'd remove it. could maybe do something better around that trait, but eh

jnh5y:
Merged as 9f91bccf6c87258d6ca54d7f417b716dcac88d52

PR #2572, opened 2020-06-11
GEOMESA-2873 Converters - support attribute names with dots

elahrvivaz:
* Supports NiFi flow attributes
* Update NiFi docs

### DIFF
--- a/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/ExpressionParser.scala
+++ b/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/ExpressionParser.scala
@@ -67,7 +67,7 @@ private [transforms] class ExpressionParser extends BasicParser {
   }
 
   private def field: Rule1[Expression] = rule("$field") {
-    "$" ~ unquotedString ~~> { i => FieldLookup(i) }
+    "$" ~ (unquotedString | "{" ~ oneOrMore(char | "." | " ") ~> { c => c } ~ "}") ~~> { name => FieldLookup(name) }
   }
 
   private def tryFunction: Rule1[Expression] = rule("try") {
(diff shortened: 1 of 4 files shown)


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/ExpressionParser.scala
```diff
@@ -67,7 +67,7 @@ private [transforms] class ExpressionParser extends BasicParser {
   }
 
   private def field: Rule1[Expression] = rule("$field") {
-    "$" ~ unquotedString ~~> { i => FieldLookup(i) }
+    "$" ~ (unquotedString | "{" ~ oneOrMore(char | "." | " ") ~> { c => c } ~ "}") ~~> { name => FieldLookup(name) }
```
jnh5y, line 70: Ok, time to ask about corner cases....

the Rule/Expression  "{" ~ oneOrMore(char | "." | " ") ~> { c => c } ~ "}" would match input like 

"{.}"
"{ . }"
"{foo...bar}"
"{ foo . bar }"
right?  

Do we want to be that permissive?  Am I missing a part of the syntax that'd avoid really weird things like that?

jnh5y: Oh, also, we could have multiple spaces and multiple dots.  Is that going to work every where with GeoTools/OGC?  

Or is this just inside the GeoMesa converters?

elahrvivaz: this is just inside the geomesa converters. yeah, you could use `${.}` and it would parse ok, but then presumably it would fail with an error that the field `.` doesn't exist... i could make it check for stricter syntax but seems like it should be handled by checks later on.

jnh5y: Cool.  Just makings sure I understand.  

If people start hurting themselves on this, we can revisit:).

jnh5y:
Merged 8e764ffad507ef4aa9a116f623874ecce22d8fa8

PR #2576, opened 2020-06-15
GEOMESA-2875 Fix WKB kryo encoding to not always store 3 dims

elahrvivaz:

### DIFF
--- a/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/WkbSerialization.scala
+++ b/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/WkbSerialization.scala
@@ -64,36 +64,37 @@ trait WkbSerialization[T <: NumericWriter, V <: NumericReader] {
   }
 
   private def writePoint(out: T, g: Point): Unit = {
+    val twoD = WkbSerialization.is2d(g)
     val coords = g.getCoordinateSequence
-    val (flag, writeDims) = if (coords.getDimension == 2) { (Point2d, false) } else { (Point, true) }
+    val (flag, writeDims) = if (twoD) { (Point2d, false) } else { (Point, true) }
     out.writeInt(flag, optimizePositive = true)
-    writeCoordinateSequence(out, coords, writeLength = false, writeDims)
+    writeCoordinateSequence(out, coords, twoD, writeLength = false, writeDims)
   }
 
   private def readPoint(in: V, dims: Option[Int]): Point =
     factory.createPoint(readCoordinateSequence(in, Some(1), dims))
 
   private def writeLineString(out: T, g: LineString): Unit = {
+    val twoD = WkbSerialization.is2d(g)
     val coords = g.getCoordinateSequence
-    val (flag, writeDims) = if (coords.getDimension == 2) { (LineString2d, false) } else { (LineString, true) }
+    val (flag, writeDims) = if (twoD) { (LineString2d, false) } else { (LineString, true) }
     out.writeInt(flag, optimizePositive = true)
-    writeCoordinateSequence(out, coords, writeLength = true, writeDims)
+    writeCoordinateSequence(out, coords, twoD, writeLength = true, writeDims)
   }
 
   private def readLineString(in: V, dims: Option[Int]): LineString =
     factory.createLineString(readCoordinateSequence(in, None, dims))
 
   private def writePolygon(out: T, g: Polygon): Unit = {
+    val twoD = WkbSerialization.is2d(g)
     val exterior = g.getExteriorRing.getCoordinateSequence
-    val twoD = exterior.getDimension == 2 &&
-        (0 until g.getNumInteriorRing).forall(i => g.getInteriorRingN(i).getCoordinateSequence.getDimension == 2)
     val (flag, writeDims) = if (twoD) { (Polygon2d, false) } else { (Polygon, true) }
     out.writeInt(flag, optimizePositive = true)
-    writeCoordinateSequence(out, exterior, writeLength = true, writeDims)
+    writeCoordinateSequence(out, exterior, twoD, writeLength = true, writeDims)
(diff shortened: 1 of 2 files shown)


geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/WkbSerialization.scala
```diff
@@ -191,4 +194,15 @@ object WkbSerialization {
...
   val Polygon: Int            = 10
+
+  private def is2d(geometry: Geometry): Boolean = {
+    // don't trust coord.getDimensions - it always returns 3 in jts
+    // instead, check for NaN for the z dimension
+    // note that we only check the first coordinate - if a geometry is written with different
+    // dimensions in each coordinate, some information may be lost
+    val coord = geometry.getCoordinate
+    // check for dimensions - use NaN != NaN to verify z coordinate
+    // TODO check for M coordinate when added to JTS
```
jnh5y, line 205: JTS already supports measure dimensions somewhat.  Folks would likely have to go to lengths to work around some deficiencies in the WKTReader/WKBReader.

Since this TODO would be a decent size change, it is worth entering a JIRA ticket to the tune of "Support M dimensions"?  (This would give us a clear place to signal that GeoMesa doesn't do it, and give a place to discuss it.)

geomesa-features/geomesa-feature-kryo/src/test/scala/org/locationtech/geomesa/features/kryo/KryoGeometrySerializerTest.scala
```diff
@@ -36,12 +37,20 @@ class KryoGeometrySerializerTest extends Specification {
...
-          val deserialized = KryoGeometrySerialization.deserialize(in)
-          deserialized mustEqual geom
+          foreach(serializers) { serializer =>
+            out.clear()
+            serializer(geom)
+            val bytes = out.toBytes
+            // ensure we didn't write 3 dimensions
+            bytes.length must beLessThan(geom.getCoordinates.length * 24)
+            KryoGeometrySerialization.deserialize(new Input(bytes)) mustEqual geom
+          }
```
jnh5y, line 53: Would it make sense to add a unit test to show that we can serialize POINTZs and LINESTRINGZs in Kryo?

jnh5y:
Is WKBSerialization used in Spark as well?  E.g., does this resolve GM-2834?

elahrvivaz:
no, spark uses the jts WKBWriter, so this doesn't resolve that issue.

geomesa-features/geomesa-feature-kryo/src/test/scala/org/locationtech/geomesa/features/kryo/KryoGeometrySerializerTest.scala
elahrvivaz: we do have unit tests for those somewhere... possibly in the accumulo datastore tests

elahrvivaz: here it is: https://github.com/locationtech/geomesa/blob/master/geomesa-features/geomesa-feature-kryo/src/test/scala/org/locationtech/geomesa/features/kryo/KryoFeatureSerializerTest.scala#L137

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/WkbSerialization.scala
elahrvivaz: this TODO was copied over from the original check. How does jts support M, since all coords have 3 dimensions? I think it would be pretty easy for us to implement if jts supports it.

elahrvivaz:
our kryo wkb isn't 'true' wkb, it's slightly simplified, so we might not want to use it in spark if we want the serialized geometries to be more interoperable. it also might complicate the classpath, since spark uses kryo. and, we'd have to bring in more depedencies to the spark-jts module, or have an alternate serialization path that kicks in when including spark-core (or whatever). all of which is to say, we could use it, but we don't atm.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/WkbSerialization.scala
jnh5y: https://github.com/locationtech/jts/blob/jts-1.16.0/modules/core/src/test/java/org/locationtech/jts/io/WKTReaderTest.java#L90

https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/test/jts/GeometryTestCase.java#L178-L201

Those might help give some ideas.

elahrvivaz: nice, looks like it was added in 1.16. guess we should read release notes :P
opened https://geomesa.atlassian.net/browse/GEOMESA-2877

jnh5y: Yeah, are you talking about the release notes that I evidently wrote here: https://github.com/locationtech/jts/releases/tag/jts-1.16.0 ?  That's why I had a vague recollection of JTS supporting ZM;).

jnh5y:
Merged as 7046cbc4c13828d7c070996ecb85ad6d652cb944

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/WkbSerialization.scala
elahrvivaz: :joy:

PR #2581, opened 2020-06-15
GEOMESA-2876 Check geometries for 2 or 3 dims when encoding with WKB

elahrvivaz:

### DIFF
--- a/geomesa-features/geomesa-feature-avro/src/test/scala/org/locationtech/geomesa/features/avro/AvroFeatureSerializerTest.scala
+++ b/geomesa-features/geomesa-feature-avro/src/test/scala/org/locationtech/geomesa/features/avro/AvroFeatureSerializerTest.scala
@@ -160,7 +160,7 @@ class AvroFeatureSerializerTest extends Specification with LazyLogging {
           }
         }
       }
-    }.pendingUntilFixed()
+    }
 
     "correctly serialize and deserialize collection types" in {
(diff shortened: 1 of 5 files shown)


elahrvivaz:
@netanel246 after chatting with @jnh5y and the author of JTS, we decided to add a wrapper around the WKB encoding to detect the geometry dimension. That way you don't pay the cost of serializing empty Z coords unless you are using them.

geomesa-features/geomesa-feature-avro/src/test/scala/org/locationtech/geomesa/features/avro/AvroFeatureSerializerTest.scala
```diff
@@ -160,7 +160,7 @@ class AvroFeatureSerializerTest extends Specification with LazyLogging {
           }
         }
       }
-    }.pendingUntilFixed()
```
jnh5y, line 163: Doesn't removing a pendingUntilFixed() just make you feel good inside?

jnh5y:
Merged as 35825301439d97bcfad0eeff14255c9520c6bf77.

@netanel246 thanks again for contributing and bringing up this topic!  Shout if you have any feedback on this change.  (I'm merging it today, but am happy to continue discussing the topic of how to best handle this going forward.  I hope this work is an improvement for GeoMesa across the board.)

PR #2593, opened 2020-07-05
GEOMESA-2884 ST_Transform functionality to convert geometries to different coordinate reference systems

danfran:
* Supports conversions for CRS like EPSG:*
* SQL and DataFrame supports added
* Added unit tests for SQL and DataFrame
* Updated documentation including the API description

Signed-off-by: Daniele Francesconi <dfrancesconi12@gmail.com>

### DIFF
--- a/geomesa-spark/geomesa-spark-jts/pom.xml
+++ b/geomesa-spark/geomesa-spark-jts/pom.xml
@@ -41,6 +41,18 @@
             <groupId>org.locationtech.spatial4j</groupId>
             <artifactId>spatial4j</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-shapefile</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-epsg-hsql</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-opengis</artifactId>
+        </dependency>
 
         <!-- test -->
         <dependency>
--- a/geomesa-spark/geomesa-spark-jts/pom.xml
+++ b/geomesa-spark/geomesa-spark-jts/pom.xml
@@ -41,18 +41,6 @@
             <groupId>org.locationtech.spatial4j</groupId>
             <artifactId>spatial4j</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-shapefile</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-epsg-hsql</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-opengis</artifactId>
-        </dependency>
 
         <!-- test -->
         <dependency>
--- a/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
+++ b/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
@@ -354,6 +354,9 @@ object DataFrameFunctions extends SpatialEncoders {
 
     def st_difference(geom1:Column, geom2:Column): TypedColumn[Any, Geometry] =
       udfToColumn(ST_Difference, relationNames, geom1, geom2)
+
+    def st_transform(geom: Column, fromCRS: Column, toCRS: Column): TypedColumn[Any, Geometry] =
+      udfToColumn(ST_Transform, relationNames, geom, fromCRS, toCRS)
   }
 
   /** Stack of all DataFrame DSL functions. */
--- a/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
+++ b/geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
@@ -354,9 +354,6 @@ object DataFrameFunctions extends SpatialEncoders {
 
     def st_difference(geom1:Column, geom2:Column): TypedColumn[Any, Geometry] =
       udfToColumn(ST_Difference, relationNames, geom1, geom2)
-
-    def st_transform(geom: Column, fromCRS: Column, toCRS: Column): TypedColumn[Any, Geometry] =
-      udfToColumn(ST_Transform, relationNames, geom, fromCRS, toCRS)
   }
 
   /** Stack of all DataFrame DSL functions. */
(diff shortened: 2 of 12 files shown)


geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/DataFrameFunctions.scala
```diff
@@ -354,6 +354,9 @@ object DataFrameFunctions extends SpatialEncoders {
 
     def st_difference(geom1:Column, geom2:Column): TypedColumn[Any, Geometry] =
       udfToColumn(ST_Difference, relationNames, geom1, geom2)
+
+    def st_transform(geom: Column, fromCRS: Column, toCRS: Column): TypedColumn[Any, Geometry] =
```
elahrvivaz, line 358: i'm a little confused... i thought all the st_transform methods and code would need to be defined in the spark-sql module instead of the spark-jts module, since they require the `org.geotools` dependencies, which we don't want to introduce to the jts module. but i see that `ST_LengthSphere` is in the jts module, even though the function `def st_lengthSphere` is defined in the spark-sql module and also uses geotools? @jnh5y can you clarify?
but I think the bottom line is that we don't want to include the changes you made in the geomesa-spark-jts/pom.xml to add geotools to the jts module.

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
```diff
@@ -58,6 +61,14 @@ object SpatialRelationFunctions {
 
   val ST_Difference: (Geometry, Geometry) => Geometry = nullableUDF((geom1, geom2) => geom1.difference(geom2))
 
+  val ST_Transform: (Geometry, String, String) => Geometry = nullableUDF { (geometry, sourceCRScode, targetCRScode) =>
+    System.setProperty("org.geotools.referencing.forceXY", "true")
```
elahrvivaz, line 65: yeah, it doesn't seem good to have this here. You might need to use a `GeometryCoordinateSequenceTransformer` to do the transform instead of using JTS directly, like we do here: https://github.com/locationtech/geomesa/blob/main/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/utils/Reprojection.scala#L41-L42

danfran: Yes I see your point, I have just followed the "natural" path suggested from existent code. I will try to reuse the "Reprojection" from above. For where to place the 2 methods (SQL and DataFrame) I wait on you guys to let me know.

geomesa-spark/geomesa-spark-jts/pom.xml
```diff
@@ -41,6 +41,18 @@
...
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-shapefile</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-epsg-hsql</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-opengis</artifactId>
```
jnh5y, line 54: Sorry if I wasn't clear previously in Gitter.  This is what @elahrvivaz was mentioning in another comment.  This module will not include GeoTools (to allow license decoupling).  Admittedly, this is a somewhat arbitrary distinction.

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/udf/SpatialRelationFunctions.scala
jnh5y: The non-JTS functions are in here: https://github.com/locationtech/geomesa/blob/main/geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeometricDistanceFunctions.scala and then a DataFrame binding for them is added in https://github.com/locationtech/geomesa/blob/main/geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/DataFrameFunctions.scala

Admittedly, the first non-JTS were about distance, so we're slightly abusing the filename, but I think I adding the functions to the first file will be ok.  

@elahrvivaz do you have a better suggestion?

elahrvivaz: no, that seems good. I think the length_sphere (vs spheroid) might be confusing and/or in the wrong place, but however the current methods are working, the classes you point out seem like the right place for new ones.

danfran: I have turned the code into:
`  val ST_Transform: (Geometry, String, String) => Geometry = nullableUDF { (geometry, fromCRSCode, toCRSCode) =>
    val transformer = new GeometryCoordinateSequenceTransformer
    val fromCode = CRS.decode(fromCRSCode, true)
    val toCode = CRS.decode(toCRSCode, true)
    transformer.setMathTransform(CRS.findMathTransform(fromCode, toCode, true))
    transformer.transform(geometry)
  }`

in the `org.locationtech.geomesa.spark.GeometricDistanceFunctions.scala` file as suggested. The problem is in the `DataFrameFunctions.scala` as I have to do this:

`trait SpatialRelations {

    import org.locationtech.geomesa.spark.jts.udf.SpatialRelationFunctions._
    import org.locationtech.geomesa.spark.DataFrameFunctions._

    def st_distanceSpheroid(left: Column, right: Column): TypedColumn[Any, lang.Double] =
      udfToColumn(ST_DistanceSphere, relationNames, left, right)

    def st_lengthSphere(line: Column): TypedColumn[Any, lang.Double] =
      udfToColumn(ST_LengthSphere, relationNames, line)

    def st_transform(geom: Column, fromCRS: Column, toCRS: Column): TypedColumn[Any, Geometry] =
      udfToColumn(ST_Transform, distanceNames, geom, fromCRS, toCRS)
  }`.

As you can see `st_transform` is using `distanceNames` map, instead of `relationNames` which is in the core module. Also I add to change visibility in `distanceNames` in `GeometricDistanceFunctions`:

`//  private[geomesa] val distanceNames = Map(
  val distanceNames = Map(`

Still this create some "encoding" issues when I try to run the test:

`@RunWith(classOf[JUnitRunner])
class DataFrameFunctionsTest extends Specification with LazyLogging with SpatialRelations {
  type DFRelation = (Column, Column) => TypedColumn[Any, jl.Boolean]
  sequential

  "DataFrame functions" should {

    val spark = SparkSQLTestUtils.createSparkSession()
    val sc = spark.sqlContext
    sc.withJTS

    var dfBlank: DataFrame = null

    // before
    step {
      dfBlank = spark.createDataFrame(spark.sparkContext.makeRDD(Seq(Row())), StructType(Seq.empty))
    }

    "st_transform" >> {
      val pointWGS84 = "POINT(-0.871722 52.023636)"
      val expectedOSGB36 = WKTUtils.read("POINT(477514.0081191745 236736.03179982008)")
      val transf = dfBlank.select(st_transform(st_geomFromWKT(pointWGS84), lit("EPSG:4326"), lit("EPSG:27700"))).first
      transf must not(throwAn[Exception])
      transf mustEqual expectedOSGB36
    }

    // after
    step {
      spark.stop()
    }
  }
}`

Am I putting code in the wrong parts?

elahrvivaz: I'm not sure at this point... I will try to check out the branch and see. @jnh5y any thoughts?

jnh5y:
@danfran looks great!  Thanks for working with us to get the code into the right spot in the repo and to document it well!

jnh5y:
Merged as 79496a59d3206aef860b56387e4a5496ac408644

PR #2596, opened 2020-07-09
GEOMESA-2889,GEOMESA-2882 Deploy gs-plugins to maven central, fix geo…

elahrvivaz:
…mesa-web deployment

### DIFF
--- a/geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
+++ b/geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
     <parent>
         <artifactId>geomesa-archetypes_2.11</artifactId>
         <groupId>org.locationtech.geomesa</groupId>
         <version>3.0.1-SNAPSHOT</version>
     </parent>
-    <modelVersion>4.0.0</modelVersion>
 
     <artifactId>geomesa-archetypes-gs-plugin_2.11</artifactId>
-    <name>GeoMesa Geoserver Plugin Archetype</name>
+    <name>GeoMesa GeoServer Plugin Archetype</name>
     <packaging>pom</packaging>
 
     <dependencyManagement>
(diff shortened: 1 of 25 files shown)


elahrvivaz:
This PR publishes the poms for -gs-plugins and -dist modules, but doesn't attach the binary files. It changes the gs-plugins to type 'pom' instead of type 'jar'

geomesa-accumulo/geomesa-accumulo-dist/src/main/assemblies/component.xml
```diff
@@ -40,10 +41,7 @@
             <useProjectArtifact>false</useProjectArtifact>
             <useTransitiveDependencies>false</useTransitiveDependencies>
             <includes>
-                <include>org.locationtech.geomesa:geomesa-accumulo-gs-plugin*</include>
                 <include>org.locationtech.geomesa:geomesa-process-wps*</include>
-                <include>org.locationtech.geomesa:geomesa-stream-gs-plugin*</include>
```
jnh5y, line 45: Ok, slow morning questions...  

1.  Why are the gs-plugin out?
2.  Why is the geomesa-process still in?

Basically, what does this block do since we aren't creating a tar.gz?

geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
```diff
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
```
jnh5y, line 5: Does this change matter?  (Just curious)

geomesa-arrow/geomesa-arrow-dist/src/main/assemblies/component.xml
```diff
@@ -22,7 +22,6 @@
             <useTransitiveDependencies>false</useTransitiveDependencies>
             <useStrictFiltering>true</useStrictFiltering>
             <includes>
-                <include>org.locationtech.geomesa:geomesa-arrow-gs-plugin*</include>
                 <include>org.locationtech.geomesa:geomesa-process-wps*</include>
```
jnh5y, line 25: If these blocks are all standardized now, can the configuration be factored up?

geomesa-accumulo/geomesa-accumulo-dist/src/main/assemblies/component.xml
elahrvivaz: the tar.gz's are still being built, and they now get included as a fileset a few lines down, instead of as a dependency set here. had to make that change as they aren't 'attached' any more, so you can't retrieve them as a regular dependency. they aren't attached any more so that they don't get published to maven after being built.

geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
elahrvivaz: it was there before, the poms were just inconsistent in the ordering of elements so i cleaned it up to make it easier to compare them.

geomesa-arrow/geomesa-arrow-dist/src/main/assemblies/component.xml
elahrvivaz: possibly... it's theoretically possibly to share common 'component' descriptors and import them in different places. could be used for e.g. readmes, etc that we include lower down as well. we can add a ticket to do that, i'd rather not do it as part of this pr though.

PR #2599, opened 2020-07-27
GEOMESA-2896 Fix case-insensitive checks for user data

elahrvivaz:


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -276,17 +296,17 @@ object RichSimpleFeatureType {
       Seq(GeomesaPrefix) ++ userData[String](UserDataPrefix).map(_.split(",")).getOrElse(Array.empty)
 
     def setZShards(splits: Int): Unit = sft.getUserData.put(IndexZShards, splits.toString)
-    def getZShards: Int = userData[String](IndexZShards).map(_.toInt).getOrElse(4)
+    def getZShards: Int = userData(IndexZShards).map(int).getOrElse(4)
 
     def setAttributeShards(splits: Int): Unit = sft.getUserData.put(IndexAttributeShards, splits.toString)
-    def getAttributeShards: Int = userData[String](IndexAttributeShards).map(_.toInt).getOrElse(4)
+    def getAttributeShards: Int = userData(IndexAttributeShards).map(int).getOrElse(4)
```
jnh5y, line 302: Part of me is nervous about having a 'int' as a function like this, but I'm going to go for it!

PR #2601, opened 2020-08-05
HBase TTL (GEOMESA-2264)

alex-shen1:
Added functionality to the `HBaseIndexWriter.write()` function in `HBaseIndexAdapter.scala` that sets a time-to-live (TTL) for each newly-written feature if its type's userdata contains a value for the `geomesa.feature.expiry` key (same one used for Accumulo; e.g. "4 seconds"). Importantly, the TTL is *not* simply determined by when the feature itself was created, but by the feature's dtg attribute. If the key does not exist, the feature will be added normally without a TTL, and if the time elapsed since the feature's dtg exceeds the expiration time, the feature will not be added at all. Once a feature is added with a TTL into HBase, it will automatically removed when enough time has passed.

Unit tests have been included in `HBaseTTLTest.scala`  for writing features and making sure they are removed with the proper timing, as well as refusing to add features that have already expired based on that feature's dtg and the type's expiration time.

Associated JIRA ticket: [https://geomesa.atlassian.net/browse/GEOMESA-2264](https://geomesa.atlassian.net/browse/GEOMESA-2264)


alex-shen1:
Messed up this branch doing rebasing, opened a new PR w/ the same text

PR #2602, opened 2020-08-05
GEOMESA-2264 HBase TTL

alex-shen1:
Added functionality to the `HBaseIndexWriter.write()` function in `HBaseIndexAdapter.scala` that sets a time-to-live (TTL) for each newly-written feature if its type's userdata contains a value for the `geomesa.feature.expiry` key (same one used for Accumulo; e.g. "4 seconds"). Importantly, the TTL is not simply determined by when the feature itself was created, but by the feature's dtg attribute. If the key does not exist, the feature will be added normally without a TTL, and if the time elapsed since the feature's dtg exceeds the expiration time, the feature will not be added at all. Once a feature is added with a TTL into HBase, it will automatically removed when enough time has passed.

Unit tests have been included in `HBaseTTLTest.scala` for writing features and making sure they are removed with the proper timing, as well as refusing to add features that have already expired based on that feature's dtg and the type's expiration time.

Associated JIRA ticket: [https://geomesa.atlassian.net/browse/GEOMESA-2264](https://geomesa.atlassian.net/browse/GEOMESA-2264)

### DIFF
--- a/docs/user/hbase/index_config.rst
+++ b/docs/user/hbase/index_config.rst
@@ -21,3 +21,26 @@ are ``snappy``, ``lzo``, ``gz``, ``bzip2``, ``lz4`` or ``zstd``.
     geomesa-hbase create-schema --compression snappy ...
 
 For more information on how to set schema options, see :ref:`set_sft_options`.
+
+Feature Expiration
+------------------
+
+HBase supports setting a per-feature time-to-live. Expiration can be set in the ``SimpleFeatureType`` user data,
+using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details on configuring the user data.
+Expiration can be set before calling ``createSchema``, or can be added to an existing schema by calling
+``updateSchema``. However, note that if added through ``updateSchema``, any existing features will not be expired.
+
+Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
+specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
+a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
+``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
+
+Features are actively expired by a background process running in each data store instance. By default, features
+are not created with an expiration time. This can be controlled by configuring the system property
+``geomesa.redis.age.off.interval``. To disable the background process, the property can be set to ``Inf``. If
+there are multiple active data store instances, they will synchronize among themselves to avoid duplicating work.
+
+Note that because the expiration is an active process, expired features may still be returned until the
+expiration process purges them. If precise expiration is required, use attribute-based expiration and apply a
+default CQL filter for each query. For example, in GeoServer you can set a default layer filter to something like
+``dtg > currentDate('-P1D')``. See :ref:`filter_functions` for an explanation of the ``currentDate`` function.
--- a/docs/user/hbase/index_config.rst
+++ b/docs/user/hbase/index_config.rst
@@ -30,4 +30,9 @@ using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details
 Expiration can be set before calling ``createSchema``, or can be added to an existing schema by calling
 ``updateSchema``. However, note that if added through ``updateSchema``, any existing features will not be expired.
 
+Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
+specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
+a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
+``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
+
 Feature expiration is based on HBase's time-to-live functionality. See https://hbase.apache.org/book.html#ttl for details.
--- a/docs/user/hbase/index_config.rst
+++ b/docs/user/hbase/index_config.rst
@@ -30,17 +30,4 @@ using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details
 Expiration can be set before calling ``createSchema``, or can be added to an existing schema by calling
 ``updateSchema``. However, note that if added through ``updateSchema``, any existing features will not be expired.
 
-Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
-specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
-a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
-``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
-
-Features are actively expired by a background process running in each data store instance. By default, features
-are not created with an expiration time. This can be controlled by configuring the system property
-``geomesa.redis.age.off.interval``. To disable the background process, the property can be set to ``Inf``. If
-there are multiple active data store instances, they will synchronize among themselves to avoid duplicating work.
-
-Note that because the expiration is an active process, expired features may still be returned until the
-expiration process purges them. If precise expiration is required, use attribute-based expiration and apply a
-default CQL filter for each query. For example, in GeoServer you can set a default layer filter to something like
-``dtg > currentDate('-P1D')``. See :ref:`filter_functions` for an explanation of the ``currentDate`` function.
+Feature expiration is based on HBase's time-to-live functionality. See https://hbase.apache.org/book.html#ttl for details.
--- a/docs/user/redis/index_config.rst
+++ b/docs/user/redis/index_config.rst
@@ -15,15 +15,7 @@ using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details
 Expiration can be set before calling ``createSchema``, or can be added to an existing schema by calling
 ``updateSchema``. However, note that if added through ``updateSchema``, any existing features will not be expired.
 
-Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
-specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
-a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
-``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
-
-Features are actively expired by a background process running in each data store instance. By default, features
-will be expired every 10 minutes. This can be controlled by configuring the system property
-``geomesa.redis.age.off.interval``. To disable the background process, the property can be set to ``Inf``. If
-there are multiple active data store instances, they will synchronize among themselves to avoid duplicating work.
+Feature expiration is based on HBase's time-to-live functionality. See https://hbase.apache.org/book.html#ttl for details.
 
 Note that because the expiration is an active process, expired features may still be returned until the
 expiration process purges them. If precise expiration is required, use attribute-based expiration and apply a
--- a/docs/user/redis/index_config.rst
+++ b/docs/user/redis/index_config.rst
@@ -15,7 +15,15 @@ using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details
 Expiration can be set before calling ``createSchema``, or can be added to an existing schema by calling
 ``updateSchema``. However, note that if added through ``updateSchema``, any existing features will not be expired.
 
-Feature expiration is based on HBase's time-to-live functionality. See https://hbase.apache.org/book.html#ttl for details.
+Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
+specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
+a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
+``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
+
+Features are actively expired by a background process running in each data store instance. By default, features
+will be expired every 10 minutes. This can be controlled by configuring the system property
+``geomesa.redis.age.off.interval``. To disable the background process, the property can be set to ``Inf``. If
+there are multiple active data store instances, they will synchronize among themselves to avoid duplicating work.
 
 Note that because the expiration is an active process, expired features may still be returned until the
 expiration process purges them. If precise expiration is required, use attribute-based expiration and apply a
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
@@ -1,16 +1,16 @@
-/***********************************************************************
+/** *********************************************************************
  * Copyright (c) 2013-2020 Commonwealth Computer Research, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Apache License, Version 2.0
  * which accompanies this distribution and is available at
  * http://www.opensource.org/licenses/apache2.0.php.
- ***********************************************************************/
+ * ********************************************************************* */
 
 package org.locationtech.geomesa.hbase.data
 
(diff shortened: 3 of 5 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -67,9 +69,9 @@ class HBaseIndexAdapter(ds: HBaseDataStore) extends IndexAdapter[HBaseDataStore]
   import scala.collection.JavaConverters._
 
   override def createTable(
-      index: GeoMesaFeatureIndex[_, _],
-      partition: Option[String],
-      splits: => Seq[Array[Byte]]): Unit = {
+                            index: GeoMesaFeatureIndex[_, _],
```
aheyne, line 72: It's typically good practice to only include code changes, even styling, that are only relevant to the content of the PR. @elahrvivaz What's your opinion?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -554,18 +580,19 @@ object HBaseIndexAdapter extends LazyLogging {
...
-      ds: HBaseDataStore,
-      indices: Seq[GeoMesaFeatureIndex[_, _]],
-      wrapper: FeatureWrapper[WritableFeature],
-      partition: Option[String]
-    ) extends BaseIndexWriter(indices, wrapper) {
+                          ds: HBaseDataStore,
+                          indices: Seq[GeoMesaFeatureIndex[_, _]],
+                          wrapper: FeatureWrapper[WritableFeature],
+                          partition: Option[String],
+                          sft: Some[SimpleFeatureType]
```
aheyne, line 594: @jnh5y Points out we're changing the API for the HBaseIndexWriter. Is this find as is, should we wrap this in a compatibility layer or is there a better way to address this?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
alex-shen1: This was probably just IntelliJ autoformatting, I can just revert it if necessary

alex-shen1:
I'm not sure why unit tests I never even touched are failing now... looking into it

alex-shen1:
> I'm not sure why unit tests I never even touched are failing now... looking into it

Broke something in the HBaseIndexAdapter, fixed now.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
elahrvivaz: yeah, looks like a lot of auto-formatting changes. @alex-shen1  could you revert those? thanks.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
elahrvivaz: i think this class is 'internal' so we can change it. i don't expect anyone would be instantiating it directly...
however, the feature indices have a reference to the sft, so you don't need to pass it in explicitly. you could instead do something like: `indices.headOption.flatMap(_.sft.getDtgIndex).getOrElse(-1)`

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -579,6 +607,22 @@ object HBaseIndexAdapter extends LazyLogging {
...
+    private val duration : Duration = sft.map{ sft =>
+      val userData = sft.getUserData
+      if (userData.containsKey("geomesa.feature.expiry")) {
+        Duration(userData.get("geomesa.feature.expiry").toString)
+      } else null
+    }.getOrElse(null)
+
+    private val expiration = if (duration != null && dtgIndex != -1) {
+      FeatureTimeExpiration("", dtgIndex, duration)
+    } else null
```
elahrvivaz, line 622: you could replace this whole block with the existing convenience method:
```
val expiration = indices.headOption.flatMap(_.sft.getFeatureExpiration).orNull
```

elahrvivaz: that will support ingest time based expiration too, if someone wanted to use it (instead of only supporting based on an attribute time)

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseTTLTest.scala
```diff
@@ -0,0 +1,135 @@
...
+          sf.setAttribute(1, dateFormat.format(featureTime))
+          sf
+        }
+
+        val ids = fs.addFeatures(new ListFeatureCollection(sft, toAdd))
+        ids.asScala.map(_.getID) must containTheSameElementsAs((0 until numFeatures).map(_.toString))
+
+        getElements(ds, typeName).size mustEqual numFeatures
+
+        Thread.sleep(ttl) // wait until one feature times out
```
elahrvivaz, line 77: hmm, i wish there was a better way to test these without relying on timing, but nothing is coming to mind...

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
elahrvivaz: this will change the configuration slightly - to use an attribute based expiration you'll need to configure it like `dtg(4 minutes)`. but that will be consistent with the docs and with accumulo

elahrvivaz:
could you add a section to the docs on the new feature? you should be able to copy/paste e.g. https://github.com/locationtech/geomesa/blob/main/docs/user/redis/index_config.rst#feature-expiration with minimal changes to https://github.com/locationtech/geomesa/blob/main/docs/user/hbase/index_config.rst. thanks!

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBaseTTLTest.scala
aheyne: My terrible idea was using AOP to intercept the system get time call with aspectj or something.

elahrvivaz: there is an 'eventually' call in specs that will re-try something a configurable number of times, that may make this more robust in case the hbase ttl doesn't always get processed right away - but maybe it always would? i'm not sure on the internals of how hbase processes ttls. and you'd still need to sleep to let it expire.
for reference: https://github.com/locationtech/geomesa/blob/main/geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreTest.scala#L170

elahrvivaz: some frameworks also let you inject a clock that you can control, but i doubt that would be feasible using the hbase mini cluster

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
alex-shen1: I don't think this works because the goal is to have the TTL be based on a feature's dtg attribute; as far as I can tell this method's returned FeatureExpiration always returns a fixed TTL without considering feature dtg. I could be mistaken, but at the very least a simple one-to-one replacement of the code block breaks the tests.

elahrvivaz: it should work if you configure the expiry in the unit test as `attribute(expiry)` i.e. `dtg(4 minutes)`

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
alex-shen1: Will the reference to sft always be there? I need to access `sft.getUserData`, but since that just returns data and not an Option, flatMap isn't playing nicely with it. It looks like I _could_ just use `indices.headOption.get.sft.getUserData`, but that feels like it'd break on some edge case that I can't think of at the moment.

elahrvivaz: i think this may be moot if you use the helper function i mentioned below - but since user data is a java map, you'd have to wrap the result in an option to use flat map. So something like:
`indices.headOption.flatMap(i => Option(i.sft.getUserData.get("foo")))`

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
alex-shen1: Should be reverted in latest commit

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
alex-shen1: Yeah, thanks - I only saw your first comment when I wrote my response; after editing the tests it passes now.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
alex-shen1: Yep, once I got the other function working it's not an issue. In fact, as far as I can tell even dtgIndex becomes irrelevant because the schema definition specifies the attribute that's being used for TTL, so that simplifies things.

alex-shen1:
One thing to note - the docs for redis say that the default expiration is 10 minutes, whereas my implementation in HBase has no expiration by default (and I've adjusted the copypasted documentation for the HBase docs accordingly). Should I make the HBase default 10 minutes as well for consistency?

docs/user/hbase/index_config.rst
```diff
@@ -21,3 +21,26 @@ are ``snappy``, ``lzo``, ``gz``, ``bzip2``, ``lz4`` or ``zstd``.
...
+using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details on configuring the user data.
+Expiration can be set before calling ``createSchema``, or can be added to an existing schema by calling
+``updateSchema``. However, note that if added through ``updateSchema``, any existing features will not be expired.
+
+Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
+specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
+a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
+``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
+
+Features are actively expired by a background process running in each data store instance. By default, features
```
elahrvivaz, line 38: Could you remove these 2 paragraphs and just add something like 'Feature expiration is based on HBase's time-to-live functionality. See https://hbase.apache.org/book.html#ttl for details.`

alex-shen1:
Actually, I think I misunderstood the redis docs - does "features will be expired every 10 minutes" simply mean that's how often data gets _deleted?_ Because if that's the case, that's also not true for HBase, seeing how the tests depend on having it check to delete consistently.

elahrvivaz:
right, the ttl in redis is slightly different - those parts of the docs can be deleted in the hbase section.

elahrvivaz:
this is looking good, thanks. could you try to revert the rest of the formatting-only changes in HBaseIndexAdapter?

alex-shen1:
What about the last paragraph (starting with "Note that...")?

alex-shen1:
Are the formatting changes not reverted yet? I thought I fixed most of them in https://github.com/locationtech/geomesa/pull/2602/commits/6c50e1aca201ed499a61735884ee9507a9df615f

docs/user/hbase/index_config.rst
```diff
@@ -21,3 +21,26 @@ are ``snappy``, ``lzo``, ``gz``, ``bzip2``, ``lz4`` or ``zstd``.
...
+are not created with an expiration time. This can be controlled by configuring the system property
+``geomesa.redis.age.off.interval``. To disable the background process, the property can be set to ``Inf``. If
+there are multiple active data store instances, they will synchronize among themselves to avoid duplicating work.
+
+Note that because the expiration is an active process, expired features may still be returned until the
```
elahrvivaz, line 43: yeah this can be deleted too.

docs/user/redis/index_config.rst
```diff
@@ -15,15 +15,7 @@ using the key ``geomesa.feature.expiry``. See :ref:`set_sft_options` for details
...
-Expiration can be based on either ingest time or a feature attribute. To set expiration based on ingest time,
-specify a time-to-live as a duration string, e.g. ``24 hours`` or ``180 days``. To set expiration based on
-a feature attribute, specify the attribute along with a time-to-live in parentheses, e.g. ``dtg(24 hours)`` or
-``event-time(30 days)`` (where ``dtg`` and ``event-time`` are ``Date``-type attributes in the schema).
-
-Features are actively expired by a background process running in each data store instance. By default, features
-will be expired every 10 minutes. This can be controlled by configuring the system property
-``geomesa.redis.age.off.interval``. To disable the background process, the property can be set to ``Inf``. If
-there are multiple active data store instances, they will synchronize among themselves to avoid duplicating work.
+Feature expiration is based on HBase's time-to-live functionality. See https://hbase.apache.org/book.html#ttl for details.
```
elahrvivaz, line 18: oops, that's the wrong file ;)

alex-shen1: darn

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -87,17 +87,26 @@ class HBaseIndexAdapter(ds: HBaseDataStore) extends IndexAdapter[HBaseDataStore]
 
         val cols = groups.apply(index.sft).map(_._1)
         val bloom = Some(BloomType.NONE)
-        val encoding = if (index.name == IdIndex.name) { None } else { Some(DataBlockEncoding.FAST_DIFF) }
+        val encoding = if (index.name == IdIndex.name) {
```
elahrvivaz, line 90: i mean all these changes. i think pretty much everything above line 606 can be reverted.

alex-shen1: Yeah, I was manually fixing tabs, but I'll just go into the latest commit from main and copypaste to overwrite.

elahrvivaz: in intellij you can do a right-click->git->compare with branch->main, that will bring up a diff view and let you review and copy over each change.

elahrvivaz:
fixed merge as 0673d6b8e1f9c7491dfca6d6179f80ca03a503c7

PR #2603, opened 2020-08-28
GEOMESA-2902 Arrow dictionary encoding for list-type attributes

elahrvivaz:
* List items are dictionary encoded, not the list itself
* Add sorting for list-type attributes
* Move ObjectType to geomesa-utils

### DIFF
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
@@ -13,8 +13,8 @@ import java.util.Date
 import com.typesafe.scalalogging.LazyLogging
 import org.geotools.geometry.jts.ReferencedEnvelope
 import org.locationtech.geomesa.arrow.features.ArrowSimpleFeature
-import org.locationtech.geomesa.arrow.vector.ArrowAttributeReader.{ArrowDateReader, ArrowLineStringReader, ArrowPointReader}
-import org.locationtech.geomesa.arrow.vector.{ArrowDictionary, ArrowDictionaryReader, GeometryVector}
+import org.locationtech.geomesa.arrow.vector.ArrowAttributeReader._
+import org.locationtech.geomesa.arrow.vector.{ArrowDictionary, GeometryVector}
 import org.locationtech.geomesa.filter.checkOrderUnsafe
 import org.locationtech.geomesa.filter.factory.FastFilterFactory
 import org.locationtech.geomesa.utils.geotools.CRS_EPSG_4326
@@ -34,6 +34,7 @@ import scala.util.control.NonFatal
   */
 object ArrowFilterOptimizer extends LazyLogging {
 
+  import org.locationtech.geomesa.utils.geotools.RichAttributeDescriptors.RichAttributeDescriptor
   import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 
   import scala.collection.JavaConversions._
@@ -110,7 +111,11 @@ object ArrowFilterOptimizer extends LazyLogging {
       case Some(dictionary) =>
         val attrIndex = sft.indexOf(props.name)
         val numericValue = dictionary.index(props.literal.evaluate(null))
-        ArrowDictionaryEquals(attrIndex, numericValue)
+        if (sft.getDescriptor(attrIndex).isList) {
+          ArrowListDictionaryEquals(attrIndex, numericValue)
+        } else {
+          ArrowDictionaryEquals(attrIndex, numericValue)
+        }
     }
   }
 
@@ -189,6 +194,14 @@ object ArrowFilterOptimizer extends LazyLogging {
     override def evaluate(o: AnyRef): Boolean = {
       val arrow = o.asInstanceOf[ArrowSimpleFeature]
       arrow.getReader(i).asInstanceOf[ArrowDictionaryReader].getEncoded(arrow.getIndex) == value
+    }
+  }
+
+  case class ArrowListDictionaryEquals(i: Int, value: Int) extends Filter {
+    override def accept(visitor: FilterVisitor, extraData: AnyRef): AnyRef = extraData
+    override def evaluate(o: AnyRef): Boolean = {
+      val arrow = o.asInstanceOf[ArrowSimpleFeature]
+      arrow.getReader(i).asInstanceOf[ArrowListDictionaryReader].getEncoded(arrow.getIndex).contains(value)
     }
   }
 }
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
@@ -10,13 +10,13 @@ package org.locationtech.geomesa.arrow.io
 
 import java.io.{ByteArrayOutputStream, Closeable, OutputStream}
 import java.nio.channels.Channels
-import java.util.PriorityQueue
 import java.util.concurrent.ThreadLocalRandom
+import java.util.{Collections, PriorityQueue}
 
 import com.google.common.collect.HashBiMap
 import com.typesafe.scalalogging.StrictLogging
 import org.apache.arrow.memory.BufferAllocator
-import org.apache.arrow.vector.complex.StructVector
+import org.apache.arrow.vector.complex.{ListVector, StructVector}
 import org.apache.arrow.vector.dictionary.DictionaryProvider.MapDictionaryProvider
 import org.apache.arrow.vector.ipc.ArrowStreamWriter
 import org.apache.arrow.vector.ipc.message.IpcOption
@@ -25,12 +25,11 @@ import org.apache.arrow.vector.util.TransferPair
 import org.apache.arrow.vector.{FieldVector, IntVector}
 import org.locationtech.geomesa.arrow.ArrowAllocator
 import org.locationtech.geomesa.arrow.io.records.{RecordBatchLoader, RecordBatchUnloader}
-import org.locationtech.geomesa.arrow.vector.ArrowAttributeReader.ArrowDateReader
+import org.locationtech.geomesa.arrow.vector.ArrowAttributeReader.{ArrowDateReader, ArrowDictionaryReader, ArrowListDictionaryReader}
 import org.locationtech.geomesa.arrow.vector.SimpleFeatureVector.SimpleFeatureEncoding
 import org.locationtech.geomesa.arrow.vector._
-import org.locationtech.geomesa.features.serialization.ObjectType
 import org.locationtech.geomesa.utils.collection.CloseableIterator
-import org.locationtech.geomesa.utils.geotools.{SimpleFeatureOrdering, SimpleFeatureTypes}
+import org.locationtech.geomesa.utils.geotools.{AttributeOrdering, ObjectType, SimpleFeatureOrdering, SimpleFeatureTypes}
 import org.locationtech.geomesa.utils.index.ByteArrays
 import org.locationtech.geomesa.utils.io.CloseWithLogging
 import org.locationtech.jts.geom.Geometry
@@ -61,6 +60,7 @@ class DeltaWriter(
   ) extends Closeable with StrictLogging {
 
   import DeltaWriter._
+  import org.locationtech.geomesa.utils.geotools.RichAttributeDescriptors.RichAttributeDescriptor
 
   import scala.collection.JavaConverters._
 
@@ -74,37 +74,41 @@ class DeltaWriter(
 
   private val vector = StructVector.empty(sft.getTypeName, allocator)
 
-  private val ordering = sort.map { case (field, reverse) =>
-    val o = SimpleFeatureOrdering(sft.indexOf(field))
-    if (reverse) { o.reverse } else { o }
-  }
+  private val ordering = sort.map { case (field, reverse) => SimpleFeatureOrdering(sft, field, reverse) }
 
   private val idWriter = ArrowAttributeWriter.id(sft, encoding, vector)
 
   private val writers = sft.getAttributeDescriptors.asScala.map { descriptor =>
     val name = descriptor.getLocalName
+    val isList = descriptor.isList
     val bindings = ObjectType.selectType(descriptor)
     val metadata = Map(SimpleFeatureVector.DescriptorKey -> SimpleFeatureTypes.encodeDescriptor(sft, descriptor))
-
     if (dictionaryFields.contains(name)) {
-      val dictMetadata = Map(SimpleFeatureVector.DescriptorKey -> s"$name:Integer")
-      val attribute = ArrowAttributeWriter(name, Seq(ObjectType.INT), None, dictMetadata, encoding, VectorFactory(vector))
-      val dictionary = {
-        val attribute = ArrowAttributeWriter(name, bindings, None, metadata, encoding, VectorFactory(allocator))
-        val writer = new BatchWriter(attribute.vector, ipcOpts)
-        attribute.vector.setInitialCapacity(initialCapacity)
-        attribute.vector.allocateNew()
-        Some(DictionaryWriter(sft.indexOf(name), attribute, writer, scala.collection.mutable.Map.empty))
+      // for list types, we encode the list items, not the whole list
+      val attribute = {
+        val desc = if (isList) { s"$name:List[Integer]" } else { s"$name:Integer" }
+        val encodedMetadata = Map(SimpleFeatureVector.DescriptorKey -> desc)
+        val encodedBindings = if (isList) { Seq(ObjectType.LIST, ObjectType.INT) } else { Seq(ObjectType.INT) }
+        ArrowAttributeWriter(name, encodedBindings, None, encodedMetadata, encoding, VectorFactory(vector))
       }
-      FieldWriter(name, sft.indexOf(name), attribute, dictionary)
+      val dictBindings = if (isList) { bindings.tail } else { bindings }
+      val ordering = AttributeOrdering(dictBindings)
+      val dict = ArrowAttributeWriter(name, dictBindings, None, metadata, encoding, VectorFactory(allocator))
+      dict.vector.setInitialCapacity(initialCapacity)
+      dict.vector.allocateNew()
+      if (isList) {
+        new DictionaryListFieldWriter(sft.indexOf(name), attribute, dict, ordering, ipcOpts)
+      } else {
+        new DictionaryFieldWriter(sft.indexOf(name), attribute, dict, ordering, ipcOpts)
+      }
     } else {
       val attribute = ArrowAttributeWriter(name, bindings, None, metadata, encoding, VectorFactory(vector))
-      FieldWriter(name, sft.indexOf(name), attribute, None)
+      new AttributeFieldWriter(sft.indexOf(name), attribute)
     }
   }
 
   // writer per-dictionary
-  private val dictionaryWriters = dictionaryFields.map(f => writers.find(_.name == f).get.dictionary.get)
+  private val dictionaryWriters = writers.collect { case d: DictionaryFieldWriter => d }
 
   // single writer to write out all vectors at once (not including dictionaries)
   private val writer = new BatchWriter(vector, ipcOpts)
@@ -120,7 +124,7 @@ class DeltaWriter(
     val last = threadingKey
     threadingKey = math.abs(ThreadLocalRandom.current().nextLong)
     logger.trace(s"$last resetting to $threadingKey")
-    writers.foreach(writer => writer.dictionary.foreach(_.values.clear()))
+    dictionaryWriters.foreach(_.resetDictionary())
   }
 
   /**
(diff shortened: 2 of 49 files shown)


geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
```diff
@@ -191,4 +196,12 @@ object ArrowFilterOptimizer extends LazyLogging {
       arrow.getReader(i).asInstanceOf[ArrowDictionaryReader].getEncoded(arrow.getIndex) == value
     }
   }
+
+  case class ArrowListDictionaryEquals(i: Int, value: Int) extends Filter {
```
jnh5y, line 200: (*Non* blocking question)

Is there a way to check for actual list equality?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -188,16 +175,12 @@ class DeltaWriter(
 
     // set attributes in the vector
```
jnh5y, line 176: Does this comment need to change?  (Proof that I read things somewhat...;))

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -213,22 +196,19 @@ class DeltaWriter(
...
   import scala.collection.JavaConverters._
 
   // empty provider
-  private val provider = new MapDictionaryProvider()
-
-  private val dictionaryOrdering: Ordering[AnyRef] = new Ordering[AnyRef] {
-    override def compare(x: AnyRef, y: AnyRef): Int =
-      SimpleFeatureOrdering.nullCompare(x.asInstanceOf[Comparable[Any]], y)
-  }
+  private val Provider = new MapDictionaryProvider()
```
jnh5y, line 211: Any particular reason for the capital P? (No need to change it.)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -289,16 +269,38 @@ object DeltaWriter extends StrictLogging {
           val name = fromVector.getField.getName
           val toVector = result.underlying.getChild(name)
           val transfer: (Int, Int, java.util.Map[Integer, Integer]) => Unit =
```
jnh5y, line 271: Just to make sure I'm understanding....

Dictionaries are paired up IntVectors.  So List[String] turns into a Dictionary + a ListVector[Int].  Further (assertion) since this is about transfers between dictionary-encoded columns, the only cases to worry with are IntVectors and List[Int]Vectors.

Is that right?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -572,21 +599,30 @@ object DeltaWriter extends StrictLogging {
...
       val descriptor = sft.getDescriptor(name)
-      val bindings = ObjectType.selectType(descriptor)
-      val metadata = Map(SimpleFeatureVector.DescriptorKey -> SimpleFeatureTypes.encodeDescriptor(sft, descriptor))
+      val bindings = {
+        val base = ObjectType.selectType(descriptor)
+        if (descriptor.isList) { base.tail } else { base }
+      }
+      val ordering = AttributeOrdering(bindings)
+      val desc = if (descriptor.isList) { s"$name:${descriptor.getListType().getSimpleName}" } else {
+        SimpleFeatureTypes.encodeDescriptor(sft, descriptor)
```
jnh5y, line 608: Should this logic just be rolled into the SFTs.encodeDescriptor?  Or is that function too general for this specific, Arrow-related use?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
```diff
@@ -832,14 +974,15 @@ object DeltaWriter extends StrictLogging {
   private class AttributeBatchMerger(
       vector: SimpleFeatureVector,
       transfers: Seq[(Int, Int) => Unit],
-      sort: ArrowAttributeReader
+      sort: ArrowAttributeReader,
+      ordering: Ordering[AnyRef]
     ) extends BatchMerger[AttributeBatchMerger](vector, transfers) {
 
-    private var value: Comparable[Any] = sort.apply(0).asInstanceOf[Comparable[Any]]
+    private var value: AnyRef = sort.apply(0)
```
jnh5y, line 981: The 'Comparable' drops out 'cause ....  List isn't comparable?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DictionaryBuildingWriter.scala
```diff
@@ -106,7 +108,12 @@ class DictionaryBuildingWriter(
 
     val dictionaries = attributeWriters.collect { case w: ArrowDictionaryWriter =>
       val name = s"dict-${w.dictionary.encoding.getId}"
-      val descriptor = SimpleFeatureTypes.renameDescriptor(sft.getDescriptor(w.vector.getField.getName), name)
+      val original = sft.getDescriptor(w.vector.getField.getName)
+      val descriptor = if (original.isList) {
```
jnh5y, line 112: Does this logic happen enough that it'd make sense to extract an ArrowSimpleFeatureTypes helper thingy?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
elahrvivaz: I think geotools supports that using the filter 'match action' of 'all', but I don't think that's supported through ECQL so effectively there isn't. Maybe there is a filter function, but I'm not aware of one. We could always add one, but again I don't think there is any way to specify a list in ECQL, other than through the default converters (e.g. we could convert the string `"a,b,c"` to `List("a", "b", "c")`)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
```diff
@@ -89,14 +93,21 @@ object SimpleFeatureArrowFileReader {
...
+      precision: SimpleFeatureEncoding): Map[String, ArrowDictionary] = {
     fields.flatMap { field =>
-      Option(field.getDictionary).toSeq.map { encoding =>
+      // check top-level dictionaries plus nested (i.e. for list-type attributes)
+      val encodings = Seq(field.getDictionary) ++ field.getChildren.asScala.map(_.getDictionary)
+      encodings.collect { case encoding if encoding != null =>
         val descriptor = SimpleFeatureTypes.createDescriptor(field.getMetadata.get(DescriptorKey))
+        val bindings = {
+          val main = ObjectType.selectType(descriptor)
+          if (descriptor.isList) { main.tail } else { main }
```
jnh5y, line 107: What's the type of main here?  What is its head in the case of isList==true?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: ah, no i think it's still valid. this foreach block is setting the attributes in the vector, then they get written out to bytes after that.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeReader.scala
```diff
@@ -150,22 +141,65 @@ object ArrowAttributeReader {
...
-        val dictionaryType = TypeBindings(bindings, encoding)
         vector match {
-          case v: TinyIntVector  => new ArrowDictionaryByteReader(v, dict, dictionaryType)
-          case v: SmallIntVector => new ArrowDictionaryShortReader(v, dict, dictionaryType)
-          case v: IntVector      => new ArrowDictionaryIntReader(v, dict, dictionaryType)
+          case v: TinyIntVector  => new ArrowDictionaryByteReader(v, dict)
+          case v: SmallIntVector => new ArrowDictionaryShortReader(v, dict)
+          case v: IntVector      => new ArrowDictionaryIntReader(v, dict)
+
+          case v: ListVector =>
```
jnh5y, line 149: Going back to my above question.  Is this covering the cases where one could have List[TinyInt/SmallInt] which are separate from Dictionary encoded situations?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: just scala convention for singleton object vals...

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: correct!

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DeltaWriter.scala
elahrvivaz: yep! that was the error thrown when you tried to dictionary encode a list-type attribute.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DictionaryBuildingWriter.scala
elahrvivaz: yeah, maybe... it's fairly simple and as I said in the other comment most places were slightly different. I could go either way, what do you think?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
elahrvivaz: for list types the result will be `Seq(LIST, STRING)` (or whatever the list item encoding is). so taking the tail is an easy way to get the item encoding as a seq.

elahrvivaz: for e.g. a string attribute main would be `Seq(STRING)`

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeReader.scala
elahrvivaz: right - if you know the number of dictionary entries up front, you can create a smaller encoding space. for the delta writer, since we don't know the sizes up front, we just encode everything as 32-bit ints. so the logic in that class doesn't have to deal with tiny/small ints.

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/GeoJsonSerializer.scala
```diff
@@ -88,6 +88,8 @@ class GeoJsonSerializer(sft: SimpleFeatureType) extends LazyLogging {
 
 object GeoJsonSerializer extends LazyLogging {
 
+  import org.locationtech.geomesa.utils.geotools.ObjectType
```
jnh5y, line 91: Why is this import needed?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -168,11 +168,8 @@ object RichAttributeDescriptors extends Conversions {
 
     private def tryClass(value: AnyRef): Class[_] = Try(Class.forName(value.asInstanceOf[String])).getOrElse(null)
 
-    def isList: Boolean = ad.getUserData.containsKey(UserDataListType)
```
jnh5y, line 171: Is this change going to mess with any AttributeDescriptor data that's serialized somewhere?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureOrdering.scala
```diff
@@ -10,64 +10,66 @@ package org.locationtech.geomesa.utils.geotools
...
 import org.locationtech.geomesa.utils.collection.TieredOrdering
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
-import org.opengis.filter.expression.PropertyName
 import org.opengis.filter.sort.{SortBy, SortOrder}
 
 import scala.math.Ordering
 
 /**
-  * Ordering for simple features. Assumes that any attributes implement `Comparable`
+  * Ordering for simple features
```
jnh5y, line 18: This change probably lines up perfectly with the change above between Comparable[Any] and AnyRef, right?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/GeoJsonSerializer.scala
elahrvivaz: i moved ObjectType out of the geomesa-features-kryo and into geomesa-utils so that i could use it in arrow

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
elahrvivaz: it should be functionally equivalent... it should be impossible to have a list-type attribute without the user-data-list-type key, and vice-versa. it just seemed more logical to check based on the attribute type.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureOrdering.scala
elahrvivaz: yeah, i switched to using an ordering in DeltaWriter so that we could compare non-comparable lists, and implemented the ordering here.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/AttributeOrdering.scala
```diff
@@ -0,0 +1,147 @@
...
+      Integer.compare(x.length, y.length)
+    }
+  }
+
+  private object ListOrdering {
+    def apply[T](delegate: Ordering[T]): Ordering[java.util.List[T]] = new NullOrdering(new ListOrdering(delegate))
+  }
+
+  private class ListOrdering[T](delegate: Ordering[T]) extends Ordering[java.util.List[T]] {
+    override def compare(x: java.util.List[T], y: java.util.List[T]): Int = {
```
elahrvivaz, line 122: @jnh5y what do you think about this as an ordering for lists? compare elements up to `min(x.length, y.length)`, then take the smaller list. there doesn't seem to be any canonical list ordering in java that i could find.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/filter/ArrowFilterOptimizer.scala
jnh5y: Yeah, I tried to look for use of the Match.ALL.  I believe you are correct!

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/DictionaryBuildingWriter.scala
jnh5y: Hmm....  I'm fine either way.  It isn't the end of the world if things are a little WET.  Just wanted to ask about it.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
jnh5y: Wow.  That's sufficiently clever that I'd love to see some comments added to explain that for future us.  Is there a quick place you could add a line or two about that?

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/GeoJsonSerializer.scala
jnh5y: Sorry, I should I mentioned that I was asking since the same import is at the top of the file.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
jnh5y: Cool!  Just making sure we weren't changing the 'API' in some sense.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
elahrvivaz: it's described here: https://github.com/locationtech/geomesa/blob/main/geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/ObjectType.scala#L40-L49

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/GeoJsonSerializer.scala
elahrvivaz: ah, I could probably remove the top-level one now. the old ObjectTypes (which is still there but deprecated) is in this same package, and it seemed to override the top-level import so i had to also add it here to get it to compile. let me try removing the other one.

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/AttributeOrdering.scala
jnh5y: Let's try this by an analogy...  If a string is a List[Char], it sounds like your list ordering would function the same as the ordering for String?  (More or less?)

Seems reasonable to me!

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/GeoJsonSerializer.scala
elahrvivaz: oh, the top-level on is the type ObjectType.ObjectType, while this one is the object ObjectType itself. they do both need to be there for it to compile. i moved the top-level one down here for consistency.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileReader.scala
elahrvivaz: added a comment

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/AttributeOrdering.scala
elahrvivaz: good point!

geomesa-features/geomesa-feature-common/src/main/scala/org/locationtech/geomesa/features/serialization/GeoJsonSerializer.scala
jnh5y: Ah!  *blinks*  I cannot read.  Thank you!

PR #2604, opened 2020-09-02
GEOMESA-2907 Remove Arrow Memory leak in ComplexArrowManualVisitor

jnh5y:
* Exposes ArrowAllocator.getAllocatedMemory and getPeakMemoryAllocation
methods.
* Addresses memory leak in ComplexArrowManualVisitor


geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileWriter.scala
```diff
@@ -24,6 +24,7 @@ import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
   * For writing simple features to an arrow file.
   *
   * Uses arrow streaming format (no footer).
+  * Closing the SimpleFeatureArrowFileWriter closes the given os.
```
jnh5y, line 27: @elahrvivaz do you believe this statement?  (I tried to trace through the indirection here, etc.)

geomesa-process/geomesa-process-vector/src/test/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcessTest.scala
```diff
@@ -42,20 +44,30 @@ class ArrowConversionProcessTest extends Specification {
...
       val bytes = process.execute(collection, null, null, null, null, null, null, null, null, null).reduce(_ ++ _)
       WithClose(SimpleFeatureArrowFileReader.streaming(() => new ByteArrayInputStream(bytes))) { reader =>
         reader.sft mustEqual sft
         SelfClosingIterator(reader.features()).map(ScalaSimpleFeature.copy).toSeq must
             containTheSameElementsAs(features)
       }
+      ArrowAllocator.getAllocatedMemory mustEqual 0
+    }
+
+    "encode a generic empty feature collection with dictionary values without leaking memory" in {
```
jnh5y, line 67: This case was the crux of this smaller, corner case memory leak.

geomesa-process/geomesa-process-vector/src/test/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcessTest.scala
```diff
@@ -42,20 +44,30 @@ class ArrowConversionProcessTest extends Specification {
...
+      // This returns an empty iterator.
+      process.execute(new ListFeatureCollection(sft), null, null, null, Seq("name"), null, null, null, null, null)
+      ArrowAllocator.getAllocatedMemory mustEqual 0
```
jnh5y, line 70: @elahrvivaz I'm curious if you can suggest places where we could test for this kind of allocation in unit tests where queries time out....

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileWriter.scala
elahrvivaz: tracing through it looks like yes, although there appears to be a small possibility that an exception during the close would cause it to not be invoked: https://github.com/apache/arrow/blob/master/java/vector/src/main/java/org/apache/arrow/vector/ipc/ArrowWriter.java#L202

jnh5y: Given that, would it be better to use a WithClose around the allocation of the outputstream as well?  Would closing the outstream multiple times cause trouble?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
```diff
@@ -262,25 +262,28 @@ object ArrowConversionProcess {
...
-          } else {
-            CloseWithLogging(writer)
-          }
-          out.toByteArray
+          writer.flush()
+          bytes.append(out.toByteArray)
+          out.reset()
         }
+
+        bytes.append(out.toByteArray)
```
elahrvivaz, line 285: i think these 2 lines (and the declaration of `bytes`) should be moved outside the WithClose block, so that any footer bytes written by the SimpleFeatureWriter get returned correctly. It might not matter if the footer is missing, but seems safer.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
```diff
@@ -262,25 +262,28 @@ object ArrowConversionProcess {
         case Some(o) => features.sorted(o).iterator
       }
 
+      buildResult(dictionaries, sorted)
+    }
+
+    private def buildResult(dictionaries: Map[String, ArrowDictionary], sorted: Iterator[SimpleFeature]) = {
```
elahrvivaz, line 268: could you add the return type to the method signature?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
jnh5y: Hmm... if ```bytes.append(out.toByteArray)``` is moved outside of the WithClose block, wouldn't ```out``` be closed by the WithClose?

geomesa-process/geomesa-process-vector/src/test/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcessTest.scala
```diff
@@ -42,20 +44,30 @@ class ArrowConversionProcessTest extends Specification {
 
   "ArrowConversionProcess" should {
     "encode an empty feature collection" in {
+      ArrowAllocator.getAllocatedMemory mustEqual 0
```
elahrvivaz, line 47: hmm, i don't think this is a safe check - even with `sequential`, the root allocator is shared across all unit tests, which often run in parallel. We could maybe modify the process method signature or constructor to pass in an optional allocator so that we can check it?

jnh5y: Good eye!

Would we need to thread the optional allocator through to the SimpleFeatureVector / Writer classes?  (Or is that mostly already plumbed?)

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/io/SimpleFeatureArrowFileWriter.scala
elahrvivaz: hmm, generally i don't think closing multiple times would be an issue, but not sure if that's 100% true. probably any exceptions during the close would be infrequent and not result in massive memory leaks so maybe it's ok as is.

elahrvivaz (approved):

geomesa-process/geomesa-process-vector/src/test/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcessTest.scala
```diff
@@ -58,6 +60,12 @@ class ArrowConversionProcessTest extends Specification {
       }
     }
 
+    "encode a generic empty feature collection with dictionary values without leaking memory" in {
+      // This returns an empty iterator.
+      process.execute(new ListFeatureCollection(sft), null, null, null, Seq("name"), null, null, null, null, null)
+      ArrowAllocator.getAllocatedMemory(sft.getTypeName) must equalTo(0)
```
elahrvivaz, line 66: we should make the type name more unique to avoid interference from any other tests

elahrvivaz (approved):

elahrvivaz:
fixed up as 42fe5a44751bfee9455d3267e937f5749ec0d175

PR #2608, opened 2020-09-04
GEOMESA-2910 Harden QueryPlanFilterVisitor against stackoverflows for…

jnh5y:
… leaning trees


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -56,7 +56,8 @@ class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends DuplicatingFilterVi
     val children = new java.util.ArrayList[Filter](f.getChildren.size)
     var i = 0
     while (i < f.getChildren.size) {
-      val child = f.getChildren.get(i).accept(this, data).asInstanceOf[Filter]
+      // Simplify the children first to avoid leaning trees patterns causing StackOverflows
+      val child = FilterHelper.simplify(f.getChildren.get(i)).accept(this, data).asInstanceOf[Filter]
```
elahrvivaz, line 60: could maybe just use `FilterHelper.flatten` on the original filter before passing into the filter visitor (i.e. in org.locationtech.geomesa.filter.visitor.QueryPlanFilterVisitor#apply). having it here will sometimes end up simplifying child filters multiple times, if for example a filter gets run through here and then it has a nested OR that gets run through again (i think).

elahrvivaz (approved):

PR #2609, opened 2020-09-04
GEOMESA-2777 Support negative numbers in the default table splitter

elahrvivaz:


jnh5y:
Any concerns about backwards-compatibility here?

elahrvivaz:
nope, it just adds some an additional feature but all the old patterns should still work (and the ones we have unit tested do, at least).

PR #2613, opened 2020-09-17
GEOMESA-1605 Temporal query guard

elahrvivaz:

### DIFF
--- a/docs/user/datastores/index_config.rst
+++ b/docs/user/datastores/index_config.rst
@@ -617,6 +617,24 @@ Interceptors will be invoked in the order they are declared in the user data. In
 on the results of query interceptors, you can enable ``TRACE``-level logging on the class
 ``org.locationtech.geomesa.index.planning.QueryRunner$``.
 
+Configuring Temporal Query Guards
+---------------------------------
+
+GeoMesa provides a basic mechanism for blocking overly broad queries (which can overwhelm the system). A
+maximum temporal duration can be specified, and any query which attempts to return a larger time period will be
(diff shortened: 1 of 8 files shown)


jahhulbert:
Seems like a good first pass to me. One thing I wonder though is if we should have it set per index like perhaps geomesa.filter.max.duration.z3 geomesa.filter.max.duration.z2 geomesa.filter.max.duration.attr so that we could specify something different per index but if course then you immediately get to something more complex like the real thing you want to do is this:

geomesa.filter.max.duration.attr = numAttrs match {
  case Many : 3 mo
  case Few: 6 mo
  case 1-5: unlimited
}

geomesa.filter.max.duration.z3 =  area (bbox) match {
   case Large: 1 week
   case Medium: 3 week
   case Small: 3 mon
   case Tiny: unlimited
}

You already did the part where it is per SFT so that is really good.

I think this will handle the basic case...but really that may be the best thing to do in the future is provide an SPI for this. @tkunicki and @jnh5y what do you think?

@elahrvivaz question: This will currently apply to the attr indexes, correct? I think that's the only main concern. I am OK limiting the Z3 but for the attrs I would have to allow longer queries.

elahrvivaz:
this won't affect the attribute index, just z3/xz3/s3.
the more complicated config is what @jnh5y wants to do - if you would use it, we can go that route. It does become more complicated to configure - any ideas there? we have jammed typesafe config into user data before, but that is kind of a PITA when you're defining sfts and have to escape everything.

elahrvivaz:
replaced with https://github.com/locationtech/geomesa/pull/2619

PR #2615, opened 2020-09-23
GEOMESA-2914 Provide simplified schema management

elahrvivaz:
* `GeoMesaDataStore.checkSchemaCompatibility` can be used to create
  and update schemas with complicated checks

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
@@ -33,7 +33,6 @@ import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes.{AttributeOpti
 import org.locationtech.geomesa.utils.geotools.converters.FastConverter
 import org.locationtech.geomesa.utils.index.IndexMode
 import org.locationtech.geomesa.utils.io.CloseWithLogging
-import org.locationtech.geomesa.utils.stats.IndexCoverage
 import org.opengis.feature.`type`.AttributeDescriptor
 import org.opengis.feature.simple.SimpleFeatureType
 import org.opengis.filter.Filter
@@ -506,7 +505,7 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
   def checkSchemaCompatibility(typeName: String, sft: SimpleFeatureType): SchemaCompatibility = {
     val previous = getSchema(typeName)
     if (previous == null) {
-      new SchemaCompatibility.Missing(this, sft)
+      new SchemaCompatibility.DoesNotExist(this, sft)
     } else {
       Try(validateSchemaUpdate(previous, sft)).failed.map(SchemaCompatibility.Incompatible).getOrElse {
         val copy = SimpleFeatureTypes.copy(sft)
@@ -679,12 +678,12 @@ object GeoMesaDataStore extends LazyLogging {
      * @param ds data store
      * @param sft schema
      */
-    class Missing(ds: GeoMesaDataStore[_], val sft: SimpleFeatureType) extends SchemaCompatibility {
+    class DoesNotExist(ds: GeoMesaDataStore[_], val sft: SimpleFeatureType) extends SchemaCompatibility {
       override def apply(): Unit = ds.createSchema(sft)
     }
 
-    object Missing {
-      def unapply(arg: Missing): Option[SimpleFeatureType] = Some(arg.sft)
+    object DoesNotExist {
+      def unapply(arg: DoesNotExist): Option[SimpleFeatureType] = Some(arg.sft)
     }
 
     /**
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
@@ -12,17 +12,16 @@ import java.io.IOException
 import java.util.Collections
 import java.util.concurrent.TimeUnit
 
-import com.github.benmanes.caffeine.cache.{CacheLoader, Caffeine}
+import com.github.benmanes.caffeine.cache.{AsyncCacheLoader, CacheLoader, Caffeine}
 import com.typesafe.scalalogging.LazyLogging
 import org.geotools.data._
 import org.locationtech.geomesa.index.FlushableFeatureWriter
 import org.locationtech.geomesa.index.api.{IndexManager, _}
 import org.locationtech.geomesa.index.conf.partition.TablePartition
-import org.locationtech.geomesa.index.geotools.GeoMesaDataStore.VersionKey
+import org.locationtech.geomesa.index.geotools.GeoMesaDataStore.{SchemaCompatibility, VersionKey}
 import org.locationtech.geomesa.index.geotools.GeoMesaDataStoreFactory.GeoMesaDataStoreConfig
 import org.locationtech.geomesa.index.index.attribute.AttributeIndex
 import org.locationtech.geomesa.index.index.id.IdIndex
-import org.locationtech.geomesa.index.metadata.GeoMesaMetadata.AttributesKey
 import org.locationtech.geomesa.index.planning.QueryPlanner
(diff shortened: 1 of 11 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
```diff
@@ -153,4 +156,22 @@ trait Runner extends LazyLogging {
...
 }
 
-case class AutocompleteInfo(path: String, commandName: String)
+object Runner {
+
+  private var added = false
+
+  /**
+   * Add jcommander converter factories. Note that the implementation is a static,
+   * shared, non-thread-safe list, so we have to synchronize ourselves.
```
jnh5y, line 165: Ok, how did you run into this?

elahrvivaz: CI kept failing on the export test... dug into it and found this.

elahrvivaz: in normal use i don't think it would ever happen since we only load 1 runner at a time.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/MetadataBackedDataStore.scala
```diff
@@ -407,5 +433,14 @@ object MetadataBackedDataStore {
...
-    Set(TableSharing, TableSharingPrefix, IndexVisibilityLevel, IndexZ3Interval, IndexXzPrecision)
+    Set(
+      TableSharing,
+      TableSharingPrefix,
+      IndexVisibilityLevel,
+      IndexZ3Interval,
+      IndexXzPrecision,
+      IndexZShards,
+      IndexIdShards,
+      IndexAttributeShards
```
jnh5y, line 444: Should one be able to change:
1. DefaultDtgField?
2. FidsAreUuidEncoded?
3. S3_INTERVAL_KEY?
4. TableCompressionType?
5. TableLogicalTime?
6. TableSharing?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
```diff
@@ -217,6 +220,57 @@ class GeoMesaDataStoreTest extends Specification {
         ds.stats.getMinMax[String](sft, "n", exact = false).map(_.max) must beSome("name0")
       }
     }
+    "update compatible schemas from typesafe config changes" in {
```
jnh5y, line 223: Ok, could we add tests for:
1.  Showing that one cannot remove a field?
2.  Showing that an UnmodifiableUserDataKeys cannot be changed?
And maybe optional.... check the conditions on geometry field / column type changes?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -162,34 +163,13 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
...
-
+    updateSchemaUserData(sft, previous)
     // try to create the new indices to ensure they are valid for the sft
-    val previousIndices = previous.getIndices.map(GeoMesaFeatureIndex.identifier)
-    val newIndices = sft.getIndices.filterNot(i => previousIndices.contains(GeoMesaFeatureIndex.identifier(i)))
-    if (newIndices.nonEmpty) {
-      try { GeoMesaFeatureIndexFactory.create(this, sft, newIndices) } catch {
-        case NonFatal(e) => throw new IllegalArgumentException(s"Error configuring new feature index:", e)
-      }
+    try { GeoMesaFeatureIndexFactory.create(this, sft, sft.getIndices) } catch {
```
jnh5y, line 168: This seems like it'll call create on existing indexes after this change.  Is that operation a no-op if the tables exist?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/MetadataBackedDataStore.scala
elahrvivaz: 1. dtg is checked here (it can be changed if the field is renamed): https://github.com/locationtech/geomesa/pull/2615/files#diff-67b90d2402343d0802fec2ee9a973023R378
2. this one is ok to update
3. no, i'll add that one
3. that should be ok... it's only used in create schema and it's transparent to us
4. also should be ok
5. in the list already

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
elahrvivaz: this just creates the configured classes in memory (i.e. Z3Index), the actual tables are created later on here: https://github.com/locationtech/geomesa/pull/2615/files/e031f94d0c1cea62f8970d53c0a8e0f5d1a0fb35#diff-06656c6f4398571e1b268f89abc06664R223
(although calling create table is supposed to be coded as a no-op if the table already exists).

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -586,6 +655,64 @@ object GeoMesaDataStore extends LazyLogging {
...
+     * @param typeName type name
+     * @param update the updated schema with all appropriate metadata
+     */
+    class Compatible(ds: GeoMesaDataStore[_], val typeName: String, val update: SimpleFeatureType)
+        extends SchemaCompatibility {
+      override def apply(): Unit = ds.updateSchema(typeName, update)
+    }
+
+    object Compatible {
+      def unapply(arg: Compatible): Option[(String, SimpleFeatureType)] = Some(arg.typeName, arg.update)
```
jnh5y, line 703: Hehehe, I find the function name 'unapply' amusing here.  (I know it is to have a destructor or whatever, but still...:))

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
elahrvivaz: this call ensures that they are SPI loadable.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -586,6 +655,64 @@ object GeoMesaDataStore extends LazyLogging {
...
+      override def apply(): Unit = {}
+    }
+
+    /**
+     * Indicates that the schema does not exist in the data store
+     *
+     * @param ds data store
+     * @param sft schema
+     */
+    class Missing(ds: GeoMesaDataStore[_], val sft: SimpleFeatureType) extends SchemaCompatibility {
```
jnh5y, line 682: Would calling this something like "NewSchema" be ok with you? 

When I was reading the unit test, I thought it was something about missing data.

elahrvivaz: how about MissingSchema or DoesNotExist? this is supposed to represent a status, so NewSchema is more of the action you'd take than the status.

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
elahrvivaz: i think those are covered already by https://github.com/locationtech/geomesa/blob/main/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala#L349 (the schema tests are somewhat scattered around the codebase...)

elahrvivaz:
@jnh5y i think i've addressed all the comments, thanks for reviewing.

PR #2616, opened 2020-09-28
GEOMESA-2915 Fix classloading of property accessors in Accumulo

elahrvivaz:


jnh5y:
Was this happening on Accumulo 1.9 and 2.0?

jnh5y:
Any chance this would impact HBase as well?

elahrvivaz:
I noticed it on accumulo 1.9 - didn't test 2.0. i don't think it affects hbase since we've never seen it there, but either way it should use the same code paths for IteratorCache.
The issue is that in geotools PropertyAccessors it caches the results of the first time it's called - so if it's called outside our classloading mojo our accessors never get loaded. i think the bug may end up being somewhat inconsistent depending on which code path is executed first. at any rate, i believe all access goes through the iterator cache so hopefully that will fix it...

PR #2619, opened 2020-09-30
GEOMESA-1605 Temporal query guard

elahrvivaz:
* Add guard method to QueryInterceptor lifecycle

### DIFF
--- a/docs/user/datastores/index_config.rst
+++ b/docs/user/datastores/index_config.rst
@@ -618,8 +618,8 @@ Interceptors must have a default, no-arg constructor. The interceptor lifecycle 
 
 1. The instance is instantiated via reflection, using its default constructor
 #. The instance is initialized via the ``init`` method, passing in the data store containing the simple feature type
-#. ``rewrite`` and ``guard`` are called repeatedly, once for each query
-#. The instance is cleaned up via the ``close`` method
+#. ``rewrite`` and ``guard`` are called once for each query
+#. When the data store is disposed, the instance is cleaned up via the ``close`` method
 
 Interceptors will be invoked in the order they are declared in the user data. In order to see detailed information
 on the results of query interceptors, you can enable ``TRACE``-level logging on the class
--- a/docs/user/datastores/index_config.rst
+++ b/docs/user/datastores/index_config.rst
@@ -633,9 +633,8 @@ GeoMesa provides a basic implementation of a query interceptor that will block o
 to return a larger time period will be stopped. This will not affect queries against indices that do not have
 a temporal component (for example, feature ID and attribute queries).
 
-To enable the guard, set ``geomesa.query.interceptors`` to
-``org.locationtech.geomesa.index.planning.QueryInterceptor$TemporalQueryGuard`` and set the duration using
-``geomesa.filter.max.duration``:
+To enable the guard, add ``org.locationtech.geomesa.index.planning.QueryInterceptor$TemporalQueryGuard``
+to ``geomesa.query.interceptors`` as indicated above, and set the duration using ``geomesa.filter.max.duration``:
 
 .. code-block:: java
 
--- a/docs/user/datastores/index_config.rst
+++ b/docs/user/datastores/index_config.rst
@@ -604,18 +604,50 @@ the trait ``org.locationtech.geomesa.index.planning.QueryInterceptor``:
         * @param query query
         */
       def rewrite(query: Query): Unit
+
+      /**
+       * Hook to allow interception of a query after extracting the query values
+       *
+       * @param filter full query filter
+       * @param values index values extracted from the query, will vary by index. E.g. for the
+       *               Z3 index will contain extracted time intervals and bounding boxes.
+       *               See `org.locationtech.geomesa.index.api.IndexKeySpace#getIndexValues()`
+       * @return
+       */
+      def guard(filter: Filter, values: Option[_]): Option[IllegalArgumentException] = None
     }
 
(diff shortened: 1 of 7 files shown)


docs/user/datastores/index_config.rst
```diff
@@ -604,19 +604,48 @@ the trait ``org.locationtech.geomesa.index.planning.QueryInterceptor``:
...
+       */
+      def guard(strategy: QueryStrategy): Option[IllegalArgumentException] = None
     }
 
 Interceptors must have a default, no-arg constructor. The interceptor lifecycle consists of:
 
 1. The instance is instantiated via reflection, using its default constructor
 #. The instance is initialized via the ``init`` method, passing in the data store containing the simple feature type
-#. ``rewrite`` is called repeatedly
+#. ``rewrite`` and ``guard`` are called repeatedly, once for each query
```
jnh5y, line 621: I think it'd be clearer as ....

 ``rewrite`` and ``guard`` are called once for each query

Maybe that's not capturing the idea that the methods are called multiple times after the implementation is instantiated though?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/z3/package.scala
```diff
@@ -25,17 +25,21 @@ package object z3 {
...
-                            intervals: FilterValues[Bounds[ZonedDateTime]],
-                            temporalBounds: Map[Short, (Double, Double)],
-                            temporalUnbounded: Seq[(Short, Short)])
+  case class XZ3IndexValues(
+      sfc: XZ3SFC,
+      geometries: FilterValues[Geometry],
+      spatialBounds: Seq[(Double, Double, Double, Double)],
+      intervals: FilterValues[Bounds[ZonedDateTime]],
+      temporalBounds: Map[Short, (Double, Double)],
+      temporalUnbounded: Seq[(Short, Short)]
```
jnh5y, line 43: What does "temporalUnbounded" mean?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
```diff
@@ -40,6 +46,14 @@ trait QueryInterceptor extends Closeable {
...
     */
   def rewrite(query: Query): Unit
+
+  /**
+   * Hook to allow interception of a query after extracting the query values
+   *
+   * @param strategy query strategy
+   * @return an exception if the query should be stopped
+   */
+  def guard(strategy: QueryStrategy): Option[IllegalArgumentException] = None
```
jnh5y, line 56: If guard has a default implementation, should rewrite have one as well?

Or is the default implementation added to improve compatibility with existing implementations of the QueryInterceptor interface?

jnh5y, line 56: Separately, I wonder if the QueryGuard should be its own interface?

elahrvivaz: right, just for compatibility

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/z3/package.scala
elahrvivaz: those are epochs that don't have any temporal endpoints, e.g. if you query `dtg > today` you'd end up with `(today's epoch, Short.MaxValue)`. that way we can scan just with a start (or end) row, instead of trying to calculate z-values for every single epoch. this is mainly used for secondary attribute index scans, and was added in https://geomesa.atlassian.net/browse/GEOMESA-2402

docs/user/datastores/index_config.rst
elahrvivaz: changed to your wording, and noted that close is called when the datastore is disposed

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/z3/package.scala
elahrvivaz: i'll add some scala docs

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
elahrvivaz: seems easier to combine them since we already have all the configuration wired up and they are very related.

docs/user/datastores/index_config.rst
```diff
@@ -604,19 +604,48 @@ the trait ``org.locationtech.geomesa.index.planning.QueryInterceptor``:
...
 
+Temporal Query Guard
+++++++++++++++++++++
+
+GeoMesa provides a basic implementation of a query interceptor that will block overly broad queries
+(which can overwhelm the system). A maximum temporal duration must be specified, and any query which attempts
+to return a larger time period will be stopped. This will not affect queries against indices that do not have
+a temporal component (for example, feature ID and attribute queries).
+
+To enable the guard, set ``geomesa.query.interceptors`` to
```
jnh5y, line 636: This setting is a comma-separated list of classes, right?  

Would it be more precise to state that one ought to be adding this to the list of interceptors?

(I'm being very precise to make sure I understand.  This may be adding complexity to the docs, so feel free to push back.)

elahrvivaz: right, i will clarify.

PR #2621, opened 2020-10-05
GEOMESA-2917 Docs for NiFi record-based processors

elahrvivaz:

### DIFF
--- a/docs/user/nifi.rst
+++ b/docs/user/nifi.rst
@@ -390,6 +390,10 @@ and output them as NiFi records for further processing.
 +-------------------------------+---------------------------------------------------------------------------------------+
 | Geometry Serialization Format | The format to use for serializing geometries, either text or binary                   |
 +-------------------------------+---------------------------------------------------------------------------------------+
+| Include Visibilities          | Include a column with visibility expressions for each row                             |
++-------------------------------+---------------------------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


docs/user/nifi.rst
```diff
@@ -192,21 +245,60 @@ are available:
...
++-----------------------------------+-----------------------------------------------------------------------------------------------------+
+| ``Feature type name``             | Name to use for the simple feature type schema. If not specified, will use the name                 |
+|                                   | from the record schema                                                                              |
++-----------------------------------+-----------------------------------------------------------------------------------------------------+
+| ``Feature ID column``             | Column that will be used as the feature ID. If not specified, a random ID will be used              |
++-----------------------------------+-----------------------------------------------------------------------------------------------------+
+| ``Geometry columns``              | Column(s) that will be deserialized as geometries, and their type, as a                             |
+|                                   | SimpleFeatureType specification string (e.g. ``the_geom:Point``). A '*' can be used to                   |
```
jnh5y, line 266: Since you say "can" be used, what are the other options? 

If one uses just the column names, what happens?

elahrvivaz: it would be the same as not specifying a default geom in a simple feature type... i think it uses the first one in that case? i can check and add a clarification if you think it's worthwhile

jnh5y: Eh, I think it is worth us being clear.  Otherwise, future us will have to read the code;).

PR #2624, opened 2020-10-06
GEOMESA-2921 Implement a graduated spatial-temporal query guard

jnh5y:

### DIFF
--- a/docs/user/datastores/query_interceptor.rst
+++ b/docs/user/datastores/query_interceptor.rst
@@ -81,12 +81,11 @@ to ``geomesa.query.interceptors`` as indicated above, and set the duration using
     sft.getUserData().put("geomesa.guard.temporal.max.duration", "28 days");
 
 Graduated Query Guard
-++++++++++++++++++++
++++++++++++++++++++++
 
 The graduated query guard applies different duration limits based on the spatial extent of the query.
 As a query becomes larger in space, it can be limited to shorter and shorter time ranges.
-The spatial extent of a query is measured in square degrees.  A series of rules limit the duration for queries
-which are at most a given size in square degrees.
+A series of rules limit the duration for queries which are at most a given size in square degrees.
 
 To enable the guard, add ``org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard``
 to ``geomesa.query.interceptors`` as indicated above.  Configuration is managed via TypeSafe config.
--- a/docs/user/datastores/query_interceptor.rst
+++ b/docs/user/datastores/query_interceptor.rst
@@ -88,7 +88,12 @@ As a query becomes larger in space, it can be limited to shorter and shorter tim
 A series of rules limit the duration for queries which are at most a given size in square degrees.
 
 To enable the guard, add ``org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard``
-to ``geomesa.query.interceptors`` as indicated above.  Configuration is managed via TypeSafe config.
+to ``geomesa.query.interceptors`` as indicated above.  Configuration is managed via
+`TypeSafe Config <https://github.com/lightbend/config>`_ which will look for files named
+``reference.conf`` and/or ``application.conf`` on the classpath.
+For use in GeoServer, a file name ``reference.conf`` can be added to ``WEB-INF/classes``.
+The configuration is under the key ``geomesa.guard.graduated``.
+
 The configuration must satisfy a few conditions:
 
 * there must be a limit on unbounded queries,
@@ -110,5 +115,3 @@ An example is given here.  Durations can be given in a number of days, hours, or
         }
       }
     }
-
-For use in GeoServer, a file name ``reference.conf`` can be added to ``WEB-INF/classes``.
--- a/docs/user/datastores/query_interceptor.rst
+++ b/docs/user/datastores/query_interceptor.rst
@@ -1,0 +1,115 @@
+.. _query_interceptors:
+
+Query Interceptors and Guards
+=============================
+
+GeoMesa provides a chance for custom logic to be applied to a query before executing it. Query interceptors must
+be specified through user data in the simple feature type, and may be set before calling ``createSchema``, or
+updated by calling ``updateSchema``. To indicate query interceptors, use the key ``geomesa.query.interceptors``:
+
+.. code-block:: java
+
+    sft.getUserData().put("geomesa.query.interceptors", "com.example.MyQueryInterceptor");
+
+The value must be a comma-separated string consisting of the names of one or more classes implementing
+the trait ``org.locationtech.geomesa.index.planning.QueryInterceptor``:
+
+.. code-block:: scala
+
+    /**
+      * Provides a hook to modify a query before executing it
+      */
+    trait QueryInterceptor extends Closeable {
+
+      /**
+        * Called exactly once after the interceptor is instantiated
+        *
+        * @param ds data store
+        * @param sft simple feature type
+        */
+      def init(ds: DataStore, sft: SimpleFeatureType): Unit
+
+      /**
+        * Modifies the query in place
+        *
+        * @param query query
+        */
+      def rewrite(query: Query): Unit
+
+      /**
+       * Hook to allow interception of a query after extracting the query values
+       *
+       * @param strategy query strategy
+       * @return an exception if the query should be stopped
+       */
+      def guard(strategy: QueryStrategy): Option[IllegalArgumentException] = None
+    }
+
+Interceptors must have a default, no-arg constructor. The interceptor lifecycle consists of:
+
+1. The instance is instantiated via reflection, using its default constructor
+#. The instance is initialized via the ``init`` method, passing in the data store containing the simple feature type
+#. ``rewrite`` and ``guard`` are called once for each query
+#. When the data store is disposed, the instance is cleaned up via the ``close`` method
+
+Interceptors will be invoked in the order they are declared in the user data. In order to see detailed information
+on the results of query interceptors, you can enable ``TRACE``-level logging on the class
+``org.locationtech.geomesa.index.planning.QueryRunner$``.
+
+Provided Query Guards
+^^^^^^^^^^^^^^^^^^^^^
+
+GeoMesa provides some basic query guards that will block overly broad queries (which can overwhelm the system).
+For additional controls, see ``geomesa.query.timeout`` and ``geomesa.scan.block-full-table`` in
+:ref:`geomesa_site_xml`.
+
+Temporal Query Guard
+++++++++++++++++++++
+
+The temporal query guard will block queries which exceed a maximum temporal duration.
+Any query which attempts to return a larger time period will be stopped. This will not affect queries against indices that do not have
+a temporal component (for example, feature ID and attribute queries), nor queries that do not have any temporal
+component at all (which can usually be blocked by ``geomesa.scan.block-full-table``).
+
+To enable the guard, add ``org.locationtech.geomesa.index.planning.guard.TemporalQueryGuard``
+to ``geomesa.query.interceptors`` as indicated above, and set the duration using ``geomesa.guard.temporal.max.duration``:
+
+.. code-block:: java
+
+    sft.getUserData().put("geomesa.query.interceptors",
+      "org.locationtech.geomesa.index.planning.guard.TemporalQueryGuard");
+    sft.getUserData().put("geomesa.guard.temporal.max.duration", "28 days");
+
+Graduated Query Guard
+++++++++++++++++++++
+
+The graduated query guard applies different duration limits based on the spatial extent of the query.
+As a query becomes larger in space, it can be limited to shorter and shorter time ranges.
+The spatial extent of a query is measured in square degrees.  A series of rules limit the duration for queries
+which are at most a given size in square degrees.
+
+To enable the guard, add ``org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard``
+to ``geomesa.query.interceptors`` as indicated above.  Configuration is managed via TypeSafe config.
+The configuration must satisfy a few conditions:
+
+* there must be a limit on unbounded queries,
+* as the size increases, the duration must decrease,
+* and a given size limit may not be repeated.
+
+An example is given here.  Durations can be given in a number of days, hours, or minutes.
+
+.. code-block:: none
+
+    geomesa {
+      guard {
+        graduated {
+          "sftName" = [
+            { size = 1,  duration = "60 days" }
+            { size = 10, duration = "3 days"  }
+            { duration = "1 day" }
+          ]
+        }
+      }
+    }
+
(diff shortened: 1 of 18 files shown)


jnh5y:
TODOs:
1.  Refactor two query guards out
2.  Test in a GeoServer
3.  Add documentation

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/z2/package.scala
```diff
@@ -16,9 +16,9 @@ package object z2 {
 
   case class Z2IndexValues(sfc: Z2SFC,
                            geometries: FilterValues[Geometry],
-                           bounds: Seq[(Double, Double, Double, Double)])
+                           spatialBounds: Seq[(Double, Double, Double, Double)]) extends SpatialIndexValues
 
   case class XZ2IndexValues(sfc: XZ2SFC,
                             geometries: FilterValues[Geometry],
-                            bounds: Seq[(Double, Double, Double, Double)])
+                            spatialBounds: Seq[(Double, Double, Double, Double)]) extends SpatialIndexValues
```
elahrvivaz, line 23: not sure if this is in the 'public api', but we could leave the `bounds` the same, mark it deprecated, and add in an `override def spatialBounds = bounds` method to fulfill the new SpatialIndexValues trait

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
```diff
@@ -160,14 +164,135 @@ object QueryInterceptor extends LazyLogging {
...
+     * 1. Sizes must not be repeated.
+     * 2. Smaller sizes must allow for longer durations.
+     * @param limits Sequence of limits
+     * @return Sorted list of limits or throws an IllegalArgumentException
+     */
+    def evaluateLimits(limits: Seq[SizeAndDuration]): Seq[SizeAndDuration] = {
+      val candidate: Seq[SizeAndDuration] = limits.sortBy(_.sizeLimit)
+      candidate.sliding(2).foreach { seq =>
+        val first = seq.head
+        val second = seq.tail.head
```
elahrvivaz, line 207: might be slightly nicer to pattern match like `.foreach { case Seq(first, second) =>`
one other gotcha is that if the list only has one element, `sliding(2)` will give you a list with one element in it...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
```diff
@@ -160,14 +164,135 @@ object QueryInterceptor extends LazyLogging {
...
+      val spatialExtent = spatialBoundsOption match {
+        case None => 360.0 * 180.0  // Whole world
+        case Some(seq) => seq.map { case  (lx: Double, ly: Double, ux: Double, uy: Double) =>
+          (ux-lx) * (uy-ly)
+        }.sum
+      }
+
+      intervalsOption match {
+        case None => // Q: What does None here mean again?!
+          None       // Returning None means that the query continues.  Is this correct?
```
elahrvivaz, line 272: right

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
```diff
@@ -160,14 +164,135 @@ object QueryInterceptor extends LazyLogging {
...
+
+      val spatialExtent = spatialBoundsOption match {
+        case None => 360.0 * 180.0  // Whole world
+        case Some(seq) => seq.map { case  (lx: Double, ly: Double, ux: Double, uy: Double) =>
+          (ux-lx) * (uy-ly)
+        }.sum
+      }
+
+      intervalsOption match {
+        case None => // Q: What does None here mean again?!
```
elahrvivaz, line 271: we might want to re-evaluate the `strategy.values.collect { case v: TemporalIndexValues => v.intervals }`, b/c it makes None ambiguous - it could mean that the index doesn't have a temporal component (the collect doesn't match), or that the temporal component is empty (v.intervals is None). those should probably be handled differently...

elahrvivaz, line 271: deleted my previous comment... it is ambiguous b/c for full table scans, strategy.values will be None. So it means that either: it was not against a temporal index, or: it was against a temporal index, but was a full table scan. Since that ambiguity can't be easily resolved without refactoring our query plans, we probably would need to also configure block full table scans to catch that case.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
```diff
@@ -160,14 +164,135 @@ object QueryInterceptor extends LazyLogging {
...
+          None       // Returning None means that the query continues.  Is this correct?
+        case Some(intervals) =>
+          if (!intervals.forall(_.isBoundedBothSides) || intervals.isEmpty) {
+            Some(new IllegalArgumentException("At least one part of the query is temporally unbounded.  Add bounded temporal constraints."))
+          } else {
+            val queryDuration = duration(intervals.values)
+            guardLimits.find( _.sizeLimit > spatialExtent).flatMap {
```
elahrvivaz, line 278: should this be `>=`? also, seems like there might not be a matching limit, in that case the query would go through, but seems like it shouldn't? since you would have configured a limit at a smaller size. (although also if the config is wrong the guards would be empty...)

elahrvivaz: i was thinking `>=` would be needed to match full world scans

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/z2/package.scala
jnh5y: Good catch.  Just pushed a fix for this.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
jnh5y: I like the pattern, and was thinking about the first time through; just didn't bother to figure out the syntax:).  I've added a little size check to handle the one limit case.  Maybe that case is degenerate and should throw some kind of error...  Thoughts?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
jnh5y: Alright, this is complex enough and I'm forgetful enough that I'm writing a comment to note this information..

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
elahrvivaz: i guess you were kind of assuming that there would be a config for the upper limit whole world bounding box, but it doesn't seem required... that probably breaks the logic in the query guard, as it would return None and let the query through (where you probably would expect it to fail?). so we should probably enforce something in the initialization or modify the logic in the guard.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryInterceptor.scala
jnh5y: You are correct.  I've added unit test cases to help with the corner cases.
The size of whole world guards is Max.Int:).

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/TemporalQueryGuard.scala
```diff
@@ -0,0 +1,50 @@
...
+        s"Query exceeds maximum allowed filter duration of $max: ${filterToString(strategy.filter.filter)}")
+    }
+  }
+
+  override def close(): Unit = {}
+
+}
+
+object TemporalQueryGuard {
+  val Config = "geomesa.filter.max.duration"
```
jnh5y, line 49: @elahrvivaz after the refactoring, should this config path be different?

elahrvivaz: sure, geomesa.guards.something.something?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -0,0 +1,145 @@
...
+  case class SizeAndDuration(sizeLimit: Int, durationLimit: Duration)
+
+  def loadConfiguration(sft: SimpleFeatureType): Seq[SizeAndDuration]  = {
+    val conf = ConfigFactory.load()
+    try {
+      val guardConfig = conf.getConfigList(s"${ConfigPath}.${sft.getTypeName}")
+      buildLimits(guardConfig)
+    } catch {
+      case e: ConfigException =>
+        logger.error(s"Configuration for GraduatedQueryGuard not available or correct for type ${sft.getTypeName}.  " +
```
elahrvivaz, line 104: this is good, but FYI if the init method throws an exception, the framework will catch it and the guard won't even be invoked, so you don't need to ensure it still works if there's an error.

docs/user/datastores/query_interceptor.rst
```diff
@@ -0,0 +1,114 @@
...
+      def rewrite(query: Query): Unit
+
+      /**
+       * Hook to allow interception of a query after extracting the query values
+       *
+       * @param strategy query strategy
+       * @return an exception if the query should be stopped
+       */
+      def guard(strategy: QueryStrategy): Option[IllegalArgumentException] = None
+    }
```
elahrvivaz, line 46: i wonder if we should make this a java interface... and if that would break existing impls or not.

docs/user/datastores/query_interceptor.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Graduated Query Guard
++++++++++++++++++++++
+
+The graduated query guard applies different duration limits based on the spatial extent of the query.
+As a query becomes larger in space, it can be limited to shorter and shorter time ranges.
+A series of rules limit the duration for queries which are at most a given size in square degrees.
+
+To enable the guard, add ``org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard``
+to ``geomesa.query.interceptors`` as indicated above.  Configuration is managed via TypeSafe config.
```
elahrvivaz, line 91: should probably clarify here that standard config loading is used and the config is expected under the key `geomesa.guard.graduated`

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -0,0 +1,135 @@
...
+        if (first.sizeLimit == second.sizeLimit) {
+          throw new IllegalArgumentException(s"Graduated query guard configuration " +
+            s"has repeated size: ${first.sizeLimit}")
+        } else if (first.durationLimit.compareTo(second.durationLimit) <= 0) {
+          throw new IllegalArgumentException(s"Graduated query guard configuration " +
+            s"has durations out of order: ${first.durationLimit} is less than ${second.durationLimit}")
+        }
+      }
+    }
+    if (candidate.last.sizeLimit != Int.MaxValue) {
```
elahrvivaz, line 116: should probably make this .lastOption.exists or add a .isEmpty check for the degenerate case of an empty Seq

geomesa-tools/conf/application.conf
```diff
@@ -9,6 +9,31 @@ geomesa {
...
+    "example-csv-graduated-guard" = {
+      attributes = [
+        { name = "fid",      type = "Integer",         index = false }
+        { name = "name",     type = "String",          index = true  }
+        { name = "age",      type = "Integer",         index = false }
+        { name = "lastseen", type = "Date",            index = false, default = true }
+        { name = "geom",     type = "Point",           index = true, srid = 4326, default = true }
+      ]
+      user-data {
+        geomesa.query.interceptors = "org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard"
```
elahrvivaz, line 34: this is going to fail if you don't have the config file right? is that going to be confusing?

jnh5y: Yeah, I don't know what to do about it though...  I mean, having a user-data key-value would get around that....

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
jnh5y: Good call.  Adding an .isEmpty right ahead of this.

docs/user/datastores/query_interceptor.rst
jnh5y: Adding some docs for that.

docs/user/datastores/query_interceptor.rst
jnh5y: Hmm... do we have folks implementing the GeoMesa Java APIs?  Also, it should still be possible to do so...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -0,0 +1,145 @@
...
+        if (first.sizeLimit == second.sizeLimit) {
+          throw new IllegalArgumentException(s"Graduated query guard configuration " +
+            s"has repeated size: ${first.sizeLimit}")
+        } else if (first.durationLimit.compareTo(second.durationLimit) <= 0) {
+          throw new IllegalArgumentException(s"Graduated query guard configuration " +
+            s"has durations out of order: ${first.durationLimit} is less than ${second.durationLimit}")
+        }
+      }
+    }
+    candidate
```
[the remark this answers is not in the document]
jnh5y: Added this.

geomesa-tools/conf/application.conf
elahrvivaz: actually... this is a typesafe config file. you could add it in here.

jnh5y: I thought about that.  It seemed weird to toss this file into GeoServer.  But yeah, lemme do that.

elahrvivaz: true... but at least it will work with the CLI tools...

elahrvivaz (approved):

PR #2631, opened 2020-10-14
GEOMESA-2928 Implement full table scan query guard

jnh5y:

### DIFF
--- a/geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
+++ b/geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
@@ -72,12 +72,13 @@ class GeoMesaDataStoreTest extends Specification {
   }
 
   "GeoMesaDataStore" should {
-    "block queries with an excessive duration using custom query guard" in {
+    "block queries with an excessive duration and spatial extent using graduated query guard" in {
       val sft = SimpleFeatureTypes.createType("test",
         "name:String,age:Int,dtg:Date,*geom:Point:srid=4326;geomesa.indices.enabled='id,z3,attr:name'")
       // NB: Uses configuration in the test reference.conf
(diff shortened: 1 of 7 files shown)


geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
```diff
@@ -72,12 +72,13 @@ class GeoMesaDataStoreTest extends Specification {
...
   "GeoMesaDataStore" should {
-    "block queries with an excessive duration using custom query guard" in {
+    "block queries with an excessive duration and spatial extent using graduated query guard" in {
       val sft = SimpleFeatureTypes.createType("test",
         "name:String,age:Int,dtg:Date,*geom:Point:srid=4326;geomesa.indices.enabled='id,z3,attr:name'")
       // NB: Uses configuration in the test reference.conf
       sft.getUserData.put("geomesa.query.interceptors",
-        "org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard")
+        "org.locationtech.geomesa.index.planning.guard.GraduatedQueryGuard," +
+          "org.locationtech.geomesa.index.planning.guard.FullTableScanQueryGuard")
```
elahrvivaz, line 81: i'm wondering if there is a scenario where you'd want a temporal guard but also to allow full table scans? maybe we should just roll the full table scan logic into the temporal guards as well, so make config easier.

jnh5y: Yeah, I could imagine query guards which derive from / delegate to other guards.  E.g., the two temporal guards could also run the full table scan guard.

elahrvivaz: is it worth making that change, or are you happy as is?

PR #2633, opened 2020-10-16
GEOMESA-2933 Ensure temporal query guards only trigger for temporal indices

elahrvivaz:

### DIFF
--- a/docs/user/datastores/query_interceptor.rst
+++ b/docs/user/datastores/query_interceptor.rst
@@ -67,8 +67,8 @@ Full Table Scan Query Guard
 +++++++++++++++++++++++++++
 
 The full table scan query guard will block queries which would cause a full table scan to be performed.
-This query guard is provided so that per feature type configuration can be managed easily.  Alternatively, one can use the
-``geomesa.scan.block-full-table`` system property to disable full table scans across all feature types.
+The query guard will be loaded in all environments. Alternatively, one can use the ``geomesa.scan.block-full-table``
+system property to disable full table scans per environment (see :ref:`geomesa_site_xml`).
 
 Just like the ``geomesa.scan.block-full-table`` property, this guard respects the ``geomesa.scan.block-full-table.threshold``
 system property.  This allows for preview queries which can be helpful to show a system is working.
@@ -87,8 +87,7 @@ Temporal Query Guard
 
 The temporal query guard will block queries which exceed a maximum temporal duration.
 Any query which attempts to return a larger time period will be stopped.
-This guard also applies the full table scan guard.
-The temporal query guard will not affect queries which do not execute a full table scan against indices that do not have
+The temporal query guard will not affect queries against indices that do not have
 a temporal component (for example, feature ID and attribute queries).
 
 To enable the guard, add ``org.locationtech.geomesa.index.planning.guard.TemporalQueryGuard``
@@ -121,7 +120,8 @@ The configuration must satisfy a few conditions:
 * as the size increases, the duration must decrease,
 * and a given size limit may not be repeated.
 
-An example is given here.  Durations can be given in a number of days, hours, or minutes.
+If no size is provided, it is equivalent to an unbounded size. Durations can be given in a number of days,
+hours, or minutes. For example:
 
 .. code-block:: none
 
--- a/docs/user/datastores/query_interceptor.rst
+++ b/docs/user/datastores/query_interceptor.rst
@@ -49,7 +49,7 @@ Interceptors must have a default, no-arg constructor. The interceptor lifecycle 
 
 1. The instance is instantiated via reflection, using its default constructor
 #. The instance is initialized via the ``init`` method, passing in the data store containing the simple feature type
-#. ``rewrite`` and ``guard`` are called once for each query
+#. ``rewrite`` and ``guard`` are called once for each query (these two methods must be thread-safe)
 #. When the data store is disposed, the instance is cleaned up via the ``close`` method
 
 Interceptors will be invoked in the order they are declared in the user data. In order to see detailed information
@@ -81,6 +81,8 @@ to ``geomesa.query.interceptors`` as indicated above.
     sft.getUserData().put("geomesa.query.interceptors",
       "org.locationtech.geomesa.index.planning.guard.FullTableScanQueryGuard");
 
+To disable the guard on a per-environment basis, set the system property ``geomesa.scan.<typeName>.block-full-table``
+to ``false``, where ``<typeName>`is the name of your feature type.
 
 Temporal Query Guard
 ++++++++++++++++++++
@@ -98,6 +100,9 @@ to ``geomesa.query.interceptors`` as indicated above, and set the duration using
     sft.getUserData().put("geomesa.query.interceptors",
       "org.locationtech.geomesa.index.planning.guard.TemporalQueryGuard");
     sft.getUserData().put("geomesa.guard.temporal.max.duration", "28 days");
+
+To disable the guard on a per-environment basis, set the system property ``geomesa.guard.temporal.<typeName>.disable``
+to ``true``, where ``<typeName>`is the name of your feature type.
 
 Graduated Query Guard
 +++++++++++++++++++++
@@ -136,3 +141,6 @@ hours, or minutes. For example:
         }
       }
     }
+
+To disable the guard on a per-environment basis, set the system property ``geomesa.guard.graduated.<typeName>.disable``
+to ``true``, where ``<typeName>`is the name of your feature type.
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/s2/S2Index.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/s2/S2Index.scala
@@ -6,7 +6,8 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.index.index.s2
+package org.locationtech.geomesa.index.index
+package s2
 
 import org.locationtech.geomesa.index.api.ShardStrategy.ZShardStrategy
 import org.locationtech.geomesa.index.api.{GeoMesaFeatureIndex, IndexKeySpace}
@@ -25,7 +26,8 @@ import org.opengis.feature.simple.SimpleFeatureType
   */
 class S2Index protected (ds: GeoMesaDataStore[_], sft: SimpleFeatureType, version: Int, geom: String, mode: IndexMode)
     extends GeoMesaFeatureIndex[S2IndexValues, Long](ds, sft, S2Index.name, version, Seq(geom), mode)
-        with SpatialFilterStrategy[S2IndexValues, Long] {
+        with SpatialFilterStrategy[S2IndexValues, Long]
+        with SpatialIndex[S2IndexValues, Long] {
 
   def this(ds: GeoMesaDataStore[_], sft: SimpleFeatureType, geom: String, mode: IndexMode) =
     this(ds, sft, S2Index.version, geom, mode)
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
@@ -145,6 +145,13 @@ class QueryPlanner[DS <: GeoMesaDataStore[DS]](ds: DS) extends QueryRunner with 
           val qs = strategy.getQueryStrategy(hints, output)
           // query guard hook
           interceptors(sft).foreach(_.guard(qs).foreach(e => throw e))
+          if (qs.values.isEmpty) {
+            qs.filter.secondary.foreach { f =>
+              logger.warn(
+                s"Running full table scan on ${qs.index.name} index for schema " +
+                  s"${sft.getTypeName} with filter ${filterToString(f)}")
+            }
+          }
           ds.adapter.createQueryPlan(qs)
         }
       }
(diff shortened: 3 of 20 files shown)


docs/user/datastores/query_interceptor.rst
```diff
@@ -49,7 +49,7 @@ Interceptors must have a default, no-arg constructor. The interceptor lifecycle
 
 1. The instance is instantiated via reflection, using its default constructor
 #. The instance is initialized via the ``init`` method, passing in the data store containing the simple feature type
-#. ``rewrite`` and ``guard`` are called once for each query
+#. ``rewrite`` and ``guard`` are called once for each query (these two methods must be thread-safe)
```
jnh5y, line 52: Did you hit a non-thread safe case as you thinking through things?

docs/user/datastores/query_interceptor.rst
```diff
@@ -81,14 +81,15 @@ to ``geomesa.query.interceptors`` as indicated above.
     sft.getUserData().put("geomesa.query.interceptors",
       "org.locationtech.geomesa.index.planning.guard.FullTableScanQueryGuard");
 
+To disable the guard on a per-environment basis, set the system property ``geomesa.scan.<typeName>.block-full-table``
+to ``false``, where ``<typeName>`is the name of your feature type.
```
jnh5y, line 85: Neat idea.  Good docs here!

docs/user/datastores/query_interceptor.rst
```diff
@@ -136,3 +141,6 @@ An example is given here.  Durations can be given in a number of days, hours, or
         }
       }
     }
+
+To disable the guard on a per-environment basis, set the system property ``geomesa.guard.graduated.<typeName>.disable``
```
jnh5y, line 145: I wonder how hard it'd be to sort out common configuration code to make it 'easy' / abstract for implementations to have this kind of capability....

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/s2/S2Index.scala
```diff
@@ -6,7 +6,8 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.index.index.s2
+package org.locationtech.geomesa.index.index
+package s2
```
jnh5y, line 10: Any particular reason to break out the package name like this?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/QueryPlanner.scala
```diff
@@ -145,6 +145,13 @@ class QueryPlanner[DS <: GeoMesaDataStore[DS]](ds: DS) extends QueryRunner with
           val qs = strategy.getQueryStrategy(hints, output)
           // query guard hook
           interceptors(sft).foreach(_.guard(qs).foreach(e => throw e))
+          if (qs.values.isEmpty) {
```
jnh5y, line 148: Nice, nice!  Good improvement that I missed.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -8,76 +8,61 @@
...
-              } else {
-                None
-              }
+  override def guard(strategy: QueryStrategy): Option[IllegalArgumentException] = {
+    val msg = if (disabled || !strategy.index.isInstanceOf[SpatioTemporalIndex[_, _]]) { None } else {
+      val values = strategy.values.collect {
+        case v: SpatialIndexValues with TemporalIndexValues => (v.spatialBounds, v.intervals)
+      }
+      values match {
+        case None => Some("Query does not have a temporal filter")
```
jnh5y, line 46: The query could be lacking a spatial filter, no?  E.g., wouldn't this be hit when you pass "dtg during date1/date2" to a Z3 index?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -8,76 +8,61 @@
...
+        case Some((s, i)) =>
+          val spatialExtent = s.map { case (lx, ly, ux, uy) => (ux - lx) * (uy - ly) }.sum
+          val limit = guardLimits.find(_.sizeLimit >= spatialExtent).getOrElse {
+            throw new IllegalStateException(
+              s"Invalid extents/limits: ${s.mkString(", ")} / ${guardLimits.mkString(", ")}")
```
jnh5y, line 51: Shouldn't the guardLimits have an unbounded limit?  Does this happen?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
```diff
@@ -199,7 +123,7 @@ class GeoMesaDataStoreTest extends Specification {
       results must haveLength(10)
     }
     "block queries which would cause a full table scan" in {
-      val sft = SimpleFeatureTypes.createType("test",
+      val sft = SimpleFeatureTypes.createType("61b44359ddb84822983587389d6a28a4",
```
jnh5y, line 126: Catchy type name!

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/TemporalQueryGuardTest.scala
```diff
@@ -0,0 +1,79 @@
...
+        "name = 'bob'",
+        "IN('123')",
+        "bbox(geom,-10,-10,10,10) AND dtg during 2020-01-01T00:00:00.000Z/2020-01-01T23:59:59.000Z",
+        "bbox(geom,-10,-10,10,10) AND (dtg during 2020-01-01T00:00:00.000Z/2020-01-01T00:59:59.000Z OR dtg during 2020-01-01T12:00:00.000Z/2020-01-01T12:59:59.000Z)"
+      )
+
+      val invalid = Seq(
+        "INCLUDE",
+        "bbox(geom,-10,-10,10,10)",
+        "bbox(geom,-180,-90,180,90)",
```
jnh5y, line 46: Ok, what changed that these are getting blocked now?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuardTest.scala
```diff
@@ -0,0 +1,119 @@
...
+        "bbox(geom,-180,-90,180,90) AND dtg during 2020-01-01T00:00:00.000Z/P1D",
+        "bbox(geom,0,0,2,4) AND dtg during 2020-01-01T00:00:00.000Z/2020-01-02T00:00:00.000Z",
+        "bbox(geom,-10,-10,10,10) AND dtg during 2020-01-01T00:00:00.000Z/2020-01-01T23:00:00.000Z",
+        "bbox(geom,-10,-10,10,10) AND (dtg during 2020-01-01T00:00:00.000Z/2020-01-01T00:59:59.000Z OR dtg during 2020-01-01T12:00:00.000Z/2020-01-01T12:59:59.000Z)"
+      )
+
+      val invalid = Seq(
+        "INCLUDE",
+        "bbox(geom,-10,-10,10,10)",
+        "bbox(geom,-180,-90,180,90)",
```
jnh5y, line 55: If you update this PR at all, could you add a time range filter to this list?

jnh5y: Maybe a dtg after, and a dtg during?

jnh5y (approved):

docs/user/datastores/query_interceptor.rst
elahrvivaz: just clarifying that if multiple queries come in at once that method might be invoked from 2 different threads at the same time. i think all our impls were fine.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/s2/S2Index.scala
elahrvivaz: it just gets the `index` package object in scope so we don't have to import things :)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
elahrvivaz: i think this would only be None if there is no spatial AND no temporal filter. since it's a temporal guard, i was thinking the error message should reference the temporal part only. what do you think?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
elahrvivaz: yeah i don't think this should ever happen, so i used the IllegalStateException to indicate something when quite wrong. seemed better than just a raw `.get`. i could maybe envision some really complex series of polygon filters that, in aggregate, exceeded Int.MaxValue degrees and flipped negative?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala
elahrvivaz: haha, i made it `UUID.randomUUID` but then had to replace it with a constant value so that the typesafe conf file lined up :)

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/TemporalQueryGuardTest.scala
elahrvivaz: i swapped the order of the indices in the user data definition - the first index declared gets used for full table scans (which may be an unintentional implementation detail...). i left a comment on that change up above; that made the full table scans hits the z3 index and get blocked by the guard, instead of the feature id index where they weren't.

not sure if we should do something to prioritize which index gets used for full table scans - when we switched to SPI loaded indices we lost the ability to easily configure that, but we could add something like a 'priority' flag to each index. or just document that you should declare them in order... in practice i think almost all serious installations would block full tables scans though?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuardTest.scala
elahrvivaz: sure

elahrvivaz: which case did you mean, exactly? you already had during and after examples in the list.

elahrvivaz: oh you mean without a bbox?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/s2/S2Index.scala
jnh5y: Yeah, figured it might be something clever like that.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
jnh5y: I think of the graduate guard as a spatial temporal guard; that's why I'm asking about it. 

It seems like the other errors show what filter limits are allowed.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
jnh5y: +1; just making sure I understood.

Seems like we could add the "//  CANNOT GET HERE" comment:).

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/TemporalQueryGuardTest.scala
jnh5y: Wow.  That's an interesting set of choices.

Do we have an analog to this page: https://docs.geoserver.org/stable/en/user/production/index.html ?

Any thoughts about which of these would be easiest for the end users?

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuardTest.scala
jnh5y: Yes!

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/planning/guard/TemporalQueryGuardTest.scala
elahrvivaz: i don't think we have anything like that - seems like a good idea.

i'm not sure which would be better... on the one hand, being able to specify a table to use for full table scans seems like it might be nice. on the other, having inconsistent tables being scanned seems potentially confusing if you don't know what's going on.

PR #2637, opened 2020-10-20
GEOMESA-2935 Optimize inArray filter function use

jnh5y:


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -143,12 +145,50 @@ protected class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends Duplicati
...
 
+  private def handleInArrayEquals(filter: PropertyIsEqualTo, extraData: AnyRef): Option[Filter] = {
+    var inArrayFunction: InArrayFunction = null
+    var hasTrue: Boolean = false
+
+    def inspectExpression(expression: Expression) = {
+      expression match {
+        case function: InArrayFunction =>
+          inArrayFunction = function
+        case literal: Literal if literal.getValue.equals(java.lang.Boolean.TRUE) =>
```
elahrvivaz, line 156: is it possible that literal.getValue might be null? i'm not sure... it would be safer to flip the `.equals` to `java.lang.Boolean.TRUE.equals(...)` or just use scala `==`, which both should just return false for a null but not throw an NPE

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
```diff
@@ -143,12 +145,50 @@ protected class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends Duplicati
...
+
+    inspectExpression(filter.getExpression1)
+    inspectExpression(filter.getExpression2)
+
+    if (inArrayFunction != null && hasTrue) {
+      val attribute = inArrayFunction.getParameters.get(0)
+      val array = inArrayFunction.getParameters.get(1)
+      import scala.collection.JavaConversions._
+      val filters = array.evaluate(null).asInstanceOf[util.List[_]].toSet[Any].map {
+        o => ff.equals(attribute, ff.literal(o))
```
elahrvivaz, line 170: instead of using `ff` here and below, it would be good to use `val ff = getFactory(extraData)` right above, to allow users to pass in their own factory hints. I'm not sure that's even possible, given our workflow, but seems safer to follow the pattern used in the rest of the class

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
```diff
@@ -41,6 +44,31 @@ class FilterHelperTest extends Specification {
 
   "FilterHelper" should {
 
+    "optimize the inArray function" >> {
+      val filter = ff.function("inArray", ff.literal("foo"), ff.literal(Array("foo", "bar")))
```
elahrvivaz, line 48: doesn't seem like this filter is checked at all? curious if the literal array will work ok or if you need the array list

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
```diff
@@ -41,6 +44,31 @@ class FilterHelperTest extends Specification {
...
+      val filter2 = ff.equals(
+        ff.function("inArray", ff.property("name"), ff.literal(new util.ArrayList(util.Arrays.asList("foo", "bar")))),
+        ff.literal(java.lang.Boolean.TRUE)
+      )
+
+      val sf = new ScalaSimpleFeature(sft, "1",
+        Array(new Date(), new Integer(1), new Integer(2), new Integer(3), new Integer(4), WKTUtils.read("POINT(1 5)"), "foo"))
```
elahrvivaz, line 55: could we add an `.evaluate(sf)` on some of the test filters to verify that they return correctly?

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
```diff
@@ -41,6 +44,31 @@ class FilterHelperTest extends Specification {
...
+      val filter4 = ff.equals(
+        ff.literal(java.lang.Boolean.TRUE),
+        ff.function("inArray",
+          ff.property("name"),
+          ff.literal(new util.ArrayList(util.Arrays.asList("foo", "bar"))))
+      )
+      val filter5 = updateFilter(filter4)
+
+      filter5.isInstanceOf[OrImpl] mustEqual(true)
+      ok
```
elahrvivaz, line 69: would it also be possible to show an ECQL example? or is that not relevant? I forget now, would a user ever need to invoke this themselves through e.g. geoserver where it would have to be ECQL?

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
jnh5y: +1.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
jnh5y: Will do!

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
jnh5y: https://github.com/geoserver/geoserver/blob/2.17.3/src/extension/querylayer/src/main/java/org/geoserver/filter/function/QueryFunction.java#L124-L151

An ArrayList is going to be generated.  Of course, this doesn't seem like a public API / fixed thing, so yeah....

geomesa-filter/src/test/scala/org/locationtech/geomesa/filter/FilterHelperTest.scala
jnh5y: (To my knowledge) There's no ECQL support for arrays.  

ECQL.toFilter will print out something like...
```inArray(name,'[foo, bar]') = true```

If you run that through ECQL.toCQL, the [foo, bar] is turned into a literal, not an (Array)List.

I think the ECQL 'support' for this query pattern is the "optimized" version that we are creating with the Visitor.

File: geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala [hunk 6]
```diff
@@ -143,12 +145,53 @@ protected class QueryPlanFilterVisitor(sft: SimpleFeatureType) extends Duplicati
...
+    inspectExpression(filter.getExpression1, extraData)
+    inspectExpression(filter.getExpression2, extraData)
+
+    if (inArrayFunction != null && hasTrue) {
+      val attribute = inArrayFunction.getParameters.get(0)
+      val array = inArrayFunction.getParameters.get(1)
+      import scala.collection.JavaConversions._
+      // The important part of this optimization is to handle ArrayLists generated by the
+      // GeoServer cross-layer querylayer extension:  E.g.
+      // https://github.com/geoserver/geoserver/blob/2.17.3/src/extension/querylayer/src/main/java/org/geoserver/filter/function/QueryFunction.java#L124-L151
```
elahrvivaz, line 171: ah, thanks for this!

elahrvivaz (approved):

elahrvivaz:
will merge when the CI build finishes

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/visitor/QueryPlanFilterVisitor.scala
jnh5y: Yeah, seemed like this was an important piece of the puzzle to comment on.  

Also, it'll invariably change later:).

PR #2642, opened 2020-10-26
2.1.x query guard

jnh5y:


elahrvivaz:
looks reasonable to me

PR #2646, opened 2020-11-04
GEOMESA-2944: Add unit test for jsonPath applied to an array

jrs53:


elahrvivaz:
Thanks for the test! I think that even though it test passes, the geomesa json serialization only works with objects: https://github.com/locationtech/geomesa/blob/main/geomesa-features/geomesa-feature-kryo/src/main/scala/org/locationtech/geomesa/features/kryo/json/KryoJsonSerialization.scala#L89

so you're likely to hit an error if you write the feature to a data store first...

we'd need to update that serialization to work with arrays as well.

PR #2647, opened 2020-11-10
GEOMESA-2947 Kafka listener with offset tracking

elahrvivaz:

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
@@ -8,7 +8,7 @@
 
 package org.locationtech.geomesa.kafka.data
 
-import java.io.IOException
+import java.io.{Closeable, IOException}
 import java.util.concurrent.{ConcurrentHashMap, ScheduledExecutorService}
 import java.util.{Collections, Properties, UUID}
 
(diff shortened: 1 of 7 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
```diff
@@ -235,7 +272,7 @@ object KafkaDataStore extends LazyLogging {
 
   val MetadataPath = "metadata"
 
-  val LoadIntervalProperty: SystemProperty = SystemProperty("geomesa.kafka.load.interval", "100ms")
+  val LoadIntervalProperty: SystemProperty = SystemProperty("geomesa.kafka.load.interval", "1s")
```
jnh5y, line 275: Any particular reason to change this?  (I'm fine either way; just asking.)

elahrvivaz: this prop controls the amount of time the consumer will wait for a message when calling poll. i was reading some things about how if the data isn't processed within the poll interval of the consumer, then kafka might consider the consumer dead. i'm not really sure if this matters or not, but i don't think there's any downside to increasing it, b/c we just loop and poll again after the timeout.

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreTest.scala
```diff
@@ -525,6 +526,66 @@ class KafkaDataStoreTest extends Specification with Mockito with LazyLogging {
...
+        WithClose(consumer.createConsumer(sft.getTypeName, "mygroup", processor)) { _ =>
+          writeUpdates()
+          eventually(seen.get must beTrue)
+          eventually(results must haveLength(numUpdates))
+        }
+
+        // verify that we can read a second batch
+        writeUpdates()
+        WithClose(consumer.createConsumer(sft.getTypeName, "mygroup", processor)) { _ =>
+          eventually(results must haveLength(numUpdates * 2))
```
jnh5y, line 581: Is this showing that the records were processed twice?

elahrvivaz: it's verifying that the second batch of records was also read - we write one batch, read the messages, close the consumer, write a second batch, create a new consumer with the same group id and verify that it picked up from where it left off.

PR #2658, opened 2020-11-30
WIP to show what the alignment project ought to be able to do.

jnh5y:


geomesa-spark/geomesa-spark-jts/src/main/scala/org/apache/spark/sql/jts/JTSTypes.scala
```diff
@@ -44,26 +52,389 @@ object JTSTypes {
...
+          holes(i - 1) = line
+        }
+        i += 1
+      }
+      gf.createPolygon(shell, holes)
+    }
+  }
+
+  // case ObjectType.POLYGON =>
+  //   group.id(GeometryBytes.TwkbPolygon)
```
jnh5y, line 323: @elahrvivaz what's with the Twkb bit of these variable names?

geomesa-spark/geomesa-spark-jts/src/main/scala/org/locationtech/geomesa/spark/jts/rules/GeometryLiteralRules.scala
```diff
@@ -22,20 +24,30 @@ object GeometryLiteralRules {
...
       plan.transform {
         case q: LogicalPlan => q.transformExpressionsDown {
-          case s: ScalaUDF =>
-            // TODO: Break down by GeometryType
-            Try {
-                s.eval(null) match {
-                  case row: GenericInternalRow =>
-                    val ret = GeometryUDT.deserialize(row)
-                    GeometryLiteral(row, ret)
-                  case other: Any =>
```
jnh5y, line 32: @elahrvivaz Hmm....  do we have an UDFs where the s.dataType is not a JTS Geometry?  

I'm trying to figure out if this 'fall back' case is still being applied after the refactoring.

elahrvivaz: i believe that the 'if' guards in the case statements will handle that. since transformExpressionsDown takes a partial function, it will just skip over non-geometry data types. since previously we matched on any scalaUDF, we had to handle the non-geometry types as well

elahrvivaz: but to answer your question, yes there are scala udfs with literals in the unit tests that started failing before i fixed this block.

geomesa-spark/geomesa-spark-jts/src/main/scala/org/apache/spark/sql/jts/JTSTypes.scala
elahrvivaz: good question... those came from our SimpleFeatureParquetSchema. In checking the code, they correspond to the WKB constants for the same types. We probably don't need to have both pairs of constants, but meh.

PR #2659, opened 2020-12-03
GEOMESA-2958 Exit with non-zero code if CLI jobs fail

elahrvivaz:
* Simplify ingest job classes


geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/hbase-libjars.list
```diff
@@ -1,5 +1,8 @@
 hbase-client
 hbase-common
+hbase-hadoop-compat
+hbase-hadoop2-compat
+hbase-metrics
```
jnh5y, line 5: Did you test this?  (Or is their addition just obvious?)

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseBulkLoadCommand.scala
```diff
@@ -49,7 +50,11 @@ class HBaseBulkLoadCommand extends HBaseDataStoreCommand {
     val locator = ds.connection.getRegionLocator(tableName)
     val config = new Configuration
     config.set("hbase.loadincremental.validate.hfile", params.validate.toString)
-    new LoadIncrementalHFiles(config).doBulkLoad(input, ds.connection.getAdmin, table, locator)
+    // use reflection to get around 1.4 vs 2.2 API differences (method return type void vs Map)
+    val load = new LoadIncrementalHFiles(config)
+    val method =
+      load.getClass.getMethod("doBulkLoad", classOf[Path], classOf[Admin], classOf[Table], classOf[RegionLocator])
+    method.invoke(load, input, ds.connection.getAdmin, table, locator)
```
jnh5y, line 57: I'm assuming you tested this?  (Otherwise, noticing the difference would be tough:)....)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
```diff
@@ -34,9 +35,10 @@ trait Runner extends LazyLogging {
         logger.error(msg, e)
         Command.user.error(msg)
         environmentErrorInfo().foreach(Command.user.error)
-        sys.exit(-1)
-      case e: ParameterException => Command.user.error(e.getMessage); sys.exit(-1)
-      case NonFatal(e) => Command.user.error(e.getMessage, e); sys.exit(-1)
+        sys.exit(1)
+      case e: ParameterException => Command.user.error(e.getMessage); sys.exit(1)
+      case e: CommandException => Command.user.error(e.getMessage); sys.exit(1)
+      case NonFatal(e) => Command.user.error(e.getMessage, e); sys.exit(1)
```
jnh5y, line 41: Hmm... any reasons to go with +1 over -1?  

I mean, I like being a positive person...

geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/hbase-libjars.list
elahrvivaz: i had to add these when running the hbase offline ingest. we don't normally smoke test that (we probably should start), so it's probably been broken for a while.

geomesa-hbase/geomesa-hbase-tools/src/main/scala/org/locationtech/geomesa/hbase/tools/ingest/HBaseBulkLoadCommand.scala
elahrvivaz: yeah, the method is the same so there was no compiler error, but the return value is different so it breaks at runtime

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
elahrvivaz: bash exit codes have to be positive numbers. that fact bit me at some point in the past, so i tend to remember it now...

geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/hbase-libjars.list
jnh5y: +1 to both!  Can you upload the smoke test?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
jnh5y: Yeah, bash must a byte for the return code?  Looks like the values have to be between 0 and 255.

geomesa-hbase/geomesa-hbase-tools/src/main/resources/org/locationtech/geomesa/hbase/tools/hbase-libjars.list
elahrvivaz: updated the checklist

PR #2660, opened 2020-12-10
 reset metedata cache when delete catalog

Susheldon:
The original logic will cause the failure to delete the catalog after calling HBaseDataStore.delete() after removing all schemas through datastore.remove(), and failing to create a new typename through datastore.create() after calling HBaseDataStore.delete()

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStore.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStore.scala
@@ -68,6 +68,10 @@ class HBaseDataStore(con: ConnectionWrapper, override val config: HBaseDataStore
 
   override def dispose(): Unit = try { super.dispose() } finally { con.close() }
 
+  override def delete():Unit={
+    super.delete()
+    metadata.asInstanceOf[HBaseBackedMetadata[String]].resetCache()
+  }
   override protected def loadIteratorVersions: Set[String] = {
     import org.locationtech.geomesa.utils.conversions.ScalaImplicits.RichIterator
 
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
@@ -85,7 +85,11 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
     * (index tables and catalog table)
     * NB: We are *not* currently deleting the query table and/or query information.
     */
-  def delete(): Unit = adapter.deleteTables(getTypeNames.flatMap(getAllTableNames).distinct)
+  def delete(): Unit = {
+    val types = getTypeNames
+    val tables = if (types.length == 0) Array(config.catalog) else types
+    adapter.deleteTables(types.flatMap(getAllTableNames).distinct)
+  }
 
   // hooks to allow extended functionality
 
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
@@ -99,7 +99,7 @@ trait TableBasedMetadata[T] extends GeoMesaMetadata[T] with LazyLogging {
   protected def scanKeys(): CloseableIterator[(String, String)]
 
   // only synchronize if table doesn't exist - otherwise it's ready only and we can avoid synchronization
-  private val tableExists: MaybeSynchronized[Boolean] =
+  private var tableExists: MaybeSynchronized[Boolean] =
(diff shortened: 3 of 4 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStore.scala
```diff
@@ -68,6 +68,10 @@ class HBaseDataStore(con: ConnectionWrapper, override val config: HBaseDataStore
 
   override def dispose(): Unit = try { super.dispose() } finally { con.close() }
 
+  override def delete():Unit={
+    super.delete()
+    metadata.asInstanceOf[HBaseBackedMetadata[String]].resetCache()
```
elahrvivaz, line 73: this is going to affect all data store impls - could we remove this change here, and put it in `GeoMesaDataStore.delete`?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -85,7 +85,11 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
     * (index tables and catalog table)
     * NB: We are *not* currently deleting the query table and/or query information.
     */
-  def delete(): Unit = adapter.deleteTables(getTypeNames.flatMap(getAllTableNames).distinct)
+  def delete(): Unit = {
+    val types = getTypeNames
+    val tables = if (types.length == 0) Array(config.catalog) else types
```
elahrvivaz, line 90: seems like this isn't being used, did you mean to pass it in to adapter.deleteTables? it's a good catch, since previously if there was an empty catalog table it wouldn't delete it (although generally the catalog table isn't empty unless you've gone through and deleted all the feature types)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
```diff
@@ -231,6 +231,11 @@ trait TableBasedMetadata[T] extends GeoMesaMetadata[T] with LazyLogging {
   // checks that the table is already created, and creates it if not
   def ensureTableExists(): Unit = tableExists.set(true, false, createTable())
 
+
+  def resetCache():Unit={
+    tableExists = if (checkIfTableExists) { new NotSynchronized[Boolean](true)} else {new IsSynchronized[Boolean](false)}
```
elahrvivaz, line 236: per my above comment, we could move this method out to `GeoMesaMetadata` with a no-op default implementation, and then override it here.

this piece is potentially problematic though... making tableExists into a var makes it not thread safe anymore. we should probably just always synchronize the value. the `MaybeSynchronized` code here was an attempt to avoid synchronization calls in the common case of the table already existing, assuming that it wouldn't be deleted ever, but that doesn't hold when `delete` is called on the data store.

elahrvivaz (changes requested):
Thanks for the fix! I'd like to get this added for all the data store impls, not just hbase, if you have time to update the PR.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
Susheldon: sorry，my mistake,`tables` should be passed to adapter.deleteTables。

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStore.scala
Susheldon: > this is going to affect all data store impls - could we remove this change here, and put it in `GeoMesaDataStore.delete`?

I didn't catch you, why HBase impls is going to affect all impls? or what you mean is to affect all implementations so it should be moved to geomesadata?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
Susheldon: > seems like this isn't being used, did you mean to pass it in to adapter.deleteTables? it's a good catch, since previously if there was an empty catalog table it wouldn't delete it (although generally the catalog table isn't empty unless you've gone through and deleted all the feature types)

sorry，my mistake,tables should be passed to adapter.deleteTables。

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseDataStore.scala
elahrvivaz: right, sorry, the bug is going to affect all the data stores so I'd like to fix it generally instead of just for hbase

Susheldon:
For my wrong operation, I will close this pull request and restart a new pull request

Susheldon:
For my wrong operation, I will close this pull request and restart a new pull request

PR #2663, opened 2020-12-14
set Charset for deserialize in CqlTransformFilter

Susheldon:
set Charset for deserialize in CqlTransformFilter,as the same as cql
Signed-off-by: xxxxs <Susheldon@users.noreply.github.com>


elahrvivaz:
Thanks for fixing this, could you update the PR with just the first commit so we don't mix in the table metadata changes?

Susheldon:
start a new PR from a other branch

PR #2665, opened 2020-12-15
reset table cache when delete catalog.

Susheldon:
A new PR for https://github.com/locationtech/geomesa/pull/2660
This change making tableExists into a var makes it not thread safe anymore.

Signed-off-by: xxxxs <Susheldon@users.noreply.github.com>

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
@@ -22,7 +22,6 @@ import org.locationtech.geomesa.index.geotools.GeoMesaDataStore.{SchemaCompatibi
 import org.locationtech.geomesa.index.geotools.GeoMesaDataStoreFactory.GeoMesaDataStoreConfig
 import org.locationtech.geomesa.index.index.attribute.AttributeIndex
 import org.locationtech.geomesa.index.index.id.IdIndex
-import org.locationtech.geomesa.index.metadata.TableBasedMetadata
 import org.locationtech.geomesa.index.planning.QueryPlanner
 import org.locationtech.geomesa.index.stats.HasGeoMesaStats
 import org.locationtech.geomesa.index.utils.{ExplainLogging, Explainer}
@@ -88,9 +87,11 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
     */
   def delete(): Unit = {
     val types = getTypeNames
-    val tables = if (types.length == 0) Array(config.catalog) else types
-    adapter.deleteTables(tables.flatMap(getAllTableNames).distinct)
-    metadata.asInstanceOf[TableBasedMetadata[String]].resetCache()
+    val tables = if (types.length == 0) Array(config.catalog) else {
+      types.flatMap(getAllTableNames).distinct
+    }
+    adapter.deleteTables(tables)
+    metadata.resetCache()
   }
 
   // hooks to allow extended functionality
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
@@ -22,6 +22,7 @@ import org.locationtech.geomesa.index.geotools.GeoMesaDataStore.{SchemaCompatibi
 import org.locationtech.geomesa.index.geotools.GeoMesaDataStoreFactory.GeoMesaDataStoreConfig
 import org.locationtech.geomesa.index.index.attribute.AttributeIndex
 import org.locationtech.geomesa.index.index.id.IdIndex
+import org.locationtech.geomesa.index.metadata.TableBasedMetadata
 import org.locationtech.geomesa.index.planning.QueryPlanner
 import org.locationtech.geomesa.index.stats.HasGeoMesaStats
 import org.locationtech.geomesa.index.utils.{ExplainLogging, Explainer}
@@ -85,7 +86,12 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
     * (index tables and catalog table)
     * NB: We are *not* currently deleting the query table and/or query information.
     */
-  def delete(): Unit = adapter.deleteTables(getTypeNames.flatMap(getAllTableNames).distinct)
+  def delete(): Unit = {
+    val types = getTypeNames
+    val tables = if (types.length == 0) Array(config.catalog) else types
+    adapter.deleteTables(tables.flatMap(getAllTableNames).distinct)
+    metadata.asInstanceOf[TableBasedMetadata[String]].resetCache()
+  }
 
   // hooks to allow extended functionality
 
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
@@ -99,7 +99,7 @@ trait TableBasedMetadata[T] extends GeoMesaMetadata[T] with LazyLogging {
   protected def scanKeys(): CloseableIterator[(String, String)]
 
   // only synchronize if table doesn't exist - otherwise it's ready only and we can avoid synchronization
-  private val tableExists: MaybeSynchronized[Boolean] =
+  private var tableExists: MaybeSynchronized[Boolean] =
     if (checkIfTableExists) { new NotSynchronized(true) } else { new IsSynchronized(false) }
 
   private val expiry = TableBasedMetadata.Expiry.toDuration.get.toMillis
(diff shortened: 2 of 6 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -85,7 +86,12 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
     * (index tables and catalog table)
     * NB: We are *not* currently deleting the query table and/or query information.
     */
-  def delete(): Unit = adapter.deleteTables(getTypeNames.flatMap(getAllTableNames).distinct)
+  def delete(): Unit = {
+    val types = getTypeNames
+    val tables = if (types.length == 0) Array(config.catalog) else types
```
elahrvivaz, line 91: I think this logic is still slightly wrong...
```suggestion
    val tables = if (types.length == 0) { Array(config.catalog) } else { types.flatMap(getAllTableNames).distinct }
```

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -85,7 +86,12 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
...
+    adapter.deleteTables(tables.flatMap(getAllTableNames).distinct)
```
elahrvivaz, line 92:
```suggestion
    adapter.deleteTables(tables)
```

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -85,7 +86,12 @@ abstract class GeoMesaDataStore[DS <: GeoMesaDataStore[DS]](val config: GeoMesaD
...
+    metadata.asInstanceOf[TableBasedMetadata[String]].resetCache()
```
elahrvivaz, line 93: this cast is unnecessary now, right?
```suggestion
    metadata.resetCache()
```

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStore.scala
```diff
@@ -22,6 +22,7 @@ import org.locationtech.geomesa.index.geotools.GeoMesaDataStore.{SchemaCompatibi
 import org.locationtech.geomesa.index.geotools.GeoMesaDataStoreFactory.GeoMesaDataStoreConfig
 import org.locationtech.geomesa.index.index.attribute.AttributeIndex
 import org.locationtech.geomesa.index.index.id.IdIndex
+import org.locationtech.geomesa.index.metadata.TableBasedMetadata
```
elahrvivaz, line 25: if the cast is unnecessary below, we can remove this line too

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
```diff
@@ -99,7 +99,7 @@ trait TableBasedMetadata[T] extends GeoMesaMetadata[T] with LazyLogging {
   protected def scanKeys(): CloseableIterator[(String, String)]
 
   // only synchronize if table doesn't exist - otherwise it's ready only and we can avoid synchronization
-  private val tableExists: MaybeSynchronized[Boolean] =
+  private var tableExists: MaybeSynchronized[Boolean] =
```
elahrvivaz, line 102: to keep this thread safe, you can just make this an AtomicBoolean, and then call `.set` on it in reset cache and ensureTableExists:
```
def ensureTableExists(): Unit = if (tableExists.compareAndSet(false, true)) { createTable() }
```

Susheldon: > to keep this thread safe, you can just make this an AtomicBoolean, and then call `.set` on it in reset cache and ensureTableExists:
> 
> ```
> def ensureTableExists(): Unit = if (tableExists.compareAndSet(false, true)) { createTable() }
> ```

Using AtomicBoolean  caused a new problem:   `==`  is not work for value and expected in class IsSynchronized --`equals` method also not work because AtomicBoolean  does not override equlas method Object. Do we need to define IsSynchronized[AtomicBoolean] ? This does not seem to be a good idea.

elahrvivaz: you can get rid of the whole MaybeSynchronized class. that was an optimization we can't use now that we are setting it back from true to false. the original idea was that in the usual case, where the table exists, that val would be `true` when created and then never change, so we don't need to synchronize on it.
so you'd do something like:

```
private val tableExists = new AtomicBoolean(checkIfTableExists)
...
def ensureTableExists(): Unit = if (tableExists.compareAndSet(false, true)) { createTable() }
...
override def resetCache(): Unit= {
  tableExists.set(checkIfTableExists)
  metaDataCache.invalidateAll()
}
```

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/metadata/TableBasedMetadata.scala
```diff
@@ -229,7 +229,18 @@ trait TableBasedMetadata[T] extends GeoMesaMetadata[T] with LazyLogging {
   }
 
   // checks that the table is already created, and creates it if not
-  def ensureTableExists(): Unit = tableExists.set(true, false, createTable())
+  def ensureTableExists(): Unit = {
+//    do not use compareAndSet because unsafe.compareAndSwapInt is called in compareAndSet，may make problem for jdk11
```
elahrvivaz, line 233: hmm, `compareAndSet` doesn't work in jdk11? is there an equivalent function we could use? the problem with having a separate `get` and `set` calls is that multiple threads might end up trying to create the table if this method is called simultaneously, whereas `compareAndSet` ensures that only one thread will actually try to create the table. (also, we currently require jdk8, and there are likely to be other issues with using jdk11, so we could punt the jdk11 problems to later if we wanted)

Susheldon: > hmm, `compareAndSet` doesn't work in jdk11? is there an equivalent function we could use? the problem with having a separate `get` and `set` calls is that multiple threads might end up trying to create the table if this method is called simultaneously, whereas `compareAndSet` ensures that only one thread will actually try to create the table. (also, we currently require jdk8, and there are likely to be other issues with using jdk11, so we could punt the jdk11 problems to later if we wanted)

unsafe method is not visible to developers by default.If I remember correctly, this will get an error. As your world ，it looks like I can change this to `compareAndSet`

Susheldon: > hmm, `compareAndSet` doesn't work in jdk11? is there an equivalent function we could use? the problem with having a separate `get` and `set` calls is that multiple threads might end up trying to create the table if this method is called simultaneously, whereas `compareAndSet` ensures that only one thread will actually try to create the table. (also, we currently require jdk8, and there are likely to be other issues with using jdk11, so we could punt the jdk11 problems to later if we wanted)

ohhhhh,My mistake. I put the jdk8 code implementation into jdk11 to consider. In fact in jdk11 the implementation of compareAndSet has been modified to use the native method instead of the unsafe method, so I'm actually ok, I've changed the code.

Susheldon:
@elahrvivaz  Hi Emilio , I have a question, why does build-and-test always fail in my PR? I have not modified the content of the test code.

elahrvivaz (approved):
thanks for working through things!

elahrvivaz (changes requested):
oh, the build is failing due to one of the metadata impls not implementing the new method you added:
```
/home/runner/work/geomesa/geomesa/geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentMetadata.scala:24: class ConfluentMetadata needs to be abstract, since method resetCache in trait GeoMesaMetadata of type ()Unit is not defined
Error:  class ConfluentMetadata(val schemaRegistry: SchemaRegistryClient) extends GeoMesaMetadata[String] with LazyLogging {
```
(the error about needing to be abstract is wrong, you just need to add an impl for the resetCache() method)

Susheldon:
> oh, the build is failing due to one of the metadata impls not implementing the new method you added:
> 
> ```
> /home/runner/work/geomesa/geomesa/geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentMetadata.scala:24: class ConfluentMetadata needs to be abstract, since method resetCache in trait GeoMesaMetadata of type ()Unit is not defined
> Error:  class ConfluentMetadata(val schemaRegistry: SchemaRegistryClient) extends GeoMesaMetadata[String] with LazyLogging {
> ```
> 
> (the error about needing to be abstract is wrong, you just need to add an impl for the resetCache() method)

Sorry, my project only includes geomesa-hbase and related content. I have modified to rewrite the build with a complete project, and now it can be successful.

elahrvivaz (approved):

elahrvivaz:
merged as https://github.com/locationtech/geomesa/commit/9daf410159906b2a5bd7d59cc6e837c4f028e4f9

PR #2667, opened 2020-12-16
check duplicate attr before create schema.

Susheldon:
check duplicate attr before create schema.

Signed-off-by: xxxxs Susheldon@users.noreply.github.com

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
@@ -127,7 +127,7 @@ object PropertyCheck {
     val names = sft.getAttributeDescriptors.asScala.map(descriptor => descriptor.getLocalName)
     val duplicateProperties = names.diff(names.distinct)
     if ( duplicateProperties.nonEmpty ){
-      throw new IllegalArgumentException(s"duplicate attr: '${duplicateProperties.mkString(",")}' ")
+      throw new IllegalArgumentException(s"Schema contains duplicate attribute names: '${duplicateProperties.mkString(", ")}' ")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/index/GeoMesaSchemaValidator.scala
```diff
@@ -120,3 +121,13 @@ object IndexConfigurationCheck {
     require(sft.getAttributeShards > 0 && sft.getAttributeShards < 128, "Attribute shards must be between 1 and 127")
   }
 }
+
+object PropertyCheck {
+  def validateDuplicateProperty(sft: SimpleFeatureType):Unit = {
+    val names = sft.getAttributeDescriptors.asScala.map(descriptor => descriptor.getLocalName)
+    val duplicateProperties = names.diff(names.distinct)
+    if ( duplicateProperties.nonEmpty ){
+      throw new IllegalArgumentException(s"duplicate attr: '${duplicateProperties.mkString(",")}' ")
```
elahrvivaz, line 130: Could you make the error message a little clearer?

```suggestion
      throw new IllegalArgumentException(s"Schema contains duplicate attribute names: '${duplicateProperties.mkString(", ")}' ")
```

elahrvivaz:
Thanks for adding this, could you also add a quick unit test in https://github.com/locationtech/geomesa/blob/main/geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala?

Susheldon:
> Thanks for adding this, could you also add a quick unit test in https://github.com/locationtech/geomesa/blob/main/geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/geotools/GeoMesaDataStoreTest.scala?

Sorry, I’m a little confused, why add unit test in GeoMesaDataStoreTest? You means write a unit test for the entire process of createSchema?

elahrvivaz:
yes, for things like this it's good to have a unit test to verify that the fix is working, and to prevent any regression due to code changes in the future. in this case, it should be pretty simple, i.e. something like:
```
val sft = SimpleFeatureTypes.createType("foo:String,bar:Int,foo:Double")
val ds = new TestGeoMesaDataStore(true)
ds.createSchema(sft) must throwAn[IllegalArgumentException]
```

elahrvivaz (approved):

elahrvivaz:
merged as https://github.com/locationtech/geomesa/commit/13e53e7bbb3afafc550aed3a9d644056b369c4c9

PR #2669, opened 2020-12-18
GEOMESA-2971 Support configurable file sizes in FSDS

elahrvivaz:
* Add set/get kvs to FS metadata
* Add target file size to sft/create schema
* Update metadata serialization
* Deprecated and updated file system compaction api to allow for size
* Move FileSizeEstimator to geomesa-utils

### DIFF
--- a/docs/user/filesystem/index_config.rst
+++ b/docs/user/filesystem/index_config.rst
@@ -108,6 +108,35 @@ Leaf storage can be specified through the user data key ``geomesa.fs.leaf-storag
         // or set directly in the user data as a string
         sft.getUserData.put("geomesa.fs.leaf-storage", "false")
 
+Configuring Target File Size
+----------------------------
+
+By default data files can grow to unlimited size as more data is written and files are compacted. This may lead
+to poor performance, if a file becomes too large. To manage this, a target file size can be configured through
+the user data key ``geomesa.fs.file-size``:
+
+.. tabs::
+
+    .. code-tab:: java
+
+        import org.locationtech.geomesa.fs.storage.common.interop.ConfigurationUtils;
+
+        SimpleFeatureType sft = ...
+        // use the utility method
+        ConfigurationUtils.setTargetFileSize(sft, false);
+        // or set directly in the user data as a string
+        sft.getUserData().put("geomesa.fs.file-size", "1GB");
+
+    .. code-tab:: scala
+
+        import org.locationtech.geomesa.fs.storage.common.RichSimpleFeatureType
+
+        val sft: SimpleFeatureType = ???
+        // use the implicit method from RichSimpleFeatureType
+        sft.setTargetFileSize("1GB")
+        // or set directly in the user data as a string
+        sft.getUserData.put("geomesa.fs.file-size", "1GB")
+
 .. _fsds_metadata_config:
 
 Configuring Metadata Persistence
--- a/docs/user/filesystem/index_config.rst
+++ b/docs/user/filesystem/index_config.rst
@@ -108,6 +108,8 @@ Leaf storage can be specified through the user data key ``geomesa.fs.leaf-storag
         // or set directly in the user data as a string
         sft.getUserData.put("geomesa.fs.leaf-storage", "false")
 
+.. _fsds_file_size_config:
+
 Configuring Target File Size
 ----------------------------
 
@@ -138,7 +140,9 @@ the user data key ``geomesa.fs.file-size``:
         sft.getUserData.put("geomesa.fs.file-size", "1GB")
 
 Note that target file size can also be specified in some operations, which will override any default configured
-in the feature type. See :ref:`fsds_compact_command` and :ref:`fsds_ingest_command` for details.
+in the feature type. See :ref:`fsds_compact_command` and :ref:`fsds_ingest_command` for details. See
+:ref:`fsds_size_threshold_prop` for controlling the file size error margin.
+
 .. _fsds_metadata_config:
 
 Configuring Metadata Persistence
--- a/docs/user/filesystem/index_config.rst
+++ b/docs/user/filesystem/index_config.rst
@@ -137,6 +137,8 @@ the user data key ``geomesa.fs.file-size``:
         // or set directly in the user data as a string
         sft.getUserData.put("geomesa.fs.file-size", "1GB")
 
+Note that target file size can also be specified in some operations, which will override any default configured
+in the feature type. See :ref:`fsds_compact_command` and :ref:`fsds_ingest_command` for details.
 .. _fsds_metadata_config:
 
 Configuring Metadata Persistence
@@ -239,4 +241,3 @@ Observers can be specified through the user data key ``geomesa.fs.observers``:
         sft.setObservers(factories)
         // or set directly in the user data as a comma-delimited string
         sft.getUserData.put("geomesa.fs.observers", factories.mkString(","))
-
--- a/geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/data/FileSystemDataStore.scala
+++ b/geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/data/FileSystemDataStore.scala
@@ -76,11 +76,13 @@ class FileSystemDataStore(
           }
           deprecated.getOrElse(true)
         }
+        val fileSize = sft.removeTargetFileSize()
 
         val path = manager.defaultPath(sft.getTypeName)
         val context = FileSystemContext(fc, conf, path, namespace)
 
-        val metadata = StorageMetadataFactory.create(context, meta, Metadata(sft, encoding, scheme, leafStorage))
+        val metadata =
+          StorageMetadataFactory.create(context, meta, Metadata(sft, encoding, scheme, leafStorage, fileSize))
         try { manager.register(path, FileSystemStorageFactory(context, metadata)) } catch {
           case NonFatal(e) => CloseQuietly(metadata).foreach(e.addSuppressed); throw e
         }
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/package.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/package.scala
@@ -40,11 +40,33 @@ package object api {
     * Holder for the metadata defining a storage instance
     *
     * @param sft simple feature type
-    * @param encoding file encoding
     * @param scheme partition scheme configuration
-    * @param leafStorage leaf storage configuration
+    * @param config key-value configurations
     */
-  case class Metadata(sft: SimpleFeatureType, encoding: String, scheme: NamedOptions, leafStorage: Boolean)
+  case class Metadata(sft: SimpleFeatureType, scheme: NamedOptions, config: Map[String, String]) {
+    def encoding: String = config(Metadata.Encoding)
+    def leafStorage: Boolean = config(Metadata.LeafStorage).toBoolean
+    def targetFileSize: Option[Long] = config.get(Metadata.TargetFileSize).map(_.toLong)
+  }
+
+  object Metadata {
+
+    val Encoding       = "encoding"
+    val LeafStorage    = "leaf-storage"
+    val TargetFileSize = "target-file-size"
+
+    def apply(
+        sft: SimpleFeatureType,
+        encoding: String,
+        scheme: NamedOptions,
+        leafStorage: Boolean,
+        fileSize: Option[Long] = None): Metadata = {
+      val config: Map[String, String] =
+        Map(Encoding -> encoding, LeafStorage -> java.lang.Boolean.toString(leafStorage)) ++
+            fileSize.map(f => TargetFileSize -> java.lang.Long.toString(f)).toMap
+      Metadata(sft, scheme, config)
+    }
+  }
 
   /**
     * Case class holding a filter and partitions
@@ -65,6 +87,23 @@ package object api {
       * @param partition partition to compact, or all partitions
       * @param threads suggested threads to use for file system operations
       */
-    def compact(partition: Option[String], threads: Int = 1)
+    @deprecated("replaced with compact(Option[String], Option[Long], Int)")
+    def compact(partition: Option[String], threads: Int): Unit
+
+    /**
+     * Compact a partition - merge multiple data files into a single file.
+     *
+     * Care should be taken with this method. Currently, there is no guarantee for correct behavior if
+     * multiple threads or storage instances attempt to compact the same partition simultaneously.
+     *
+     * @param partition partition to compact, or all partitions
+     * @param fileSize approximate target size of files, in bytes
+     * @param threads suggested threads to use for file system operations
+     */
+    // noinspection ScalaDeprecation
+    def compact(partition: Option[String], fileSize: Option[Long] = None, threads: Int = 1): Unit = {
+      // default impl to prevent API breakage
+      compact(partition, threads)
+    }
   }
 }
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/java/org/locationtech/geomesa/fs/storage/common/interop/ConfigurationUtils.java
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/java/org/locationtech/geomesa/fs/storage/common/interop/ConfigurationUtils.java
@@ -11,6 +11,7 @@ package org.locationtech.geomesa.fs.storage.common.interop;
 import com.typesafe.config.ConfigFactory;
 import com.typesafe.config.ConfigRenderOptions;
 import com.typesafe.config.ConfigValueFactory;
+import org.locationtech.geomesa.utils.text.Suffixes.Memory$;
 import org.opengis.feature.simple.SimpleFeatureType;
 
 import java.util.List;
@@ -33,6 +34,11 @@ public class ConfigurationUtils {
 
     public static void setLeafStorage(SimpleFeatureType sft, boolean leafStorage) {
         sft.getUserData().put("geomesa.fs.leaf-storage", String.valueOf(leafStorage));
+    }
+
+    public static void setTargetFileSize(SimpleFeatureType sft, String size) {
+        Memory$.MODULE$.bytes(size).get(); // validate input
+        sft.getUserData().put("geomesa.fs.file-size", size);
     }
 
     public static void setScheme(SimpleFeatureType sft, String scheme, Map<String, String> options) {
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
@@ -17,14 +17,15 @@ import org.locationtech.geomesa.fs.storage.api.FileSystemStorage.{FileSystemUpda
 import org.locationtech.geomesa.fs.storage.api.StorageMetadata.StorageFileAction.StorageFileAction
 import org.locationtech.geomesa.fs.storage.api.StorageMetadata._
 import org.locationtech.geomesa.fs.storage.api._
-import org.locationtech.geomesa.fs.storage.common.AbstractFileSystemStorage.{FileSystemPathReader, MetadataObserver}
+import org.locationtech.geomesa.fs.storage.common.AbstractFileSystemStorage.{FileSystemPathReader, MetadataObserver, WriterConfig}
 import org.locationtech.geomesa.fs.storage.common.observer.FileSystemObserverFactory.CompositeObserver
 import org.locationtech.geomesa.fs.storage.common.observer.{FileSystemObserver, FileSystemObserverFactory}
 import org.locationtech.geomesa.fs.storage.common.utils.StorageUtils.FileType
+import org.locationtech.geomesa.fs.storage.common.utils.StorageUtils.FileType.FileType
 import org.locationtech.geomesa.fs.storage.common.utils.{PathCache, StorageUtils}
 import org.locationtech.geomesa.index.planning.QueryRunner
 import org.locationtech.geomesa.utils.collection.CloseableIterator
-import org.locationtech.geomesa.utils.io.{CloseQuietly, FlushQuietly, WithClose}
+import org.locationtech.geomesa.utils.io.{CloseQuietly, FileSizeEstimator, FlushQuietly, WithClose}
 import org.locationtech.geomesa.utils.stats.MethodProfiling
 import org.locationtech.jts.geom.{Envelope, Geometry}
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
@@ -44,7 +45,7 @@ abstract class AbstractFileSystemStorage(
     val context: FileSystemContext,
     val metadata: StorageMetadata,
     extension: String
-  ) extends FileSystemStorage with MethodProfiling with LazyLogging {
+  ) extends FileSystemStorage with SizeableFileSystemStorage with MethodProfiling with LazyLogging {
 
   // don't require observers if we never write any data
   lazy private val observers = {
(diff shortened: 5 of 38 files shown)


docs/user/filesystem/index_config.rst
```diff
@@ -108,6 +108,35 @@ Leaf storage can be specified through the user data key ``geomesa.fs.leaf-storag
...
+
+    .. code-tab:: scala
+
+        import org.locationtech.geomesa.fs.storage.common.RichSimpleFeatureType
+
+        val sft: SimpleFeatureType = ???
+        // use the implicit method from RichSimpleFeatureType
+        sft.setTargetFileSize("1GB")
+        // or set directly in the user data as a string
+        sft.getUserData.put("geomesa.fs.file-size", "1GB")
```
jnh5y, line 138: Ok, so if the SFT has one size and the compact command is run with a different one, the one specified with the command will be used, right?

While that's obvious, I wonder if it is worth calling that out specifically.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/package.scala
```diff
@@ -65,6 +87,23 @@ package object api {
       * @param partition partition to compact, or all partitions
       * @param threads suggested threads to use for file system operations
       */
-    def compact(partition: Option[String], threads: Int = 1)
+    @deprecated("replaced with compact(Option[String], Option[Long], Int)")
+    def compact(partition: Option[String], threads: Int): Unit
```
jnh5y, line 91: Should the signature remain 
```def compact(partition: Option[String], threads: Int = 1)```?

If it does not, wouldn't this break folks calling out with like
`compact(partition)`?

File: geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/java/org/locationtech/geomesa/fs/storage/common/interop/ConfigurationUtils.java [hunk 3]
```diff
@@ -35,6 +36,11 @@ public static void setLeafStorage(SimpleFeatureType sft, boolean leafStorage) {
         sft.getUserData().put("geomesa.fs.leaf-storage", String.valueOf(leafStorage));
     }
 
+    public static void setTargetFileSize(SimpleFeatureType sft, String size) {
+        Memory$.MODULE$.bytes(size).get(); // validate input
+        sft.getUserData().put("geomesa.fs.file-size", size);
```
jnh5y, line 41: While I see this is consistent, is there any way to avoid the magic string and reference the static variable instead?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/java/org/locationtech/geomesa/fs/storage/common/interop/ConfigurationUtils.java
```diff
@@ -35,6 +36,11 @@ public static void setLeafStorage(SimpleFeatureType sft, boolean leafStorage) {
         sft.getUserData().put("geomesa.fs.leaf-storage", String.valueOf(leafStorage));
     }
 
+    public static void setTargetFileSize(SimpleFeatureType sft, String size) {
+        Memory$.MODULE$.bytes(size).get(); // validate input
```
jnh5y, line 40: Hmm.... looks like the size bit may be should be factored out into a common place/name (rather than being under "Memory")?

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/FileSizeEstimator.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Estimate how many features to write to hit our target size
+   *
+   * @param written number of bytes written so far
+   * @return
+   */
+  def estimate(written: Long): Int = {
+    val e = math.round((target - written) / estimate)
+    if (e < 1) { 1 } else if (e.isValidInt) { e.intValue() } else { Int.MaxValue}
```
jnh5y, line 36: Two small things:
1.  Space after `Int.MaxValue`....:)
2.  Does this return an Int so that it can be used in control flow?  (E.g., there's no reasonable way to write a Longs-worth of features?)

docs/user/filesystem/index_config.rst
```diff
@@ -108,6 +108,35 @@ Leaf storage can be specified through the user data key ``geomesa.fs.leaf-storag
...
+
```
elahrvivaz, line 139:
```suggestion

Note that target file size can also be specified in some operations, which will override any default configured in the feature type. See :ref:`fsds_compact_command` and :ref:`fsds_ingest_command` for details.
```

docs/user/filesystem/index_config.rst
elahrvivaz: added a note

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/package.scala
elahrvivaz: I think scala won't let you have default values for overloaded functions. But since the new method has defaults for both `threads` and `fileSize` that invocation will still work. I think all of the following will still work:
`compact(partition)`, `compact(partition, 1)`, `compact(partition, threads = 1)`

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/java/org/locationtech/geomesa/fs/storage/common/interop/ConfigurationUtils.java
elahrvivaz: i think there may be some semantic differences when discussing size in memory vs disk, but not entirely sure...

elahrvivaz: i didn't think the difference was worth changing anything

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/java/org/locationtech/geomesa/fs/storage/common/interop/ConfigurationUtils.java
elahrvivaz: i spent about 15 minutes trying to get this to work and couldn't... supposedly `org.locationtech.geomesa.fs.storage.common.package$StorageKeys$.MODULE$.FileSizeKey()` should work but that doesn't compile, even though the class shows up in scalap and javap. meh

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/io/FileSizeEstimator.scala
elahrvivaz:
```suggestion
    if (e < 1) { 1 } else if (e.isValidInt) { e.intValue() } else { Int.MaxValue }
```

elahrvivaz: right, it's used as a loop control variable. most of the time, it will re-calculate after looping for that many results. So you can write a longs-worth of values, but you have to reset the loop counter a few times. that's not true for some file formats (like parquet), because you can't get an accurate file size until you actually close the file. so in those cases, you wouldn't be able to write more than Int.MaxValue features. I would guess most of the uses of this are in while loops, so we could probably make it a long instead if you think it's worthwhile. At an pretty conservative 10 bytes per feature, using an int would still give you about 20GB max file size.

elahrvivaz: i went ahead and updated it to a long

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/data/FileSystemDataStore.scala
```diff
@@ -76,11 +76,13 @@ class FileSystemDataStore(
           }
           deprecated.getOrElse(true)
         }
+        val fileSize = sft.removeTargetFileSize()
```
jnh5y, line 79: Out of curiosity, why remove the filesize from the SFT?  

I can imagine it is reasonable; just want to discuss it.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/package.scala
jnh5y: Ah!  I missed that the new function has two default parameters (which Scala likely desugars to 3 different functions).  Makes sense!

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
```diff
@@ -187,32 +211,87 @@ abstract class AbstractFileSystemStorage(
...
+    override def write(feature: SimpleFeature): Unit = {
+      if (writer == null) {
+        val config = paths.next
+        path = config.path
+        writer = createWriter(config)
+      }
+      writer.write(feature)
+      count += 1
+      remaining -= 1
+      if (remaining == 0) {
```
jnh5y, line 277: Just to make sure I'm reading this correctly... we are only updating the estimate after we've written how many ever records that the estimator initially suggested that we write, correct?

jnh5y: The question I have is if we should check earlier or more often?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/SizeableFileSystemStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+}
+
+object SizeableFileSystemStorage {
+
+  val BytesPerFeature  = "bytes-per-feature"
+  val UseDynamicSizing = "use-dynamic-sizing"
+
+  val FileSizeErrorThreshold: SystemProperty = SystemProperty("geomesa.fs.size.threshold", "0.05")
```
jnh5y, line 75: Is this system property documented?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/jobs/PartitionOutputFormat.scala
```diff
@@ -51,49 +54,135 @@ class PartitionOutputFormat(delegate: SingleFileOutputFormat) extends OutputForm
...
+      def write(key: Void, value: SimpleFeature): Unit = {
+        val geom = value.getDefaultGeometry.asInstanceOf[Geometry]
+        if (geom != null) {
+          bounds.expandToInclude(geom.getEnvelopeInternal)
+        }
+        count += 1L
+      }
+
+      def meta: PartitionMetadata = {
+        val f = files.map(StorageFile(_, System.currentTimeMillis()))
```
jnh5y, line 119: Do we care about calling `System.currentTimeMills()` repeatedly?  (It seems like it could return multiple different times if the operation spans more than a millisecond.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/jobs/PartitionOutputFormat.scala
```diff
@@ -51,49 +54,135 @@ class PartitionOutputFormat(delegate: SingleFileOutputFormat) extends OutputForm
...
+        count += 1
+        remaining -= 1
+        if (remaining == 0) {
+          writer.close(context)
+          writer = null
+          // adjust our estimate to account for the actual bytes written
+          total += storage.context.fc.getFileStatus(path).getLen
+          estimator.update(total, count)
+          remaining = estimator.estimate(0L)
+        }
```
jnh5y, line 176: I'm only skimming.  Seems like this logic is in the ChunkedFileWriter as well?  Is there a way to refactor it?  (Says the guy who cannot quite find things anymore after all the clever refactoring.)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
```diff
@@ -65,21 +67,34 @@ import org.opengis.feature.simple.SimpleFeatureType
...
+
+  override val scheme: PartitionScheme = PartitionSchemeFactory.load(sft, meta.scheme)
+  override val encoding: String = meta.config(Metadata.Encoding)
+  override val leafStorage: Boolean = meta.config(Metadata.LeafStorage).toBoolean
+
+  private val kvs = new ConcurrentHashMap[String, String](meta.config.asJava)
+
+  override def get(key: String): Option[String] = Option(kvs.get(key))
+
+  override def set(key: String, value: String): Unit = {
```
jnh5y, line 91: Sanity check question:  Do we have a unit tests which cover changes to the FSDS JDBC metdata, etc, etc?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/MetadataSerialization.scala
```diff
@@ -77,16 +87,19 @@ object MetadataSerialization extends MethodProfiling {
...
+    case class StoragePersistenceV2(
+        featureType: Config, partitionScheme: PartitionSchemeConfig, encoding: String, leafStorage: Boolean)
     case class StoragePersistenceV1(featureType: Config, partitionScheme: PartitionSchemeConfig, encoding: String)
     case class PartitionSchemeConfig(scheme: String, options: Map[String, String])
 
     implicit val PartitionSchemeConfigConvert: ConfigConvert[PartitionSchemeConfig] =
       deriveConvert[PartitionSchemeConfig]
     implicit val StoragePersistenceV1Convert: ConfigConvert[StoragePersistenceV1] =
       deriveConvert[StoragePersistenceV1]
+    implicit val StoragePersistenceV2Convert: ConfigConvert[StoragePersistenceV2] =
```
jnh5y, line 101: Since this PR is introducing a new metadata storage format version do we need to provide any notes about it in the upgrade guide?  Any utilities or notes about updating the metadata?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/test/scala/org/locationtech/geomesa/parquet/CompactionTest.scala
```diff
@@ -93,7 +101,18 @@ class CompactionTest extends Specification with AllExpectations {
...
+      // compact to a given file size
+      // verify if file is appropriately sized, it won't be modified
+      val paths = fsStorage.getFilePaths(partition).map(_.path)
+      val size = paths.map(f => fc.getFileStatus(f).getLen).sum
+      fsStorage.compact(Some(partition), Some(size))
+      fsStorage.getFilePaths(partition).map(_.path) mustEqual paths
+      // verify files are split into smaller ones
+      fsStorage.compact(Some(partition), Some(size / 2))
+      fsStorage.metadata.getPartition(partition).map(_.files.length).getOrElse(0) must beGreaterThan(1)
+      SelfClosingIterator(fsStorage.getReader(Query.ALL, Some(partition))).toList must haveSize(299)
```
jnh5y, line 115: I didn't read closely enough.  What's going on with the feature count (seemingly) going from 300 to 299?

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/compact/FileSystemCompactionJob.scala
```diff
@@ -77,7 +80,14 @@ trait FileSystemCompactionJob extends StorageConfiguration with JobWithLibJars {
     configureOutput(storage.metadata.sft, job)
 
     // save the existing files so we can delete them afterwards
-    val existingDataFiles = partitions.map(p => (p, storage.getFilePaths(p.name).toList)).toList
+    // mimic the filtering done in PartitionInputFormat
+    val sizeable = Option(storage).collect { case s: SizeableFileSystemStorage => s }
+    val sizeCheck = sizeable.flatMap(s => s.targetSize(targetFileSize).map(t => (p: Path) => s.fileIsSized(p, t)))
+    val existingDataFiles = partitions.toList.flatMap { p =>
+      val files = storage.getFilePaths(p.name).filterNot(f => sizeCheck.exists(_.apply(f.path)))
+      // TODO get counts right... use m/r counters?
```
jnh5y, line 88: Is there a ticket we could enter for this?

geomesa-fs/geomesa-fs-tools/src/test/scala/org/locationtech/geomesa/fs/tools/ingest/CompactCommandTest.scala
```diff
@@ -12,143 +12,182 @@ import java.nio.file.{Files, Path}
...
-  val features2: Seq[ScalaSimpleFeature] = Seq.tabulate(10) { j =>
-    val i = j+10
-    ScalaSimpleFeature.create(sft, s"$i", s"test$i", 100 + i, s"2017-06-0${5 + (j % 3)}T04:03:02.0001Z", s"MULTILINESTRING((0 0, 10 10.$j))",
-      pt, line, polygon, mpt, mline, mpolygon)
-  }
+  val numFeatures = 10000
+  val targetFileSize = 12000L // kind of a magic number, in that it divides up the features into files fairly evenly with no remainder
 
-  def getDataStore: DataStore = {
+  lazy val ds = {
```
jnh5y, line 66: Can the one FSDS read from both Orc and Parquet formats?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
```diff
@@ -466,12 +467,13 @@ object ExportCommand extends LazyLogging {
...
       exporter.start(sft)
-      count = 0
+      count = 0L
     }
 
     @tailrec
     private def export(features: Iterator[SimpleFeature], result: Option[Long]): Option[Long] = {
-      val counter = features.take(estimator.estimate(exporter.bytes)).map { f => count += 1; f }
+      var estimate = estimator.estimate(exporter.bytes)
+      val counter = features.takeWhile { _ => count += 1; estimate -= 1; estimate >= 0 }
```
jnh5y, line 476: Neat!  The export command can take a file size limit now?  

Is that documented?  What happens when the file overflows?

geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/data/FileSystemDataStore.scala
elahrvivaz: this is a pattern we use for several of the fs config values. for convenience, we allow them to be set in the simple feature type, but then they end up getting persisted separately in the metadata class, so we don't want to keep 2 (potentially conflicting) values around. there's a comment at the top of the section that you can't see in the pr: https://github.com/locationtech/geomesa/blob/3273e45b6c06760bc7adcf9f1a4612a3315d8f68/geomesa-fs/geomesa-fs-datastore/src/main/scala/org/locationtech/geomesa/fs/data/FileSystemDataStore.scala#L62

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
elahrvivaz: in the case of parquet and orc, we can't actually check the file size until the file is closed, as there is a bunch of metadata/overhead that gets written at that point. but to answer your broader question, there's a trade-off between the getting the file the exact size and the cost of checking the file size as you go, which requires some I/O. so the approach taken here is to estimate an initial number of features, and then check the file size after that, and refine the estimate for the next iteration. the fs storage system will also persist the estimate (unless you disable it) so that files will stay a consistent size between invocations.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/SizeableFileSystemStorage.scala
elahrvivaz: nope, will add, thanks

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/jobs/PartitionOutputFormat.scala
elahrvivaz: it's ok if the times aren't the same, but yeah seems like we could save a few calls. I'll update.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/jobs/PartitionOutputFormat.scala
elahrvivaz: it is very similar, but there are enough differences in how the paths are generated, the writers are closed (the output format requires `close(context)`), etc that i didn't think it was worthwhile to wrap a new trait around the two. the logic is also not terribly complex, although not entirely trivial either. i could be persuaded to unify it if you feel strongly...

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
elahrvivaz: we do actually have a unit test, i will add some new test cases for these changes.

elahrvivaz: and... the test failed so good catch :)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/MetadataSerialization.scala
elahrvivaz: no, this should be transparent to the end user.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/test/scala/org/locationtech/geomesa/parquet/CompactionTest.scala
elahrvivaz: there's a feature being deleted up on lines 96-100

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/compact/FileSystemCompactionJob.scala
elahrvivaz: i'm not sure it matters at the moment - we should probably revisit this more holistically as part of GEOMESA-2961, 2962 and 2963 (i.e. if/when we start using the metadata values for query file pruning). I updated 2962 with a note about this.

geomesa-fs/geomesa-fs-tools/src/test/scala/org/locationtech/geomesa/fs/tools/ingest/CompactCommandTest.scala
elahrvivaz: yes!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
elahrvivaz: yeah, that's been a feature for a while. it is documented, look for 'chunk-size'. it will spill over to a new file when it hits the limit.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/jobs/PartitionOutputFormat.scala
jnh5y: Since the code is slightly different, I don't feel terribly strongly about a unification.  (If there were a trivial refactoring, yeah, maybe....)

Seems reasonable to leave it as is!

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
jnh5y: Whew!  Touchdown reviewer!  Give me enough time and even I'll get lucky and turn over a bug!

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/MetadataSerialization.scala
jnh5y: Any chance of issues with mixing and matching formats in the same directory?  E.g. if I wrote to the same feature type with multiple versions, would that work?

geomesa-fs/geomesa-fs-tools/src/test/scala/org/locationtech/geomesa/fs/tools/ingest/CompactCommandTest.scala
jnh5y: Neat!

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/MetadataSerialization.scala
elahrvivaz: no issue - if you write data with a newer version you won't be able to read it with an older version, which is (for better or worse) consistent with the rest of the project. i'm not sure of any way to change the format and still have older clients be able to read it, but i'm open to suggestions.

jnh5y: Hmm... this change is additive right?  It only adds configuration to the metadata.  So it seems surprising that a reader would not be able to read data that's been written using a new, optional write-path configuration.  *blinks*

I don't have any particular suggestions at the minute.  I suppose I'd have to see the error which an older reader throws while trying to read metadata from this PR.

In some sense, I'd love to see 3.x GM FSDS tools which are "mix and match"....  (I suppose I'd argue that the gm cli tools give a kind of API and not being able to mix and match represents an API break that'd be great to capture with a major version bump.  I'm not firm on this stance; just worth discussing for a moment.)

elahrvivaz: the 'encoding' and 'leaf storage' options were relocated, so that's why earlier versions can't read the new files. given our current data compatibility guarantees (https://www.geomesa.org/documentation/stable/user/upgrade.html#data-compatibility), we don't support mix-and-match between minor versions. we could change that and enforce things more strictly, but that's a larger discussion. i've added a note to the upgrade guide about the breaking change.

PR #2676, opened 2021-01-07
GEOMESA-2977 Option to query z-filters in 1.3 Accumulo installs

elahrvivaz:
* Use QueryHints.FILTER_COMPAT = "1.3"
* Supports basic non-strict date time queries against z3 index
* Supports basic queries against xz3 index
* Does not support transforms or advanced queries

### DIFF
--- a/docs/user/datastores/query_config.rst
+++ b/docs/user/datastores/query_config.rst
@@ -106,6 +106,38 @@ QueryHints.EXACT_COUNT ``Boolean`` ``true`` or ``false``
 
         ...&viewparams=EXACT_COUNT:true
 
+Filter Compatibility
+--------------------
+
+GeoMesa provides a limited compatibility mode, which allows for using a newer client version with
+an older distributed-runtime version, for back-ends that have a distributed installation. Currently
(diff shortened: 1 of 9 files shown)


jnh5y:
Ok, so the z2 index is not supported, right?  

How hard would it be to throw an exception for non-supported combinations?

jnh5y:
Would it be worth adding an optional bit to the Query Explainer output to show that a back compatible scan is being run?

jnh5y:
If we were gonna be nice, I could imagine running the remote version code during DataStore creation and automatically setting the hint:).  *shrugs*  That's probably a ton of work going forward.

elahrvivaz:
> If we were gonna be nice, I could imagine running the remote version code during DataStore creation and automatically setting the hint:). _shrugs_ That's probably a ton of work going forward.

Unfortunately getting the remote version is kind of slow... we used to load it up when creating a data store but now it's lazily done I think. but i think we'd have to provide better back-compatibility support than this to justify that.

elahrvivaz:
> Ok, so the z2 index is not supported, right?
> 
> How hard would it be to throw an exception for non-supported combinations?

Across all data stores, or just Accumulo? it would be a lot of checks...

PR #2677, opened 2021-01-08
GEOMESA-2979 Pass QUERY_INDEX hint to stats and getCount

jnh5y:

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreStatsTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreStatsTest.scala
@@ -19,6 +19,7 @@ import org.geotools.geometry.jts.ReferencedEnvelope
 import org.geotools.util.factory.Hints
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.accumulo.TestWithMultipleSfts
+import org.locationtech.geomesa.accumulo.index.AccumuloJoinIndex
 import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.locationtech.geomesa.index.conf.QueryHints.{EXACT_COUNT, QUERY_INDEX}
 import org.locationtech.geomesa.index.index.z2.Z2Index
@@ -399,9 +400,10 @@ class AccumuloDataStoreStatsTest extends Specification with TestWithMultipleSfts
         // this doesn't seem to happen with regular accumulo
         // so only test one of each query type here
 
-        // deleting the gm.AccumuloDataStoreStatsTest_AccumuloDataStoreStatsTest0_join_name_geom_dtg_v8 table
-        // to show that the QUERY_INDEX hint is being passed through
-        ds.connector.tableOperations().delete("gm.AccumuloDataStoreStatsTest_AccumuloDataStoreStatsTest0_join_name_geom_dtg_v8")
+        // deleting the "name" index table to show that the QUERY_INDEX hint is being passed through
+        ds.manager.indices(sft).collectFirst {
+          case i: AccumuloJoinIndex if i.attributes.head == "name" => i.getTableNames().head
+        }.foreach { ds.connector.tableOperations().delete(_) }
 
         val filters = Seq("bbox(geom,0,0,10,5)", "name < '7'")
         forall(filters.map(ECQL.toFilter)) { filter =>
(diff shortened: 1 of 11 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
```diff
@@ -59,7 +61,11 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
     import org.locationtech.geomesa.index.conf.QueryProperties.QueryExactCount
 
     val useExactCount = query.getHints.isExactCount.getOrElse(QueryExactCount.get.toBoolean)
-    val count = ds.stats.getCount(getSchema, query.getFilter, useExactCount).getOrElse(-1L)
+    val hints = new Hints()
+    query.getHints.getRequestedIndex.foreach { index => hints.put(QUERY_INDEX, index) }
+    hints.put(COST_EVALUATION, query.getHints.getCostEvaluation)
```
jnh5y, line 66: I don't know if this should be passed through or not.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreStatsTest.scala
```diff
@@ -389,6 +392,37 @@ class AccumuloDataStoreStatsTest extends Specification with TestWithMultipleSfts
...
+
+      "calculate exact counts with index query hints" >> {
+        // mock accumulo seems to really struggle with setting up the stat iterator
+        // the call to batchScanner.iterator can take several seconds
+        // this doesn't seem to happen with regular accumulo
+        // so only test one of each query type here
+
+        // deleting the gm.AccumuloDataStoreStatsTest_AccumuloDataStoreStatsTest0_join_name_geom_dtg_v8 table
+        // to show that the QUERY_INDEX hint is being passed through
+        ds.connector.tableOperations().delete("gm.AccumuloDataStoreStatsTest_AccumuloDataStoreStatsTest0_join_name_geom_dtg_v8")
```
elahrvivaz, line 404: it would probably be good to create a new schema for just this test, so that other tests don't expect the table and die. also, to insulate from table naming / version changes, you should be able to get the table name like:
```
ds.manager.indices(sft).collectFirst {
  case i: AccumuloJoinIndex if i.attributes.head == "foo" => i.getTableNames().head
}
```

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
elahrvivaz: seems like a good idea

elahrvivaz: what do you think of having a Seq of hints to copy in the companion object? So something like:
```
val CountHints = Seq(QUERY_INDEX, COST_EVALUATION)
...
CountHints.foreach(key => if (query.getHints.contains(key)) { hints.put(key, query.getHints.get(key)) })
```
would make it easier to add/modify the hints we copy

jnh5y: Sold!

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreStatsTest.scala
jnh5y: I like the ds.manager call!  I agree it'd be good to create a new schema just for this; that said, this test is already relying on the order of the blocks, so I don't feel bad about leaving it as is.  Thoughts?

elahrvivaz: :shrug:

elahrvivaz (approved):
shall i go ahead and merge?

jnh5y:
Yes, I think this PR can be merged now.

elahrvivaz:
9f8e833686ced341c23845bb38157bda46c5d48c

PR #2683, opened 2021-01-15
GEOMESA-2985 FeatureSource.getCount implementations should represent …

jnh5y:
…maxFeatures

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
@@ -19,6 +19,7 @@ import org.geotools.data._
 import org.geotools.data.simple.{SimpleFeatureCollection, SimpleFeatureIterator, SimpleFeatureSource}
 import org.geotools.feature.collection.SortedSimpleFeatureCollection
 import org.geotools.geometry.jts.ReferencedEnvelope
+import org.locationtech.geomesa.index.conf.QueryProperties.QueryExactCountMaxFeatures
 import org.locationtech.geomesa.index.geotools.GeoMesaFeatureSource.{DelegatingResourceInfo, GeoMesaQueryCapabilities}
 import org.locationtech.geomesa.index.planning.QueryRunner
 import org.locationtech.geomesa.index.stats.HasGeoMesaStats
@@ -61,10 +62,10 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
 
     val useExactCount = query.getHints.isExactCount.getOrElse(QueryExactCount.get.toBoolean)
 
-    val count = if (query.getMaxFeatures < 1000) {
+    val count = if (useExactCount && query.getMaxFeatures < QueryExactCountMaxFeatures.get.toInt) {
       SelfClosingIterator(getFeatures(query)).size
     } else {
-      ds.stats.getCount(getSchema, query.getFilter, useExactCount).getOrElse(-1L)
+      math.min(ds.stats.getCount(getSchema, query.getFilter, useExactCount).getOrElse(-1L), query.getMaxFeatures)
     }
 
     if (count > Int.MaxValue) {
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
@@ -62,10 +62,17 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
 
     val useExactCount = query.getHints.isExactCount.getOrElse(QueryExactCount.get.toBoolean)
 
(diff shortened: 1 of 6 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
```diff
@@ -59,7 +60,12 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
     import org.locationtech.geomesa.index.conf.QueryProperties.QueryExactCount
 
     val useExactCount = query.getHints.isExactCount.getOrElse(QueryExactCount.get.toBoolean)
-    val count = ds.stats.getCount(getSchema, query.getFilter, useExactCount).getOrElse(-1L)
+
+    val count = if (query.getMaxFeatures < 1000) {
+      SelfClosingIterator(getFeatures(query)).size
+    } else {
+      ds.stats.getCount(getSchema, query.getFilter, useExactCount).getOrElse(-1L)
```
elahrvivaz, line 67: we should have a math.min here right?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
```diff
@@ -59,7 +60,12 @@ class GeoMesaFeatureSource(val ds: DataStore with HasGeoMesaStats,
     import org.locationtech.geomesa.index.conf.QueryProperties.QueryExactCount
 
     val useExactCount = query.getHints.isExactCount.getOrElse(QueryExactCount.get.toBoolean)
-    val count = ds.stats.getCount(getSchema, query.getFilter, useExactCount).getOrElse(-1L)
+
+    val count = if (query.getMaxFeatures < 1000) {
```
elahrvivaz, line 64: maybe make the threshold a sys prop?

jnh5y: Either a system property or just change this to `!query.isMaxFeaturesUnlimited`.

Thoughts?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
jnh5y: Technically yes.  That'd avoid a situation where GM estimates a count larger than the maxFeatures.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureSource.scala
elahrvivaz: i'd go with sys prop, just to avoid regression with someone asking for a large but not unlimited number of features

jnh5y: geomesa.exact.count.max.features - When EXACT_COUNT=true and maxFeatures is lower than this setting, GeoMesa will run the query to determine how many records are in the result set.  Otherwise, GeoMesa will use the Stats API and respect the EXACT_COUNT setting.

elahrvivaz (approved):

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedFeatureSourceView.scala
```diff
@@ -45,8 +45,14 @@ class MergedFeatureSourceView(
 
   override def getSchema: SimpleFeatureType = sft
 
-  override def getCount(query: Query): Int =
-    sources.map { case (source, filter) => source.getCount(mergeFilter(query, filter)) }.sum
+  override def getCount(query: Query): Int = {
+    val total = sources.map { case (source, filter) => source.getCount(mergeFilter(query, filter)) }.sum
```
elahrvivaz, line 49: just thought of another issue... if some of these return -1 (indicating count is expensive), the result might be unexpected (i.e. -1 + 1 = 0). should we just return -1 in that case, or filter out the -1's and return the stores that return counts? or something else?

jnh5y: Maybe something like
`val total = sources.map { case (source, filter) => source.getCount(mergeFilter(query, filter)) }.filter(_<0).sum` ?

jnh5y: This is a good corner case to think through!  Good eye!

elahrvivaz: since I think you're out, i made an executive decision to return -1 if any of the stores return -1, on the theory that `undefined + n = undefined`...

elahrvivaz:
merged as 6a286d6929a16b11d304316f3b2b0f5edf31d7e0

PR #2685, opened 2021-01-18
GEOMESA-2986 Check for histogram bins when creating stats

elahrvivaz:
* Separate binning logic so we don't have to allocate an array to check bounds

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Histogram.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Histogram.scala
@@ -13,10 +13,12 @@ import java.util.Date
 import com.typesafe.scalalogging.LazyLogging
 import org.locationtech.jts.geom.{Coordinate, Geometry}
 import org.locationtech.geomesa.utils.geotools.GeometryUtils
+import org.locationtech.geomesa.utils.stats.BinnedArray.Binning
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.collection.immutable.ListMap
 import scala.reflect.ClassTag
+import scala.util.Try
 
 /**
   * The histogram's state is stored in an indexed array, where the index is the bin number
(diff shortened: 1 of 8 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/stats/Histogram.scala
```diff
@@ -205,6 +207,17 @@ object Histogram {
...
+  /**
+   * Checks if 2 values map to the same bin
+   *
+   * @param o1 object 1
+   * @param o2 object 2
+   * @param bins number of binds
+   * @tparam T type binding
+   * @return
+   */
+  def equivalent[T: ClassTag](o1: T, o2: T, bins: Int): Boolean = Try(Binning(bins, (o1, o2))).isFailure
```
jnh5y, line 219: Ok....  *blinks*  What is this doing?

Binning can fail for multiple reasons, right?  Does this definitely check equivalence?

elahrvivaz: ah, i think the only way it would fail is if the endpoints end up mapping to the same bin. this is admittedly not the most elegant solution... but because you have to handle each binding differently, it's the easiest. previously we just compared the endpoints, but that doesn't handle the case where they are different but map to the same bin (which was the underlying bug).

elahrvivaz: the rest of the changes are to separate out the logic for this with actually allocating the binned array, so that we're not allocating an array here.

elahrvivaz:
```suggestion
  def equivalent[T: ClassTag](o1: T, o2: T, bins: Int): Boolean =
     Try(Binning(bins, (o1, o2))).isFailure // this will fail if the two values map to the same bin
```

PR #2689, opened 2021-01-21
GEOMESA-2990 Fix update writes that only modify visibility

elahrvivaz:


michaelchehreh-ccri:
This looks good to me and Joseph. We haven't necessarily used this exact commit, but we used his version of it (more or less the same) when fixing those visibilities

PR #2692, opened 2021-01-30
GEOMESA-2996 Fix FSDS covering partitions for montly partition scheme

elahrvivaz:

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/DateTimeScheme.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/DateTimeScheme.scala
@@ -8,8 +8,8 @@
 
 package org.locationtech.geomesa.fs.storage.common.partitions
 
-import java.time.format.DateTimeFormatter
-import java.time.temporal.{ChronoUnit, TemporalAdjusters}
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
+import java.time.temporal.{ChronoField, ChronoUnit, TemporalAdjusters, WeekFields}
 import java.time.{ZoneOffset, ZonedDateTime}
 import java.util.Date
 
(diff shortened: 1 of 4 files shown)


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/DateTimeScheme.scala
```diff
@@ -172,20 +174,37 @@ object DateTimeScheme {
...
 
-    object Minute       extends Format("minute",        "yyyy/MM/dd/HH/mm", ChronoUnit.MINUTES)
-    object Hourly       extends Format("hourly",        "yyyy/MM/dd/HH",    ChronoUnit.HOURS  )
-    object Daily        extends Format("daily",         "yyyy/MM/dd",       ChronoUnit.DAYS   )
-    object Weekly       extends Format("weekly",        "yyyy/ww",          ChronoUnit.WEEKS  )
-    object Monthly      extends Format("monthly",       "yyyy/MM",          ChronoUnit.MONTHS )
-    object JulianMinute extends Format("julian-minute", "yyyy/DDD/HH/mm",   ChronoUnit.MINUTES)
-    object JulianHourly extends Format("julian-hourly", "yyyy/DDD/HH",      ChronoUnit.HOURS  )
-    object JulianDaily  extends Format("julian-daily",  "yyyy/DDD",         ChronoUnit.DAYS   )
+    val Minute       : Format = Format("minute",        "yyyy/MM/dd/HH/mm", ChronoUnit.MINUTES)
```
jnh5y, line 186: Would changing from an object to val cause a small, external "api" change?

elahrvivaz: possibly? i'm not sure if there are any actual differences between a val and an object in this case. but i don't think this qualifies as part of the public api - the configuration is still the same. if you object, I can change it back, it just seemed a little odd that they were objects.

jnh5y: We can leave it.  I'm on a streak where I try to think through when an API _may_ be changing.  :)

PR #2694, opened 2021-02-02
GEOMESA-3000 Use Nailgun for CLI

elahrvivaz:
* Set GEOMESA_NG_ENABLED=true to enable
* Nailgun sever is started automatically and will shutdown with inactivity
* Refactor binary dist assembly to common archetype
* Upgrade jcommander to 1.78 to support repeated invocations


build/cqs.tsv
```diff
@@ -2,12 +2,13 @@ asm:asm	3.1	compile
 com.amazonaws:aws-java-sdk-cloudwatch	1.11.179	compile
 com.amazonaws:aws-java-sdk-core	1.11.179	compile
 com.amazonaws:jmespath-java	1.11.179	compile
-com.beust:jcommander	1.48	compile
+com.beust:jcommander	1.78	compile
```
elahrvivaz, line 5: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22738

build/cqs.tsv
```diff
@@ -116,6 +117,8 @@ javax.measure:unit-api	1.0	compile
 javax.xml.bind:jaxb-api	2.2.7	compile
 joda-time:joda-time	2.3	compile
 log4j:log4j	1.2.17	compile
+net.java.dev.jna:jna	4.5.2	compile
```
elahrvivaz, line 120: 4.5.1 is approved... is that close enough? https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15644

build/cqs.tsv
```diff
@@ -116,6 +117,8 @@ javax.measure:unit-api	1.0	compile
...
+net.java.dev.jna:jna-platform	4.5.2	compile
```
elahrvivaz, line 121: 4.5.1: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15645

build/cqs.tsv
elahrvivaz: https://clearlydefined.io/definitions/maven/mavencentral/net.java.dev.jna/jna/4.5.2

build/cqs.tsv
elahrvivaz: https://clearlydefined.io/definitions/maven/mavencentral/net.java.dev.jna/jna-platform/4.5.2

build/cqs.tsv
```diff
@@ -2,12 +2,13 @@ asm:asm	3.1	compile
...
 com.amazonaws:aws-java-sdk-core	1.11.179	compile
 com.amazonaws:jmespath-java	1.11.179	compile
-com.beust:jcommander	1.48	compile
+com.beust:jcommander	1.78	compile
 com.chuusai:shapeless_2.11	2.3.3	compile
 com.clearspring.analytics:stream	2.9.2	compile
 com.conversantmedia:disruptor	1.2.13	compile
 com.esotericsoftware:kryo-shaded	3.0.3	compile
 com.esotericsoftware:minlog	1.3.0	compile
+com.facebook:nailgun-server	1.0.1	compile
```
elahrvivaz, line 11: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23058

elahrvivaz:
having some weird issues with tty that i wasn't seeing before, need to fix it.

docs/user/cli/nailgun.rst
```diff
@@ -0,0 +1,31 @@
...
+==============
+
+GeoMesa can use a persistent server to eliminate the overhead of bootstrapping a JVM for each command, which can
+make commands much faster to start. The server must be enabled by setting the environment variable
+``GEOMESA_NG_ENABLED=true``.
+
+.. warning::
+
+  The Nailgun server provides no security, and should only be used in a trusted environment. By default
+  it will bind to all host addresses, allowing unauthenticated connections from any client on the same network.
```
jnh5y, line 13: This is great.  If I were ever going to push for a Docker, this would be a great reason to make one:).

docs/user/upgrade.rst
```diff
@@ -99,6 +99,12 @@ FileSystem Data Store Metadata Format Change
 The metadata format for the FileSystem data store has been changed to support storing arbitrary key-value pairs.
 Any data written with version 3.2.0 or later will not be readable by earlier GeoMesa versions.
 
+Lambda Data Store Binary Distribution Change
```
jnh5y, line 102: Hmm... any reason this changed with this PR?

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
```diff
@@ -23,116 +29,87 @@
...
+            <version>${project.version}</version>
         </dependency>
         <dependency>
             <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-geojson-gs-plugin_${scala.binary.version}</artifactId>
-            <type>pom</type>
+            <artifactId>geomesa-spark-converter_${scala.binary.version}</artifactId>
         </dependency>
 
-        <!-- Web Services (dist) -->
```
jnh5y, line 69: Hmm... we were packing the web services in the Accumulo dist?  Was it the only bundle like that?

geomesa-accumulo/geomesa-accumulo-tools/conf-filtered/accumulo-env.sh
```diff
@@ -54,3 +54,5 @@ function get_accumulo_classpath() {
     echo "$accumulo_cp" | sed 's/[^:]*geomesa-accumulo-distributed-runtime[^:]*jar//'
   fi
 }
+
+get_accumulo_classpath
```
jnh5y, line 58: What does adding this do?

geomesa-archetypes/geomesa-archetypes-binary-dist/assembly.xml
```diff
@@ -0,0 +1,189 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/component/1.1.2"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/component/1.1.2 http://maven.apache.org/xsd/component-1.1.2.xsd">
```
jnh5y, line 4: When you change the archetype, do you clean out and respin various part of the project to test it out?

I'm trying to understand the workflow with Maven archetypes.

geomesa-arrow/geomesa-arrow-dist/pom.xml
```diff
@@ -23,107 +29,40 @@
...
             <artifactId>geomesa-arrow-tools_${scala.binary.version}</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-tools_${scala.binary.version}</artifactId>
-            <classifier>data</classifier>
-        </dependency>
 
-        <!-- GeoServer Plugins -->
+        <!-- GeoServer plugin - needed as a dependency for reactor ordering, but don't bring in transitive deps -->
```
jnh5y, line 33: Hmm... I wonder if there's something else we could be depending on.  This is slick which is both cool and worrying;).

geomesa-arrow/geomesa-arrow-dist/pom.xml
```diff
@@ -23,107 +29,40 @@
...
-        </dependency>
-        <dependency>
-            <groupId>log4j</groupId>
-            <artifactId>log4j</artifactId>
-            <scope>compile</scope>
-        </dependency>
-
-        <!-- manage these out from the libdir -->
-        <dependency>
-            <groupId>org.scala-lang</groupId>
```
jnh5y, line 96: For the tools are packaging the Scala compiler or requiring folks to download it?

Also, do you happen to know how NG will have a scala console?

geomesa-arrow/geomesa-arrow-tools/src/main/scala/org/locationtech/geomesa/arrow/tools/ArrowRunner.scala
```diff
@@ -8,36 +8,29 @@
...
-    new ScalaConsoleCommand
-  )
+  override protected def commands: Seq[Command] = {
+    super.commands ++ Seq(
+      new ArrowDescribeSchemaCommand,
+      new ArrowExportCommand,
+      new ArrowIngestCommand,
+      new ArrowGetTypeNamesCommand,
+      new ArrowGetSftConfigCommand,
+      new GenerateAvroSchemaCommand,
```
jnh5y, line 29: Ok, I'll bite.  Why isn't `GenerateAvroSchemaCommand` one of the general commands?

geomesa-cassandra/geomesa-cassandra-tools/conf-filtered/cassandra-env.sh
```diff
@@ -33,3 +33,6 @@ function get_cassandra_classpath() {
     echo "$classpath"
   fi
 }
+
+get_cassandra_classpath
+
```
jnh5y, line 38: Extra whitespace (just to show that I'm reading). ;)

geomesa-fs/geomesa-fs-dist/pom.xml
```diff
@@ -42,76 +40,34 @@
...
-            <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-process-wps_${scala.binary.version}</artifactId>
-        </dependency>
 
-        <!-- additional deps to manage in for deploying to lib dir-->
-        <dependency>
-            <groupId>org.locationtech.geomesa</groupId>
-            <artifactId>geomesa-fs-datastore_${scala.binary.version}</artifactId>
-        </dependency>
+        <!--spark  runtime -->
```
jnh5y, line 57: Does depending on the Spark runtime get this pom to the right spot in the reactor?  Either way, it may be worth sticking the pattern you've established.

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
```diff
@@ -8,45 +8,38 @@
...
+  override protected def commands: Seq[Command] = {
+    super.commands ++ Seq(
+      new FsCreateSchemaCommand,
+      new FsDescribeSchemaCommand,
+      new FsExportCommand,
+      new FsPlaybackCommand,
+      new FsIngestCommand,
+      new FsGetTypeNamesCommand,
+      new FsGetSftConfigCommand,
+      new ManageMetadataCommand,
```
jnh5y, line 33: Is the `ManageMetadataCommand` an Fs command?  (E.g. should it be renamed?)

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
```diff
@@ -8,45 +8,38 @@
...
+      new ScalaConsoleCommand,
```
jnh5y, line 34: Is the ScalaConsole command not common to all?

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
```diff
@@ -64,7 +64,7 @@ object GeoToolsDataStoreCommand {
       description = "Parameter for DataStoreFinder to load the data store, in the form key=value",
       variableArity = true,
       converter = classOf[KeyValueConverter])
-    val params: java.util.List[(String, String)] = new java.util.ArrayList()
+    var params: java.util.List[(String, String)] = new java.util.ArrayList()
```
jnh5y, line 67: *blinks*  Why did this change to be a var?

geomesa-kafka/geomesa-kafka-tools/src/main/scala/org/locationtech/geomesa/kafka/tools/KafkaRunner.scala
```diff
@@ -8,37 +8,27 @@
...
-    new export.KafkaListenCommand,
-    new export.KafkaExportCommand,
-    new ingest.KafkaIngestCommand,
-    new status.KafkaDescribeSchemaCommand,
-    new status.KafkaGetTypeNamesCommand,
-    // common commands, placeholders for script functions
-    new HelpCommand(this, jc),
-    new ConfigureCommand,
-    new ClasspathCommand,
-    new ScalaConsoleCommand,
```
jnh5y, line 36: Ok, looks like the Scala console is bouncing around a bit.  Which tool bundles should it be in?

geomesa-lambda/geomesa-lambda-tools/src/main/scala/org/locationtech/geomesa/lambda/tools/LambdaRunner.scala
```diff
@@ -19,36 +18,28 @@ import org.locationtech.geomesa.lambda.tools.data._
...
+  override protected def commands: Seq[Command] = {
+    super.commands ++ Seq(
+      new LambdaCreateSchemaCommand,
+      new LambdaDeleteFeaturesCommand,
+      new AccumuloDescribeSchemaCommand,
+      new AccumuloExplainCommand,
+      new LambdaExportCommand,
+      new AccumuloGetTypeNamesCommand,
+      new LambdaRemoveSchemaCommand,
+      new AccumuloVersionRemoteCommand,
```
jnh5y, line 35: Probably not a major concern, but this suggests that the LDS version command only checks the version of Accumulo.  Maybe we should do something with Kafka as well?  (Or does that not make sense as there is no distributed code for Kafka?)

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
jnh5y: Looks like it is.

geomesa-arrow/geomesa-arrow-tools/src/main/scala/org/locationtech/geomesa/arrow/tools/ArrowRunner.scala
jnh5y: Looks like it is also common.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
```diff
@@ -148,32 +167,55 @@ trait Runner extends LazyLogging {
...
 
-  private var added = false
-
-  /**
-   * Add jcommander converter factories. Note that the implementation is a static,
-   * shared, non-thread-safe list, so we have to synchronize ourselves.
-   *
-   * TODO revisit this if we upgrade jcommander versions
-   */
-  def addConverterFactories(jc: JCommander): Unit = synchronized {
```
jnh5y, line 171: What happened to this?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/help/ScalaConsoleCommand.scala
```diff
@@ -6,21 +6,23 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.tools.status
+package org.locationtech.geomesa.tools.help
```
jnh5y, line 9: Hmm... 'help' (and 'status') both seem like odd package names for the Scala Console command.  (I'm not opposed just mentioning it in case there's an obvious better option after some reflection.)

Separately, I wonder if there are split packages here that we will need to think about.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/package.scala
```diff
@@ -0,0 +1,94 @@
...
+      }
+      _console
+    }
+
+    def setConsole(c: Prompt.SystemConsole): Unit = _console = c
+  }
+
+  trait DistributedCommand {
+
+    def libjarsFiles: Seq[String]
```
jnh5y, line 87: Oh man, lib jar files....  have we taken a spin through those recently?  I wonder if anything needs to be updated....

docs/user/upgrade.rst
elahrvivaz: i had to make a bunch of changes to the dists for nailgun support (adding the ng client, modifying the scripts, updating the -env classes), so I made an archetype that standardizes everything into one place. i forget why, exactly, but that made including the accumulo runtime in the lambda dist problematic. plus, this makes our artifacts smaller :) if you feel strongly, i can try to include it again

geomesa-accumulo/geomesa-accumulo-dist/pom.xml
elahrvivaz: yeah, it was the only one. and it doesn't seem relevant to include, since you'd generally pull it down through maven.

geomesa-accumulo/geomesa-accumulo-tools/conf-filtered/accumulo-env.sh
elahrvivaz: this makes it so that when you source the script it will print out the classpath (vs having to call the function after sourcing the script). this was done so that we can just source all the *-env.sh files in the conf folder in order to build up the classpath without having to specify them per back-end, which means we can use the same geomesa-* script for all of them.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/NailgunServer.scala
```diff
@@ -0,0 +1,227 @@
...
+  private val start = System.currentTimeMillis()
+  private val lastRequest = new AtomicLong(System.currentTimeMillis())
+
+  private val timers: LoadingCache[String, (AtomicInteger, Timer)] = Caffeine.newBuilder().build(
+    new CacheLoader[String, (AtomicInteger, Timer)]() {
+      override def load(k: String): (AtomicInteger, Timer) = (new AtomicInteger(0), registry.timer(k))
+    }
+  )
+
+  private val idle: Future[_] =
```
jnh5y, line 44: Could we use a ScheduledExecutorService to check the state (e.g., stop when !executingSomething and lastCall is "old")?

The Thread.sleeps worry me for no good reason.

geomesa-archetypes/geomesa-archetypes-binary-dist/assembly.xml
elahrvivaz: archetypes aren't really a maven thing, it's just the nomenclature i adopted. it's just used as the parent of all the -dist modules now, instead of e.g. geomesa-accumulo being the parent of geomesa-accumulo-dist, etc. i did test out all the binary dists and compared them to the previous way to ensure there weren't any unintentional differences

elahrvivaz: i should say, maven archetypes are a thing, but not this thing ;)

maybe i should have named it something else...

geomesa-arrow/geomesa-arrow-dist/pom.xml
elahrvivaz: the problem is due to us not attaching the gs-plugin tars to maven anymore... so we have to include the gs-plugin as a <file> instead of a <dependency> in the maven assembly plugin. but we still need the reactor order to be correct... so yeah, i'm not sure of a better way without re-attaching the gs-plugin tars (which would mean the get deployed to central again)

geomesa-arrow/geomesa-arrow-dist/pom.xml
elahrvivaz: we don't bundle it - the `scala-console` command will check for `scala` on the path, and download it if not. it doesn't go through NG, it spins up separately.

geomesa-arrow/geomesa-arrow-tools/src/main/scala/org/locationtech/geomesa/arrow/tools/ArrowRunner.scala
elahrvivaz: ah, probably an oversight. i'll remove it here.

elahrvivaz: removed

geomesa-cassandra/geomesa-cassandra-tools/conf-filtered/cassandra-env.sh
elahrvivaz: i think we try to always have a trailing empty line to fix sed commands or something like that?

geomesa-fs/geomesa-fs-dist/pom.xml
elahrvivaz: i think all the modules depend on the associated spark runtime, as they get bundled into the dist/spark dir. do you mean we shouldn't attach them?

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
elahrvivaz: yeah it's an fs command... probably should be renamed. want me to do that as part of this pr?

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
elahrvivaz: removed, thanks

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
elahrvivaz: the jcommander update didn't like having it be a val...

geomesa-kafka/geomesa-kafka-tools/src/main/scala/org/locationtech/geomesa/kafka/tools/KafkaRunner.scala
elahrvivaz: should be in the common one

geomesa-lambda/geomesa-lambda-tools/src/main/scala/org/locationtech/geomesa/lambda/tools/LambdaRunner.scala
elahrvivaz: yeah, there's no remote code for kafka so no remote version

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
elahrvivaz: my TODO above came true! :D

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/help/ScalaConsoleCommand.scala
elahrvivaz: `help` _is_ kind of confusing haha... these are all the "placeholder" commands that are actually implemented in the bash script, but we want them to show up in the jcommander help. if you've got a suggestion i'd be happy to change it.

re: split packages, i'm not going to worry about that until we need to, but you're probably right.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/package.scala
elahrvivaz: testing the distributed ingest during the smoke testing should shake out any major issues...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/NailgunServer.scala
elahrvivaz: i did initially implement this with a scheduled executor service and re-submitting the job each time, but i ran into issues with it dying appropriately when the nailgun was shut down explicitly. iirc, scheduled executors will keep the jvm up if they have a scheduled job still. hmm, although now that i'm looking at it the executor should probably be shut down after submitting the job... i'll test that out real quick. but having a daemon thread mostly sleeping in the background seemed acceptable. i suspect a scheduled executor service would be doing something similar with sleeping/idle threads internally.

geomesa-archetypes/geomesa-archetypes-binary-dist/assembly.xml
jnh5y: That clears it up!

geomesa-arrow/geomesa-arrow-dist/pom.xml
jnh5y: Great!

geomesa-cassandra/geomesa-cassandra-tools/conf-filtered/cassandra-env.sh
jnh5y: Eh, I think I saw one without.... not a big deal though!

geomesa-fs/geomesa-fs-dist/pom.xml
jnh5y: No the spark runtimes should be in the dist bundles.  I was trying to suggest that maybe depending on the gs-plugin and then excluding it was an unnecessary trick?

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
jnh5y: Yes, let's do that here; sounds good to me.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/Runner.scala
jnh5y: Yeah, I noticed the change in JCommander version at the end of the PR!

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/help/ScalaConsoleCommand.scala
jnh5y: Yeah, that is great for now.   We'll figure it out when we sort out Java 11 support;).

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/package.scala
jnh5y: Yeah, makes sense.  I kinda wish they were not potentially environment dependent (which is a separate issue).

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/NailgunServer.scala
jnh5y: If you are right about the ScheduledExecutors.... I imagine there's either a way to set the Service to shutdown on JVM exit or we could try adding a shutdownhook to remove the scheduled task.

At least one of those options ought to work...

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
jnh5y: Oooookkkk... Is JCommander mutating state?

docs/user/upgrade.rst
jnh5y: No need.

geomesa-cassandra/geomesa-cassandra-tools/conf-filtered/cassandra-env.sh
elahrvivaz: it's aspirational :)

geomesa-fs/geomesa-fs-dist/pom.xml
elahrvivaz: i think i had to exclude the transitive deps from the gs-plugin b/c it pulls in some unnecessary jars, but didn't have to do the same with spark i guess... there was definitely a reason i did it haha

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/FsRunner.scala
elahrvivaz: done

geomesa-gt/geomesa-gt-tools/src/main/scala/org/locationtech/geomesa/geotools/tools/GeoToolsDataStoreCommand.scala
elahrvivaz: yeah, pretty much all our param classes use vars. i think b/c this was a list it was ok before since the list was mutable? but the way jcommander works is you pass in your param class and it fills in all the vars based on the command line args.

elahrvivaz: it's mutated once up front, and then theoretically not touched again although there's nothing preventing you i guess...

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/NailgunServer.scala
elahrvivaz: here's the initial impl that tried to do pretty much that, but it didn't quite work correctly: https://github.com/locationtech/geomesa/blob/3c053e11f3079dbeb8751e5f8da7b222d7baca77/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/NailgunServer.scala#L32-L70

elahrvivaz: from what i can tell the difference is `TIMED_WAITING (sleeping)` vs `TIMED_WAITING (parking)`, which doesn't really seem to matter... if there were multiple tasks or threads it would be better to use the executor, but for one thread and task it seems equivalent. the current approach seems simpler and clearer to me, as well, but maybe that's just me.

I'm nervous to change it since i had issues with the original scheduled executor and this seems to be working correctly.

geomesa-fs/geomesa-fs-dist/pom.xml
jnh5y: :shipit:

PR #2697, opened 2021-02-10
GEOMESA-2995 Explicit support for tools ingest from stdin

elahrvivaz:


elahrvivaz:
Stacks on top of #2694 - requires the jcommander upgrade from that PR

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -56,29 +57,21 @@ trait IngestCommand[DS <: DataStore]
...
-      lists.flatMap { file =>
-        WithClose(file.open) { iter =>
-          iter.flatMap { case (_, is) => IOUtils.lineIterator(is, StandardCharsets.UTF_8).asScala }.toList
-        }
-      }
-    } else {
-      params.files.asScala
+    val inputs: Inputs = {
+      val files = Inputs(params.files.asScala)
+      if (params.srcList) { files.asSourceList } else { files }
```
jnh5y, line 65: Wow, let me make sure I've got this...  One could pass in via stdin data or a list of files to ingest?  Neat.

elahrvivaz: yeah, i think you could do this before too

PR #2701, opened 2021-02-15
GEOMESA-3006 Option to suppress empty files on export

elahrvivaz:


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
```diff
@@ -187,12 +185,27 @@ trait ExportCommand[DS <: DataStore] extends DataStoreCommand[DS]
...
   }
 
-  protected def export(ds: DS, query: Query, exporter: FeatureExporter): Option[Long] = {
+  /**
+   * Hook for overriding export
+   *
+   * @param ds data store
+   * @param query query
+   * @param exporter exporter
+   * @param empty export empty files or no
```
jnh5y, line 194: Hmm... I wonder if there's a better variable name like `writeEmptyOutput` or `allowEmptyOutput`.  Admittedly, that's me being verbose.

It took me a few minutes of reading this PR to figure out what the `empty` flag was doing...  Thoughts?

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/GeoJsonExporter.scala
```diff
@@ -40,10 +42,12 @@ class GeoJsonExporter(os: OutputStream, counter: ByteCounter) extends ByteCounte
       if (serializer != null) {
         serializer.endFeatureCollection(writer)
         writer.flush()
-        os.write('\n')
+        stream.os.write('\n')
       }
     } finally {
-      writer.close() // also closes underlying writer and output stream
+      if (writer != null) {
```
jnh5y, line 48: Hmm... I wonder how the writer could been cleared....  The check is probably sensibly though.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/GeoJsonExporter.scala
```diff
@@ -8,19 +8,21 @@
...
-class GeoJsonExporter(os: OutputStream, counter: ByteCounter) extends ByteCounterExporter(counter) {
-
-  private val writer = GeoJsonSerializer.writer(new OutputStreamWriter(os))
+class GeoJsonExporter(stream: ExportStream) extends ByteCounterExporter(stream) {
 
+  private var writer: JsonWriter = _
   private var serializer: GeoJsonSerializer = _
 
   override def start(sft: SimpleFeatureType): Unit = {
+    writer = GeoJsonSerializer.writer(new OutputStreamWriter(stream.os))
```
jnh5y, line 25: Reading through this makes me wonder if we could use a lazy val for writer?

Also, this start method is not thread safe.  I am guessing we use it in a single threaded manner.

geomesa-tools/src/test/scala/org/locationtech/geomesa/tools/export/ExportCommandTest.scala
```diff
@@ -165,6 +165,25 @@ class ExportCommandTest extends Specification {
         readFeatures(format, file) mustEqual features.take(1)
       }
     }
+    "suppress empty output files" in {
```
jnh5y, line 168: This test is great.  Since the first case does return empty files would it be worth changing the title to say something about testing both settings of the option?

elahrvivaz:
```suggestion
    "suppress or allow empty output files" in {
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ExportCommand.scala
elahrvivaz: I'll update it

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/GeoJsonExporter.scala
elahrvivaz: if the exporter is never `start`ed, then the writer will be null

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/GeoJsonExporter.scala
elahrvivaz: yeah, all the exporters are single threaded. we could use a lazy val, but we'd still have to track whether the val was initialized. closing the writer will write the closing `}` to the json stream, so we don't want to instantiate it until we know that there is data to export, otherwise you'll end up with an empty feature collection but you'll still get an output file.

PR #2704, opened 2021-02-19
GEOMESA-3011 Fix match for CLI dependency jars

elahrvivaz:

### DIFF
--- a/geomesa-tools/conf-filtered/geomesa-env.sh
+++ b/geomesa-tools/conf-filtered/geomesa-env.sh
@@ -143,7 +143,7 @@ function check_classpath() {
     group="${gav%%:*}"
     artifact="${gav#$group:}"
     artifact="${artifact%%:*}"
-    if [[ $classpath != *"$artifact"* ]]; then
+    if [[ ! $classpath =~ (^|:|/)${artifact}(-[^:]*)*\.jar ]]; then
       missing+=("$gav")
     fi
   done


geomesa-tools/conf-filtered/geomesa-env.sh
```diff
@@ -143,7 +143,7 @@ function check_classpath() {
     group="${gav%%:*}"
     artifact="${gav#$group:}"
     artifact="${artifact%%:*}"
-    if [[ $classpath != *"$artifact"* ]]; then
+    if [[ ! $classpath =~ (^|:|/)${artifact}(-[^:]*)*\.jar ]]; then
```
jnh5y, line 146: Ok, what corner case does this handle?

elahrvivaz: basically stricter substring matching. iirc this was to fix commons-math/commons-math3 - previously `foo.jar:commons-math3.jar` would match for artifact `commons-math`, this makes it so it doesn't.

elahrvivaz: and yes, that did cause me a 10 minute headache :P

jnh5y: That only took 10 minutes to figure out?  I probably would have been stuck for much longer;).

PR #2705, opened 2021-02-19
GEOMESA-3010 Make CLI dependency update use unique temp file

elahrvivaz:

### DIFF
--- a/geomesa-tools/conf-filtered/geomesa-env.sh
+++ b/geomesa-tools/conf-filtered/geomesa-env.sh
@@ -182,8 +182,9 @@ function check_classpath() {
 
 function disable_classpath_checks() {
   if [[ -n "$(grep '^export GEOMESA_CHECK_DEPENDENCIES' "${GEOMESA_CONF_DIR}/geomesa-env.sh")" ]]; then
+    local tmpfile="$(mktemp)"
     sed 's/^export GEOMESA_CHECK_DEPENDENCIES.*/export GEOMESA_CHECK_DEPENDENCIES="false"/' \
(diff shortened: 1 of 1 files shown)


geomesa-tools/conf-filtered/geomesa-env.sh
```diff
@@ -182,8 +182,9 @@ function check_classpath() {
 
 function disable_classpath_checks() {
   if [[ -n "$(grep '^export GEOMESA_CHECK_DEPENDENCIES' "${GEOMESA_CONF_DIR}/geomesa-env.sh")" ]]; then
+    local tmpfile="$(mktemp)"
     sed 's/^export GEOMESA_CHECK_DEPENDENCIES.*/export GEOMESA_CHECK_DEPENDENCIES="false"/' \
-      "${GEOMESA_CONF_DIR}/geomesa-env.sh" > tmp && mv tmp "${GEOMESA_CONF_DIR}/geomesa-env.sh" \
+      "${GEOMESA_CONF_DIR}/geomesa-env.sh" >> $tmpfile && mv $tmpfile "${GEOMESA_CONF_DIR}/geomesa-env.sh" \
```
jnh5y, line 187: What goes wrong otherwise?  Does the temp file get used by multiple installs / invocations?

elahrvivaz: no, but if you have a file name tmp in the bin directory it will overwrite it, and if you have a directory named tmp it will fail to write and barf out an error message. this way we guarantee there's no conflict with existing files

elahrvivaz: it's also possible that two invocations would conflict, as you say, but the odds of that are probably pretty low since this is only done once per install.

PR #2706, opened 2021-02-19
GEOMESA-3012 Remove unused Kerberos credentials from Accumulo Spark job

elahrvivaz:
* Refactor Accumulo input format configuration to not require Job instance


geomesa-accumulo/geomesa-accumulo-spark/src/main/scala/org/locationtech/geomesa/spark/accumulo/AccumuloSpatialRDDProvider.scala
```diff
@@ -48,38 +45,9 @@ class AccumuloSpatialRDDProvider extends SpatialRDDProvider with LazyLogging {
...
-        // assertNotStopped()
-
-        // From sc.newAPIHadoopRDD
-        // Add necessary security credentials to the JobConf. Required to access secure HDFS.
-        val jconf = job.getConfiguration.asInstanceOf[JobConf]
-        SparkHadoopUtilShim.addCredentials(jconf)
-
-        // Iterate over tokens in credentials and add the Accumulo one to the configuration directly
-        // This is because the credentials seem to disappear between here and the YARN executor
-        // See https://stackoverflow.com/questions/44525351/delegation-tokens-with-accumulo-spark
```
jnh5y, line 67: Ok, so whatever this approach was doing is unused?

elahrvivaz: Yeah, I asked James S about this in gitter. We removed the code that read this token back out a while ago, but never removed the part that puts it in here. I believe accumulo sets the delegation token and reads it out without us having to explicitly do anything.

elahrvivaz: that stack overflow link has disappeared too, so i'm not sure what was the original intent here :(

PR #2707, opened 2021-02-19
GEOMESA-3013,GEOMESA-3009,GEOMESA-2998 Support Kerberos in HBase Spark

elahrvivaz:
* Fix synchronization of multiple confs if there are multiple workers in a single JVM
* Fix resolution of _HOST in principal names

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
@@ -17,6 +17,7 @@ import com.typesafe.scalalogging.LazyLogging
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.hbase.client.{Connection, ConnectionFactory}
 import org.apache.hadoop.hbase.security.User
+import org.apache.hadoop.hbase.security.token.AuthenticationTokenIdentifier
 import org.apache.hadoop.hbase.{HBaseConfiguration, HConstants}
 import org.apache.hadoop.security.authentication.util.KerberosUtil
 import org.apache.hadoop.security.{SecurityUtil, UserGroupInformation}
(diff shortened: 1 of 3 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
```diff
@@ -146,47 +147,57 @@ object HBaseConnectionPool extends LazyLogging {
...
-          logger.warn(
-            s"Changing global authenticated Hadoop user from '${currentUser.getUserName}' to '$principal' -" +
-                "this will affect any connections still using the old user")
+        val keytab = conf.get(HBaseGeoMesaKeyTab)
+        val rawPrincipal = conf.get(HBaseGeoMesaPrincipal)
+
+        if (keytab == null || rawPrincipal == null) {
+          lazy val missing =
+            Seq(HBaseGeoMesaKeyTab -> keytab, HBaseGeoMesaPrincipal -> rawPrincipal).collect { case (k, null) => k }
+          logger.warn(s"Security is enabled but missing credentials under '${missing.mkString("' and '")}'")
```
jnh5y, line 163: Are the docs clear about setting both of these? 

And is it required?

geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
```diff
@@ -37,6 +40,14 @@ class HBaseSpatialRDDProvider extends SpatialRDDProvider {
 
     val ds = DataStoreConnector[HBaseDataStore](dsParams)
 
+    if (User.isSecurityEnabled) {
+      // configure auth token so that the HBase TokenUtil can pick it up for
+      // authentication in the remote worker processes - the keytab won't be accessible there
+      // needs to be in UserGroupInformation.getCurrentUser().getCredentials()
+      val token = TokenUtil.obtainToken(ds.connection, User.getCurrent)
+      UserGroupInformation.getCurrentUser.getCredentials.addToken(token.getService, token)
```
jnh5y, line 48: What does setting this token do for us?  

Is it getting added to the Hadoop configuration object?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
elahrvivaz: yeah the docs are clear: https://www.geomesa.org/documentation/stable/user/hbase/kerberos.html#kerberos
they are both required. if kerberos is enabled, you need a keytab for anything to work. previously this 
 block would fail with an NPE (or maybe just silently, i forget exactly), so this way you at least get a warning.

geomesa-hbase/geomesa-hbase-spark/src/main/scala/org/locationtech/geomesa/spark/hbase/HBaseSpatialRDDProvider.scala
elahrvivaz: it doesn't get added to the configuration, but it will get added the to hadoop Job (as hadoop keeps credentials there instead of the conf for whatever reason). there is some code in spark that looks in `UserGroupInformation.getCurrentUser().getCredentials()` for hbase delegation tokens, as that's where hbase will set the tokens in their input formats. i was hoping it would just work without us doing anything, but for some reason it wasn't added unless i did so explicitly.

elahrvivaz: the HBase TokenUtil referenced in the comment is the class that gets invoked and expects the token to be there.

PR #2708, opened 2021-02-22
GEOMESA-3014 Compact rendering of FSDS metadata files

elahrvivaz:

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
@@ -145,16 +145,20 @@ object StorageMetadata {
   }
 
   /**
-    * Holds a storage file
-    *
-    * @param name file name (relative to the root path)
-    * @param timestamp timestamp for the file
-    * @param action type of file (append, modify, delete)
-    */
+   * Holds a storage file
+   *
+   * @param name file name (relative to the root path)
+   * @param timestamp timestamp for the file
+   * @param action type of file (append, modify, delete)
+   * @param sort sort fields, if any, as feature type attribute number
+   * @param bounds known bounds, if any, keyed by feature type attribute number
+   */
   case class StorageFile(
       name: String,
       timestamp: Long,
-      action: StorageFileAction.StorageFileAction = StorageFileAction.Append
+      action: StorageFileAction.StorageFileAction = StorageFileAction.Append,
+      sort: Seq[Int] = Seq.empty,
+      bounds: Seq[(Int, String, String)] = Seq.empty
     )
 
   /**
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/package.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/package.scala
@@ -8,20 +8,23 @@
 
 package org.locationtech.geomesa.fs.storage.common
 
-import com.typesafe.config.{ConfigValue, ConfigValueFactory}
-import org.locationtech.geomesa.fs.storage.api.StorageMetadata.StorageFileAction.StorageFileAction
+import java.util.Collections
+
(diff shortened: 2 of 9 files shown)


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
```diff
@@ -145,16 +145,20 @@ object StorageMetadata {
...
+   * @param sort sort fields, if any, as feature type attribute number
+   * @param bounds known bounds, if any, keyed by feature type attribute number
+   */
   case class StorageFile(
       name: String,
       timestamp: Long,
-      action: StorageFileAction.StorageFileAction = StorageFileAction.Append
+      action: StorageFileAction.StorageFileAction = StorageFileAction.Append,
+      sort: Seq[Int] = Seq.empty,
+      bounds: Seq[(Int, String, String)] = Seq.empty
```
jnh5y, line 161: Does this change the format for the metadata?  

If so, any migration or compatibility considerations?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/package.scala
```diff
@@ -55,15 +58,20 @@ package object metadata {
...
 
     def +(other: PartitionConfig): PartitionConfig = {
       require(action == PartitionAction.Add, "Can't aggregate into non-add actions")
       val ts = math.max(timestamp, other.timestamp)
-      PartitionConfig(name, action, files ++ other.files, count + other.count, envelope.merge(other.envelope), ts)
+      val env = if (envelope.isEmpty) { other.envelope } else if (other.envelope.isEmpty) { envelope } else {
+        val Seq(xmin1, ymin1, xmax1, ymax1) = envelope
+        val Seq(xmin2, ymin2, xmax2, ymax2) = other.envelope
+        Seq(math.min(xmin1, xmin2), math.min(ymin1, ymin2), math.max(xmax1, xmax2), math.max(ymax1, ymax2))
+      }
```
jnh5y, line 73: This logic makes sense.  Should it be here? 

Definitely non-blocking, I'm just surprised that it hasn't been needed before here or elsewhere.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
jnh5y: Assuming this changes the stored metadata, are there unit tests covering the new data?

elahrvivaz: it does, but the readers can handle both formats. if you write data with a newer version, you won't be able to read it back out. currently these are unused, but i wanted to add them as placeholders. there are unit tests checking the old schemas that cover this.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/package.scala
elahrvivaz: i don't think we store envelopes as untyped Seqs like this anywhere else... the impetus here is to make the json rendering as compact as possible.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
jnh5y: Do we need to mention that in the upgrade guide?  (E.g. not to mix and match new and old writers?)

elahrvivaz: oh, we actually already have it in there due to the key-value metadata change :D

PR #2709, opened 2021-02-23
GEOMESA-2970 Deprecate Arrow multi-file and double pass scans

elahrvivaz:

### DIFF
--- a/docs/user/datastores/analytic_queries.rst
+++ b/docs/user/datastores/analytic_queries.rst
@@ -263,14 +263,6 @@ following query hints:
 +-------------------------------------+--------------------+------------------------------------+
 | QueryHints.ARROW_DICTIONARY_FIELDS  | String (optional)  | dictionaryFields                   |
 +-------------------------------------+--------------------+------------------------------------+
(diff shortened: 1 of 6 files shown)


docs/user/datastores/analytic_queries.rst
```diff
@@ -263,14 +263,6 @@ following query hints:
 +-------------------------------------+--------------------+------------------------------------+
 | QueryHints.ARROW_DICTIONARY_FIELDS  | String (optional)  | dictionaryFields                   |
 +-------------------------------------+--------------------+------------------------------------+
-| QueryHints.ARROW_DICTIONARY_VALUES  | String (optional)  | N/A                                |
```
jnh5y, line 266: Hmm....  Should we wait on removing the documentation?

elahrvivaz: ah, not sure. i assume we don't want people to use it since it's going away. i guess we could optionally add some kind of deprecated flag to the docs?

jnh5y: I doubt that anyone is using these features, so yeah, I'll merged this as is!

PR #2712, opened 2021-03-08
GEOMESA-3019 Add support for Spark 3.1.1

jnh5y:


jnh5y:
To test Spark 3.1.1 support, one can use the change-scala-version.sh script, and change the Spark version to 3.1.1.

As a note, the test for GEOMESA-2454: Spark: cannot resolve 'CAST(`geom` AS STRUCT<`wkb`: BINARY>)' due to data type mismatch: cannot cast point to geometryDONE  “breaks” since Spark now does the thing which Seth Fitzsimmons expected previously.

elahrvivaz (approved):
lgtm - is this ready to merge?

jnh5y:
Yeah, I think it is good to go.  The tough thing is that the unit tests do fail since a "pendingUntilFixed" style test is reversed by 3.1.1.  

*shrugs*  It *ought* to work as a binary in a Spark 3.1.1 environment.  I have not tested that though.  I'm fine with shipping it (the change is a generalization and shouldn't negatively impact anything) or waiting until more testing can occur.

elahrvivaz:
i'll ship it and let's just not advertise it until we test it some more.

jnh5y:
> i'll ship it and let's just not advertise it until we test it some more.

Super secret Spark support! ;)

PR #2713, opened 2021-03-08
GEOMESA-3021,GEOMESA-3022 Consistency check for FSDS metadata

elahrvivaz:
* Fix scanning of more than 65535 partitions

### DIFF
--- a/docs/user/filesystem/commandline.rst
+++ b/docs/user/filesystem/commandline.rst
@@ -144,6 +144,7 @@ sub-commands:
 * ``compact`` - compact multiple metadata files down to a single file
 * ``register`` - create a new metadata entry for an existing data file
 * ``unregister`` - remove a metadata entry for an existing data file
+* ``check-consistency`` - check consistency between the metadata and data files
 
 To invoke the command, use the command name followed by the sub-command, then any arguments. For example::
 
@@ -157,13 +158,13 @@ Argument                 Description
 ======================== =============================================================
 
 ``compact``
-^^^^^^^^^^^
+~~~~~~~~~~~
 
 The ``compact`` sub-command will rewrite multiple metadata files as a single file. Note that this does
 not change the data files; that is accomplished by the top-level ``compact`` command, as described above.
 
 ``register/unregister``
-^^^^^^^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~~~~~~~
 
 The ``register`` and ``unregister`` sub-commands will add or delete metadata associated with a particular file.
 The files must already exist under the appropriate partition path. If new data files are created through some
@@ -181,4 +182,18 @@ Argument                 Description
 ``--bounds``             Geographic bounds of the data files being registered, in the
                          form ``xmin,ymin,xmax,ymax``. This is not required, but can
                          be used later for estimating query bounds
+======================== =============================================================
+
+``check-consistency``
+~~~~~~~~~~~~~~~~~~~~~
+
+The ``check-consistency`` sub-command will check the metadata against the data files. It will
+find data files that are not referenced in the metadata, and metadata entries that do not
+correspond to data files.
+
+======================== =============================================================
+Argument                 Description
+======================== =============================================================
+``--partitions``         The name of partitions to check, or none for all partitions
+``--repair``             Update the metadata to fix any inconsistencies found
 ======================== =============================================================
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
@@ -100,6 +100,11 @@ trait StorageMetadata extends Compactable with Closeable {
     */
   def removePartition(partition: PartitionMetadata): Unit
 
+  /**
+   * Invalidate any cached state
+   */
+  def invalidate(): Unit = throw new NotImplementedError() // TODO remove default impl in next major release
+
   @deprecated("deprecated with no replacement")
   def reload(): Unit = {}
 }
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/FileSystemThreadedReader.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/FileSystemThreadedReader.scala
@@ -16,6 +16,7 @@ import org.locationtech.geomesa.features.ScalaSimpleFeature
 import org.locationtech.geomesa.fs.storage.api.StorageMetadata.{StorageFileAction, StorageFilePath}
 import org.locationtech.geomesa.fs.storage.common.AbstractFileSystemStorage.FileSystemPathReader
 import org.locationtech.geomesa.utils.collection.CloseableIterator
+import org.locationtech.geomesa.utils.concurrent.PhaserUtils
 import org.locationtech.geomesa.utils.io.WithClose
 import org.opengis.feature.simple.SimpleFeature
 
@@ -99,6 +100,8 @@ object FileSystemThreadedReader extends StrictLogging {
       }
 
       try {
+        var child = new Phaser(phaser) // ensure that we don't register too many parties on this phaser
+        var parties = 0 // track registered parties
         readers.foreach { case (reader, files) =>
           // group our files by actions which can be parallelized
           val groups = scala.collection.mutable.ListBuffer.empty[Seq[StorageFilePath]]
@@ -119,7 +122,14 @@ object FileSystemThreadedReader extends StrictLogging {
             groups += group // add the last group
           }
 
-          es.submit(new ChainedReaderTask(es, phaser, reader, groups.head, groups.tail, queue))
+          // each chained reader task will register at most groups.length parties
+          parties += groups.length
+          if (parties > PhaserUtils.MaxParties) {
+            parties = groups.length
+            child = new Phaser(phaser)
+          }
+          child.register() // register new task
+          es.submit(new ChainedReaderTask(es, child, reader, groups.head, groups.tail, queue))
         }
       } catch {
         case NonFatal(e) => es.shutdownNow(); throw e
@@ -171,9 +181,7 @@ object FileSystemThreadedReader extends StrictLogging {
       chain: Seq[Seq[StorageFilePath]],
       queue: BlockingQueue[SimpleFeature],
       mods: scala.collection.mutable.Set[String] = scala.collection.mutable.HashSet.empty[String]
-  ) extends Runnable {
-
-    phaser.register()
+    ) extends Runnable {
 
     override def run(): Unit = {
       val child = new Phaser(1) {
@@ -181,6 +189,7 @@ object FileSystemThreadedReader extends StrictLogging {
           // when this group is done, submit the next group for processing
           try {
             if (chain.nonEmpty) {
+              phaser.register() // register new task
               es.submit(new ChainedReaderTask(es, phaser, reader, chain.head, chain.tail, queue, mods))
             }
             true // return true to indicate the phaser should terminate
@@ -190,7 +199,10 @@ object FileSystemThreadedReader extends StrictLogging {
         }
       }
       try {
-        group.foreach(file => es.submit(new ReaderTask(child, queue, file.path, read(reader, file, mods))))
+        group.foreach { file =>
+          child.register() // register new task
+          es.submit(new ReaderTask(child, queue, file.path, read(reader, file, mods)))
+        }
       } finally {
         child.arriveAndDeregister()
       }
@@ -211,8 +223,6 @@ object FileSystemThreadedReader extends StrictLogging {
       path: Path,
       iter: => CloseableIterator[SimpleFeature]
     ) extends Runnable {
-
-    phaser.register()
 
     override def run(): Unit = {
       try {
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
@@ -22,7 +22,7 @@ import org.apache.hadoop.fs._
 import org.locationtech.geomesa.fs.storage.api.StorageMetadata.PartitionMetadata
 import org.locationtech.geomesa.fs.storage.api._
 import org.locationtech.geomesa.fs.storage.common.utils.PathCache
-import org.locationtech.geomesa.utils.concurrent.CachedThreadPool
+import org.locationtech.geomesa.utils.concurrent.{CachedThreadPool, PhaserUtils}
 import org.locationtech.geomesa.utils.io.WithClose
 import org.locationtech.geomesa.utils.stats.MethodProfiling
 import org.locationtech.geomesa.utils.text.StringSerialization
@@ -182,6 +182,11 @@ class FileBasedMetadata(
     partitions.invalidate(BoxedUnit.UNIT)
   }
 
+  override def invalidate(): Unit = {
+    partitions.invalidateAll()
+    metadata.invalidateAll()
+  }
+
   override def close(): Unit = {}
 
   /**
@@ -251,7 +256,7 @@ class FileBasedMetadata(
       val pool = new CachedThreadPool(threads)
       // use a phaser to track worker thread completion
       val phaser = new Phaser(2) // 1 for the initial directory worker + 1 for this thread
-      pool.submit(new DirectoryWorker(pool, phaser, directory, result))
+      pool.submit(new DirectoryWorker(pool, phaser, fc.listStatus(directory), result))
       // wait for the worker threads to complete
       try {
         phaser.awaitAdvanceInterruptibly(phaser.arrive())
@@ -331,33 +336,40 @@ class FileBasedMetadata(
   private class DirectoryWorker(
       es: ExecutorService,
       phaser: Phaser,
-      dir: Path,
+      listDirectory: => RemoteIterator[FileStatus],
       result: ConcurrentHashMap[String, PartitionFiles]
     ) extends Runnable {
 
     override def run(): Unit = {
       try {
-        val iter = fc.listStatus(dir)
-        while (iter.hasNext) {
+        var i = phaser.getRegisteredParties + 1
+        val iter = listDirectory
+        while (iter.hasNext && i < PhaserUtils.MaxParties) {
           val status = iter.next
           val path = status.getPath
           lazy val name = path.getName
           if (status.isDirectory) {
-            phaser.register() // register the new worker thread
-            es.submit(new DirectoryWorker(es, phaser, path, result))
+            i += 1
+            // use a tiered phaser on each directory avoid the limit of 65535 registered parties
+            es.submit(new DirectoryWorker(es, new Phaser(phaser, 1), fc.listStatus(path), result))
           } else if (name.startsWith(UpdatePartitionPrefix)) {
             // pull out the partition name but don't parse the file yet
             val encoded = name.substring(8, name.length - 42) // strip out prefix and suffix
             val partition = StringSerialization.decodeAlphaNumericSafeString(encoded)
             result.merge(partition, PartitionFiles(unparsed = Seq(path)), addFiles)
           } else if (name == CompactedHocon || name == CompactedJson) {
+            i += 1
             phaser.register() // register the new worker thread
             es.submit(new CompactedParser(phaser, path, result))
           } else if (name.startsWith(UpdateFilePrefix) && name.endsWith(JsonPathSuffix)) {
             // old update files - have to parse them to get the partition name
+            i += 1
             phaser.register() // register the new worker thread
             es.submit(new UpdateParser(phaser, path, result))
           }
+        }
+        if (iter.hasNext) {
+          es.submit(new DirectoryWorker(es, new Phaser(phaser, 1), iter, result))
         }
       } catch {
         case _: FileNotFoundException => // the partition dir was deleted... just return
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
@@ -127,6 +127,8 @@ class JdbcMetadata(
     }
   }
 
+  override def invalidate(): Unit = {}
+
   override def close(): Unit = pool.close()
 }
 
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PathCache.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PathCache.scala
@@ -27,10 +27,10 @@ object PathCache {
   // cache for checking existence of files
   private val pathCache =
     Caffeine.newBuilder().expireAfterWrite(duration, TimeUnit.MILLISECONDS).build(
-          new CacheLoader[(FileContext, Path), java.lang.Boolean]() {
-            override def load(key: (FileContext, Path)): java.lang.Boolean = key._1.util.exists(key._2)
-          }
-        )
+      new CacheLoader[(FileContext, Path), java.lang.Boolean]() {
+        override def load(key: (FileContext, Path)): java.lang.Boolean = key._1.util.exists(key._2)
+      }
+    )
 
   // cache for individual file status
   private val statusCache =
@@ -43,11 +43,11 @@ object PathCache {
   // cache for checking directory contents
   private val listCache =
     Caffeine.newBuilder().expireAfterWrite(duration, TimeUnit.MILLISECONDS).build(
-          new CacheLoader[(FileContext, Path), Stream[FileStatus]]() {
-            override def load(key: (FileContext, Path)): Stream[FileStatus] =
-              RemoteIterator(key._1.listStatus(key._2)).toStream
-          }
-        )
+      new CacheLoader[(FileContext, Path), Stream[FileStatus]]() {
+        override def load(key: (FileContext, Path)): Stream[FileStatus] =
+          RemoteIterator(key._1.listStatus(key._2)).toStream
+      }
+    )
 
   /**
     * * Register a path as existing
@@ -89,7 +89,12 @@ object PathCache {
     * @param path path
     * @return
     */
-  def status(fc: FileContext, path: Path): FileStatus = statusCache.get((fc, path))
+  def status(fc: FileContext, path: Path, reload: Boolean = false): FileStatus = {
+    if (reload) {
+      invalidate(fc, path)
+    }
+    statusCache.get((fc, path))
+  }
 
   /**
     * List the children of a path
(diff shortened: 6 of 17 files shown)


docs/user/filesystem/commandline.rst
```diff
@@ -182,3 +183,24 @@ Argument                 Description
...
+======================== ==================================================================
+``--partitions``         The name of partitions to check, or none for all partitions
+``--repair``             Incrementally update the metadata to fix any inconsistencies found
+``--rebuild``            Rebuild the entire metadata based on the data files found
+``--force``              Force execution without prompting
+======================== ==================================================================
+
+.. warning::
+
+  Using both ``--partitions`` and ``--rebuild`` will cause any data files not in the specified
```
jnh5y, line 205: Nice warning!  Hopefully it'll prevent the 😱s...

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
```diff
@@ -100,6 +100,19 @@ trait StorageMetadata extends Compactable with Closeable {
     */
   def removePartition(partition: PartitionMetadata): Unit
 
+  /**
+   * Overwrite any existing partitions
+   *
+   * @param partitions partitions
+   */
+  def setPartitions(partitions: Seq[PartitionMetadata]): Unit =
+    throw new NotImplementedError() // TODO remove default impl in next major release
```
jnh5y, line 109: Is there already a deprecation on the interface?  Also, do we already have a ticket for that particular deprecation?

elahrvivaz: this is a new method, not a deprecation. i added a default impl so that any existing classes would still compile against it.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/FileSystemThreadedReader.scala
```diff
@@ -212,8 +224,6 @@ object FileSystemThreadedReader extends StrictLogging {
       iter: => CloseableIterator[SimpleFeature]
     ) extends Runnable {
 
-    phaser.register()
```
jnh5y, line 215: I think I followed the other Phasers.  Where did this one go?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -140,6 +140,25 @@ class FileBasedMetadata(
...
+  override def setPartitions(partitions: Seq[PartitionMetadata]): Unit = {
+    val map = new ConcurrentHashMap[String, PartitionFiles]
+    this.partitions.put(BoxedUnit.UNIT, map)
+    this.metadata.invalidateAll()
+
+    val configs = partitions.map { partition =>
+      val action = PartitionAction.Add
+      val envelope = partition.bounds.map(b => EnvelopeConfig(b.envelope)).getOrElse(Seq.empty)
+      val config = PartitionConfig(partition.name, action, partition.files, partition.count, envelope, System.currentTimeMillis())
+      // note: side effects in map
```
jnh5y, line 152: Can you say more about the side effects here?  I'm not seeing which of these calls has a side effect.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -328,37 +340,65 @@ class FileBasedMetadata(
...
+   */
+  private def delete(paths: Iterable[Path], threads: Int): Unit = {
+    if (threads < 2) {
+      paths.foreach(fc.delete(_, false))
+    } else {
+      val ec = ExecutionContext.fromExecutorService(new CachedThreadPool(threads))
+      try {
+        val parPaths = paths.par
+        parPaths.tasksupport = new ExecutionContextTaskSupport(ec)
+        parPaths.foreach(fc.delete(_, false))
```
jnh5y, line 357: Do we need any more error handling here? 

It seems like some deletes may need to be retried.  

Also, is the `paths` variable just the files?  Or does it include directories?  (I.e., can we get a delete called on a directory before it is empty?  And if deleting a file fails the first time, is any attempt to retry it?)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -328,37 +340,65 @@ class FileBasedMetadata(
...
+        while (iter.hasNext && i < PhaserUtils.MaxParties) {
           val status = iter.next
           val path = status.getPath
           lazy val name = path.getName
           if (status.isDirectory) {
-            phaser.register() // register the new worker thread
-            es.submit(new DirectoryWorker(es, phaser, path, result))
+            i += 1
+            // use a tiered phaser on each directory avoid the limit of 65535 registered parties
+            es.submit(new DirectoryWorker(es, new Phaser(phaser, 1), fc.listStatus(path), result))
```
jnh5y, line 382: Does this cover the situation where there is more than one directory with over 65K children?  And what happens if there are 65K+ files in the root directory?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -328,37 +340,65 @@ class FileBasedMetadata(
...
             phaser.register() // register the new worker thread
             es.submit(new CompactedParser(phaser, path, result))
           } else if (name.startsWith(UpdateFilePrefix) && name.endsWith(JsonPathSuffix)) {
             // old update files - have to parse them to get the partition name
+            i += 1
             phaser.register() // register the new worker thread
             es.submit(new UpdateParser(phaser, path, result))
           }
         }
+        if (iter.hasNext) {
```
jnh5y, line 399: Ok, if I'm understanding the strategy, as the iterator is looped through, tasks are submitted up to the 65K limit for the Phaser to manage.  

How quickly does this happen?  Basically, could there be a situation where the JVM heap has millions of objects on it representing the work that needs to be done and that causes an OOM?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
```diff
@@ -108,10 +108,19 @@ class JdbcMetadata(
...
+      PartitionsTable.clear(connection, root)
+      partitions.foreach(PartitionsTable.insert(connection, root, _))
+    }
+  }
+
   // noinspection ScalaDeprecation
   override def compact(partition: Option[String], threads: Int): Unit = compact(partition, None, threads)
 
   override def compact(partition: Option[String], fileSize: Option[Long], threads: Int): Unit = {
+    // TODO execute as a transaction
```
jnh5y, line 123: As usual, I'm on the fence about a ticket here.  To your point, if we aren't planning on it, no need.  Other the other hand, if we really ought to do it, it may be worth being in the backlog.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PathCache.scala
```diff
@@ -98,7 +103,12 @@ object PathCache {
     * @param dir directory path
     * @return
     */
-  def list(fc: FileContext, dir: Path): Iterator[FileStatus] = listCache.get((fc, dir)).iterator
+  def list(fc: FileContext, dir: Path, reload: Boolean = false): Iterator[FileStatus] = {
```
jnh5y, line 106: Hmm.... neat.  This is a clear example of a function which accepts a boolean that controls whether or not there's a side effect and potentially any threads-safety issues.

Not sure I've see such a clear example of that.  (I'm not necessarily suggesting a change here.)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/StorageUtils.scala
```diff
@@ -43,10 +45,20 @@ object StorageUtils {
...
 
+  /**
+   * Extract the 'leaf' part of a partition name from a file name.
+   *
+   * If file is not from a leafed partition scheme, result is indeterminate and may error.
+   *
+   * @param name file name
+   * @return
+   */
+  def leaf(name: String): String = name.substring(0, name.indexOf(LeafSeparator))
```
jnh5y, line 60: Can typenames have '_'s?

Even if that's the case, the typename is completely to the right of the leaf, so that'd be ok?

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FsManageMetadataCommand.scala
```diff
@@ -82,6 +97,229 @@ object FsManageMetadataCommand {
...
   }
 
+  class CheckConsistencyCommand extends FsDataStoreCommand with LazyLogging {
+
+    override val name = "check-consistency"
+    override val params = new CheckConsistencyParams()
+
+    override def validate(): Option[ParameterException] = {
+      if (params.repair && params.rebuild) {
+        Some(new ParameterException("Please specify at most one of --repair and --rebuild"))
```
jnh5y, line 107: We will repair and we will rebuild!  Separately though...

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FsManageMetadataCommand.scala
```diff
@@ -82,6 +97,229 @@ object FsManageMetadataCommand {
...
+              } else {
+                val leafPartition =
+                  if (storage.metadata.leafStorage) { s"$partition/${StorageUtils.leaf(name)}" } else { partition }
+                if (partitions.forall(_.contains(leafPartition))) {
+                  onDisk.computeIfAbsent(leafPartition, computeFunction).put(name, java.lang.Boolean.TRUE)
+                }
+              }
+            }
+            if (iter.hasNext) {
+              pool.submit(new ListWorker(new Phaser(phaser, 1), partition, iter))
```
jnh5y, line 308: How should I be thinking about Phasers?  Is the "1" passed into the constructor a count representing the calling thread?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/FileSystemThreadedReader.scala
elahrvivaz: went up to line 203 here: https://github.com/locationtech/geomesa/pull/2713/files/eaf4f4eb2535bc5b12861df6146b4bf398ba4e06#diff-72f9fb4eace5357104fd4f3412a896d6891987d26b26e47b9e4940de76e51126R203

I thought it was clearer to register the phaser where it's created, instead of using the constructor of the reader to register it.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
elahrvivaz: oh, it's the 2 lines after this comment

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FsManageMetadataCommand.scala
```diff
@@ -82,6 +97,229 @@ object FsManageMetadataCommand {
...
+        }
+      }
+
+      private class ListWorker(phaser: Phaser, partition: String, listDirectory: => RemoteIterator[FileStatus])
+          extends Runnable {
+        override def run(): Unit = {
+          try {
+            var i = phaser.getRegisteredParties + 1
+            val iter = listDirectory
+            while (iter.hasNext && i < PhaserUtils.MaxParties) {
```
jnh5y, line 288: Is this pattern for having multiple Phasers worth abstracting?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
elahrvivaz: paths is just the files in this case, and the recursive flag is passed as false.

i don't think there are transient delete failures that you would be expected to catch and retry... 

generally, `compact` and similar commands are not very robust with regards to failures, probably something we should try to address better, but it's tricky without having anything like transactions.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
```diff
@@ -328,37 +340,65 @@ class FileBasedMetadata(
...
-      dir: Path,
+      listDirectory: => RemoteIterator[FileStatus],
       result: ConcurrentHashMap[String, PartitionFiles]
     ) extends Runnable {
 
     override def run(): Unit = {
       try {
-        val iter = fc.listStatus(dir)
-        while (iter.hasNext) {
+        var i = phaser.getRegisteredParties + 1
```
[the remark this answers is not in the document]
elahrvivaz: i think registered parties will always be 1 in this case. there are various checks to ensure that it does not end up being MaxParties, if it did then we wouldn't be able to create a new child phaser and things would break...

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
elahrvivaz: yeah, it will list out the first 65k and then create a new phaser to handle the next batch

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
elahrvivaz: ah, potentially... but these objects are fairly lightweight, all the state is mostly shared among workers (the iterator of results, the concurrent map to return results). you'd end up with at most a task per file or directory (and briefly some duplicates when/if it needs to pass off to a new phaser to list out a whole directory), so it's limited by the number of files you have on disk and how fast it parses through them. in practice it definitely consumes a decent amount of memory overall when you have a lot of files, but likely a lot of that is just keeping the data for each file in the cache.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
elahrvivaz: yeah, we already had a bunch of TODOs in here to note places where a transaction should be added...

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PathCache.scala
elahrvivaz: there shouldn't be any thread safety issues b/c everything is going back to guava caches, which are thread safe

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/StorageUtils.scala
elahrvivaz: i don't think typenames would be in these file names, they're usually `leaf + type char + uuid`. but yeah, the first _ should delimit the leaf so if there are other ones it should be ok. now, if the last piece of the partition has underscores we're screwed :P

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FsManageMetadataCommand.scala
elahrvivaz: that's the number of parties you want to pre-register with the phaser. it's equivalent to `val p = new Phaser(); p.register()`

geomesa-fs/geomesa-fs-tools/src/main/scala/org/locationtech/geomesa/fs/tools/ingest/FsManageMetadataCommand.scala
elahrvivaz: possibly, but it didn't seem particularly easy to do... we're using it in 2 places, if we get a third we can abstract it then ;)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
jnh5y: Ah, so the TODO is to add implementations to the other classes which inherit from this interface?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
jnh5y: Sounds good!  This as-is is probably ok.

Would it be the case that some operations could sensibly be repeated?  E.g., if a delete hit an issue, could be it re-run and work ok?  (Or maybe I want to that a compact/metadata clean up is repeated?)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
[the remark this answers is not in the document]
jnh5y: What all would be needed for that to happen?  Would we have to have 65K * 65K files or something like that?

Separate question, do we need to help people with 4 billion+ files in a directory?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/JdbcMetadata.scala
jnh5y: +1.  Yeah, if all this is for transaction support, yeah, no need for a ticket.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/StorageUtils.scala
jnh5y: Right, the filename doesn't have the typename, so we are fine...

Wait, what's the last piece of partition?  Ok, I'm confused.  Did I find a bug or not:)?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
elahrvivaz: all of our classes already have impls, it's just in case external people have added impls. with the default NotimplementedError, it doesn't break semver, but once we remove that it will (at least that's my interpretation...) so the TODO is for the api breaking change that we can't make until a major release

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
elahrvivaz: i can't really think of any transient issues that would likely resolve with a retry, but it's possible... if something does go wrong, then you can use the consistency check commands here to resolve the discrepancies. but yeah, in general we need to be a lot more robust, but that's very hard to pull off

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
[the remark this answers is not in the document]
elahrvivaz: i don't think it can happen as currently coded, was my point, but it's possible there's a bug in the logic.

re: 4B, are you thinking we'd hit some memory limit or something? i haven't considered that, but it seems untenable with the current architecture.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/StorageUtils.scala
elahrvivaz: well, partitions can technically be anything since it's extensible. i think we have some TODOs around making them be more robust (i.e. we determine the depth of a partition by counting `/`s). we support attribute partitions, so it's possible someone could create a partition scheme like 'yyyy/MM/dd/<attr-foo>' and the encoded attribute value could have underscores in it. it could also have slashes in it, which would also break things. so, yeah it could potentially be a bug but in practice probably not, and we'd need to change some things across the code base to fix it.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
jnh5y: Ah!  Got it!

Some day, I want to see a minimal GitHub/Java/Scala project to show all these little version / API upgrade situations.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
jnh5y: Sounds good.  Just asking 8000 questions to see if there are easy ways to make things more robust:).

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PathCache.scala
jnh5y: Yeah, I wasn't thinking that there were any particular thread safety issues; good to know that you've thought through it!

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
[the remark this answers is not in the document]
jnh5y: Eh, at this point in code reviews I'm just trying to get a sense for places where things will break.  Since this PR about a 65K limit with Phasers, it seems natural to ask where the next boundary is.

It seems like we have a Phaser of Phasers, right?  If that's correct, then if we spun up 65K+ Phasers, the outer Phaser would break, right?  And to get there, we'd 65K * 65 ~= 4 billion+ files in a directory?  

If my statement above is wrong, then that's fine.  My second question/point is basically, if someone gets 4 billion files in a directory, they may be beyond our assistance:) and a GeoMesa bug will be the least of their concerns.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
elahrvivaz: i think we'd have to re-architect things a bit for that to work. with a finite queue you'd have to use a different thread pool for the producer and the consumer, otherwise your threads could all be used up by producers waiting to put things in the queue, with no consumers able to read things back off. we'd need to come up with a way to be notified when all the threads are done, as well. yeah, it might be worthwhile to re-think this approach at some point, and that might be a good replacement.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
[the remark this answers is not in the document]
elahrvivaz: ah, gotcha. no, it shouldn't hit any phaser limits at the moment. iirc if it hits the limit it will use the last registration spot to create a child phaser, and then start using that (which will give another 65k registrations), and if it hits that limit it will keep creating more levels of child phasers, etc.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/metadata/FileBasedMetadata.scala
jnh5y: Cool.

PR #2715, opened 2021-03-09
GEOMESA-3015 Migrate to ejml 0.34

jnh5y:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,6 +68,8 @@
         <geoserver.version>2.17.3</geoserver.version>
         <jts.version>1.17.0</jts.version> <!-- needs to track with gt/gs version -->
         <spatial4j.version>0.8</spatial4j.version> <!-- needs to track with jts version -->
+        <!-- match dependency version used by gt-referencing for convenience -->
+        <ejml.version>0.34</ejml.version>
         <avro.version>1.8.2</avro.version> <!-- see also confluent.avro.version -->
         <parquet.version>1.9.0</parquet.version>
         <orc.version>1.5.4</orc.version>
@@ -953,6 +955,11 @@
                 </exclusions>
             </dependency>
             <dependency>
+                <groupId>org.ejml</groupId>
+                <artifactId>ejml-all</artifactId>
+                <version>${ejml.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>org.geotools</groupId>
                 <artifactId>gt-referencing</artifactId>
                 <version>${gt.version}</version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -68,6 +68,8 @@
         <geoserver.version>2.17.3</geoserver.version>
         <jts.version>1.17.0</jts.version> <!-- needs to track with gt/gs version -->
         <spatial4j.version>0.8</spatial4j.version> <!-- needs to track with jts version -->
+        <!-- match dependency version used by gt-referencing for convenience -->
```
elahrvivaz, line 71: is org.ejml actually used by gt-referencing now? previously they were using the com.googlecode.efficient-java-matrix-library ejml library.

pom.xml
```diff
@@ -952,6 +954,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.ejml</groupId>
+                <artifactId>ejml-all</artifactId>
```
elahrvivaz, line 959: did you run the build cqs script? i assume we'll need to check the license.

pom.xml
jnh5y: Looks like: https://github.com/geotools/geotools/blob/main/pom.xml#L1085-L1089
https://github.com/geotools/geotools/blob/main/modules/library/referencing/pom.xml#L173-L177

build/cqs.tsv
```diff
@@ -200,6 +199,7 @@ org.eclipse.xsd:org.eclipse.xsd	2.12.0	compile
 org.ehcache:sizeof	0.3.0	compile
 org.ejml:ejml-core	0.34	compile
 org.ejml:ejml-ddense	0.34	compile
+org.ejml:ejml-simple	0.34	compile
```
jnh5y, line 202: Other versions of ejml-simple are on ClearlyDefined: https://clearlydefined.io/?sort=releaseDate&sortDesc=true&name=ejml-simple

Thoughts?

pom.xml
elahrvivaz: i wonder, do you know if they switched or if the old comment was wrong? good catch though

build/cqs.tsv
elahrvivaz: looks like 0.34 is on there too, declared as apache 2.0. good enough i guess? https://clearlydefined.io/definitions/maven/mavencentral/org.ejml/ejml-simple/0.34

elahrvivaz: well actually it's also in that link you sent with 'discovered' as apache 2.0

pom.xml
jnh5y: Lemme see.... if I recall, GT was using vecmath and then switched to com.googlecode.efficient-java-matrix-library 0.26 and then upgrade to 0.34.  

The move from vecmath was motivated by licenses back in the day; I think I helped with that at FOSS4G NA 2015...  Can't remember when the 0.26 -> 0.34 transition happened.

PR #2717, opened 2021-03-09
GEOMESA-3018 Update Maven release plugin to drop Scala ABI version fr…

jnh5y:
…om release tag


jnh5y:
> did you test this out at all?

I think I tried the other day with -DdryRun=true, and it didn't say what Git tags it would make...

I'm not sure how to try it out without releasing.  Ideas?

elahrvivaz:
nope, sounds good. it asks you to confirm the tag name anyway, right? so we can double-check when we make the release.

PR #2718, opened 2021-03-10
GEOMESA-3016 Update Nifi docs

elahrvivaz:


docs/user/nifi/avro.rst
```diff
@@ -1,17 +1,14 @@
...
 
-GeoAvro processors (``AvroToPutGeoMesa``) accept the following configuration parameters for specifying the
+GeoAvro processors (``AvroToPutGeoMesa*``) accept the following configuration parameters for specifying the
 input source. Each datastore-specific processor also has additional parameters for connecting to the datastore,
 detailed in the following sections.
 
-+-------------------------+-------------------------------------------------------------------------------------------+
-| Property                | Description                                                                               |
-+=========================+===========================================================================================+
-| ``Avro SFT match mode`` | Determines how Avro SimpleFeatureType mismatches are handled.                             |
```
jnh5y, line 11: What's the replacment for the Avro match mode in the GeoFile processors?

elahrvivaz: it's outlined in the common data store ingest params, 'CompatibilityMode'

jnh5y: Ok, I tried to look and I'm not finding it quickly.  Can you linkslap to the nightly docs or the line in GeoMesa source?

jnh5y: Wait, wait.... the Avro SFT Match mode is different from schema updates, right?!

The former is for handling out of order attributes in Avro files and the other is for adding new columns on the end....  Did the Avro match mode go away?

elahrvivaz: it's the same idea, but schema compatibility also lets you update the existing schema. if you use 'exact' it will have to match exactly (strict match mode), if you use 'existing' it will adapt the incoming data to match the existing schema (lenient match mode), you can also use 'update' to update the existing schema to match the avro file (which wasn't possible before)

jnh5y: But wait.... when we ingest an Avro file that's got out of order attributes, we wouldn't want that to update the schema in GeoMesa.

Is that handling still present?

elahrvivaz: yes, that's the 'existing' mode. it will only update the schema if you use 'update'.

elahrvivaz: previously we had 2 ways to handle discrepancies between the input and the schema on disk - require exact match, or map the input to the existing schema. we still have those 2 (strict and lenient, now exact and existing) but we also added the ability to update the existing schema to match the input. this matches the behavior of the other ingest processors.

jnh5y: Ooooooh.....  ok, so the "match by name" is covered by the "existing" option, "exact" will reject mismatched data, and "update" will add columns?

docs/user/nifi/processors.rst
```diff
@@ -79,6 +80,12 @@ All types of input processors have some common configuration parameters for cont
 | ``Identifying Attribute``     | When using a modifying writer, the attribute used to uniquely identify the feature.     |
 |                               | If not specified, will use the feature ID                                               |
 +-------------------------------+-----------------------------------------------------------------------------------------+
+| ``Schema Compatibility``      | Controls how differences between the configured schema and the existing schema in the   |
+|                               | data store (if any) will be handled. ``Existing`` will use the existing schema and drop |
+|                               | any additional fields in the configured schema. ``Update`` will update the existing     |
+|                               | schema to match the configured schema. ``Exact`` requires the configured schema to      |
+|                               | match the existing schema.                                                              |
```
jnh5y, line 87: Is it easy to add break out bullets for each of these options inside this table?

docs/user/nifi/avro.rst
```diff
@@ -1,17 +1,14 @@
...
-+-------------------------+-------------------------------------------------------------------------------------------+
-
-The SimpleFeatureTypes in GeoAvro may or may not match the SimpleFeatureType in the target datastore.
-To address this, the AvroToPut processors have a property to set the SFT match mode. It can either be set to
-an exact match ("by attribute number and order") or a more lenient one ("by attribute name"). The latter setting
-will not write fields which are not in the target SFT.
```
jnh5y, line 17: Do the AvroToPutGeoMesa* Processors inherit the schema compatibility modes from the converter processors?

If so, should the docs link to them?  (Thanks for answering all these questions!  I find this to be some subtle stuff we are keeping track of.)

elahrvivaz: they use the same property, but they are handled slightly differently... i'd have to go check exactly how the converters handle `existing` mode. i'm not sure a cross-link would be useful though, other than in a "look you can also use converters" way.

docs/user/nifi/processors.rst
elahrvivaz: ah, good question. i think we do that other places, so ti should be possible. i'll try to give it a shot this afternoon.

PR #2719, opened 2021-03-12
GEOMESA-3023 Fix download of scala-console

elahrvivaz:


geomesa-tools/conf-filtered/geomesa-env.sh
```diff
@@ -362,7 +362,7 @@ function geomesa_scala_console() {
   # Check if we already downloaded scala
   if [[ -d "${%%tools.dist.name%%_HOME}/dist/scala-%%scala.version%%/" ]]; then
     scalaCMD="${%%tools.dist.name%%_HOME}/dist/scala-%%scala.version%%/bin/scala"
-  elif [[ $(which scala > /dev/null) -eq 0 && -n "$(scala -version 2>&1 | grep %%scala.binary.version%%)" ]]; then
```
jnh5y, line 365: What corner case / situation does this change cover?

elahrvivaz: it was just a bug in the logic... the check doesn't actually work correctly. if you have scala on the path already then it proceeds fine, but if not it will fail and not download it. we can back-port this to 3.x, it was broken in some refactoring starting with 3.0.0

PR #2720, opened 2021-03-12
GEOMESA-2903 Log HBase zookeeper connection

elahrvivaz:

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
@@ -101,6 +101,7 @@ object HBaseConnectionPool extends LazyLogging {
       ProvidedConnection(ConnectionParam.lookup(params))
     } else {
       val conf = getConfiguration(params)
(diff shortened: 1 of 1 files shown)


elahrvivaz:
from what I can tell, there's no good way to see what resources are loaded through a configuration object. the resources that it contains may or may not point to actual files, it only determines that when it tries to parse them. they are also relative names (i.e. `hbase-site.xml`), the full path is only known internally when they are parsed. so the only other sane option i see is to print out the entire conf, but that seems very verbose.

elahrvivaz:
kerberos info is already logged

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseConnectionPool.scala
```diff
@@ -101,6 +101,7 @@ object HBaseConnectionPool extends LazyLogging {
       ProvidedConnection(ConnectionParam.lookup(params))
     } else {
       val conf = getConfiguration(params)
+      logger.debug(s"Connecting to HBase instance at ${conf.get(HConstants.ZOOKEEPER_QUORUM)}")
```
jnh5y, line 104: I wonder if we should add logic to add warning if the ZooKeeper is localhost.  

Meh, probably too much.

elahrvivaz: we do have a warning about that already actually.

PR #2722, opened 2021-03-15
GEOMESA-3028 Split Github actions for Scala 2.11 and 2.12

elahrvivaz:


elahrvivaz:
I'm not quite sure why this isn't triggering the builds, but it seems to if I push directly to LT. I assume it will work correctly after it gets merged.

PR #2735, opened 2021-03-19
GEOMESA-3001 Upgrade install version for httpclient

elahrvivaz:

### DIFF
--- a/geomesa-fs/geomesa-fs-tools/conf-filtered/dependencies.sh
+++ b/geomesa-fs/geomesa-fs-tools/conf-filtered/dependencies.sh
@@ -53,8 +53,9 @@ function dependencies() {
     "com.amazonaws:aws-java-sdk-s3:${aws_sdk_version}:jar"
     # joda-time required for aws sdk
     "joda-time:joda-time:2.8.1:jar"
-    "org.apache.httpcomponents:httpclient:4.3.4:jar"
-    "org.apache.httpcomponents:httpcore:4.3.3:jar"
+    # these are the versions used by hadoop 2.8 and 3.1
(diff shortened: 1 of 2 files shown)


jnh5y:
The file `docs/user/filesystem/install.rst` also mentions the 4.3.4 version of these libraries.  That should be updated.

geomesa-fs/geomesa-fs-tools/conf-filtered/dependencies.sh
```diff
@@ -53,8 +53,9 @@ function dependencies() {
     "com.amazonaws:aws-java-sdk-s3:${aws_sdk_version}:jar"
     # joda-time required for aws sdk
     "joda-time:joda-time:2.8.1:jar"
-    "org.apache.httpcomponents:httpclient:4.3.4:jar"
-    "org.apache.httpcomponents:httpcore:4.3.3:jar"
+    # these are the versions used by hadoop 2.8 and 3.1
+    "org.apache.httpcomponents:httpclient:4.5.2:jar"
+    "org.apache.httpcomponents:httpcore:4.4.4:jar"
```
jnh5y, line 58: I wonder if we should switch these to being filtered?  😬   Not sure what's best here.  Your call.

elahrvivaz: i considered it but we'd have to pull the version out into a property and seemed like overkill since we don't do that for most jars. i'm not sure where these versions came from tbh, maybe a really old version of hadoop.

jnh5y: Cool.  Shipping it!

PR #2737, opened 2021-03-19
GEOMESA-2991 Configuration to require visibilities on write

elahrvivaz:
* New index configuration `geomesa.vis.required`
* Writes are blocked in the API, so not 100% foolproof
* Accumulo will also apply the ReqVisFilter on all tables

### DIFF
--- a/docs/user/datastores/index_config.rst
+++ b/docs/user/datastores/index_config.rst
@@ -638,6 +638,9 @@ Note that this configuration will prevent missing visibility labels in normal wr
 it is still possible to write unlabelled data through older clients, bulk loading, or direct access
 to the underlying database.
 
+In Accumulo data stores, setting this configuration will also set the Accumulo ``ReqVisFilter`` on all data
+tables, which will prevent any unlabelled data from being returned in queries.
+
 Mixed Geometry Types
 --------------------
 
--- a/docs/user/datastores/index_config.rst
+++ b/docs/user/datastores/index_config.rst
@@ -619,6 +619,25 @@ user-data key ``geomesa.temporal.priority``:
 
 This may be configured before calling ``createSchema``, or updated by calling ``updateSchema``.
 
+.. _required_vis_config:
+
+Configuring Required Visibilities
+---------------------------------
+
+GeoMesa supports :ref:`data_security` through the use of visibility labels to secure access to sensitive data.
+To help prevent data spills, a schema can be configured to reject any writes that don't contain valid visibilities.
+To enable this setting, use the user-data key ``geomesa.vis.required``:
+
+.. code-block:: java
+
+    sft.getUserData().put("geomesa.vis.required", "true");
+
+This may be configured before calling ``createSchema``, or updated by calling ``updateSchema``.
+
+Note that this configuration will prevent missing visibility labels in normal write paths, but that
+it is still possible to write unlabelled data through older clients, bulk loading, or direct access
+to the underlying database.
+
 Mixed Geometry Types
 --------------------
 
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
@@ -18,9 +18,10 @@ import org.opengis.feature.simple.SimpleFeatureType
 object VisibilityIterator {
 
   val Name = "ReqVisFilter"
+  val Priority = 15 // run before the accumulo versioning iterator at 20, and before any of our custom iterators
 
   def set(tableOps: TableOperations, table: String): Unit =
-    tableOps.attachIterator(table, new IteratorSetting(30, Name, classOf[ReqVisFilter]))
+    tableOps.attachIterator(table, new IteratorSetting(Priority, Name, classOf[ReqVisFilter]))
 
   def set(ds: AccumuloDataStore, sft: SimpleFeatureType): Unit = {
     val tableOps = ds.connector.tableOperations()
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
@@ -1,0 +1,38 @@
+/**********************************************************************
+ * Copyright (c) 2013-2021 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ **********************************************************************/
+
+package org.locationtech.geomesa.accumulo.iterators
+
+import org.apache.accumulo.core.client.IteratorSetting
+import org.apache.accumulo.core.client.admin.TableOperations
+import org.apache.accumulo.core.iterators.IteratorUtil.IteratorScope
+import org.apache.accumulo.core.iterators.user.ReqVisFilter
+import org.locationtech.geomesa.accumulo.data.AccumuloDataStore
+import org.opengis.feature.simple.SimpleFeatureType
+
+object VisibilityIterator {
+
+  val Name = "ReqVisFilter"
+
+  def set(tableOps: TableOperations, table: String): Unit =
+    tableOps.attachIterator(table, new IteratorSetting(30, Name, classOf[ReqVisFilter]))
+
+  def set(ds: AccumuloDataStore, sft: SimpleFeatureType): Unit = {
+    val tableOps = ds.connector.tableOperations()
+    ds.getAllIndexTableNames(sft.getTypeName).filter(tableOps.exists).foreach(set(tableOps, _))
+  }
+
+  def clear(ds: AccumuloDataStore, sft: SimpleFeatureType): Unit = {
+    val tableOps = ds.connector.tableOperations()
+    ds.getAllIndexTableNames(sft.getTypeName).filter(tableOps.exists).foreach { table =>
+      if (IteratorScope.values.exists(scope => tableOps.getIteratorSetting(table, Name, scope) != null)) {
+        tableOps.removeIterator(table, Name, java.util.EnumSet.allOf(classOf[IteratorScope]))
+      }
+    }
+  }
+}
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
@@ -1,10 +1,10 @@
-/**********************************************************************
+/***********************************************************************
  * Copyright (c) 2013-2021 Commonwealth Computer Research, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Apache License, Version 2.0
  * which accompanies this distribution and is available at
  * http://www.opensource.org/licenses/apache2.0.php.
- **********************************************************************/
+ ***********************************************************************/
 
 package org.locationtech.geomesa.accumulo.iterators
 
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
@@ -31,8 +31,6 @@ import scala.util.control.NonFatal
 
 trait GeoMesaFeatureWriter[DS <: GeoMesaDataStore[DS]] extends SimpleFeatureWriter with Flushable with LazyLogging {
 
-  import scala.collection.JavaConverters._
-
   def ds: DS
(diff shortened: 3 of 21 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
```diff
@@ -0,0 +1,38 @@
+/**********************************************************************
+ * Copyright (c) 2013-2021 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ **********************************************************************/
```
jnh5y, line 7: The license check is dying on this file for some reason:

`Missing header in: /home/runner/work/geomesa/geomesa/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala`

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.accumulo.core.iterators.user.ReqVisFilter
+import org.locationtech.geomesa.accumulo.data.AccumuloDataStore
+import org.opengis.feature.simple.SimpleFeatureType
+
+object VisibilityIterator {
+
+  val Name = "ReqVisFilter"
+
+  def set(tableOps: TableOperations, table: String): Unit =
+    tableOps.attachIterator(table, new IteratorSetting(30, Name, classOf[ReqVisFilter]))
```
jnh5y, line 23: The stats iterator has priority 30.  This iterator should be fired earlier than that so the stats for records with empty visibilities are not analyzed.

jnh5y, line 23: Does HBase have a similar filter that could be used?  Maybe it is worth a "todo" or ticket to indicate that adding something like this in HBase is not done / is future work?

jnh5y:
Would this prevent writing to a KDS sans vis?

elahrvivaz:
> Would this prevent writing to a KDS sans vis?

no, we'd have to add that separately as it doesn't use the index adapters.

jnh5y:
> > Would this prevent writing to a KDS sans vis?
> 
> no, we'd have to add that separately as it doesn't use the index adapters.

Hmm... ok, do we need to do that for the KDS as part of this PR?  Or should we add caveats about which datastores work with the limited writes?

I'm asking since the documentation is about a configuration on GeoMesa datastores generally.

elahrvivaz:
the ticket was only for accumulo, but it probably makes sense to add it to the KDS too.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
elahrvivaz: seems like this could safely go before the default accumulo versioning iterator, no?

jnh5y: Assuming that this is settings for scan time only, yes, I think that is reasonable.  

In terms of minc/majc, I don't have a clear rule that'd help determine what to do here.

docs/user/datastores/index_config.rst
```diff
@@ -619,6 +619,25 @@ user-data key ``geomesa.temporal.priority``:
 
 This may be configured before calling ``createSchema``, or updated by calling ``updateSchema``.
 
+.. _required_vis_config:
+
+Configuring Required Visibilities
```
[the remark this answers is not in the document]
elahrvivaz: it's enabled for all scopes by default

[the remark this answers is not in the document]
elahrvivaz: added note to the docs about ReqVisFilter

jnh5y:
> the ticket was only for accumulo, but it probably makes sense to add it to the KDS too.

Yeah, I'm just imagining how this would play out in NiFi.  For users with a lambda architecture writing to the KDS and Accumulo, they may be surprised if data sans vis ended up elsewhere.  (Although, the KDS does "fail closed" on empty vis, so maybe it is ok?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/VisibilityIterator.scala
elahrvivaz: not sure... given that hbase vis are tacked on as coprocessors that might get messy :P

jnh5y: Ayup, I know; that's why I was thinking docs or tickets to let people know that HBase support for this is more work.  The docs changes may have covered this.

elahrvivaz: it's applied to all scopes, but really it doesn't seem like it would matter for different scopes - since vis is part of the key, the record would be killed out regardless.

docs/user/datastores/index_config.rst
```diff
@@ -638,6 +638,9 @@ Note that this configuration will prevent missing visibility labels in normal wr
 it is still possible to write unlabelled data through older clients, bulk loading, or direct access
 to the underlying database.
 
+In Accumulo data stores, setting this configuration will also set the Accumulo ``ReqVisFilter`` on all data
+tables, which will prevent any unlabelled data from being returned in queries.
```
jnh5y, line 642: Since we are setting the ReqVisFilter on scan, majc, and minc, this will do more than prevent data from being returned in queries.  It ought to be prevent data without vises from being persisted in the database or remaining in the database (if the configuration is added).

Part of me wonders if we should separate the GeoMesa writing concern from the stronger guarantees in Accumulo.

elahrvivaz: it will only delete things when they get compacted though, which potentially might never happen?

jnh5y: Hehehe, we are into corner cases on corner cases.  Generally, I think it is safe to assume that all tables are going to see minor and major compactions.  

From https://accumulo.apache.org/docs/2.x/configuration/server-properties there are two table.compaction.*.idle properties that say that when a tablet is idle for a period of time a compaction may happen.  (For minc, there is a note that says "There is no guarantee an idle tablet will be compacted."....)

I suppose I'm trying to figure out if there'd a situation where someone would want to have the write protection but not have the Accumulo iterators configured?  This one is obviously tough:).

elahrvivaz: i can't really think of a situation where someone would want one and not the other... if they really wanted to they could go remove the iterator configs by hand.

elahrvivaz:
> Although, the KDS does "fail closed" on empty vis, so maybe it is ok?

i don't think it does fail closed, it uses the same logic as accumulo iirc

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
```diff
@@ -51,8 +49,13 @@ trait GeoMesaFeatureWriter[DS <: GeoMesaDataStore[DS]] extends SimpleFeatureWrit
     // `write` will calculate all mutations up front in case the feature is not valid, so we don't write partial entries
     try { getWriter(writable).write(writable, update) } catch {
       case NonFatal(e) =>
-        val attributes = s"${writable.getID}:${writable.getAttributes.asScala.mkString("|")}"
-        throw new IllegalArgumentException(s"Error indexing feature '$attributes'", e)
+        val msg = s"Error indexing feature '${writable.getID}:${DataUtilities.encodeFeature(writable, false)}'"
+        if (e.isInstanceOf[IllegalArgumentException]) {
+          throw new IllegalArgumentException(msg, e)
+        } else {
+          throw new RuntimeException(msg, e)
```
jnh5y, line 56: Ooo, this is new.  What gives us a non-IllegalArgumentException?

elahrvivaz: errors in e.g. writing a mutation to hbase (table offline, timeout, whatever). i think this code originally just generated the mutations, but didn't write them, so throwing IllegalArgumentException was correct. it was updated at some point to both generate mutations and write them, but the catch/throw wasn't updated. admittedly this solution isn't great, but it requires the least change. someone on gitter reported an error thrown by this block which made me check it again.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureWriter.scala
```diff
@@ -122,4 +124,13 @@ object KafkaFeatureWriter {
       producer.send(record)
     }
   }
+
+  trait RequiredVisibilityWriter extends AppendKafkaFeatureWriter {
```
jnh5y, line 128: Just to make sure I'm following, these traits extend the class that they are mixed back into in order to override a method from the base class, right?

Would an alternative by to have this trait require a `feature: SimpleFeature =>` and then we could mix it in more easily?  (I think I'm thinking "cake pattern" ish...)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala
jnh5y: Perfect!

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureWriter.scala
elahrvivaz: i thought this was the cake pattern haha. but i wanted to keep it fairly tightly bound so that we're not executing "excess" code on each write.

jnh5y: I guess I'm wondering if we can have one `RequiredVisibilityWriter` which extends something like `FeatureWriter` and then mixes in.

Such a class may need the little syntax for requiring a "feature: SimpleFeature" to be part of it, right?

elahrvivaz: ah, maybe, but the one in index-api is hooked into IndexWriter not FeatureWriter. also, since feature is a var i'm not sure if that would work? i think it would have to be redefined with a function accessor.

jnh5y: Eh, I think a var would work.  Do you want to try or leave it as-is?

elahrvivaz: i could add it at the geomesa feature writer level, but the check here would double in size: https://github.com/locationtech/geomesa/blob/main/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala#L101-L111
otherwise, i'd have to standardize some var names here and make these public/ maybe protected: https://github.com/locationtech/geomesa/blob/main/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/GeoMesaFeatureWriter.scala#L225

i guess i'm not crazy about extracting it to the feature writer level b/c it relies on some low-level implementation details

geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/data/LambdaFeatureWriter.scala
```diff
@@ -63,12 +62,10 @@ object LambdaFeatureWriter {
     override def close(): Unit = features.close()
   }
 
-  trait RequiredVisibilityWriter extends AppendLambdaFeatureWriter {
+  trait RequiredVisibilityWriter extends AppendLambdaFeatureWriter with VisibilityChecker {
```
jnh5y, line 65: Hmm.... I was trying to avoid the repeated trait which extends the specific writers...  I may poke at this for a minute.

elahrvivaz: yeah, i just don't really like exposing the simple feature vars into a trait... seems like a pretty low-level detail. if the feature writer api methods took a feature as input, then we could just mix it in at that level, but since they use the `hasNext`, `next`, `write` api calls it doesn't work very cleanly.

PR #2739, opened 2021-03-25
GEOMESA-3031 Example for NiFi Kafka to HBase flow

elahrvivaz:


docs/user/nifi/examples.rst
```diff
@@ -0,0 +1,55 @@
+Examples
+--------
+
+Ingest a GeoMesa Kafka Topic into HBase
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A common use case is to store the most recent state of an object in a GeoMesa Kafka data store to drive a live map
+view, but to persist every state of the object into long-term storage in HBase or Accumulo for history and
```
jnh5y, line 8: I think the "but" could be an "and" here.

docs/user/nifi/install.rst
```diff
@@ -1,3 +1,5 @@
+.. _nifi_install:
+
 Installation
```
jnh5y, line 3: Can we update this to be "GeoMesa NiFi Installation" so that the ref to it reads as...

"Follow the steps under *GeoMesa NiFi Installation* to install the GeoMesa processors in your NiFi instance"

?

Basically, I'm looking for that sentence to have a bit more info/context about what one is installing.

docs/user/nifi/examples.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Follow the steps under :ref:`nifi_install` to install the GeoMesa processors in your NiFi instance. This
+tutorial requires the GeoMesa Kafka and HBase NARs, as well as the standard GeoMesa service NARs.
+
+Add and Configure the Processors
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The first step is to add two processors to your flow, a ``GetGeoMesaKafkaRecord`` processor and an
+``AvroToPutGeoMesaHBase`` processor. Connect the output of the ``Get`` processor to the input of the ``Put``
+processor, and auto-terminate all the other connections:
```
jnh5y, line 23: Hmm....  from what I've seen auto-terminating connection is very bad in production settings.  How about saying something like:

"For testing and development purposes, other connections can be auto-terminated.  In production, appropriate handling for those connections can be considered."

docs/user/nifi/examples.rst
```diff
@@ -0,0 +1,55 @@
...
+being read:
+
+.. image:: /user/_static/img/nifi-avro-record-writer-config.png
+   :align: center
+
+The ``GetGeoMesaKafkaRecord`` processor will pull data off of the GeoMesa Kafka topic and write it out using
+a NiFi record writer. The processor can output data in any format supported by the NiFi records API, but for
+this example we're going to use GeoAvro to simplify the ingest into HBase. Configure the processor by
+setting the appropriate Kafka connection parameters and setting the ``GeoAvroRecordWriterFactory`` you just
+created for the output. Set a unique Kafka group ID, to ensure that the processor reads all the data coming
```
jnh5y, line 41: We require folks to set their own Kafka group ID?  

This means that folks duplicating a GetGeoMesaKafkaRecord processor could get surprised!

docs/user/nifi/examples.rst
```diff
@@ -0,0 +1,55 @@
...
+``AvroToPutGeoMesaHBase`` processor. Connect the output of the ``Get`` processor to the input of the ``Put``
+processor, and auto-terminate all the other connections:
+
+.. image:: /user/_static/img/nifi-kafka-to-hbase-flow.png
+   :align: center
+
+Next, create a new controller service by clicking 'Configure' on the NiFi flow, then going to the Controller
+Services tab and clicking the ``+`` button, then selecting the ``GeoAvroRecordSetWriterFactory``. The
+``GeoAvroRecordSetWriterFactory`` can be left with its default configuration, which is to use NiFi expressions.
+The ``GeoGeoMesaKafkaRecord`` processor will populate the necessary attributes according to the feature type
```
jnh5y, line 31: We have 'GetGeoMesaKafkaRecord' processors (typo).

I keep getting downvoted when I suggest we name something GeoGeoGeo.

docs/user/nifi/examples.rst
```diff
@@ -0,0 +1,55 @@
...
+   :align: center
+
+The ``AvroToPutGeoMesaHBase`` processor will read most the necessary configuration from the GeoAvro file, but
+you still need to configure it with the ``hbase-site.xml`` file for your cluster, which will let it connect to HBase,
+and a catalog table, which is where it will write out data:
+
+.. image:: /user/_static/img/nifi-avro-to-hbase-config.png
+   :align: center
+
+Once all the processors are configured, enable them in the NiFi UI. If everything goes correctly, you'll start
```
jnh5y, line 54: Part of me wants to suggest expanding all the contractions (you'll, we're, etc) for some reason.  Admittedly, that's personal preference.

jnh5y:
General question: Does the Feature type name for the GeoAvroRecordSetWriterFactory need to be set?  (Either by the user or by the NiFi expressions?)

docs/user/nifi/examples.rst
```diff
@@ -0,0 +1,55 @@
...
+The first step is to add two processors to your flow, a ``GetGeoMesaKafkaRecord`` processor and an
+``AvroToPutGeoMesaHBase`` processor. Connect the output of the ``Get`` processor to the input of the ``Put``
+processor, and auto-terminate all the other connections:
+
+.. image:: /user/_static/img/nifi-kafka-to-hbase-flow.png
+   :align: center
+
+Next, create a new controller service by clicking 'Configure' on the NiFi flow, then going to the Controller
+Services tab and clicking the ``+`` button, then selecting the ``GeoAvroRecordSetWriterFactory``. The
+``GeoAvroRecordSetWriterFactory`` can be left with its default configuration, which is to use NiFi expressions.
```
jnh5y, line 30: Should we mention enabling the GeoAvroRecordSetWriterFactory?  Or is that obvious?

docs/user/nifi/examples.rst
elahrvivaz:
```suggestion
view, and to persist every state of the object into long-term storage in HBase or Accumulo for history and
```

docs/user/nifi/install.rst
elahrvivaz: man, that's not even a line I added ;)

if we update it though, it also has the effect of updating the sidebar menu to be kind of redundant and verbose...

you don't think the rest of the sentence provides enough context?

docs/user/nifi/examples.rst
elahrvivaz: well, the group id has to be the same for any given processor, but unique from all other processors. i don't know of a way to do that automatically...

docs/user/nifi/examples.rst
elahrvivaz:
```suggestion
The ``GetGeoMesaKafkaRecord`` processor will populate the necessary attributes according to the feature type
```

elahrvivaz:
> General question: Does the Feature type name for the GeoAvroRecordSetWriterFactory need to be set? (Either by the user or by the NiFi expressions?)

It will default to the name set in the schema, which will come from the avro file, which will come from the kafka topic. so it doesn't need to be set unless you want to change it

docs/user/nifi/examples.rst
elahrvivaz: added it to the line at the bottom about enabling processors

docs/user/nifi/examples.rst
elahrvivaz: this may not be much of an issue either way - generally, you probably wouldn't duplicate the processor, you would just duplicate and split the output flow. if you are consuming a different topic, then it doesn't matter if the id is the same.

docs/user/nifi/install.rst
jnh5y: Yeah, it is fine.  I prefer verbose links / descriptions, but yeah, if it'll mess with the sidebar, we can leave it as well.

jnh5y:
> > General question: Does the Feature type name for the GeoAvroRecordSetWriterFactory need to be set? (Either by the user or by the NiFi expressions?)
> 
> It will default to the name set in the schema, which will come from the avro file, which will come from the kafka topic. so it doesn't need to be set unless you want to change it

Sounds good!

PR #2740, opened 2021-03-26
GEOMESA-3033 Provide metrics reporting for Kafka data store cache

elahrvivaz:
* New geomesa-metrics modules ported from geomesa-convert


docs/user/convert/parsing_and_validation.rst
```diff
@@ -68,7 +68,7 @@ must be registered through a special
     }
 
 
-Valdiators are provided with a `Dropwizard MetricRegistry <https://metrics.dropwizard.io/>`__, which can be used
+Validators are provided with a `Dropwizard MetricRegistry <https://metrics.dropwizard.io/>`__, which can be used
```
jnh5y, line 71: Nice catch.  Makes me wonder how hard it'd be to run a spell checker on the docs directory,

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreParams.scala
```diff
@@ -237,6 +237,16 @@ trait KafkaDataStoreParams extends NamespaceParams {
       readWrite = ReadWriteFlag.ReadOnly
     )
 
+  val MetricsReporters =
+    new GeoMesaParam[String](
+      "kafka.metrics.reporters",
+      "Reporters used to publish Kafka metrics, as TypeSafe config. . To use multiple reporters, " +
+          "nest them under the key 'reporters'",
+      default = """{"type":"slf4j","logger":"org.locationtech.geomesa.kafka.metrics"}""",
```
jnh5y, line 245: Ok, so if I do nothing when publishing a store, I'll get metrics, right?  Do I need to set a log level to pick them up?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCacheWithMetrics.scala
```diff
@@ -0,0 +1,40 @@
...
+class KafkaFeatureCacheWithMetrics(sft: SimpleFeatureType, config: IndexConfig, metrics: GeoMesaMetrics)
+    extends KafkaFeatureCacheImpl(sft, config){
+
+  metrics.register(sft.getTypeName, "index-size", new Gauge[Int] { override def getValue: Int = size() })
+
+  private val updates     = metrics.meter(sft.getTypeName, "updates")
+  private val removals    = metrics.meter(sft.getTypeName, "removals")
+  private val expirations = metrics.meter(sft.getTypeName, "expirations")
+
+  override def put(feature: SimpleFeature): Unit = {
```
jnh5y, line 26: Is the use of 'clears' in the KDS going to throw things off?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreParams.scala
elahrvivaz: yeah, they'll get logged at debug level by default

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCacheWithMetrics.scala
elahrvivaz: this is just counting the rate of incoming features, so clears shouldn't affect it

jnh5y: Ok, so we do not have a metric that reports the number of records in the KDS cache, right?

elahrvivaz: we do, it's the 'index-size' one a few lines up

elahrvivaz: although this just reports the main hashmap size, the indices could potentially go out of sync. but i haven't seen that happen since the switch to using futures for expiring features.

elahrvivaz:
actually i just realized this might throw an error, let me update it real quick

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCacheWithMetrics.scala
jnh5y: Got it.   As a sanity check, 'size()' calls `state.size()` (State is a ConcurrentHashMap.)

Here: https://github.com/locationtech/geomesa/blob/main/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCacheImpl.scala#L32 we have a comment about not iterating over `state`.

Is this the best way to compute the cache's size?

elahrvivaz: size doesn't require iterating over the map. it's not just a lookup, but it should be essentially constant time: https://stackoverflow.com/a/22996395/7809538
it seems like they improved the algorithm in java 8 so that it doesn't ever need to lock, looking at the source for it.

elahrvivaz: seems like there is some book-keeping done with each put/remove that simplifies the call to size

elahrvivaz:
> actually i just realized this might throw an error, let me update it real quick

should be gtg now

PR #2741, opened 2021-03-29
GEOMESA-3034 Redis - Fix explain logging exceptions

elahrvivaz:

### DIFF
--- a/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/QueryPlan.scala
+++ b/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/QueryPlan.scala
@@ -20,6 +20,8 @@ import org.locationtech.geomesa.utils.collection.CloseableIterator
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
+import scala.util.Try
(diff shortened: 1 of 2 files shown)


elahrvivaz:
this is not an ideal fix, but it's a tricky issue that requires some more thought. I opened https://geomesa.atlassian.net/browse/GEOMESA-3035 as a follow-on

jnh5y:
Cool.  I took a spin through the code.  Since visibility is not used in the state map, why does it chuck the NotImplementedError?

For the Arrow bits, it looks like dictionaries or something non-serializable are involved.  I wonder if a filtered state map could solve this....

But yeah, this is where the follow-on ticket makes sense.

elahrvivaz:
so visibility is not used in the state map b/c it's not serializable atm. the check is to make sure that if there is a vis filter, then it will fail serialization. normally, the state map is only used for m/r job configurations, when we need to pass a query plan reduce step through to the hadoop reducers. that's only implemented for accumulo and hbase, which don't ever have a vis step there b/c the check is handled natively. however, we hijacked the serialization as a nice way to show the reduce step in the explain logging, and that's where the error is coming in.

elahrvivaz:
and if you had a vis filter, and needed to serialize the state, you would definitely not want it to be silently filtered out. so probably we just need a better toString representation of the state...

PR #2742, opened 2021-03-30
GEOMESA-3037 Support Kafka 2.7 classpath

elahrvivaz:
* Always use Zookeeper 3.4 compatibility mode
* Switch to newer Admin code for creating topics
* Requires runtime Kafka jars version 2.0+
* Should still support Kafka brokers back to version 0.10


elahrvivaz:
I've tested so far with zookeeper 3.4.12 and kafka 2.7.0. Planning to test with zookeeper 3.4.14, 3.5.x, kafka 2.4, and possibly other kafka versions.

elahrvivaz:
tested with zk 3.4.14 and 3.5.9, and kafka 2.7.0, 2.4.1 and 2.1.1

PR #2744, opened 2021-03-31
GEOMESA-3044 Support Kakfa transactional writes

elahrvivaz:

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureWriter.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureWriter.scala
@@ -9,10 +9,11 @@
 package org.locationtech.geomesa.kafka.data
 
 import java.io.Flushable
-import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.{AtomicLong, AtomicReference}
 
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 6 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureWriter.scala
```diff
@@ -130,4 +135,65 @@ object KafkaFeatureWriter {
...
+    override def flush(): Unit = producer.flush()
+  }
+
+  case class KafkaTransactionState(producer: Producer[Array[Byte], Array[Byte]])
+      extends Transaction.State with KafkaFeatureProducer {
+
+    private val tx = new AtomicReference[Transaction](null)
+    private var inTransaction = false
+
+    override def send(record: ProducerRecord[Array[Byte], Array[Byte]]): Unit = {
```
jnh5y, line 154: Are these states handled by single threaded access?

elahrvivaz: in the general use case, yes. it's definitely possible to use them in a multi-threaded manner, i.e. by creating a transaction and then passing it to different threads. but that seems like something that probably wouldn't be supported in geotools anyway? i thought about making this thread safe but decided not to.

elahrvivaz:
> How does this impact things in NiFi or other existing clients?

we'd have to add something to nifi to enable this.

PR #2745, opened 2021-04-01
GEOMESA-3045 Include wps jar in gs-plugins

elahrvivaz:

### DIFF
--- a/geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
+++ b/geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
@@ -174,6 +174,23 @@
         </dependencies>
     </dependencyManagement>
 
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-process-wps_${scala.binary.version}</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.geotools</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.locationtech.jts</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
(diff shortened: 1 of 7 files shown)


elahrvivaz:
I verified that the dist and gs plugins still have the same jars, except for the one wps jar being included now.

geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
```diff
@@ -174,6 +174,23 @@
         </dependencies>
     </dependencyManagement>
 
+    <dependencies>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-process-wps_${scala.binary.version}</artifactId>
```
jnh5y, line 180: Where does this inclusion add the jar?  Is it part of the gs-plugin tar.gz?

jnh5y:
Should this PR impact lines 18 or 29 here: https://github.com/locationtech/geomesa/blob/main/geomesa-archetypes/geomesa-archetypes-binary-dist/assembly.xml#L18-L29 ?

geomesa-archetypes/geomesa-archetypes-gs-plugin/pom.xml
elahrvivaz: right, it makes it compile scope and then it gets picked up by the dependencies section of the maven assembly plugin

elahrvivaz:
> Should this PR impact lines 18 or 29 here: https://github.com/locationtech/geomesa/blob/main/geomesa-archetypes/geomesa-archetypes-binary-dist/assembly.xml#L18-L29 ?

no, i left it in the gs-plugins folder in the binary dist for backwards compatibility (we can remove it later), and we don't need/want it in the tools themselves since it only goes in geoserver. and the gs-plugin tar.gz gets pulled in separately so it's not affected by the includes/excludes

PR #2747, opened 2021-04-07
GEOMESA-3048 Add readiness check for Kafka DataStore Initial Loads

jnh5y:

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
@@ -11,7 +11,7 @@ package org.locationtech.geomesa.kafka.data
 import java.io.Closeable
 import java.time.Duration
 import java.util.Collections
-import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicLong}
 import java.util.concurrent.{ConcurrentHashMap, CountDownLatch, Executors}
 
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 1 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
```diff
@@ -89,7 +89,24 @@ trait KafkaCacheLoader extends Closeable with LazyLogging {
...
+    private val count = new AtomicInteger(0)
+    private val firstLoadStarTime = new AtomicLong(0L)
+
+    def startLoad(): Boolean = {
+      count.incrementAndGet()
+      firstLoadStarTime.compareAndSet(0L, System.currentTimeMillis())
+    }
+    def completedLoad(): Unit = {
+      if (count.decrementAndGet() == 0) {
+        logger.info(s"Last active initial load completed.  " +
```
elahrvivaz, line 103: could we reset firstLoadStartTime here, so that if a new data store is loaded it will have an accurate time?

jnh5y: Good call.  I like it.

elahrvivaz: if so, there's some potential concurrency issues since we're changing the count and time separately - might be easier just to make both methods synchronized

jnh5y: Do we need to synchronize the methods or the entire object?

elahrvivaz: i don't think you can synchronize a whole object - you can just put `def foo(): Unit = sycnhronized {}` which will use a lock on the object so synchronize across all methods that are declared that way

elahrvivaz: there is probably a more elegant way but just brute synchronization should be good enough here...

elahrvivaz (approved):

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaCacheLoader.scala
jnh5y: Got it.  I thought that synchronized only locked that method.  That makes sense.

PR #2750, opened 2021-04-14
GEOMESA-3051 Don't split FSDS --storage-opt on commas

elahrvivaz:
* Makes it easier to set JSON in the user data

### DIFF
--- a/docs/user/filesystem/commandline.rst
+++ b/docs/user/filesystem/commandline.rst
@@ -120,7 +120,7 @@ Argument                 Description
 ``--leaf-storage``       Use leaf storage
 ``--target-file-size``   Target size for data files (e.g. 500MB or 1GB)
 ``--temp-path``          Path to a temp directory used for working files
-``--storage-opt``        Additional storage options, as ``key=value``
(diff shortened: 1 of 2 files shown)


docs/user/filesystem/commandline.rst
```diff
@@ -120,7 +120,7 @@ Argument                 Description
 ``--leaf-storage``       Use leaf storage
 ``--target-file-size``   Target size for data files (e.g. 500MB or 1GB)
 ``--temp-path``          Path to a temp directory used for working files
-``--storage-opt``        Additional storage options, as ``key=value``
+``--storage-opt``        Additional storage options to set as SimpleFeatureType user data, in the form ``key=value``
```
jnh5y, line 123: Is there a quick example of a storage-opt which has a comma in it?

elahrvivaz: some of the storage opts are json, so they have commas. did you mean for including in the docs?

jnh5y: Eh, knowing that it is json is helpful.  If adding to the docs is handy, feel free to do it later.  Merging!

PR #2751, opened 2021-04-15
GEOMESA-2696 Converters - pass nulls through type inference

elahrvivaz:


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/GeometryFunctionFactory.scala
```diff
@@ -34,55 +34,63 @@ class GeometryFunctionFactory extends TransformerFunctionFactory {
     case Array(g: Array[Byte]) => WKBUtils.read(g).asInstanceOf[Point]
     case Array(x: Float, y: Double) => gf.createPoint(new Coordinate(x, y))
     case Array(x: Double, y: Float) => gf.createPoint(new Coordinate(x, y))
+    case Array(null) | Array(null, null) => null
```
jnh5y, line 37: Idle note:  Looks like this doesn't support XYZ, XYM, XYZM, right?

elahrvivaz: right, we should probably add that

jnh5y: Are you knocking that out in this PR or another ticket?

elahrvivaz: i'll open a ticket

PR #2753, opened 2021-04-16
GEOMESA-3054 Optimize FSDS JDBC metadata SQL queries

elahrvivaz:


jnh5y:
What's the crux of the optimization?

elahrvivaz:
> What's the crux of the optimization?

it was previously doing a manual join, e.g. `execute("select id from table1").foreach(id => "select * from table2 where id = $id")`. this changes it to be a sql join instead, e.g. `execute("select * from table1, table2 where table1.id = table2.id")`.

PR #2754, opened 2021-04-20
GEOMESA-3058 Converters - don't cache global variables

elahrvivaz:


geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/Expression.scala
```diff
@@ -170,14 +170,17 @@ object Expression {
...
-    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = doEval(ctx)
+    private var ref: EvaluationContext = _
+    private var i = -1
+
+    override def eval(args: Array[Any])(implicit ctx: EvaluationContext): Any = {
+      if (!ctx.eq(ref)) {
+        ref = ctx
+        i = ctx.indexOf(n)
+      }
+      if (i == -1) { null } else { ctx.get(i) }
```
jnh5y, line 182: Is caching the evaluationContext actually saving time here?  (I'm kinda fine with it either way.)

elahrvivaz: not sure if it's saving time, but we need to keep a reference to it so we can know if the index of a named value has changed (which it can with a new evaluation context)

PR #2755, opened 2021-04-27
GEOMESA-3060 Fix FSDS Spark runtime jar classpath for EMR

elahrvivaz:
* Include guava
* Relocate jackson and paranamer


jnh5y:
Did you test with a local Apache Spark 3?

elahrvivaz:
> Did you test with a local Apache Spark 3?

I did not, but these are just relocations so shouldn't affect anything. guava was already being relocated (although not included in the jar, hence it was causing issues).

elahrvivaz:
I tested this out locally and was able to query hbase 1 and accumulo 1, and to load a csv with `spark.read.format("csv").option("header", "true").load("~/geomesa/geomesa-tools/examples/ingest/csv/example.csv")`. However, I tried the same csv load with the older jars and that worked too...

PR #2757, opened 2021-04-30
GEOMESA-3055 Add limits on deserialization recursion

elahrvivaz:


jnh5y:
Hmm... would this kind of issue impact JTS as well?

If so, I wonder if it'd be worth opening an issue on the JTS GitHub?

elahrvivaz:
> Hmm... would this kind of issue impact JTS as well?
> 
> If so, I wonder if it'd be worth opening an issue on the JTS GitHub?

hmm, not sure. @dr-jts do you think this type of thing could affect JTS? for reference, we have hit some bugs where corrupted data caused our geometry deserialization to stack overflow due to it thinking that there were thousands of nested geometry collections...

dr-jts:
> do you think this type of thing could affect JTS? for reference, we have hit some bugs where corrupted data caused our geometry deserialization to stack overflow due to it thinking that there were thousands of nested geometry collections...

Yes, I think this could be an issue for JTS.  There was a recent [fix](https://github.com/locationtech/jts/pull/675) to allow limiting the number of coordinates/rings/elements read, but it doesn't limit the nesting of `GeometryCollection`s.

PR #2758, opened 2021-05-05
GEOMESA-3065 HBase - fix infinite loop caused by HBase meta errors

elahrvivaz:


elahrvivaz:
we should probably back-port this

elahrvivaz:
I tested this against the HBase instance that was exhibiting the issue, but it would be hard to unit test without exposing some private methods and using mock objects.

elahrvivaz:
This fixed the infinite loop, but the export still eventually hung due to the inconsistencies in HBase causing scans to not work. Running `hbck` fixed the issue. So this is probably a good thing to merge, but back-porting it probably isn't necessary.

jnh5y:
Hmm... when the export failed later, was it clear that there was an HBase issue?

elahrvivaz:
> Hmm... when the export failed later, was it clear that there was an HBase issue?

I believe so, it was errors in the hbase scanner.next iirc

PR #2762, opened 2021-05-10
GEOMESA-3061 Converters - support bytes in Avro top-level union types

elahrvivaz:


geomesa-convert/geomesa-convert-avro/src/main/scala/org/locationtech/geomesa/convert/avro/AvroConverter.scala
```diff
@@ -75,15 +74,25 @@ object AvroConverter {
...
+          fields.add(new Schema.Field(field.name, field.schema, field.doc, field.defaultVal()))
+        }
+        fields.add(new Schema.Field(BytesField, Schema.create(Schema.Type.BYTES), "raw bytes", ""))
+
+        val updated = Schema.createRecord(schema.getName, schema.getDoc, schema.getNamespace, schema.isError)
+        updated.setFields(fields)
+        updated
+
+      case Schema.Type.UNION =>
+        Schema.createUnion(schema.getTypes.asScala.map(addBytes): _*)
```
jnh5y, line 90: Wait!  I just saw this movie!  What happens if there are nested Unions!?!??!

elahrvivaz: hopefully there aren't infinitely nested unions :P

jnh5y: Well, it'd be due to bad bytes on disk...  Eh, we can just remember this for when the time comes...;)

PR #2766, opened 2021-05-11
GEOMESA-3071 Move all converter state into evaluation context

elahrvivaz:
* Provided converters are thread-safe as a result of this change
* Custom Fields and TransformerFunctions should implement new methods,
  but will work without changes


jnh5y:
@elahrvivaz ready to rebase.

elahrvivaz:
@jnh5y if you do a rebase and merge that will keep the commit from 3.2 for the original fix for the evaluation context index

PR #2772, opened 2021-05-26
GEOMESA-3075 Support for z/m coordinates in converters

elahrvivaz:


jnh5y (approved):
Should we enter tickets for Z/M support for the other types?

elahrvivaz:
the other types only support WKT/WKB, so they already work with Z/M I believe.

PR #2774, opened 2021-06-02
GEOMESA-3084 Use common authorization lookup for Accumulo Spark

elahrvivaz:


geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaAccumuloInputFormat.scala
```diff
@@ -134,6 +133,22 @@ object GeoMesaAccumuloInputFormat extends LazyLogging {
...
+  }
+
+  /**
+   * Configure the input format based on a query plan
+   *
+   * @param conf configuration to update
+   * @param params data store parameters
+   * @param plan query plan
+   */
+  def configure(
```
jnh5y, line 147: What's the goal of having two different `configure` methods?  Is it so that one can query with more restricted Auths than they have available?

elahrvivaz: no, it was just to keep the old one there for api compatibility.

elahrvivaz: could maybe deprecate the old one, but i forget if it's actually still used anywhere.

PR #2775, opened 2021-06-15
GEOMESA-3091 Attribute level visibilities error with null attribute values

jrs53:
Correctly populate null bitset when merging serialised features according to visibilities

Tested via new unit test, as well as manually on an Accumulo 1.7 cluster

Signed-off-by: James Srinivasan <jsrinivasan@dstl.gov.uk>

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
@@ -90,11 +90,12 @@ class KryoVisibilityRowEncoder extends RowEncodingIterator {
       // Column qualifiers tell us which attributes are intended to be populated given visibilities
       keys.get(i).getColumnQualifier.getBytes.foreach { unsigned =>
         val index = java.lang.Byte.toUnsignedInt(unsigned)
-        val pos = metadata.setPosition(index)
-        val len = metadata.setPosition(index + 1) - pos
-        attributes(index) = (bytes, 4 + pos, len) // pos is relative to first byte of attribute offsets, see length calc
-        assert((metadata.nulls.contains(index) && len==0) || (!metadata.nulls.contains(index) && len!=0), s"Metadata null bitset (${metadata.nulls.toString}) and value length ($len) are inconsistent")
-        length += len
+        if(!metadata.nulls.contains(index)) {
+          val pos = metadata.setPosition(index)
+          val len = metadata.setPosition(index + 1) - pos
+          attributes(index) = (bytes, 4 + pos, len) // pos is relative to first byte of attribute offsets, see length calc
+          length += len
+        }
       }
       i += 1
     }
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
@@ -1,5 +1,6 @@
 /***********************************************************************
  * Copyright (c) 2013-2021 Commonwealth Computer Research, Inc.
+ * Portions Crown Copyright (c) 2016-2021 Dstl
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Apache License, Version 2.0
  * which accompanies this distribution and is available at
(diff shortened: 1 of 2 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -68,22 +69,31 @@ class KryoVisibilityRowEncoder extends RowEncodingIterator {
     * @return
     */
   private def encodeV3(keys: java.util.List[Key], values: java.util.List[Value]): Value = {
-    // 1 byte for version + 2 bytes for count + 1 byte for size + 4 bytes per attribute for offsets
-    // + 4 bytes for user data offset + null bit set bytes
-    var length = (count + 2 + IntBitSet.size(count)) * 4
 
-    var valueCursor = length // tracks our output position for copying bytes
+    // Calculate length of serialised attributes, excluding attribute values themselves
+    var length = 1 +            // version
```
elahrvivaz, line 74: does this length end up the same, or was it not calculated correctly before?

elahrvivaz (approved):
thanks for adding such detailed comments!

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
jrs53: It ends up precisely the same as before, but I thought this was a clearer way to explain it

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -68,22 +69,31 @@ class KryoVisibilityRowEncoder extends RowEncodingIterator {
...
       val input = KryoFeatureDeserialization.getInput(bytes, 1, bytes.length - 1) // skip the version byte
       val metadata = Metadata(input) // read count, size, etc
+      // Column qualifiers tell us which attributes are intended to be populated given visibilities
       keys.get(i).getColumnQualifier.getBytes.foreach { unsigned =>
         val index = java.lang.Byte.toUnsignedInt(unsigned)
         val pos = metadata.setPosition(index)
         val len = metadata.setPosition(index + 1) - pos
-        attributes(index) = (bytes, 4 + pos, len) // the offset is always 4 since we're copying into a new byte array
+        attributes(index) = (bytes, 4 + pos, len) // pos is relative to first byte of attribute offsets, see length calc
+        assert((metadata.nulls.contains(index) && len==0) || (!metadata.nulls.contains(index) && len!=0), s"Metadata null bitset (${metadata.nulls.toString}) and value length ($len) are inconsistent")
```
jnh5y, line 96: When would this assert fail?  Is this going to stop writers?

Said differently, is anyone going to hit this?

elahrvivaz: asserts usually get compiled out. if you run a unit test through intellij, it will add the `-ea` flag to the compile option to enable assertions. i'm not sure if maven does something similar but probably not since it would have to re-build to remove them.

elahrvivaz: ah, i guess it's a runtime check to enable assertions, not compile time.

jrs53: I'm not sure, that's why the assert is there. It wasn't clear to me from the Kryo encoding docs that a null value was indicated by both the corresponding null bit being set and the length of the attribute being zero (seems redundant?).

I chose assert (Vs require) so they could be disabled at runtime (this might be the default).

I don't mind losing the line tho.

elahrvivaz: hmm, actually i think the length could be zero without it being null, i.e. for an empty string. so yeah, it's probably better to remove the check, or at least the second clause. null implies zero length, but zero length does not imply null.

elahrvivaz: so, looking again, it does seem like any non-null value will write at least one byte, so this will work. however, i realized that we weren't checking for nulls in the initial loop, which would also solve the problem and would probably be more "correct" in that it doesn't rely on an implementation detail of how the bytes are serialized. I.e. `val nulls = IntBitSet(count)` should be moved up above this code block, and then this foreach should be:

```
        val index = java.lang.Byte.toUnsignedInt(unsigned)
        if (metadata.nulls.contains(index)) {
          nulls.add(index)
        } else {
          val pos = metadata.setPosition(index)
          val len = metadata.setPosition(index + 1) - pos
          attributes(index) = (bytes, 4 + pos, len) // the offset is always 4 since we're copying into a new byte array
          length += len
        }
```

then the check down below for zero length doesn't need to be done.

elahrvivaz: actually we don't even need to set the nulls in this loop, since it's set below. just need the `if (!metadata.nulls.contains(index))` guard around setting `attributes(index) = ...`.

jrs53: Yeah, that sounds a much better solution - will make change. I forget why I wasn't looking at the original null bitsets (there was a reason)

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -90,11 +90,12 @@ class KryoVisibilityRowEncoder extends RowEncodingIterator {
...
-        val pos = metadata.setPosition(index)
-        val len = metadata.setPosition(index + 1) - pos
-        attributes(index) = (bytes, 4 + pos, len) // pos is relative to first byte of attribute offsets, see length calc
-        assert((metadata.nulls.contains(index) && len==0) || (!metadata.nulls.contains(index) && len!=0), s"Metadata null bitset (${metadata.nulls.toString}) and value length ($len) are inconsistent")
-        length += len
+        if(!metadata.nulls.contains(index)) {
+          val pos = metadata.setPosition(index)
+          val len = metadata.setPosition(index + 1) - pos
+          attributes(index) = (bytes, 4 + pos, len) // pos is relative to first byte of attribute offsets, see length calc
+          length += len
```
jrs53, line 97: Assumption len=0 if null, so only need to increment in non null case

elahrvivaz:
@jrs53 do you need this back-ported to any earlier versions?

jrs53:
I think I'll use it as an excuse to move to 3.x (assuming our client can support multiple geoservers since we'll likely still have some things on 2.4 for a while)

PR #2776, opened 2021-06-17
GEOMESA-2111 Bulk ingest for Accumulo

elahrvivaz:
* New 'bulk-ingest' tools command for writing and loading RFiles directly

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/mapreduce/package.scala
+++ b/geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/mapreduce/package.scala
@@ -1,0 +1,136 @@
+/***********************************************************************
+ * Copyright (c) 2013-2021 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.jobs.accumulo
+
+import java.io.{BufferedOutputStream, DataInput, DataOutput, PrintStream}
+import java.nio.charset.StandardCharsets
+import java.util.{Base64, Scanner}
+
+import com.github.benmanes.caffeine.cache.{CacheLoader, Caffeine}
+import org.apache.accumulo.core.data.Key
+import org.apache.hadoop.conf.{Configurable, Configuration}
+import org.apache.hadoop.fs.Options.CreateOpts
+import org.apache.hadoop.fs.{CreateFlag, FileContext, FileSystem, Path}
+import org.apache.hadoop.io.{BinaryComparable, Text, Writable, WritableComparable}
+import org.apache.hadoop.mapreduce.{Job, Partitioner}
+import org.locationtech.geomesa.utils.io.WithClose
+
+import scala.collection.mutable.ArrayBuffer
+
+package object mapreduce {
+
+  object Configurator {
+
+    private val TypeNameKey   = "org.locationtech.geomesa.accumulo.typename"
+    private val PartitionsKey = "org.locationtech.geomesa.accumulo.partitions"
+
+    def setTypeName(conf: Configuration, typeName: String): Unit = conf.set(TypeNameKey, typeName)
+    def getTypeName(conf: Configuration): String = conf.get(TypeNameKey)
+    def setPartitions(conf: Configuration, partitions: Seq[String]): Unit =
+      conf.set(PartitionsKey, partitions.mkString(","))
+    def getPartitions(conf: Configuration): Option[Seq[String]] = Option(conf.get(PartitionsKey)).map(_.split(","))
+  }
+
+  class TableAndKey extends WritableComparable[TableAndKey] {
+
+    private var table: Text = _
+    private var key: Key = _
+
+    def this(table: Text, key: Key) = {
+      this()
+      this.table = table
+      this.key = key
+    }
+
+    def getTable: Text = table
+    def setTable(table: Text): Unit = this.table = table
+    def getKey: Key = key
+    def setKey(key: Key): Unit = this.key = key
+
+    override def write(out: DataOutput): Unit = {
+      table.write(out)
+      key.write(out)
+    }
+
+    override def readFields(in: DataInput): Unit = {
+      table = new Text()
+      table.readFields(in)
+      key = new Key()
+      key.readFields(in)
+    }
+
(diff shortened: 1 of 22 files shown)


geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/mapreduce/package.scala
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    override def readFields(in: DataInput): Unit = {
+      table = new Text()
+      table.readFields(in)
+      key = new Key()
+      key.readFields(in)
+    }
+
+    override def compareTo(o: TableAndKey): Int = {
```
jnh5y, line 68: Do we need to do anything with equals?  Or is that just a nice-to-have?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/ingest/AccumuloBulkIngestCommand.scala
```diff
@@ -0,0 +1,188 @@
...
+          }
+          dir
+        }
+
+        Command.user.info(s"Running bulk ingestion in distributed ${if (params.combineInputs) "combine " else "" }mode")
+        new BulkConverterIngest(ds, connection, sft, converter, inputs.paths, output, tempPath, maxSplitSize,
+          index, partitions, libjarsFiles, libjarsPaths)
+
+      case _ =>
+        throw new NotImplementedError(s"Missing implementation for mode $mode")
```
jnh5y, line 106: Do we want to call out DistributedCombine specifically?

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/ingest/AccumuloIngestCommand.scala
```diff
@@ -18,6 +18,7 @@ import org.locationtech.geomesa.tools.ingest.IngestCommand
 import org.locationtech.geomesa.tools.ingest.IngestCommand.IngestParams
 
 class AccumuloIngestCommand extends IngestCommand[AccumuloDataStore] with AccumuloDistributedCommand {
+
```
jnh5y, line 21: Meh.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/StorageJobUtils.scala
```diff
@@ -41,11 +44,12 @@ object StorageJobUtils extends LazyLogging {
 
   // hadoop 3 API
   private def distCpOptions3(src: Path, dest: Path): DistCpOptions = {
-    val clas = Class.forName("org.apache.hadoop.tools.DistCpOptions.Builder")
+    val clas = Class.forName("org.apache.hadoop.tools.DistCpOptions$Builder")
```
jnh5y, line 47: Ok, should we document that the bulk load job requires Hadoop 3?  Or is that obvious?  (Maybe Accumulo 2 requires Hadoop 3?)

jnh5y: Also, relative to this, is the distcp used on the HBase bulk import job?  E.g. does the Hadoop 2/3 difference matter for that?

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/mapreduce/package.scala
elahrvivaz: ah, i think it doesn't matter. it needs to be comparable so that the reducer can sort the output. but maybe it would be good to also implement equals and hash code...

geomesa-accumulo/geomesa-accumulo-tools/src/main/scala/org/locationtech/geomesa/accumulo/tools/ingest/AccumuloBulkIngestCommand.scala
elahrvivaz: distributed combine got removed as a standalone option, it's just distributed and local now, and you can specify `--combine-inputs` to get the effect of distributed combine.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/StorageJobUtils.scala
elahrvivaz: this should work with both hadoop 2 and 3 (although i didn't test it), there's a different method that gets called for hadoop 2. i think this hadoop 3 code has been broken all along and we just never noticed.

this is not used for the hbase bulk load.

elahrvivaz: i did document that it's only implemented for accumulo 2, although i'm sure we could do accumulo 1 if we wanted but there didn't seem to be a pressing reason.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/mapreduce/package.scala
jnh5y: Your call.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/utils/StorageJobUtils.scala
jnh5y: Makes sense.

geomesa-accumulo/geomesa-accumulo-jobs/src/main/scala/org/locationtech/geomesa/jobs/accumulo/mapreduce/package.scala
elahrvivaz: i probably should have gone back and done this but didn't get around to it... hopefully it won't bite us in some subtle way :P

jnh5y: Yeah, I don't think this'll be our shared mistake this year.....  If I did, I wouldn't have hit the merge button.  The docs on compareTo make it sound optional to think about equals/hashCode/whatever.

PR #2777, opened 2021-06-17
GEOMESA-3092 Support Lambda NiFi processor

elahrvivaz:
* Separate Lambda data store config parsing from Accumulo
* Move PartitionSchemeArgResolver to data store module
* Handle check for FeatureIdImpl to avoid proxy errors in NiFi

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PartitionSchemeArgResolver.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PartitionSchemeArgResolver.scala
@@ -6,14 +6,14 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.fs.tools.utils
+package org.locationtech.geomesa.fs.storage.common.utils
 
 import java.nio.charset.StandardCharsets
 
 import org.apache.commons.io.IOUtils
 import org.locationtech.geomesa.fs.storage.api.{NamedOptions, PartitionSchemeFactory}
 import org.locationtech.geomesa.fs.storage.common.StorageSerialization
-import org.locationtech.geomesa.fs.tools.utils.PartitionSchemeArgResolver.SchemeArgs
+import org.locationtech.geomesa.fs.storage.common.utils.PartitionSchemeArgResolver.SchemeArgs
 import org.locationtech.geomesa.utils.conf.ArgResolver
 import org.locationtech.geomesa.utils.io.{PathUtils, WithClose}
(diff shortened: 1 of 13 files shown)


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PartitionSchemeArgResolver.scala
```diff
@@ -6,14 +6,14 @@
  * http://www.opensource.org/licenses/apache2.0.php.
  ***********************************************************************/
 
-package org.locationtech.geomesa.fs.tools.utils
+package org.locationtech.geomesa.fs.storage.common.utils
```
jnh5y, line 9: Ok, out of curiosity, why did this file move?

geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/data/LambdaDataStore.scala
```diff
@@ -35,29 +34,22 @@ import org.opengis.filter.Filter
...
+    extends DataStore with HasGeoMesaStats with LazyLogging {
 
   import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 
   import scala.collection.JavaConverters._
 
   private val authProvider: Option[AuthorizationsProvider] = persistence match {
-    case ds: AccumuloDataStore => Some(ds.config.authProvider)
+    // this is a bit of a hack to work around hbase null vs empty auths
+    case ds: GeoMesaDataStore[_] if ds.config.authProvider.getAuthorizations != null => Some(ds.config.authProvider)
```
jnh5y, line 46: Ok, so ds.config.authProvider is never null here?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/utils/PartitionSchemeArgResolver.scala
elahrvivaz: mainly so i didn't need to pull in another dependency in nifi, but it seemed better to have it more generally available instead of tied to the tools module.

geomesa-lambda/geomesa-lambda-datastore/src/main/scala/org/locationtech/geomesa/lambda/data/LambdaDataStore.scala
elahrvivaz: this is scala! things are never null! :D
the other option would be to have some kind of 'empty' auth provider and match on the specific class.

PR #2778, opened 2021-06-22
GEOMESA-3093 Fix nailgun script exit after startup

elahrvivaz:
* Fix nailgun start script remaining open in the background, which
  prevents shell from exiting when run from a script


jnh5y (approved):
This is a little more shell gymnastics than I've seen before.  Cool.

Is there a good resource for thinking through / understanding techniques like this?

elahrvivaz:
> This is a little more shell gymnastics than I've seen before. Cool.
> 
> Is there a good resource for thinking through / understanding techniques like this?

Mmm, I just found it on stack overflow haha... i haven't found a good reference that covers everything you might want to do. The subtleties of bash are enormous...

PR #2779, opened 2021-06-23
GEOMESA-3094 Fix attribute plus date index equals queries

elahrvivaz:


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/attribute/DateIndexKeySpace.scala
```diff
@@ -78,8 +78,17 @@ class DateIndexKeySpace(val sft: SimpleFeatureType, dtgField: String)
...
   override def getRangeBytes(ranges: Iterator[ScanRange[Long]], tier: Boolean): Iterator[ByteRange] = {
     ranges.map {
-      case SingleRowRange(row)  => SingleRowByteRange(ByteArrays.toOrderedBytes(row))
-      case BoundedRange(lo, hi) => BoundedByteRange(ByteArrays.toOrderedBytes(lo), ByteArrays.toOrderedBytes(hi))
+      case BoundedRange(lo, hi) =>
+        BoundedByteRange(ByteArrays.toOrderedBytes(lo), ByteArrays.toOrderedBytes(hi))
+
+      case SingleRowRange(row) =>
+        val ordered = ByteArrays.toOrderedBytes(row)
+        if (tier) {
```
jnh5y, line 86: Oh whew, what is 'tier'ing here again?

elahrvivaz: "tiering" indicates that the values will be used as prefixes in a tiered index structure, i.e. the attr index will get passed `tier = true` in an attr+z3 or attr+dtg index. It wouldn't ever be true here with our current indices since this is only the second half of the attr+dtg index, but seemed worthwhile to include the logic.
the ranges end up being slightly different if you're adding additional ranges to the end of it. I.e. in an attribute index the filter `name = bob` would roughly translate to the range `bob->boc`, since the end value is exclusive. However, if you are appending a time filter to the end of it, you'd want to return `bob->bob`, and then append the time range on the end so that you end up with `bob123->bob125` instead of `bob123->boc125`, which would be incorrect. if that's clear as mud, lmk ;)

jnh5y: That explanation makes sense; I just needed the reminder of how we are calling it.:)

PR #2781, opened 2021-07-07
GEOMESA-3098 MurmurHash128 and Base64Decode converter functions

elahrvivaz:


docs/user/convert/function_usage.rst
```diff
@@ -604,15 +613,6 @@ Example: ``uuidZ3Centroid(linestring('LINESTRING(102 0, 103 1, 104 0, 105 1)', d
 
 See :ref:`customizing_z_index` for details on Z3 intervals.
 
-base64
```
jnh5y, line 607: Whew...  should we mention that 'base64' is deprecated somehow and indicate that the new function name is 'base64Encode'?

geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/transforms/IdFunctionFactory.scala
```diff
@@ -70,10 +72,22 @@ class IdFunctionFactory extends TransformerFunctionFactory with LazyLogging {
       hasher.hashString(args(0).toString, StandardCharsets.UTF_8)
   }
 
-  private val murmur3_128: TransformerFunction =
-    new NamedTransformerFunction(Seq("murmur3_128", "murmur3_64"), pure = true) {
-      private val hasher = Hashing.murmur3_128()
-      override def apply(args: Array[AnyRef]): AnyRef =
-        Long.box(hasher.hashString(args(0).toString, StandardCharsets.UTF_8).asLong())
+  private val murmur3_64 =
+    TransformerFunction.pure("murmur3_128", "murmur3_64") { args =>
```
jnh5y, line 76: Should we add some local comments about the historical reasons for the name being what it is?

docs/user/convert/function_usage.rst
```diff
@@ -541,7 +541,7 @@ ID Functions
 stringToBytes
 ^^^^^^^^^^^^^
 
-Description: Converts a string to a UTF-8 byte array (to pass to ``md5()`` or ``base64()``).
+Description: Converts a string to a UTF-8 byte array (to pass to other functions like ``md5()``).
```
jnh5y, line 544: Hmm.... do we want to continue mentioning base64Encode?  (I'm fine either way, just double-checking.)

elahrvivaz: i mainly removed this b/c i'm not sure why you would take a string, get its bytes, then base64 encode them? but if you think that's a normal use case i'll add it back in.

jnh5y: Nevermind; you are right; don't mind the reviewer asking to keep everything....

PR #2783, opened 2021-07-08
GEOMESA-3095 Generate random feature IDs when id-field is not specified in converter spec

Kontinuation:


geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaOutputFormat.scala
```diff
@@ -98,7 +98,7 @@ object GeoMesaOutputFormat {
       try {
         val sftName = value.getFeatureType.getTypeName
         val writer = writers.getOrElseUpdate(sftName, createWriter(sftName))
-        FeatureUtils.write(writer, value, useProvidedFid = true)
+        FeatureUtils.write(writer, value)
```
elahrvivaz, line 101: since the feature id has been set in the mapper, shouldn't this still be `useProvidedFid = true`?

Kontinuation: The feature id generation code was added to mapper of fsds, and distributed ingestion on fsds does not use this output format.

Kontinuation: The fix in GeoMesaOutputFormat is for datastores other than fsds. I’ve only tested this on HBase, not sure if  this could cover all the cases since testing on every supported datastore takes lots of efforts.

elahrvivaz: ah, ok. hbase should be fine, thanks. did you try both with provided fids and without?

elahrvivaz: I will run through and quickly test accumulo, just in case.

elahrvivaz: i verified it's working as expected, although i'm still confused as to why since i don't think the provided fid hints would make it to the output format.

elahrvivaz:
thanks for the fix!

geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaOutputFormat.scala
Kontinuation: I've also noticed that. Maybe because the ingestion jobs for accumulo and hbase does not have reducers so everything can be done in mapper, and it is possible to skip data serialization/deserialization between mapper output and GeoMesaOutputFormat.

PR #2784, opened 2021-07-09
GEOMESA-3100 Kafka layer views

elahrvivaz:
* Views provide filters and/or transforms without duplicating the data in Kafka


docs/user/kafka/layer_views.rst
```diff
@@ -0,0 +1,27 @@
...
+.. code-block:: java
+
+    import org.geotools.data.DataStore;
+    import org.geotools.data.DataStoreFinder;
+
+    Map<String, Serializable> parameters = new HashMap<>();
+    parameters.put("kafka.zookeepers", "localhost:2181");
+    parameters.put("kafka.brokers", "localhost:9092");
+    parameters.put("kafka.layer.views",
+        "{my-type-name=[{type-name=\"view\",filter=\"INCLUDE\",transform=[\"dtg\",\"geom\"]}]}");
```
jnh5y, line 21: I'd opine that the example should define at least two layers.

docs/user/kafka/layer_views.rst
```diff
@@ -0,0 +1,37 @@
...
+    import org.geotools.data.DataStoreFinder;
+
+    String views =
+      "{" +
+      "  type1 = [" +
+      "    { type-name = \"transformView\", transform = [ \"dtg\", \"geom\", \"name\" ] }," +
+      "    { type-name = \"filterView\", filter = \"bbox(geom,0,0,10,10)\" }" +
+      "  ]," +
+      "  type2 = [" +
+      "    { type-name = \"filterTransformView\", filter = \"bbox(geom,0,0,10,10)\", transform = [ \"dtg\", \"geom\" ] }" +
```
jnh5y, line 24: What order are the transforms and filters applied in?  Can one filter on a new / transformed column?

Hmm... actually, this doesn't show a new column being computed.  Is that possible?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
```diff
@@ -78,17 +81,22 @@ class KafkaDataStore(
     KafkaDataStore.producer(config)
   }
 
+  // view type name -> actual type name
+  private val layerViewLookup =
+    config.layerViewsConfig.flatMap { case (typeName, views) => views.map(_.typeName -> typeName).toMap }
```
jnh5y, line 86: Silly corner case:  Is there any checking that the configuration doesn't have a collision in the layernames?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
```diff
@@ -147,6 +155,26 @@ class KafkaDataStore(
...
+            throw new IllegalStateException("Inconsistent layer view config")
+          }
+          KafkaDataStore.createLayerView(parent, view).viewSft
+        }
+    }
+  }
+
+  override def getTypeNames: Array[String] = {
+    val nonViews = super.getTypeNames
+    nonViews ++ layerViewLookup.collect { case (k, v) if nonViews.contains(v) => k }
```
jnh5y, line 175: I wonder if it'd be "nice" to log a warning when the base store does *not* have the layer configured for a view.  

E.g., the case when `!nonViews.contains(v)`.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
```diff
@@ -386,6 +444,21 @@ object KafkaDataStore extends LazyLogging {
...
+   * @param sft simple feature type the view is based on
+   * @param config layer view config
+   * @return
+   */
+  private[kafka] def createLayerView(sft: SimpleFeatureType, config: LayerViewConfig): LayerView = {
+    val viewSft = SimpleFeatureTypes.renameSft(sft, config.typeName)
+    val filter = config.filter.map(FastFilterFactory.optimize(viewSft, _))
+    val transform = config.transform.map(Transforms(viewSft, _))
+    val finalSft = transform.map(Transforms.schema(viewSft, _)).getOrElse(viewSft)
+    LayerView(finalSft, filter, transform)
```
jnh5y, line 459: Do we have a cheap/quick way to verify that the filter will work against the `finalSft`?

docs/user/kafka/layer_views.rst
elahrvivaz: we always apply filters first and then transforms, i think mainly b/c you're allowed to filter on things that are not in the transform. it should be possible to use any of the normal transforms we use in our queries, it's the same code that does the transform, although I don't have an example of it here.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaListeners.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Collections
+import java.util.concurrent.ConcurrentHashMap
+import scala.util.control.NonFatal
+
+trait KafkaListeners extends StrictLogging {
+
+  import scala.collection.JavaConverters._
+
+  // use a flag instead of checking listeners.isEmpty, which is slightly expensive for ConcurrentHashMap
```
jnh5y, line 25: Ah, this is great...

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
elahrvivaz: there is, here: https://github.com/elahrvivaz/geomesa/blob/fcr_kafka_layer_filters/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreFactory.scala#L329-L332

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
elahrvivaz: i think the default case would be to not have any views, so i don't think we'd want to log a warning on that, unless i'm misunderstanding you

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
elahrvivaz: it gets applied to the original sft so there's no need to check.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureStore.scala
```diff
@@ -12,21 +12,22 @@ import org.geotools.data.{FeatureListener, Transaction}
...
   override def removeFeatures(filter: Filter): Unit = filter match {
     case Filter.INCLUDE => clearFeatures()
     case _ => super.removeFeatures(filter)
   }
 
-  override def addFeatureListener(listener: FeatureListener): Unit = cache.addListener(this, listener)
+  override def addFeatureListener(listener: FeatureListener): Unit = listeners.addListener(this, listener)
 
-  override def removeFeatureListener(listener: FeatureListener): Unit = cache.removeListener(this, listener)
+  override def removeFeatureListener(listener: FeatureListener): Unit = listeners.removeListener(this, listener)
```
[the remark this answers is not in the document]
elahrvivaz: i don't think i tried that but i imagine if you had direct access you'd probably just want to listen to the main topic and not mess with views. the views are more for exposing things through geoserver to end users who shouldn't see the original data for whatever reason.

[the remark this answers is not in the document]
jnh5y: I suppose I'm also amused by the possibility of using a view to "clean up" the output (via projection).

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCache.scala
```diff
@@ -59,7 +69,8 @@ object KafkaFeatureCache extends LazyLogging {
     *
     * @return
     */
-  def empty(): KafkaFeatureCache = EmptyFeatureCache
+  def empty(views: Seq[LayerView] = Seq.empty): KafkaFeatureCache =
+    new EmptyFeatureCache(views.map(v => KafkaFeatureCacheView.empty(v.viewSft)))
```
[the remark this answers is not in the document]
elahrvivaz: yeah, if you disable the cache (e.g. b/c you're only using feature listening) you'd still have the feature types but there just won't be any features when you query the cache.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
jnh5y: Awesome.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCache.scala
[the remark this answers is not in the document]
jnh5y: Ohohohoh, this is a cache.  

So this case goes back to "listening to a view", right?  Which you can likely do, but we haven't tested that whacky idea.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureStore.scala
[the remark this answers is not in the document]
elahrvivaz: yeah, that is a possibility, or do potentially do dictionary lookups or enrichment or something.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCache.scala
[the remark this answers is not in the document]
elahrvivaz: i think listening without caching any features for querying is the only case you'd use the empty cache, but i'm not 100% sure...

docs/user/kafka/layer_views.rst
jnh5y: Always?  In the stats DSL, I remember working to setup transforms first so that they could be filtered and have stats computed on new columns.

I can't find a unit test proving that quickly enough though...

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
jnh5y: If `nonViews` had one SFT named `foo`, and someone computed a view on `bar`.  What happens?

I think that's the case where `nonViews.contains(v)` is false, right?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaFeatureStore.scala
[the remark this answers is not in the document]
jnh5y: Ah!  That could be shiny; set up a little listener to do local enrichment....  neat.

The transforms shown in the documentation and unit tests are only projections.  Do you have the time to modify the unit test to at least rename a field?

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/index/KafkaFeatureCache.scala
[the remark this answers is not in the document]
jnh5y: That seems like the "main" use case of this code.

docs/user/kafka/layer_views.rst
elahrvivaz: pretty sure it's always filter first

jnh5y: Pretty sure I did transform, filter, aggregate...  I know aggregate is last.;)  I'll have to look sometime...

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStore.scala
elahrvivaz: right, yeah if you configure a view on a feature type that doesn't exist yet. I see your point now.

PR #2785, opened 2021-07-13
GEOMESA-3101 Fix duplicate results in Lambda store

elahrvivaz:


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
```diff
@@ -51,6 +51,8 @@ class MergedQueryRunner(ds: HasGeoMesaStats, stores: Seq[(Queryable, Option[Filt
       original: Query,
       explain: Explainer): CloseableIterator[SimpleFeature] = {
 
+    // TODO deduplicate arrow, bin, density queries...
```
jnh5y, line 54: Ok, how would updates come to the Lambda DataStore?  Would it be more sensible to pass all updates to the KDS side of things?  

Or route updates to Kafka if the feature is not in Accumulo?

Also, I think if we go this route, then we really, really ought to implement deduplication for Arrow, Bin, Density...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreViewFactory.scala
```diff
@@ -124,7 +126,16 @@ object MergedDataStoreViewFactory extends GeoMesaDataStoreInfo with NamespacePar
       readWrite = ReadWriteFlag.ReadOnly
     )
 
-  override val ParameterInfo: Array[GeoMesaParam[_ <: AnyRef]] = ConfigLoaderParam.toArray :+ ConfigParam
+  val DeduplicateParam =
+    new GeoMesaParam[java.lang.Boolean](
+      "geomesa.merged.deduplicate",
```
jnh5y, line 131: Undocumented parameter!;)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedQueryRunner.scala
elahrvivaz: all writes go to kafka first, including all updates. we could check the long-term store to see if there is an existing record, but then we'd probably have performance issues again as it would be similar to a modifying write without the lambda store.

The issue with arrow, bin and density is that we don't (usually) have enough info to deduplicate them in the client. so we'd have to either skip all server-side processing, or push down the list of fids to exclude (which may be large). I also think with those queries, it's likely to not affect the UI much if there are some duplicates (densities would be slightly off, bins/arrows would show duplicate points but they would just overlap each other and any lookup click would only return the one feature, tables and charts would be off though).

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/view/MergedDataStoreViewFactory.scala
elahrvivaz: fixed

PR #2796, opened 2021-08-13
Add up front check for exact class match in fast convert, replace set…

elahrvivaz:
…Attributes in FeatureUtils with loop to avoid creating a new list


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/FeatureUtils.scala
```diff
@@ -116,7 +116,11 @@ object FeatureUtils {
     copyToFeature(writer.next(), sf, useProvidedFid)
 
   def copyToFeature(toWrite: SimpleFeature, sf: SimpleFeature, useProvidedFid: Boolean = false): SimpleFeature = {
-    toWrite.setAttributes(sf.getAttributes)
+    var i = 0
+    while (i < sf.getAttributeCount) {
+      toWrite.setAttribute(i, sf.getAttribute(i))
```
jnh5y, line 121: Does this skip the convert call?

elahrvivaz: no this will still run the convert. it will just avoid allocating a new list in the call to getAttributes()

PR #2799, opened 2021-08-23
time partition filters

mmaclach:
ignore OR clauses with different attributes when extracting bounds

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
@@ -198,12 +198,14 @@ object FilterHelper {
   def extractAttributeBounds[T](filter: Filter, attribute: String, binding: Class[T]): FilterValues[Bounds[T]] = {
     filter match {
       case o: Or =>
+        val set = o.getChildren.foldLeft(Set.empty[Option[String]])((acc, f) => acc + getAttributeProperty(f))   
         val all = o.getChildren.flatMap { f =>
           val child = extractAttributeBounds(f, attribute, binding)
           if (child.isEmpty) { Seq.empty } else { Seq(child) }
         }
         val union = FilterValues.or[Bounds[T]](Bounds.union[T]) _
-        all.reduceLeftOption[FilterValues[Bounds[T]]](union).getOrElse(FilterValues.empty)
+        if (set.size != 1) FilterValues.empty  
+        else all.reduceLeftOption[FilterValues[Bounds[T]]](union).getOrElse(FilterValues.empty)
 
       case a: And =>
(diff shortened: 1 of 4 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/FilterHelper.scala
```diff
@@ -198,12 +198,14 @@ object FilterHelper {
   def extractAttributeBounds[T](filter: Filter, attribute: String, binding: Class[T]): FilterValues[Bounds[T]] = {
     filter match {
       case o: Or =>
+        val set = o.getChildren.foldLeft(Set.empty[Option[String]])((acc, f) => acc + getAttributeProperty(f))
```
elahrvivaz, line 201: i think the logic here is slightly more restrictive than we need it to be. for example, something like:
`(dtg DURING a/b AND geom bbox) OR (dtg DURING c/d)` would fail this check, however, we could still extract `dtg DURING a/b OR dtg DURING c/d` as a valid bounds.

I think instead, the condition that we want to give up on is if any of the children bounds are empty, e.g. in the case of `dtg DURING a/b OR geom bbox`, the geom clause would return an empty bounds. we could modify the `o.getChildren.flatMap` to be a `o.getChildren.reduce` of some sort (or maybe a .map followed by a .fold), and give up and return an empty filter values if we hit a child clause that returns an empty bounds.

It would be good to add those 2 cases as a unit test, too.

mmaclach: Ah ok, makes sense. I'll make those changes

geomesa-hbase/geomesa-hbase-datastore/src/test/scala/org/locationtech/geomesa/hbase/data/HBasePartitioningTest.scala
```diff
@@ -119,6 +119,16 @@ class HBasePartitioningTest extends Specification with LazyLogging {
           testQuery(ds, typeName, "name = 'name5'", transforms, Seq(toAdd(5)))
         }
 
+        {
+          val filter = ECQL.toFilter("(bbox(geom,38,48,52,62) and " +
+            "dtg BETWEEN 2018-01-01T00:00:00+00:00 AND 2018-01-04T00:00:00+00:00) OR " +
+            "(bbox(geom,38,48,52,61) and dtg BETWEEN 2018-01-05T00:00:00+00:00 AND 2018-01-08T00:00:00+00:00)")
+          val plans = ds.getQueryPlan(new Query(sft.getTypeName, filter))
+          println(plans)
```
elahrvivaz, line 127:
```suggestion
```

elahrvivaz (approved):

PR #2802, opened 2021-09-19
[GEOMESA-3132] Integrated Apache Sedona with GeoMesa Spark SQL

Kontinuation:
Signed-off-by: Kristin Cowalcijk <morisama@yeah.net>


geomesa-spark/geomesa-spark-test-with-sedona/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <systemPropertyVariables combine.children="append">
+                        <geomesa.use.sedona>true</geomesa.use.sedona>
```
jnh5y, line 85: Ah!  You don't have to work this hard to set the system variables.  You can set and unset them in unit test like this:

https://github.com/locationtech/geomesa/blob/main/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreAuthTest.scala#L82-L87

This would avoid needing to create the separate modules.

jnh5y:
@Kontinuation this is awesome!  Thank you!

jnh5y:
In terms of some bookkeeping, since Sedona is a new dependency (and may pull in others), Eclipse requires us to check on the IP of dependencies. 

For GeoMesa, we do that via a script that we run to build up the entire list of dependencies.  It is here `build/calculate-cqs.sh` and it modifies / rebuilds `builds/cqs.tsv`.  Can you run that script and commit the changes to this PR?  (This will allow @elahrvivaz and I to run down any info about the new dependencies!)

geomesa-spark/geomesa-spark-test-with-sedona/pom.xml
Kontinuation: The reason for placing sedona test in separated module is to ensure that it runs in its own process, since [UDT serde](https://github.com/locationtech/geomesa/pull/2802/files#diff-f7fb90df88946514556bd7ae01e8de1c3a18d4db0a8d05cd37d07a31e2c87f0fR50) only initialize once for performance reason. We can also override `reuseForks` config of surefire plugin in `geomesa-spark-sql` module to achieve this. I'll submit a patch if you prefer this way.

`geomesa-spark-test-without-sedona` have to run in its own module. This is for validating no sedona classes will be loaded when `geomesa.use.sedona` is not enabled, so we want it to run in an environment without sedona on the classpath.

Kontinuation:
> In terms of some bookkeeping, since Sedona is a new dependency (and may pull in others), Eclipse requires us to check on the IP of dependencies.
> 
> For GeoMesa, we do that via a script that we run to build up the entire list of dependencies. It is here `build/calculate-cqs.sh` and it modifies / rebuilds `builds/cqs.tsv`. Can you run that script and commit the changes to this PR? (This will allow @elahrvivaz and I to run down any info about the new dependencies!)

Newly generated `builds/cqs.tsv` was commited. I've skimmed through this file and found four dependency changes:

```
org.apache.accumulo:accumulo-hadoop-mapreduce	2.0.0	provided
org.apache.sedona:sedona-core-2.4_2.11	1.0.1-incubating	provided
org.apache.sedona:sedona-sql-2.4_2.11	1.0.1-incubating	provided
org.wololo:jts2geojson	0.16.1	test
```

build/cqs.tsv
```diff
@@ -312,16 +315,17 @@ org.apache.curator:curator-test	2.7.1	test
...
 org.apache.hbase:hbase-testing-util	1.4.12	test
 org.apache.hbase:hbase-testing-util	2.2.3	test
-org.apache.kafka:kafka_2.11	test:2.1.1	test
 org.apache.kafka:kafka-clients	test:2.1.1	test
+org.apache.kafka:kafka_2.11	test:2.1.1	test
 org.cassandraunit:cassandra-unit	2.2.2.1	test
 org.codehaus.groovy:groovy-jsr223	2.4.5	test
-org.geotools:gt-geometry	23.3	test
 org.geotools.jdbc:gt-jdbc-h2	23.3	test
+org.geotools:gt-geometry	23.3	test
```
elahrvivaz, line 323: weird, seems like the sort order changed, must be something specific to the local environment. I'll see if I can standardize it in the build script.

docs/user/spark/sparksql.rst
```diff
@@ -279,3 +281,114 @@ write to a distributed file system like HDFS.
...
+
+GeoMesa SparkSQL can work seamlessly with `Apache Sedona <http://sedona.apache.org/>`_. You can enable this feature by
+specifying ``geomesa.use.sedona`` system property to ``true`` when submitting your Spark job. Please make sure that
+Apache Sedona JAR ``sedona-python-adapter-${spark-version}_${scala-version}-${sedona-version}.jar`` is submitted with
+``--jars`` option:
+
+.. code-block:: shell
+
+   spark-submit --conf "spark.driver.extraJavaOptions=-Dgeomesa.use.sedona=true" \
+                --conf "spark.executor.extraJavaOptions=-Dgeomesa.use.sedona=true" \
```
elahrvivaz, line 296: this seems good enough for a first pass, but I wonder if it would it be possible to detect sedona on the classpath and just enable it automatically if it's found?

docs/user/spark/sparksql.rst
```diff
@@ -279,3 +281,114 @@ write to a distributed file system like HDFS.
...
+
+There are several configs to take care of when creating Spark session object:
+
+.. code-block:: scala
+
+    val spark = SparkSession.builder().
+      // ...
+      config("spark.serializer", "org.apache.spark.serializer.KryoSerializer").
+      config("spark.kryo.registrator", classOf[GeoMesaSparkKryoRegistrator].getName).
+      config("spark.geomesa.sedona.udf.prefix", "sedona_").
```
elahrvivaz, line 308: would it make sense to have this be the default value if nothing is specified for the sedona prefix?

jnh5y: I want to riff on this idea for a minute.  To clarify, I want to discuss what I am about to suggest as an idea; we should think about it and reach a decision first.

Given @Kontinuation's point about wanting to test with and without Sedona, I wonder if it'd be worth it to just generate GeoMesa-only Spark runtime jars AND GeoMesa+Sedona Spark runtime jars.

The Sedona project cannot bundle GeoTools due to Apache Software Foundation rules.  The Eclipse Foundation permits GeoMesa to bundle GeoTools (by a board-level exemption!).  Given that context, we could basically provide dependencies included options where things just work out of the box.

Thoughts?

docs/user/spark/sparksql.rst
Kontinuation: Enabling Sedona support silently could be a breaking change -- it will change the serde of geometry UDTs, which will break existing pyspark apps.

elahrvivaz: maybe we could have a flag to disable it, but enable it by default if it's available. it seems unlikely someone would have sedona on the classpath but not want to use it? unless someone was using geomesa and sedona separately, but it seems like they would conflict.

Kontinuation: Maybe you are right. I'll enable it when `org.apache.spark.sql.sedona_sql.UDT.GeometryUDT` is available, and `geomesa.use.sedona` system property is not explicitly set to `false`.

elahrvivaz: awesome!

geomesa-spark/geomesa-spark-jts/pom.xml
```diff
@@ -29,6 +29,10 @@
             <version>${spark.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-spark-core_${scala.binary.version}</artifactId>
```
elahrvivaz, line 34: @jnh5y this pulls in geotools to the spark-jts module, does that matter?

jnh5y:
@elahrvivaz Did we get the CQ sorted?

@Kontinuation as an update, I haven't forgotten about this.  I may be in a better position to test things out in another week or so.  (Rather than just merging the PR as is without testing.)

elahrvivaz:
> @elahrvivaz Did we get the CQ sorted?

No, do we need to open a CQ? there wasn't anything in clearly defined last I looked, but it's an apache project.

build/cqs.tsv
```diff
@@ -290,12 +291,14 @@ org.apache.hadoop:hadoop-mapreduce-client-core	2.8.5	provided
...
 org.apache.hadoop:hadoop-yarn-common	2.8.5	provided
 org.apache.hbase:hbase-server	2.2.3	provided
-org.apache.kafka:kafka_2.11	2.1.1	provided
+org.apache.kafka:kafka_2.12	2.1.1	provided
 org.apache.kafka:kafka-clients	2.1.1	provided
-org.apache.spark:spark-catalyst_2.11	2.4.7	provided
-org.apache.spark:spark-core_2.11	2.4.7	provided
-org.apache.spark:spark-sql_2.11	2.4.7	provided
-org.scala-lang:scala-compiler	2.11.7	provided
+org.apache.sedona:sedona-core-2.4_2.12	1.0.1-incubating	provided
```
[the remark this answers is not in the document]
jnh5y: Doh, sorry to be slow to respond.  I'll double-check.  Since Sedona is incubating, that may be a different status with Eclipse.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/sedona/package.scala
```diff
@@ -0,0 +1,49 @@
...
+      if(!sqlContext.experimental.extraStrategies.contains(s))
+        sqlContext.experimental.extraStrategies ++= Seq(s)
+    }
+  }
+
+  private def registerUdfs(sqlContext: SQLContext, prefix: String): Unit = {
+    val sparkSession = sqlContext.sparkSession
+    Catalog.expressions.foreach(f => sparkSession.sessionState.functionRegistry.createOrReplaceTempFunction(
+      prefix + f.getClass.getSimpleName.dropRight(1), f))
+    Catalog.aggregateExpressions_UDAF.foreach(f => sparkSession.udf.register(prefix + f.getClass.getSimpleName, f))
```
geoHeil, line 47: Does this also include the sedona UDAFs? https://github.com/apache/incubator-sedona/blob/5055078d18c687f92fcc67b7377203c501823b8a/sql/src/main/scala/org/apache/sedona/sql/UDF/UdfRegistrator.scala#L44

build/cqs.tsv
geoHeil, line 296: switch over to the current `1.1.1-incubating` release

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
```diff
@@ -71,6 +73,24 @@ object SQLRules extends LazyLogging {
...
 
+  def sedonaExprToGTFilter(pred: Sedona_ST_Predicate): Option[GTFilter] = {
+    val left = pred.children.head
+    val right = pred.children.last
+    (sparkExprToGTExpr(left), sparkExprToGTExpr(right)) match {
+      case (Some(expr1), Some(expr2)) => pred match {
+        case Sedona_ST_Contains(_) => Some(ff.contains(expr1, expr2))
+        case Sedona_ST_Crosses(_) => Some(ff.crosses(expr1, expr2))
+        case Sedona_ST_Overlaps(_) => Some(ff.overlaps(expr1, expr2))
+        case Sedona_ST_Intersects(_) => Some(ff.intersects(expr1, expr2))
```
geoHeil, line 84: After running the code:
```bash

git clone git@github.com:Kontinuation/geomesa.git
cd geomesa
git checkout sedona-integration-pr

build/mvn clean install -T8 -DskipTests
```

I fail to get the functions registered:
```
spark-shell --packages "org.locationtech.geomesa:geomesa-spark-sql_2.12:3.4.0-SNAPSHOT,org.apache.sedona:sedona-python-adapter-3.0_2.12:1.1.1-incubating" --conf "spark.serializer=org.apache.spark.serializer.KryoSerializer" --conf "spark.kryo.registrator=import org.locationtech.geomesa.spark.GeoMesaSparkKryoRegistrator" --conf "spark.geomesa.sedona.udf.prefix=sedona_"

Class.forName("org.apache.sedona.core.serde.SedonaKryoRegistrator")
Class.forName("org.apache.spark.sql.sedona_sql.UDT.GeometryUDT")

import org.locationtech.geomesa.spark.jts
jts.initJTS(spark.sqlContext)

spark.sessionState.functionRegistry.listFunction.map(_.funcName).sorted.filter(_.toUpperCase.contains("ST_")).foreach(println)
// all the geomesa stuff is here
spark.sessionState.functionRegistry.listFunction.map(_.funcName).sorted.filter(_.toUpperCase.contains("SEDONA")).foreach(println)
// no SEDONA things are here
```

even though:

`mvn -pl org.locationtech.geomesa:geomesa-spark-test-with-sedona_2.12 clean package`

succeeds (and thus can load) the sedona functions.

elahrvivaz: you might need one of the runtime jars, e.g. geomesa-gt-spark-runtime, those are uber shaded jars with all the required dependencies.

docs/user/spark/sparksql.rst
```diff
@@ -279,3 +281,120 @@ write to a distributed file system like HDFS.
...
+   import geomesa_pyspark
+   ...
+   from pyspark.sql import SparkSession
+   ...
+   from sedona.register import SedonaRegistrator
+
+   spark = SparkSession.builder.config(...).getOrCreate()
+   ...
+   geomesa_pyspark.init_sql(spark)
+   SedonaRegistrator.registerAll(spark)
```
geoHeil, line 398: https://github.com/locationtech/geomesa/pull/2802#discussion_r804132557 why is this not registering the functions?

geoHeil: https://github.com/locationtech/geomesa/pull/2802/files#diff-8b5907286a459c13b7639108a67cb0aadb6846f1fa79df21cb21cb72173c26b2R91 seems to do it though

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
geoHeil: https://github.com/locationtech/geomesa/pull/2802/files#diff-8b5907286a459c13b7639108a67cb0aadb6846f1fa79df21cb21cb72173c26b2R91 is looking useful

geoHeil: SQLTypes.init(spark.sqlContext)

elahrvivaz: will --packages pull from non-standard repositories? geotools is in it's own maven repo.

geoHeil: no  - you would need to specify the resolvers. However, as I want to use geomesa`s snapshot for the PR I needed to install it to the local M2 repository first. Then everything is pulled already by maven and spark directly resolves locally

geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/sedona/package.scala
geoHeil: yes:

```
spark.sessionState.functionRegistry.listFunction.map(_.funcName).sorted.filter(_.toUpperCase.contains("AGGR")).foreach(println)
sedona_st_envelope_aggr
sedona_st_intersection_aggr
sedona_st_union_aggr

```

geomesa-spark/geomesa-spark-sql/src/main/scala/org/apache/spark/sql/SQLRules.scala
Kontinuation: > SQLTypes.init(spark.sqlContext)

Yes. Sedona UDFs and UDAFs were registered in [`SQLTypes.init`](https://github.com/locationtech/geomesa/pull/2802/files#diff-9cf5352519b4798ab6e5e89f812d98784c45138e99efe3ff67350fb75eb8cd7eR17-R23).

I need to emphasize this in the documentation by adding extra explanations to the [example code for creating Spark session object](https://github.com/locationtech/geomesa/pull/2802/files#diff-b165c48f8f912cd9b059e05b8150825e34c3dfe9e38a7059e202efeaa8728de1R307-R327).

tosen1990:
I am trying to compile this branch using scala 2.11 and cannot pass the accumulo-spark-runtime test case.
When I run test case without accumulo-spark-runtime jar,`isUsingSedona` can work correctly and when with this jar,the sedona dependency fail to work.
Anyone has suggestions?

geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/sedona/package.scala
```diff
@@ -0,0 +1,49 @@
...
+    }
+    Seq(new JoinQueryDetector(sqlContext.sparkSession)).foreach { s =>
+      if(!sqlContext.experimental.extraStrategies.contains(s))
+        sqlContext.experimental.extraStrategies ++= Seq(s)
+    }
+  }
+
+  private def registerUdfs(sqlContext: SQLContext, prefix: String): Unit = {
+    val sparkSession = sqlContext.sparkSession
+    Catalog.expressions.foreach(f => sparkSession.sessionState.functionRegistry.createOrReplaceTempFunction(
```
geoHeil, line 45: What about upgrading this expression to the new API of 3.2 (here for sedona registration and in the geomesa UDF registration):

https://github.com/apache/incubator-sedona/blob/5055078d18c687f92fcc67b7377203c501823b8a/sql/src/main/scala/org/apache/sedona/sql/UDF/UdfRegistrator.scala#L38

changing:

```
sparkSession.sessionState.functionRegistry.createOrReplaceTempFunction(
```

to:

```
sparkSession.sessionState.functionRegistry.registerFunction
```

This should already ensure compatibility with 3.2 Spark.

jnh5y: Can we make that change and remain compatible with Spark 3.1 and before?  (If not, it may be possible to use reflection.)

Historically, I have been an advocate of providing backwards compatibility and limiting dependency changes.

elahrvivaz: +1 to keeping back-compatibility with spark 3.1 and earlier. supporting spark 3.2 might be a larger change that we don't want to roll into this pr.

geoHeil: Would be 3.1 compatible: https://github.com/apache/spark/blob/v3.1.2/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L47 as the method signature seems to be there as well.

geoHeil:
FYI: - see my latest post in https://gitter.im/apache/sedona:

I do not know if and how well you support Spark 2.x on scala 2.11 - but Sedona (and thus also Geomesa using this PR) might run into the following problem:

The Apache Sedona project claims 2.4 spark (legacy https://sedona.apache.org/setup/platform/) support which is built using Scala 2.11.
However, https://mvnrepository.com/artifact/org.apache.sedona/sedona-python-adapter-3.0 the main sedona-python-adapter odes not seem to be published for 2.11 or Spark 2.x.


Where/how can I obtain a 2.11 & spark 2.4.x compatible build of this project?

geoHeil:
> FYI: - see my latest post in https://gitter.im/apache/sedona:
> 
> I do not know if and how well you support Spark 2.x on scala 2.11 - but Sedona (and thus also Geomesa using this PR) might run into the following problem:
> 
> The Apache Sedona project claims 2.4 spark (legacy https://sedona.apache.org/setup/platform/) support which is built using Scala 2.11. However, https://mvnrepository.com/artifact/org.apache.sedona/sedona-python-adapter-3.0 the main sedona-python-adapter odes not seem to be published for 2.11 or Spark 2.x.
> 
> Where/how can I obtain a 2.11 & spark 2.4.x compatible build of this project?

Though: `mvn clean install -DskipTests -Dscala=2.11 -Dspark=2.4` seems to be possible (but simply not published)

geoHeil:
> > FYI: - see my latest post in https://gitter.im/apache/sedona:
> > I do not know if and how well you support Spark 2.x on scala 2.11 - but Sedona (and thus also Geomesa using this PR) might run into the following problem:
> > The Apache Sedona project claims 2.4 spark (legacy https://sedona.apache.org/setup/platform/) support which is built using Scala 2.11. However, https://mvnrepository.com/artifact/org.apache.sedona/sedona-python-adapter-3.0 the main sedona-python-adapter odes not seem to be published for 2.11 or Spark 2.x.
> > Where/how can I obtain a 2.11 & spark 2.4.x compatible build of this project?
> 
> Though: `mvn clean install -DskipTests -Dscala=2.11 -Dspark=2.4` seems to be possible (but simply not published)

https://mvnrepository.com/artifact/org.apache.sedona/sedona-python-adapter-2.4 seems to be already published

tosen1990:
I got the imcompatible problem of  `org.locationtech.jts` when using `BroadcastIndexJoin` from sedona with error:
```scala
Caused by: java.lang.NoSuchMethodError: org.locationtech.jts.index.quadtree.Quadtree.getRoot()Lorg/locationtech/jts/index/quadtree/Root;
	at org.locationtech.jts.index.quadtree.IndexSerde.write(IndexSerde.java:61)
	at org.apache.sedona.core.geometryObjects.SpatialIndexSerde.write(SpatialIndexSerde.java:66)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:628)
```
Currently sedona uses JTS 1.18 and geomesa uses 1.17,anyone knows how to make this work?

Kontinuation:
> I got the imcompatible problem of `org.locationtech.jts` when using `BroadcastIndexJoin` from sedona with error:
> 
> ```scala
> Caused by: java.lang.NoSuchMethodError: org.locationtech.jts.index.quadtree.Quadtree.getRoot()Lorg/locationtech/jts/index/quadtree/Root;
> 	at org.locationtech.jts.index.quadtree.IndexSerde.write(IndexSerde.java:61)
> 	at org.apache.sedona.core.geometryObjects.SpatialIndexSerde.write(SpatialIndexSerde.java:66)
> 	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:628)
> ```
> 
> Currently sedona uses JTS 1.18 and geomesa uses 1.17,anyone knows how to make this work?

I reproduced this problem on PySpark and found a workaround: we can place sedona jar prior to the geomesa jar when specifying `--jars` command line parameter, so that JTS 1.18 will take precedence over JTS 1.17:

```
spark-submit --jars sedona-python-adapter-3.0_2.12-1.1.1-incubating.jar,geomesa-hbase-spark-runtime-hbase2_2.12-3.5.0-SNAPSHOT.jar [other parameters]
```

If you are using Java or Scala, a similar approach can be applied I believe.

elahrvivaz:
i've merged in main, fixed conflicts, and bumped the sedona version to 1.3.1 (and fixed a few resulting errors) here: https://github.com/locationtech/geomesa/pull/2948

elahrvivaz:
merged as https://github.com/locationtech/geomesa/commit/3d4d9749f7344a0357c6fef2f20e3839ed4deb51
sorry for the long wait, thanks for the patience!

PR #2810, opened 2021-11-04
GEOMESA-3147 Fixed None push-down filter applied in GeoMesaRelation

5iFish:


geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaRelation.scala
```diff
@@ -279,8 +279,10 @@ object GeoMesaRelation extends LazyLogging {
           Some(IndexedRDD(rdd))
       }
     }
+   
+   val filter = Option(ECQL.toFilter(params.getOrElse("query", "INCLUDE")))
 
-    GeoMesaRelation(sqlContext, sft, schema, params, None, cached, partitioned)
+    GeoMesaRelation(sqlContext, sft, schema, params, filter, cached, partitioned)
```
elahrvivaz, line 285: @jnh5y do you know if this was intentionally set to None for some reason? the fix seems reasonable, but I'm not too familiar with the intricacies here.

elahrvivaz:
@5iFish thanks for the fix! would it be possible to add a unit test showing the issue and solution?

5iFish:
> @5iFish thanks for the fix! would it be possible to add a unit test showing the issue and solution?

Glad to receive your message. I will add the unit test tomorrow. Good night.

geomesa-spark/geomesa-spark-sql/src/main/scala/org/locationtech/geomesa/spark/GeoMesaRelation.scala
jnh5y: Ok, so this code could use a little refactoring.  

I believe the apply method here is only being used when GeoSpark/Sedona-inspired indexed/partitioned RDDs are being created.  

From reading through the code above, the filter is being applied to the rawRdd, and that is being used to create partitioned and/or indexed RDDs.  

Since the filter has been used there, it does not need to be set (as far as I can help tell).

I am not sure that this change would fix any bug or affect any change.  @5iFish can you say more about the issue you were seeing?  (I may be wrong about some of my assumptions.)

5iFish: The reason for push-down filter in geomesarelation is to ensure that data consistency among the rawRdd and partitioned and/or indexed RDDs.

I have to write duplicate SQL-like query, which differs from CQL, to make sure the data consistency.

For example:

`// create temporary view with query option
spark.sql(
    s"""
      |CREATE TEMPORARY VIEW wktdata
      |USING org.locationtech.geomesa.spark.GeoMesaDataSource
      |OPTIONS (
      |  fs.path '$path',
      |  geomesa.feature '$format',
      |  query 'dtg AFTER 2016-01-01T10:00:00.000Z',
      |)
      |""".stripMargin)

// must add SQL-like query "dtg > '2016-01-01T10:00:00.000Z'" to make sure the data consistency
spark.sql("select * from wktdata where dtg > '2016-01-01T10:00:00.000Z'")`

After the fixing, I don't need to worry about the data consistency. Personally, I believe the change is positive. @jnh5y  what's your opinions?

jnh5y: Ah!  I haven't seen the `USING` datasource `OPTIONS` syntax for creating a table.  That is probably calling the `apply` method directly.  

Ok, can you write a unit test which covers this?  It should be straightforward to create a table which ought to exclude some of the data using the specified query.  You could then show that there's excluded data when you query for `select *`...

5iFish: I inserted the unit test into SparkSQLDataTest.scala, please check.

jnh5y:
@5iFish thanks for finding and fixing this issue!

PR #2813, opened 2021-11-17
GEOMESA-3152 Deletes with Attribute level visibilities do not work

jnh5y:


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -88,18 +89,27 @@ class KryoVisibilityRowEncoder extends RowEncodingIterator {
...
       }
       i += 1
     }
 
+    val attributeVisString = attributeVis.mkString(",")
+    // The magic number "34" is brought to you by adding up
+    //   Class and size info for the key and value (2 bytes for the class, and 2 for the size per key and value).
+    //     2*(2+2) = 8
+    //   The size of the key ("geomesa.feature.visibility".size = 26)
+    //   And the size of the attributeVisString.
```
elahrvivaz, line 111: :100:

jnh5y:
@elahrvivaz thoughts on attribute-viz and modifying features?  There are sharp corner cases there.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/QueryPlan.scala
```diff
@@ -285,8 +286,15 @@ object QueryPlan {
     override def schema: SimpleFeatureType = sft
 
     protected def createSerializer: KryoFeatureSerializer = {
+      val hasAttributeLevelVis = sft.getVisibilityLevel == VisibilityLevel.Attribute
       val builder = KryoFeatureSerializer.builder(sft)
```
elahrvivaz, line 290:
```suggestion
      val builder = KryoFeatureSerializer.builder(sft)
      // note: attribute level vis are serialized into the user data since they can't be held by a single key
```

elahrvivaz (approved):

elahrvivaz:
> @elahrvivaz thoughts on attribute-viz and modifying features? There are sharp corner cases there.

I think deleting features brings up the same corner cases - if you can't actually see all the data, then you can't update or delete it effectively. We might be able to check the user data in the client before one of those ops and throw an error if something is not visible, however maybe that would also be a data spill?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/iterators/KryoVisibilityRowEncoder.scala
```diff
@@ -88,18 +89,27 @@ class KryoVisibilityRowEncoder extends RowEncodingIterator {
...
           val len = metadata.setPosition(index + 1) - pos
           attributes(index) = (bytes, 4 + pos, len) // pos is relative to first byte of attribute offsets, see length calc
+          attributeVis(index) = key.getColumnVisibility.toString
           length += len
         }
       }
       i += 1
     }
 
+    val attributeVisString = attributeVis.mkString(",")
```
elahrvivaz, line 106: if you can't see all the attributes, this would end up with 'null's in the string. maybe we should replace those with empty strings?

jnh5y:
@jrs53 any chance you'd be able to try something like your ingest in an existing unit test?  That may make it easier to figure out what the issue is.

elahrvivaz:
looks like kryo sometimes requires more space than the string length of the string:
```
scala> val out = new com.esotericsoftware.kryo.io.Output(Array.ofDim[Byte](500))
out: com.esotericsoftware.kryo.io.Output = com.esotericsoftware.kryo.io.Output@69497a36

scala> val longString = (0 until 100).mkString
longString: String = 0123456789101112131415161718192021222324252627282930313233343536373839404142434445464748495051525354555657585960616263646566676869707172737475767778798081828384858687888990919293949596979899

scala> out.writeString(longString)

scala> out.position
res12: Int = 192

scala> longString.length
res13: Int = 190
```

elahrvivaz:
writeAscii seems to consistently use the size of the string, and should be ok for our use case as it can still be read with readString

elahrvivaz:
hmm, will vis strings always be ascii?

elahrvivaz:
Seems like we can't rely on that: https://accumulo.apache.org/docs/2.x/security/authorizations#security-label-expression-syntax

elahrvivaz:
merged as https://github.com/locationtech/geomesa/commit/5b75b7dc0c62638b7efa6942e4dd0be35369af16

PR #2820, opened 2021-12-06
GEOMESA-3158 Seek ahead in Z scans

elahrvivaz:
* Z3/Z2 scans will seek to the next valid z-value instead of iterating
* Z range calculations will not stop based on number of recursions
* New query hint `Z_SEEK` can be used to disable seeking
* New system property `geomesa.ranges.recurse` controls level of recursion

Co-authored-by: Jim Hughes <jhughes@ccri.com>

### DIFF
--- a/docs/user/datastores/runtime_config.rst
+++ b/docs/user/datastores/runtime_config.rst
@@ -267,6 +267,13 @@ This property works in conjunction with ``geomesa.scan.block-full-table``, above
 on the number of features that are returned (through the use of ``maxFeatures``), then it will not be blocked.
 The property is specified as an integer. By default, a limit of 1000 or less is allowed.
 
+geomesa.scan.ranges.recurse
++++++++++++++++++++++++++++
+
+This property controls the max level of recursion that will be used when generating scan ranges. Higher levels of
+recursion will generate more accurate ranges at the cost of longer query planning times, and vice-versa. Generally
+this does not need to be configured.
+
 geomesa.scan.ranges.target
 ++++++++++++++++++++++++++
 
@@ -274,6 +281,12 @@ This property provides a rough upper-limit for the number of row ranges that wil
 query. It is specified as a number. In general, more ranges will result in fewer false-positive rows being
 scanned, which will speed up most queries. However, too many ranges can take a long time to generate, and
 overwhelm clients, causing slowdowns. The optimal value depends on the environment.
+
+geomesa.scan.seek
++++++++++++++++++
+
+The property controls seeking in distributed scans. By default seeking is enabled, which should provide the best
+performance.
 
 geomesa.serializer.cache.expiry
 +++++++++++++++++++++++++++++++
--- a/docs/user/datastores/runtime_config.rst
+++ b/docs/user/datastores/runtime_config.rst
@@ -273,7 +273,7 @@ geomesa.scan.ranges.recurse
 This property controls the max level of recursion that will be used when generating scan ranges. Higher levels of
 recursion will generate more accurate ranges at the cost of longer query planning times, and lower levels will do
 the opposite. By default there is no limit. Generally this does not need to be configured, and setting it may
-limit the ranges generated to substantially less than ``geomesa.scan.ranges.target``.
+limit the ranges generated to substantially less than ``geomesa.scan.ranges.target``. It is specified as a integer.
 
 geomesa.scan.ranges.target
 ++++++++++++++++++++++++++
@@ -287,7 +287,7 @@ geomesa.scan.seek
 +++++++++++++++++
 
 The property controls seeking in distributed scans. By default seeking is enabled, which should provide the best
-performance.
+performance. It is specified as ``true`` or ``false``.
 
 geomesa.serializer.cache.expiry
 +++++++++++++++++++++++++++++++
--- a/docs/user/datastores/runtime_config.rst
+++ b/docs/user/datastores/runtime_config.rst
@@ -271,8 +271,8 @@ geomesa.scan.ranges.recurse
 +++++++++++++++++++++++++++
 
 This property controls the max level of recursion that will be used when generating scan ranges. Higher levels of
-recursion will generate more accurate ranges at the cost of longer query planning times, and vice-versa. Generally
-this does not need to be configured.
+recursion will generate more accurate ranges at the cost of longer query planning times, and vice-versa. By default
+there is no limit. Generally this does not need to be configured.
 
 geomesa.scan.ranges.target
 ++++++++++++++++++++++++++
--- a/docs/user/datastores/runtime_config.rst
+++ b/docs/user/datastores/runtime_config.rst
@@ -271,8 +271,9 @@ geomesa.scan.ranges.recurse
 +++++++++++++++++++++++++++
 
 This property controls the max level of recursion that will be used when generating scan ranges. Higher levels of
-recursion will generate more accurate ranges at the cost of longer query planning times, and vice-versa. By default
-there is no limit. Generally this does not need to be configured.
+recursion will generate more accurate ranges at the cost of longer query planning times, and lower levels will do
+the opposite. By default there is no limit. Generally this does not need to be configured, and setting it may
+limit the ranges generated to substantially less than ``geomesa.scan.ranges.target``.
 
 geomesa.scan.ranges.target
 ++++++++++++++++++++++++++
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
@@ -23,6 +23,8 @@ import org.locationtech.geomesa.utils.index.ByteArrays
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
 
+import java.nio.ByteBuffer
+
 @RunWith(classOf[JUnitRunner])
 class Z3IteratorTest extends Specification {
 
@@ -48,35 +50,50 @@ class Z3IteratorTest extends Specification {
           map: java.util.Map[String, String],
           iteratorEnvironment: IteratorEnvironment): Unit = {}
       override def seek(range: Range, collection: java.util.Collection[ByteSequence], b: Boolean): Unit = {
+        staged = key
         key = null
-        staged = null
       }
       override def hasTop: Boolean = staged != null
     }
 
     val iter = new Z3Iterator
     iter.init(srcIter, Z3Iterator.configure(indexValues, 0, None, 25).getOptions, null)
+    iter.seek(new org.apache.accumulo.core.data.Range(new Key(new Text(k)), null), null, inclusive = false)
     iter
   }
 
   "Z3Iterator" should {
 
-    "iterate on points" >> {
-      "keep in bounds values" >> {
-        val test1 = indexValues.sfc.index(-76.0, 38.5, 500)
-        val row = ByteArrays.toBytes(Array[Byte](0, 0), test1)
-        val iter = iterator(row)
-        iter.next()
-        iter.hasTop must beTrue
+    "serialize iterator version" >> {
+      val buf = ByteBuffer.allocate(5)
+      val read = ByteBuffer.wrap(buf.array())
+      (-9 to -1).map(_.toByte).foreach { version =>
+        buf.clear()
+        buf.put(version).mark()
+        (0 to 10000).foreach { i =>
+          buf.reset()
+          buf.putInt(i)
+          read.clear()
+          read.getInt must beLessThan(0)
+          read.position(1)
+          read.get() must beGreaterThanOrEqualTo(0.toByte)
+        }
       }
+      ok
+    }
 
-      "drop out of bounds values" >> {
-        val test2 = indexValues.sfc.index(-70.0, 38.5, 500)
-        val row = ByteArrays.toBytes(Array[Byte](0, 0), test2)
-        val iter = iterator(row)
-        iter.next()
-        iter.hasTop must beFalse
-      }
+    "keep in bounds values" >> {
+      val test1 = indexValues.sfc.index(-76.0, 38.5, 500)
+      val row = ByteArrays.toBytes(Array[Byte](0, 0), test1)
+      val iter = iterator(row)
+      iter.hasTop must beTrue
+    }
+
+    "drop out of bounds values" >> {
+      val test2 = indexValues.sfc.index(-70.0, 38.5, 500)
+      val row = ByteArrays.toBytes(Array[Byte](0, 0), test2)
+      val iter = iterator(row)
+      iter.hasTop must beFalse
     }
 
     "configure 1.3 compatibility mode" >> {
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
@@ -12,9 +12,10 @@ import org.apache.accumulo.core.data.{ByteSequence, Key, Range, Value}
 import org.apache.accumulo.core.iterators.{IteratorEnvironment, SortedKeyValueIterator}
 import org.apache.hadoop.io.Text
 import org.geotools.filter.text.ecql.ECQL
+import org.geotools.util.factory.Hints
 import org.junit.runner.RunWith
 import org.locationtech.geomesa.index.api.ShardStrategy.NoShardStrategy
-import org.locationtech.geomesa.index.conf.FilterCompatibility
+import org.locationtech.geomesa.index.conf.{FilterCompatibility, QueryHints}
 import org.locationtech.geomesa.index.index.z3.Z3IndexKeySpace
 import org.locationtech.geomesa.index.index.z3.legacy.Z3IndexV4.Z3IndexKeySpaceV4
 import org.locationtech.geomesa.index.utils.ExplainNull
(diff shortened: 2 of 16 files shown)


docs/user/datastores/runtime_config.rst
```diff
@@ -267,6 +267,13 @@ This property works in conjunction with ``geomesa.scan.block-full-table``, above
 on the number of features that are returned (through the use of ``maxFeatures``), then it will not be blocked.
 The property is specified as an integer. By default, a limit of 1000 or less is allowed.
 
+geomesa.scan.ranges.recurse
++++++++++++++++++++++++++++
+
+This property controls the max level of recursion that will be used when generating scan ranges. Higher levels of
```
jnh5y, line 273: What's the default?

Also, does this need to be discussed relative to `geomesa.scan.ranges.target`?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z2Filter.scala
```diff
@@ -8,8 +8,9 @@
 
 package org.locationtech.geomesa.index.filters
 
-import java.nio.ByteBuffer
+import org.geotools.util.factory.Hints
```
jnh5y, line 11: Organize imports:).

docs/user/datastores/runtime_config.rst
elahrvivaz: it's not directly related to scan.ranges.target, so i think we don't need to discuss the two in conjunction.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z2Filter.scala
```diff
@@ -38,10 +39,16 @@ class Z2Filter(val xy: Array[Array[Int]]) extends RowFilter {
...
+
   private val RangeSeparator = ":"
   private val TermSeparator  = ";"
 
-  def apply(values: Z2IndexValues): Z2Filter = {
+  def apply(values: Z2IndexValues): Z2Filter = apply(values, seek = true)
+
+  def apply(values: Z2IndexValues, hints: Hints): Z2Filter = apply(values, hints.isScanSeeking)
+
+  private def apply(values: Z2IndexValues, seek: Boolean): Z2Filter = {
```
jnh5y, line 51: Ok, this is just wiring up the hint and not using it, right?  Do we want to set up seeking for Z2?  If not, should we be clear about the difference in the docs?  (That is, that one can only seek with Z3 initially?)

docs/user/datastores/runtime_config.rst
jnh5y: Ok, but before this PR, recursion only happened 7 times, right?  Meaning that one could not get more than a certain number of ranges?

(This is my understanding about how they are related; lemme know where I'm getting it wrong.)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z2Filter.scala
elahrvivaz: we should probably set up seeking for z2 eventually, but I haven't done it yet. i don't think we need to get into the specifics in the docs, given that this ideally should be transparent to users and is configurable mainly as a hedge against regression.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/QueryProperties.scala
```diff
@@ -22,6 +22,7 @@ object QueryProperties {
 
   // rough upper limit on the number of ranges we will generate per query
   val ScanRangesTarget: SystemProperty = SystemProperty("geomesa.scan.ranges.target", "2000")
+  val ScanSeek        : SystemProperty = SystemProperty("geomesa.scan.seek", "true")
```
jnh5y, line 25: Is this documented as a "system property"?  (I'm vaguely confused about how all the settings are documented, etc.)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/QueryHints.scala
```diff
@@ -186,6 +187,11 @@ object QueryHints {
       Option(hints.get(LAMBDA_QUERY_PERSISTENT).asInstanceOf[java.lang.Boolean]).forall(_.booleanValue)
     def isLambdaQueryTransient: Boolean =
       Option(hints.get(LAMBDA_QUERY_TRANSIENT).asInstanceOf[java.lang.Boolean]).forall(_.booleanValue)
+    def isScanSeeking: Boolean =
```
jnh5y, line 190: What's the query hint that someone passes?  Is that documented?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/RowFilter.scala
```diff
@@ -9,7 +9,17 @@
 package org.locationtech.geomesa.index.filters
 
 trait RowFilter {
+
+  import RowFilter.FilterResult
+
+  @deprecated("replaced with filter")
   def inBounds(buf: Array[Byte], offset: Int): Boolean
+
+  // TODO remove default impl in next major version
```
jnh5y, line 18: Is it worth tagging a ticket for 4.0.0?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala
```diff
@@ -55,10 +55,11 @@ class Z3SFC(period: TimePeriod, precision: Int = 21) extends SpaceTimeFillingCur
                       t: Seq[(Long, Long)],
                       precision: Int,
                       maxRanges: Option[Int]): Seq[IndexRange] = {
+    val maxRecursion = sys.props.get("geomesa.scan.ranges.recurse").map(_.toInt).orElse(Some(Int.MaxValue))
     val zbounds = for { (xmin, ymin, xmax, ymax) <- xy ; (tmin, tmax) <- t } yield {
       ZRange(index(xmin, ymin, tmin), index(xmax, ymax, tmax))
     }
-    Z3.zranges(zbounds.toArray, precision, maxRanges)
+    Z3.zranges(zbounds.toArray, precision, maxRanges, maxRecursion)
```
jnh5y, line 62: Should we add this to Z2?  Maybe as a separate ticket with separate testing?

docs/user/datastores/runtime_config.rst
elahrvivaz: that's true, and I should/will add something to the upgrade guide. the recursion limit was a holdover from the initial implementation, when we would not consider the number of ranges left to process when determining when to stop. with the current implementation, scan.ranges.target should be sufficient to restrict the processing to reasonable levels, and setting the recursion limit will sometimes cause less ranges to be generated than desired. I'll add a note to that effect here as well.

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
```diff
@@ -48,35 +50,50 @@ class Z3IteratorTest extends Specification {
...
   "Z3Iterator" should {
 
-    "iterate on points" >> {
-      "keep in bounds values" >> {
-        val test1 = indexValues.sfc.index(-76.0, 38.5, 500)
-        val row = ByteArrays.toBytes(Array[Byte](0, 0), test1)
-        val iter = iterator(row)
-        iter.next()
-        iter.hasTop must beTrue
+    "serialize iterator version" >> {
```
jnh5y, line 67: Do we need this? 

Also, should we have tests for verifying any backwards compatibility in the serialization format?  (Feel free to push back.)

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
```diff
@@ -60,20 +123,87 @@ class Z3Filter(
...
+  val XYKey      = "zxy"
+  val TKey       = "zt"
+  val EpochKey   = "epoch"
+  val MaxTimeKey = "mt"
+  val SeekKey    = "s"
+
 
-  def apply(values: Z3IndexValues): Z3Filter = {
+  // note: we use a negative version number to ensure that we error out with a negative length if an older version
+  // of the filter tries to deserialize a newer serialized buffer. This was verified with the code snippet below
```
jnh5y, line 184: Ok, so what does the error look like?

Also, sounds like this is not backwards compatible, right?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
```diff
@@ -104,17 +234,19 @@ object Z3Filter extends RowFilterFactory[Z3Filter] {
...
   }
 
   override def serializeToBytes(filter: Z3Filter): Array[Byte] = {
     // 4 bytes for length plus 16 bytes for each xy val (4 ints)
     val xyLength = 4 + filter.xy.length * 16
     // 4 bytes for length, then per-epoch 4 bytes for length plus 8 bytes for each t val (2 ints)
     val tLength = 4 + filter.t.map(bounds => if (bounds == null) { 4 } else { 4 + bounds.length * 8 } ).sum
-    // additional 4 bytes for min/max epoch
-    val buffer = ByteBuffer.allocate(xyLength + tLength + 4)
+    // additional 1 byte fore version, 2 bytes for epoch offset, 4 bytes for max time, 1 byte for seek
```
jnh5y, line 246: ^fore^for

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
```diff
@@ -157,24 +306,38 @@ object Z3Filter extends RowFilterFactory[Z3Filter] {
...
+    new Z3Filter(xy, t, epochOffset, maxTime, seek)
   }
+
+  /**
+   * Gets z-low and z-hi for a bounding box in normalized space
+   */
+  private def zBound(xmin: Int, ymin: Int, tmin: Int, xmax: Int, ymax: Int, tmax: Int): Array[Long] =
+    Array(Z3(xmin, ymin, tmin).z, Z3(xmax, ymax, tmax).z)
+
+  private case class ZBounds(bounds: Array[Array[Long]], min: Long)
```
jnh5y, line 342: Ok, future me wants to see a comment about what the Array[Array[Long]] means:).

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/QueryProperties.scala
elahrvivaz: yep, right here: https://github.com/locationtech/geomesa/pull/2820/files#diff-9e613b450b11691f7415f85f61ed549cceddcc634eb6d08add885189b143ec9fR286

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
```diff
@@ -8,23 +8,85 @@
...
-    val maxEpoch: Short
-  ) extends RowFilter {
+    val epochOffset: Short,
+    val maxTime: Int,
+    val seek: Boolean
+  ) extends RowFilter with StrictLogging {
+
+  private val zBoundsNoTime: ZBounds = if (!seek) { null } else {
+    val bounds = xy.map { case Array(xmin, ymin, xmax, ymax) =>
+      Z3Filter.zBound(xmin, ymin, 0, xmax, ymax, maxTime)
```
jnh5y, line 53: In the backwards compatibility case, maxTime is 0, right?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/RowFilter.scala
elahrvivaz: added it to https://geomesa.atlassian.net/browse/GEOMESA-2966

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/QueryHints.scala
elahrvivaz: it's not documented, i'd consider it an 'internal' hint at the moment. it's `SCAN_SEEKING`

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
```diff
@@ -60,20 +123,87 @@ class Z3Filter(
...
+        if (next < nextZ) {
+          nextZ = next
+        }
+      }
+      i += 1
+    }
+
+    var nextEpoch = epoch
+
+    if (nextZ == Long.MaxValue && epoch != Short.MaxValue) {
```
jnh5y, line 150: What's the bit about `epoch != Short.MaxValue`?

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala
elahrvivaz: yeah, i think z2 is less likely to suffer from this issue as the dimensions are better aligned. we could do it along with z2 seeking, but atm we're not really using z2 are we?

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
elahrvivaz: i just wanted to verify it deserialized like i expected and figured it was worth leaving. i can remove it if you want. testing back compatibility would be a little messy, i'd have to copy the old code into the test module i guess? it's not really a supported use case, but i was trying to not break things as much as possible for people who mix-and-match.

docs/user/datastores/runtime_config.rst
jnh5y: I like the addition.  Is it clear that this property is an int/short?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z2Filter.scala
jnh5y: +1 +1 + 1

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
elahrvivaz: it would be some kind of error from running `Array.ofDim(-1)`. so not exactly helpful, but it would ensure that things fail fast.

it is backwards compatible in that the newer iterator should work with older clients, but it is not backwards compatible in that older iterators work with newer clients. we don't have a good way to detect the iterator version in order to send the older serialization format...

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
elahrvivaz: :golfing:

geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala
jnh5y: The z2 index is in for spatial-only / context datasets.  This is subtle enough that a separate ticket (or waiting on it entirely) would be sensible to me!

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
elahrvivaz: seek will be false in that case so this line won't get executed (it's checked in the deserialize method).

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/filters/Z3Filter.scala
elahrvivaz: avoiding numeric wrap-around...

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/iterators/Z3IteratorTest.scala
jnh5y: Sounds great to me.  Leave it as-is and no need to test serialization compatibility (since we assume that folks are matching remote and local code).

docs/user/datastores/runtime_config.rst
elahrvivaz: clear to me ;) 
i'll add a note

jnh5y (approved):
I've read through the PR and the logic looks good to me.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/conf/QueryHints.scala
jnh5y: Hmm...  under what circumstances would it be worth documenting?  I can both appreciate having "internal" hints and also, if we don't document them, I'd worry that we wouldn't find 'em later when we need 'em.

elahrvivaz: i was mainly thinking this would be useful just for testing and otherwise people would use the sys prop (which is documented).

elahrvivaz:
@jnh5y I'm moving this to 4.0, since it has breaking api changes in the iterators, and it didn't seem to improve things much... does that sound ok to you, or should we merge it sooner?

jnh5y:
> @jnh5y I'm moving this to 4.0, since it has breaking api changes in the iterators, and it didn't seem to improve things much... does that sound ok to you, or should we merge it sooner?

Either moving it to 4.0 or not merging it are fine by me.  I was hoping for a bigger win with it, so the complexity may not be justified?

elahrvivaz:
I'm going to close this, since it didn't seem to perform better. I'm willing to re-open if anyone objects, though.

PR #2824, opened 2021-12-17
GEOMESA-3162: Geomesa client configuration file properly loaded.

jgibson:
Despite the best efforts of GEOMESA-2800 only a handful of Accumulo client settings were actually loaded at runtime. This meant that Accumulo client settings like TLS or instance.id were ignored. Worse, there was no way for a user to specify them because TLS settings are not available as command line options for tools like `geomesa-accumulo ingest`.

Other related changes (originally separate commits, but squashed together to meet coding standards):

* New license headers for MITRE 2021 contributions. This is fairly brittle, but AFAICT this is a limitation of the license-maven-plugin.
* AvroSchemaRegistryConverterTest now uses a random Jetty port, not 8089. (Necessary for the unit tests to complete successfully when port 8089 is in use by another process.)
* Error message in accumulo config parser is clearer when authentication options are missing.

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
@@ -27,7 +27,7 @@ import scala.util.Try
  * Client configuration options, loaded from a config file
  *
  * @param config Accumulo client configuration
- * @param hasInstances instance id or name set in config file
+ * @param hasInstance instance id or name set in config file
  * @param instance instance name (not actually used during configuration)
  * @param zookeepers zookeeper connect string (not actually used during configuration)
  * @param zkTimeout zookeeper timeout (not actually used during configuration)
@@ -38,7 +38,7 @@ import scala.util.Try
 @SuppressWarnings(Array("deprecated"))
 case class AccumuloClientConfig(
     config: ClientConfiguration,
-    hasInstances: Boolean,
+    hasInstance: Boolean,
     instance: Option[String],
     zookeepers: Option[String],
     zkTimeout: Option[String],
@@ -48,7 +48,7 @@ case class AccumuloClientConfig(
   ) {
   override def toString: String = {
     val values =
-      Seq(s"hasInstances=$hasInstances") ++
+      Seq(s"hasInstance=$hasInstance") ++
       instance.map(s => s"instance=$s").toSeq ++ zookeepers.map(s => s"zookeepers=$s") ++
         zkTimeout.map(s => s"zookeepers.timeout=$s") ++ authType.map(s => s"auth.type=$s") ++
         principal.map(s => s"principal=$s") ++ token.map(_ => "token=***")
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
@@ -1,6 +1,6 @@
 /***********************************************************************
  * Copyright (c) 2013-2020 Commonwealth Computer Research, Inc.
- * Portions MITRE Copyright (c) 2021
+ * Portions Copyright (c) 2021 The MITRE Corporation
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Apache License, Version 2.0
  * which accompanies this distribution and is available at
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
@@ -1,9 +1,14 @@
 /***********************************************************************
  * Copyright (c) 2013-2020 Commonwealth Computer Research, Inc.
+ * Portions MITRE Copyright (c) 2021
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Apache License, Version 2.0
  * which accompanies this distribution and is available at
  * http://www.opensource.org/licenses/apache2.0.php.
+ * This software was produced for the U. S. Government under Basic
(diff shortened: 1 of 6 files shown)


elahrvivaz:
Thanks for the fix! do you know if the new api calls will work with both accumulo 1.10 and 2.0?

jgibson:
I've only tested with 2.0, but because the `ClientConfiguration` class from Accumulo that we're using is a deprecated holdover from 1.x I'd think that it should behave correctly.

geomesa-convert/geomesa-convert-avro-schema-registry/src/test/scala/org/locationtech/geomesa/convert/avro/registry/AvroSchemaRegistryConverterTest.scala
```diff
@@ -17,7 +17,8 @@ import org.junit.runner.RunWith
 import org.locationtech.geomesa.convert2.SimpleFeatureConverter
 import org.locationtech.geomesa.utils.geotools.SimpleFeatureTypes
 import org.locationtech.geomesa.utils.io.WithClose
-import org.mortbay.jetty.{Handler, Request, Server}
+import org.mortbay.jetty.{Connector, Handler, Request, Server}
```
jnh5y, line 20: Why are there changes to this test?  Was it breaking on your machine?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
```diff
@@ -20,14 +26,19 @@ import scala.util.Try
 /**
  * Client configuration options, loaded from a config file
  *
- * @param instance instance name
- * @param zookeepers zookeeper connect string
- * @param zkTimeout zookeeper timeout
+ * @param config Accumulo client configuration
+ * @param hasInstances instance id or name set in config file
```
jnh5y, line 30: (Picky) Should this be "hasInstance"?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
```diff
@@ -20,14 +26,19 @@ import scala.util.Try
...
+ * @param instance instance name (not actually used during configuration)
+ * @param zookeepers zookeeper connect string (not actually used during configuration)
+ * @param zkTimeout zookeeper timeout (not actually used during configuration)
  * @param authType  authentication type ("password", "kerberos" or authentication token class)
  * @param principal authentication principal (username, keytab file, etc)
  * @param token authentication token (password, etc)
  */
+@SuppressWarnings(Array("deprecated"))
```
jnh5y, line 38: What warning is being deprecated?  Is ClientConfiguration deprecated?

geomesa-convert/geomesa-convert-avro-schema-registry/src/test/scala/org/locationtech/geomesa/convert/avro/registry/AvroSchemaRegistryConverterTest.scala
jgibson: Yes, I happened to have another daemon running on port 8089 at the time which caused the test fail. I figured that I'd remove that landmine for everyone else.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
jgibson: Yes, ClientConfiguration is deprecated, but because you need to support 1.x I don't think that you can avoid using it without a bunch of gymnastics.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
jgibson: Yeah I think so. I probably assumed that it was like the zookeeper setting where you could have multiple instances. Just pushed up a fix.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
```diff
@@ -20,14 +26,19 @@ import scala.util.Try
...
 case class AccumuloClientConfig(
+    config: ClientConfiguration,
+    hasInstance: Boolean,
```
elahrvivaz, line 41: technically changing this class signature will break semantic versioning in a bug fix. It's kind of an internal class though... @jnh5y do you have any thoughts on this? if we change the target branch to main then there's no problem.

geomesa-convert/geomesa-convert-avro-schema-registry/src/test/scala/org/locationtech/geomesa/convert/avro/registry/AvroSchemaRegistryConverterTest.scala
jnh5y: @elahrvivaz if you are happy, I'm happy.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/AccumuloClientConfig.scala
jgibson: Would package-private be sufficient?

jgibson: OK, I restored the original default constructor, made the new fields private, and made the accessors package-private so public API should remain unchanged.

elahrvivaz: ah, great thanks!

elahrvivaz (approved):

jgibson:
Realized that I had done the copyright headers incorrectly, so I pushed a small fix for that.

PR #2825, opened 2021-12-21
WIP to support Spark 3.2.0.

jnh5y:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
         <scalatra.version>2.6.3</scalatra.version>
         <kryo.version>3.0.3</kryo.version>
         <arrow.version>0.16.0</arrow.version>
-        <netty.version>4.1.48.Final</netty.version>
+        <netty.version>4.1.68.Final</netty.version>
         <metrics.version>3.2.6</metrics.version>
         <aws.sdk.version>1.11.421</aws.sdk.version>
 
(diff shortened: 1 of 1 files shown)


jnh5y:
Spark 3.2.0 does not have 2.11 artifacts...

elahrvivaz:
i'm not sure that our build passes against spark 3.1 either, but the artifacts do work there...

pashashiz:
Hello, we were using Geomesa with Spark 3.1, 3.2 and now with 3.3 for more than a year already. All works fine except for one issue with Databricks compatibility (I am going to propose a fix today). Any chance we can upgrade the spark to the latest 3.x version? Yet that will require dropping Scala 2.11 I suppose or we can play around with maven profiles and have different spark versions. If you need any help with that, please, let me know.

elahrvivaz:
> Hello, we were using Geomesa with Spark 3.1, 3.2 and now with 3.3 for more than a year already. All works fine except for one issue with Databricks compatibility (I am going to propose a fix today). Any chance we can upgrade the spark to the latest 3.x version? Yet that will require dropping Scala 2.11 I suppose or we can play around with maven profiles and have different spark versions. If you need any help with that, please, let me know.

We are planning on bumping up our spark version with our next major release, which will come immediately following our pending 3.5.0 release. As you say, we've been holding back b/c it requires dropping scala 2.11 support. thanks!

PR #2830, opened 2021-12-31
GEOMESA-3168 Confluent KDS schemas can be overridden per topic via config parameter

fdfea:


geomesa-kafka/geomesa-kafka-confluent/src/test/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactoryTest.scala
```diff
@@ -0,0 +1,116 @@
+package org.locationtech.geomesa.kafka.confluent
```
elahrvivaz, line 1: build is failing due to a missing license header in this file. you should be able to run `mvn license:format` to add it.

docs/user/kafka/confluent.rst
```diff
@@ -163,3 +163,62 @@ Example GeoMesa Avro Schema
...
     }
+
+Schema Overrides Config
+-----------------------
+
+The schema used to generate a ``SimpleFeatureType`` may optionally be overridden per topic by adding a data store
+configuration parameter at the key ``kafka.schema.overrides``. The value must be a Typesafe Config string with the
+top-level key ``schemas`` that is an object that contains a mapping from topic name to schema definition.
+If an override for a schema exists, it will be used instead of the schema registry. The overrides might be useful
+if you have an existing schema without the geomesa properties, or if you have an external topic that is not connected
```
jnh5y, line 174: I'm picky about 'GeoMesa' always being rendered 'GeoMesa' for consistency.:)

docs/user/kafka/confluent.rst
```diff
@@ -163,3 +163,62 @@ Example GeoMesa Avro Schema
         }
       ]
     }
+
+Schema Overrides Config
+-----------------------
+
+The schema used to generate a ``SimpleFeatureType`` may optionally be overridden per topic by adding a data store
```
jnh5y, line 170: Is the override the entire schema or does it modify things?  (The example makes it look like it is the entire schema.)

geomesa-features/geomesa-feature-avro/src/main/scala/org/locationtech/geomesa/features/avro/AvroSimpleFeatureTypeUtils.scala
```diff
@@ -13,17 +13,17 @@ import org.apache.avro.generic.GenericRecord
...
 import org.locationtech.geomesa.utils.text.{WKBUtils, WKTUtils}
-import org.locationtech.jts.geom.{Geometry, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon}
+import org.locationtech.jts.geom._
 import org.opengis.feature.simple.SimpleFeatureType
 
 import java.nio.ByteBuffer
 import java.util.{Date, Locale}
 import scala.annotation.tailrec
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
```
jnh5y, line 22: Good job on cleaning this up!

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentFeatureSerializer.scala
```diff
@@ -8,41 +8,46 @@
...
+import java.io.{InputStream, OutputStream}
+import java.net.URL
+import java.util.Date
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
 object ConfluentFeatureSerializer extends LazyLogging {
 
-  def builder(sft: SimpleFeatureType, schemaRegistryUrl: URL): Builder = new Builder(sft, schemaRegistryUrl)
+  def builder(sft: SimpleFeatureType, schemaRegistryUrl: URL, schemaOverride: Option[Schema]): Builder =
```
jnh5y, line 31: Hmm....  if we are going to implement a fluent "Builder" pattern here, then it may make sense to have methods which let one set the sft, schema registry, and overrides separately.  

On the other hand, it may be that the builder pattern is *completely* unnecessary.  If that's the case, we could rip it out.

docs/user/kafka/confluent.rst
```diff
@@ -163,3 +163,62 @@ Example GeoMesa Avro Schema
         }
       ]
     }
+
+Schema Overrides Config
```
jnh5y, line 167: @elahrvivaz any other places that things like `kafka.schema.overrides` needs to be documented?

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
```diff
@@ -54,14 +62,44 @@ object ConfluentKafkaDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogg
...
+
+  private[confluent] def parseSchemaOverrides(schemaOverridesConfig: String): Map[String, (SimpleFeatureType, Schema)] = {
+    try {
+      Option(schemaOverridesConfig)
+        .map(ConfigFactory.parseString(_).resolve())
+        .filter(_.hasPath("schemas"))
+        .map {
+          _.getObject("schemas").asScala.map { case (topic, schemaConfig) =>
+            val schema = new Schema.Parser().parse(schemaConfig.render(ConfigRenderOptions.concise()))
+            topic -> (AvroSimpleFeatureTypeUtils.schemaToSft(schema, Some(topic)), schema)
```
jnh5y, line 97: Ok, so in the case that all the configuration is correct, this is fine. 

What kind of error messages does one get if the config for one or more of the schemas is wrong?  Is there anyway to give feedback about which schema(s) could be better?

docs/user/kafka/confluent.rst
fdfea: It does override the whole thing, I could make it per field, that would just be a little more work. Should I?

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
fdfea: I'll make it so it throws errors on a per schema basis

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentFeatureSerializer.scala
fdfea: Yeah I don't really know why it was there in the first place

fdfea: I think its a builder because you can add the other serialization options. I could make the schema override part of the builder, but I'm not sure if that really makes anything better.

docs/user/kafka/confluent.rst
elahrvivaz: i think this is the only docs page that deals with confluent, so i don't think it needs to be anywhere else

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
```diff
@@ -54,14 +62,50 @@ object ConfluentKafkaDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogg
   override val DisplayName = "Confluent Kafka (GeoMesa)"
   override val Description = "Confluent Apache Kafka\u2122 distributed log"
 
-  val SchemaRegistryUrl = new GeoMesaParam[URL]("kafka.schema.registry.url", "URL to a confluent schema registry server, used to read Confluent schemas (experimental)")
+  val SchemaRegistryUrl =
+    new GeoMesaParam[URL](
+      "kafka.schema.registry.url",
+      "URL to a confluent schema registry server, used to read Confluent schemas (experimental)",
+      optional = false
```
elahrvivaz, line 69: if you have schema overrides, then you don't need a schema registry, correct? in the docs it said one use case is to not have a schema registry, but here we are requiring one.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStore.scala
```diff
@@ -8,19 +8,28 @@
...
 object ConfluentKafkaDataStore {
 
-  def apply(config: KafkaDataStoreConfig, schemaRegistryUrl: URL): KafkaDataStore = {
-    val metadata = new ConfluentMetadata(new CachedSchemaRegistryClient(schemaRegistryUrl.toExternalForm, 100))
-    val serialization = new ConfluentGeoMessageSerializerFactory(schemaRegistryUrl)
+  def apply(
+      config: KafkaDataStoreConfig,
+      schemaRegistryUrl: URL,
+      schemaOverrides: Map[String, (SimpleFeatureType, Schema)]): KafkaDataStore = {
+    val sftNameToSchema = schemaOverrides.values.map { case (sft, schema) => sft.getTypeName -> schema }.toMap
```
elahrvivaz, line 26: this strikes me as slightly problematic since there's nothing enforcing unique feature type names, so they might clobber each other. maybe it's not really an issue in regular use

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
elahrvivaz: it would probably be easiest to just update the docs to remove that reference, since we use this param to load the confluent KDS instead of the regular one.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
```diff
@@ -54,14 +62,50 @@ object ConfluentKafkaDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogg
...
+    KafkaDataStoreParams.Brokers.exists(params) &&
+      KafkaDataStoreParams.Zookeepers.exists(params) &&
+      SchemaRegistryUrl.exists(params)
+  }
+
+  private[confluent] def parseSchemaOverrides(schemaOverridesConfig: String): Map[String, (SimpleFeatureType, Schema)] = {
+    try {
+      Option(schemaOverridesConfig)
+        .map(ConfigFactory.parseString(_).resolve())
+        .filter(_.hasPath("schemas"))
```
elahrvivaz, line 93: should this be an error if there's no 'schemas' defined?

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
```diff
@@ -54,14 +62,50 @@ object ConfluentKafkaDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogg
...
+        .map {
+          _.getObject("schemas").asScala.map { case (topic, schemaConfig) =>
+            try {
+              val schema = new Schema.Parser().parse(schemaConfig.render(ConfigRenderOptions.concise()))
+              val sft = AvroSimpleFeatureTypeUtils.schemaToSft(schema, Some(topic))
+              topic -> (sft, schema)
+            } catch {
+              case NonFatal(ex) =>
+                throw new IllegalArgumentException(s"Schema override for topic '$topic' is invalid: ${ex.getMessage}")
```
elahrvivaz, line 102: generally we try to avoid throwing and immediately re-catching an exception, as it's a somewhat confusing control flow, but it's ok for now.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentMetadata.scala
```diff
@@ -8,32 +8,37 @@
...
+import org.opengis.feature.simple.SimpleFeatureType
 
+import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
-class ConfluentMetadata(val schemaRegistry: SchemaRegistryClient) extends GeoMesaMetadata[String] with LazyLogging {
-
-  import ConfluentMetadata.{SubjectPostfix, SchemaIdKey}
+class ConfluentMetadata(schemaRegistry: SchemaRegistryClient, sftOverrides: Map[String, SimpleFeatureType])
```
elahrvivaz, line 25:
```suggestion
class ConfluentMetadata(val schemaRegistry: SchemaRegistryClient, sftOverrides: Map[String, SimpleFeatureType] = Map.empty)
```
for better api compatibility

geomesa-kafka/geomesa-kafka-confluent/src/test/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactoryTest.scala
```diff
@@ -0,0 +1,124 @@
+/***********************************************************************
+ * Copyright (c) 2013-2021 Commonwealth Computer Research, Inc.
```
elahrvivaz, line 2:
```suggestion
 * Copyright (c) 2013-2022 Commonwealth Computer Research, Inc.
```

sorry, you'll probably need to update this since i updated the year

geomesa-features/geomesa-feature-avro/src/main/scala/org/locationtech/geomesa/features/avro/AvroSimpleFeatureTypeUtils.scala
```diff
@@ -13,17 +13,17 @@ import org.apache.avro.generic.GenericRecord
...
 import scala.reflect.{ClassTag, classTag}
 import scala.util.control.NonFatal
 
-object AvroSimpleFeatureTypeParser {
+object AvroSimpleFeatureTypeUtils {
```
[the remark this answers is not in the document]
fdfea: I renamed it in anticipation of my next PR where I add a new method `sftToSchema`, in which case AvroSimpleFeatureTypeParser didn't really seem like it captured all the functionality. I'll revert it for this PR.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStore.scala
fdfea: @elahrvivaz The SFTs in the schemaOverrides map are generated using schemaToSft, where the name of the SFT is the topic in the config override. Thus, the SFTs cannot have overlapping names because if a topic appears more than once in the config only one will be used since the (topic -> config) are stored in a map. I could have called schemaToSft later to avoid passing a (SFT, Schema) (as the schema is directly used to create the SFT), but I wanted to validate the schema as early as possible. Also, because the schema overrides are passed in on creation of the DS, there could not be an SFT that already exists with the same name, and if you try to create another SFT with the same name it won't let you? Does that make sense / address the issue you mentioned?

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
fdfea: I thought it would be fine for the map to just be empty, I have a unit test for that case.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
```diff
@@ -29,8 +34,11 @@ class ConfluentKafkaDataStoreFactory extends DataStoreFactorySpi {
   override def createDataStore(params: java.util.Map[String, Serializable]): KafkaDataStore = {
     val config = KafkaDataStoreFactory.buildConfig(params)
     val url = ConfluentKafkaDataStoreFactory.SchemaRegistryUrl.lookup(params)
+    val schemaOverridesConfig = ConfluentKafkaDataStoreFactory.SchemaOverrides.lookup(params)
+    val schemaOverrides = ConfluentKafkaDataStoreFactory.parseSchemaOverrides(schemaOverridesConfig)
```
elahrvivaz, line 38: can we change this to be `lookupOpt` and then pass an option into `ConfluentKafkaDataStoreFactory.parseSchemaOverrides`? that way we're clear that it may be null

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
elahrvivaz: i was thinking if you had a typo in the config or something, it would just fail silently and might be hard to find out why. maybe we could just log a warning if there is a config provided but there's no 'schemas' key present, although an exception would be more obvious ;)

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStore.scala
elahrvivaz: it makes sense, it would just be nice to enforce that with the api vs relying on the code creating the Map. maybe we could just change this line to be:
```suggestion
    val sftNameToSchema = schemaOverrides.map { case (topic, (_, schema)) => topic -> schema }
```

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
fdfea: Would it make sense to just remove the outer try-catch?

elahrvivaz: no, i think you'd want it for the `ConfigFactory.parseString` call. possibly you could refactor to use `Try` with `flatMap` and then check for `Success` or `Failure` at the end.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
fdfea: Gotcha, I'll add that

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStoreFactory.scala
```diff
@@ -54,14 +62,47 @@ object ConfluentKafkaDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogg
...
+    new GeoMesaParam[String](
+      "kafka.schema.overrides",
+      "Typesafe configuration defining a map from topic name to schema override (experimental)",
+      optional = true,
+      largeText = true
+    )
 
   override val ParameterInfo: Array[GeoMesaParam[_ <: AnyRef]] =
-    KafkaDataStoreFactory.ParameterInfo.+:(SchemaRegistryUrl)
+    KafkaDataStoreFactory.ParameterInfo ++ Array(SchemaRegistryUrl, SchemaOverrides)
```
elahrvivaz, line 81:
```suggestion
    Array(SchemaRegistryUrl) ++ KafkaDataStoreFactory.ParameterInfo ++ Array(SchemaOverrides)
```

This would probably make config easier, since the geoserver page orders the input properties according to this. Previously the `+:` was pre-pending the schema registry url, so this will keep that behavior and then add the SchemaOverrides below required things like kafka brokers, etc.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentKafkaDataStore.scala
fdfea: Yeah that makes more sense.

docs/user/kafka/confluent.rst
jnh5y: I think this is fine as is.

jnh5y: As long as the docs are clear about how it all works.

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentFeatureSerializer.scala
jnh5y: Ok, then things are likely sensible as is.

jnh5y (approved):

PR #2832, opened 2022-01-13
GEOMESA-3170 Allow for specifying different shard numbers for Z2 and …

aheyne:
…Z3 indices.

- Adds config keys geomesa.z2.splits and geomesa.z3.splits

### DIFF
--- a/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
+++ b/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
@@ -303,8 +303,13 @@ object RichSimpleFeatureType extends Conversions {
     def getUserDataPrefixes: Seq[String] =
       Seq(GeomesaPrefix) ++ userData[String](UserDataPrefix).map(_.split(",")).getOrElse(Array.empty)
 
-    def setZShards(splits: Int): Unit = sft.getUserData.put(IndexZShards, splits.toString)
-    def getZShards: Int = userData(IndexZShards).map(int).getOrElse(4)
+    def setZ2Shards(splits: Int): Unit = sft.getUserData.put(IndexZ2Shards, splits.toString)
+    def getZ2Shards: Int =
+      userData(IndexZ2Shards).map(int).getOrElse(userData(IndexZShards).map(int).getOrElse(4))
+
+    def setZ3Shards(splits: Int): Unit = sft.getUserData.put(IndexZ3Shards, splits.toString)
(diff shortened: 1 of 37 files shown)


geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/SimpleFeatureTypes.scala
```diff
@@ -64,6 +63,11 @@ object SimpleFeatureTypes {
     val UpdateBackupMetadata  = "schema.update.backup.metadata"
     val UpdateRenameTables    = "schema.update.rename.tables"
 
+    @deprecated("replaced with IndexZ2Shards and IndexZ3Shards")
```
elahrvivaz, line 66: Could we leave z.splits as undeprecated, and just make the others take precedence if present? that would simplify config in the general case, and you've already accounted for the others having higher priority.

aheyne: Sure, do you want the same done in the schema builder?

elahrvivaz:
lgtm, I'd just like to keep the original config undeprecated.

jnh5y:
Also, is it worth updating the Git commit to say "Deprecates" instead of "Depricates"?;)

geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala
```diff
@@ -303,8 +303,13 @@ object RichSimpleFeatureType extends Conversions {
     def getUserDataPrefixes: Seq[String] =
       Seq(GeomesaPrefix) ++ userData[String](UserDataPrefix).map(_.split(",")).getOrElse(Array.empty)
 
-    def setZShards(splits: Int): Unit = sft.getUserData.put(IndexZShards, splits.toString)
-    def getZShards: Int = userData(IndexZShards).map(int).getOrElse(4)
+    def setZ2Shards(splits: Int): Unit = sft.getUserData.put(IndexZ2Shards, splits.toString)
```
elahrvivaz, line 306:
```suggestion
    def setZShards(splits: Int): Unit = sft.getUserData.put(IndexZShards, splits.toString)
    @deprecated("use z2/z3 specific methods")
    def getZShards: Int = userData(IndexZShards).map(int).getOrElse(4)
    def setZ2Shards(splits: Int): Unit = sft.getUserData.put(IndexZ2Shards, splits.toString)
```

PR #2833, opened 2022-01-14
GEOMESA-3171 Postgis partitioning dialect for JDBCDataStore

elahrvivaz:
* Designed for real-time data flows
* Data is partitioned into tables based on the feature time
  * Allows effecient querying through Postgres partition pruning
* Partitioning can be enabled through data store param `dbtype=postgis-partitioned`
* Requires `pg_cron` extension to be installed

### DIFF
--- a/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreFactory.scala
+++ b/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreFactory.scala
@@ -1,0 +1,63 @@
+/***********************************************************************
+ * Copyright (c) 2013-2022 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.gt.partition.postgis
+
+import org.geotools.data.postgis.{PostGISDialect, PostGISPSDialect, PostgisNGDataStoreFactory}
+import org.geotools.jdbc.{JDBCDataStore, SQLDialect}
+import org.locationtech.geomesa.gt.partition.postgis.dialect.PartitionedPostgisDialect
+
+class PartitionedPostgisDataStoreFactory extends PostgisNGDataStoreFactory {
+
+  import PartitionedPostgisDataStoreParams.DbType
+
+  override def getDisplayName: String = "PostGIS (partitioned)"
+
+  override def getDescription: String = "PostGIS Database with time-partitioned tables"
+
+  override protected def getDatabaseID: String = DbType.sample.asInstanceOf[String]
+
+  override protected def setupParameters(parameters: java.util.Map[_, _]): Unit = {
+    super.setupParameters(parameters)
+    // override postgis dbkey
+    parameters.asInstanceOf[java.util.Map[AnyRef, AnyRef]].put(DbType.key, DbType)
+  }
+
+  override protected def createDataStoreInternal(store: JDBCDataStore, params: java.util.Map[_, _]): JDBCDataStore = {
+
+    val ds = super.createDataStoreInternal(store, params)
+    val dialect = new PartitionedPostgisDialect(ds)
+
+    // TODO copy any other configs?
+    ds.getSQLDialect match {
+      case d: PostGISDialect =>
+        dialect.setLooseBBOXEnabled(d.isLooseBBOXEnabled)
+        dialect.setEstimatedExtentsEnabled(d.isEstimatedExtentsEnabled)
+        dialect.setFunctionEncodingEnabled(d.isFunctionEncodingEnabled)
+        dialect.setSimplifyEnabled(d.isSimplifyEnabled)
+        dialect.setEncodeBBOXFilterAsEnvelope(d.isEncodeBBOXFilterAsEnvelope)
+        ds.setSQLDialect(dialect)
+
+      case d: PostGISPSDialect =>
+        dialect.setLooseBBOXEnabled(d.isLooseBBOXEnabled)
+        dialect.setFunctionEncodingEnabled(d.isFunctionEncodingEnabled)
+        // TODO dialect.setEstimatedExtentsEnabled(d.isEstimatedExtentsEnabled)
+        // TODO dialect.setSimplifyEnabled(d.isSimplifyEnabled)
+        dialect.setEncodeBBOXFilterAsEnvelope(d.isEncodeBBOXFilterAsEnvelope)
+        ds.setSQLDialect(new PostGISPSDialect(ds, dialect))
+
+      case d => throw new IllegalArgumentException(s"Expected PostGISDialect but got: ${d.getClass.getName}")
+    }
+    ds
+  }
+
+  // these will get replaced in createDataStoreInternal, above
+  override protected def createSQLDialect(dataStore: JDBCDataStore): SQLDialect = new PostGISDialect(dataStore)
+  override protected def createSQLDialect(dataStore: JDBCDataStore, params: java.util.Map[_, _]): SQLDialect =
+    new PostGISDialect(dataStore)
+}
--- a/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
+++ b/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
@@ -1,0 +1,188 @@
+/***********************************************************************
+ * Copyright (c) 2013-2022 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.gt.partition.postgis.dialect
+
+import com.typesafe.scalalogging.StrictLogging
+import org.geotools.data.postgis.PostGISDialect
+import org.geotools.jdbc.JDBCDataStore
+import org.locationtech.geomesa.gt.partition.postgis.dialect.functions.{LogCleaner, TruncateToPartition, TruncateToTenMinutes}
+import org.locationtech.geomesa.gt.partition.postgis.dialect.procedures._
+import org.locationtech.geomesa.gt.partition.postgis.dialect.tables._
+import org.locationtech.geomesa.gt.partition.postgis.dialect.triggers.{DeleteTrigger, InsertTrigger, UpdateTrigger, WriteAheadTrigger}
+import org.locationtech.geomesa.utils.geotools.{Conversions, SimpleFeatureTypes}
+import org.opengis.feature.simple.SimpleFeatureType
+
+import java.sql.{Connection, DatabaseMetaData}
+import scala.collection.mutable.ArrayBuffer
+
+class PartitionedPostgisDialect(store: JDBCDataStore) extends PostGISDialect(store) with StrictLogging {
+
+  // order of calls from JDBCDataStore during create schema:
+  //  encodeCreateTable
+  //  encodeTableName
+  //  encodePrimaryKey
+  //  encodeColumnName
+  //  encodeColumnType
+  //  encodePostColumnCreateTable
+  //  encodePostCreateTable
+  //  postCreateTable
+
+  // order of calls during remove schema:
+  //  preDropTable
+  //  "DROP TABLE" + encodeTableName
+  //  postDropTable
+
+  // state for checking when we want to use the write_ahead table in place of the main view
+  private val replaceTableName = new ThreadLocal[Boolean]() {
+    override def initialValue(): Boolean = false
+  }
+
+  // filter out the partition tables from exposed feature types
+  override def getDesiredTablesType: Array[String] = Array("VIEW")
+
+  override def encodeTableName(raw: String, sql: StringBuffer): Unit = {
+    sql.append(s""""${if (replaceTableName.get) { raw + WriteAheadTableSuffix } else { raw }}"""")
+    replaceTableName.remove()
+  }
+
+  override def encodePrimaryKey(column: String, sql: StringBuffer): Unit = {
+    encodeColumnName(null, column, sql)
+    // make our primary key a string instead of the default integer
+    sql.append(" character varying NOT NULL")
+  }
+
+  override def postCreateTable(schemaName: String, sft: SimpleFeatureType, cx: Connection): Unit = {
+    val info = TypeInfo(schemaName, sft)
+
+    super.postCreateTable(schemaName, sft, cx)
+
+    implicit val ex: ExecutionContext = new ExecutionContext(cx)
+    try {
+      val commands = ArrayBuffer.empty[Sql]
+      commands += WriteAheadTable
+      commands += WriteAheadTrigger
+      commands += PartitionTables
+      commands += MainView
+      commands += InsertTrigger
+      commands += UpdateTrigger
+      commands += DeleteTrigger
+      commands += PrimaryKeyTable
+      commands += AnalyzeQueueTable
+      commands += SortQueueTable
+      commands += TruncateToTenMinutes
+      commands += TruncateToPartition
+      commands += RollWriteAheadLog
+      commands += PartitionWriteAheadLog
+      commands += MergeWriteAheadPartitions
+      if (info.partitions.maxPartitions.isDefined) {
+        commands += DropAgedOffPartitions
+      }
+      commands += PartitionMaintenance
+      commands += AnalyzePartitions
+      commands += PartitionSort
+      commands += LogCleaner
+
+      commands.foreach(_.create(info))
+    } finally {
+      ex.close()
+    }
+  }
+
+  /**
+   * Re-create the PLGP/SQL functions and procedures associated with a feature type. This can be used
+   * to 'upgrade in place' if the functions are changed.
+   *
+   * @param schemaName database schema, e.g. "public"
+   * @param sft feature type
+   * @param cx connection
+   */
+  def recreateFunctions(schemaName: String, sft: SimpleFeatureType, cx: Connection): Unit = {
+    val info = TypeInfo(schemaName, sft)
+    implicit val ex: ExecutionContext = new ExecutionContext(cx)
+    try {
+      val commands = ArrayBuffer.empty[Sql]
+      commands += TruncateToTenMinutes
+      commands += TruncateToPartition
+      commands += RollWriteAheadLog
+      commands += PartitionWriteAheadLog
+      commands += MergeWriteAheadPartitions
+      if (info.partitions.maxPartitions.isDefined) {
+        commands += DropAgedOffPartitions
+      }
+      commands += PartitionMaintenance
+      commands += AnalyzePartitions
+      commands += PartitionSort
+      commands += LogCleaner
+      commands.foreach(_.drop(info))
+      commands.foreach(_.create(info))
+    } finally {
+      ex.close()
+    }
+  }
+
+  override def postCreateFeatureType(
+      sft: SimpleFeatureType,
+      metadata: DatabaseMetaData,
+      schemaName: String,
+      cx: Connection): Unit = {
+    // normally views get set to read-only, override that here since we use triggers to delegate writes
+    sft.getUserData.remove(JDBCDataStore.JDBC_READ_ONLY)
+  }
+
+  override def preDropTable(schemaName: String, sft: SimpleFeatureType, cx: Connection): Unit = {
+    // due to the JDBCDataStore hard-coding "DROP TABLE" we have to redirect it away from the main view
+    replaceTableName.set(true)
+    val info = TypeInfo(schemaName, sft)
+
+    implicit val ex: ExecutionContext = new ExecutionContext(cx)
+    try {
+      Seq(
+        PartitionMaintenance,
+        PartitionSort,
+        LogCleaner,
+        PrimaryKeyTable,
+        MainView,
+        InsertTrigger,
+        UpdateTrigger,
+        DeleteTrigger,
+        PartitionTables
+      ).foreach(_.drop(info))
+    } finally {
+      ex.close()
+    }
+  }
+
+  override def postDropTable(schemaName: String, sft: SimpleFeatureType, cx: Connection): Unit = {
+    val tables = Tables(sft, schemaName)
+    // rename the sft so that configuration is applied to the write ahead table
+    super.postDropTable(schemaName, SimpleFeatureTypes.renameSft(sft, tables.writeAhead.name.raw), cx)
+  }
+}
+
+object PartitionedPostgisDialect {
+
+  object Config extends Conversions {
+
+    val IntervalHours                  = "pg.partitions.interval.hours"
+    val MaxPartitions                  = "pg.partitions.max"
+    val WriteAheadTableSpace           = "pg.partitions.tablespace.wa"
+    val WriteAheadPartitionsTableSpace = "pg.partitions.tablespace.wa-partitions"
+    val MainTableSpace                 = "pg.partitions.tablespace.main"
+    val CronMinute                     = "pg.partitions.cron.minute"
+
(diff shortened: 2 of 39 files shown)


jnh5y:
effecient should be efficient in the commit message.:)

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreFactory.scala
```diff
@@ -0,0 +1,63 @@
...
+    // override postgis dbkey
+    parameters.asInstanceOf[java.util.Map[AnyRef, AnyRef]].put(DbType.key, DbType)
+  }
+
+  override protected def createDataStoreInternal(store: JDBCDataStore, params: java.util.Map[_, _]): JDBCDataStore = {
+
+    val ds = super.createDataStoreInternal(store, params)
+    val dialect = new PartitionedPostgisDialect(ds)
+
+    // TODO copy any other configs?
```
jnh5y, line 36: Is this worth a ticket?  Or resolving before merging?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  override def postCreateTable(schemaName: String, sft: SimpleFeatureType, cx: Connection): Unit = {
+    val info = TypeInfo(schemaName, sft)
+
+    super.postCreateTable(schemaName, sft, cx)
+
+    implicit val ex: ExecutionContext = new ExecutionContext(cx)
+    try {
+      val commands = ArrayBuffer.empty[Sql]
+      commands += WriteAheadTable
```
jnh5y, line 68: Does the order matter?  And seems like a Seq() of these might read better?  I don't know what syntactically easiest to read here...

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/functions/LogCleaner.scala
```diff
@@ -0,0 +1,61 @@
...
+  override def name(info: TypeInfo): String = "cron_log_cleaner"
+
+  override def jobName(info: TypeInfo): String = s"${info.name}-cron-log-cleaner"
+
+  override protected def createStatements(info: TypeInfo): Seq[String] =
+    Seq(function()) ++ super.createStatements(info)
+
+  override protected def dropStatements(info: TypeInfo): Seq[String] = Seq.empty // function is shared between types
+
+  override protected def schedule(info: TypeInfo): String = "0 0 * * *"
```
jnh5y, line 28: Errrm.... should the cron schedule be configurable?  Maybe later?

Since this pattern is common, I'd worry that it may run at the same time as other things...

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
elahrvivaz: The order does matter. I mainly did it this way to avoid having to ++ around the awkward optional procedure, but I don't mind changing it if you'd prefer.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/functions/LogCleaner.scala
```diff
@@ -0,0 +1,61 @@
...
+    val maintenanceSuffix = PartitionMaintenance.jobName(info)
+    val rollSuffix = RollWriteAheadLog.jobName(info)
+    val analyzeSuffix = AnalyzePartitions.jobName(info)
+    val logsSuffix = jobName(info)
+
+    s"""CREATE OR REPLACE FUNCTION cron_log_cleaner(name text, tokeep interval) RETURNS void LANGUAGE plpgsql AS
+       |  $$BODY$$
+       |    DECLARE
+       |      maintenance_name text;
+       |      roll_name text;
```
jnh5y, line 44: I'm only skimming at the minute.  Is this "roll" like "turn over" or "role" like "purpose"?  (In this context, I could imagine either.)

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/functions/LogCleaner.scala
elahrvivaz: yeah, this is a fairly lightweight function... maybe i'll just move it to like 1:30AM? i did make the more process-intensive schedules configurable and semi-randomized when not explicit.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/functions/LogCleaner.scala
elahrvivaz: the first!

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -0,0 +1,158 @@
...
+  val hours = 1
+  val spec =
+    "name:String,age:Int,dtg:Date,*geom:Point:srid=4326;" +
+        Seq(
+          s"pg.partitions.interval.hours=$hours",
+          "pg.partitions.cron.minute=0"/*,
+          "pg.partitions.max=2",
+          "pg.partitions.tablespace.wa=wa",
+          "pg.partitions.tablespace.wa-partitions=wapa",
+          "pg.partitions.tablespace.main=main"*/
```
jnh5y, line 42: Are these intentionally commented out?

elahrvivaz:
> effecient should be efficient in the commit message.:)

we'll probably need to fix that on the merge

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
elahrvivaz: yeah, i can remove them if you like. i need to add docs still

elahrvivaz: this test requires a postgis instance so it's skipped, it's really more just a quick way to create the schema and insert some data as it doesn't really validate anything.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreFactory.scala
elahrvivaz: mmm, probably not...

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -0,0 +1,158 @@
...
+        } else {
+          WithClose(ds.asInstanceOf[JDBCDataStore].getConnection(Transaction.AUTO_COMMIT)) { cx =>
+            val dialect = ds.asInstanceOf[JDBCDataStore].dialect match {
+              case p: PartitionedPostgisDialect => p
+              case p: PostGISPSDialect =>
+                val m = p.getClass.getDeclaredMethod("getDelegate")
+                m.setAccessible(true)
+                m.invoke(p).asInstanceOf[PartitionedPostgisDialect]
+            }
+            dialect.recreateFunctions("public", sft, cx)
```
jnh5y, line 92: *blinks* Ok, I might need more coffee or an explanation for the use of reflection here.  What's the biggest goal of this block?

elahrvivaz: the JDBC data store is very locked down. it's a `final` class, and when using prepared statements you can't even access the dialect directly. this was just a bit of hackery to expose the custom dialect, so that I can re-create the plpgsql procedures without dropping and re-creating the schema.

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -0,0 +1,158 @@
...
+      update = false,
+      delete = false,
+      remove = false
+    )
+
+  lazy val sft = SimpleFeatureTypes.createType(s"test", spec)
+
+  "PartitionedPostgisDataStore" should {
+    "work" in {
+      skipped("requires postgis instance")
```
jnh5y, line 58: Whew...  I almost want to recommend TestContainers so that we could have a PostGIS unit (well, integration, really) tests here.

https://github.com/murrayk/postgis-integration-test-example seems to work out some of the details.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/triggers/UpdateTrigger.scala
```diff
@@ -0,0 +1,60 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.gt.partition.postgis.dialect
+package triggers
+
+object UpdateTrigger extends SqlFunction {
```
jnh5y, line 12: Scaladoc?

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
elahrvivaz: mm requires docker though...

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/triggers/InsertTrigger.scala
```diff
@@ -0,0 +1,42 @@
...
+  private def function(info: TypeInfo): String =
+    s"""CREATE OR REPLACE FUNCTION "${name(info)}"() RETURNS trigger AS
+       |  $$BODY$$
+       |    BEGIN
+       |      INSERT INTO ${info.tables.writeAhead.name.full} VALUES(NEW.*);
+       |      RETURN NEW;
+       |    END;
+       |  $$BODY$$
+       |LANGUAGE plpgsql VOLATILE
+       |COST 100;""".stripMargin
```
jnh5y, line 30: What was does the cost here mean?

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
jnh5y: Yeah...  I wonder if there's way to use Docker when you have it and skip it when you don't.... that's probably too finicky.

Is it worth adding explicit testing for this datastore to the minor release punch/check list?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/triggers/InsertTrigger.scala
elahrvivaz: it can be used by the query planner, but I'm not really sure it's relevant. I copied it from an existing postgis partitioning scheme we had set up.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/tables/PrimaryKeyTable.scala
```diff
@@ -0,0 +1,49 @@
...
+         |  pk_column_idx integer,
+         |  pk_column character varying,
+         |  pk_policy character varying,
+         |  pk_sequence character varying
+         |);""".stripMargin
+    val cleanup = s"""DELETE FROM $table WHERE table_schema = ? AND table_name = ?;"""
+    val entry = s"""INSERT INTO $table(table_schema, table_name, pk_column_idx, pk_column) VALUES (?, ?, ?, ?);"""
+    ex.execute(create)
+    ex.executeUpdate(cleanup, Seq(info.schema, info.tables.view.name.raw))
+    ex.executeUpdate(entry, Seq(info.schema, info.tables.view.name.raw, 0, "fid"))
```
jnh5y, line 37: Is there any kind of locking which needs to happen to make sure these steps/actions happen "together" / atomically?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/procedures/RollWriteAheadLog.scala
```diff
@@ -0,0 +1,83 @@
...
+package org.locationtech.geomesa.gt.partition.postgis.dialect
+package procedures
+
+import org.locationtech.geomesa.gt.partition.postgis.dialect.tables.WriteAheadTable
+
+import java.util.Locale
+
+/**
+ * Moves the current write ahead table to a sequential name, and creates a new write ahead table
+ * to accept writes going forward
```
jnh5y, line 18: At a high-level, how are incoming writes handled during a roll?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/procedures/PartitionMaintenance.scala
```diff
@@ -0,0 +1,58 @@
...
+  override def jobName(info: TypeInfo): String = s"${info.name}-part-maintenance"
+
+  override protected def createStatements(info: TypeInfo): Seq[String] =
+    Seq(proc(info)) ++ super.createStatements(info)
+
+  override protected def schedule(info: TypeInfo): String = {
+    val minute = info.partitions.cronMinute.getOrElse {
+      // spread out the cron schedule so that all the feature types don't run at the exact same time
+      // also don't run at same minute as roll-write-ahead (i.e. use 0-8)
+      // TODO this will mean data left in the write ahead table longer -
```
jnh5y, line 30: Are the WA tables renamed to something?  If so, can the "new" WA table just be named what it will be named and then writes are directed to whatever table they belong to?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/tables/PrimaryKeyTable.scala
elahrvivaz: they're part of a transaction so that should happen automatically. but this table is just modified up front during createSchema, so shouldn't really have any conflicts.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/package.scala
```diff
@@ -0,0 +1,426 @@
...
+  private[dialect] val WriteAheadTableSuffix            = "_wa"
+  private[dialect] val PartitionedWriteAheadTableSuffix = "_wa_partition"
+  private[dialect] val PartitionedTableSuffix           = "_partition"
+  private[dialect] val AnalyzeTableSuffix               = "_analyze_queue"
+  private[dialect] val SortTableSuffix                  = "_sort_queue"
+
+  /**
+   * Wrapper to facilitate jdbc calls.
+   *
+   * Note: does not do any validation, inputs must be sanitized by the caller.
```
jnh5y, line 36: Which inputs?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/procedures/RollWriteAheadLog.scala
elahrvivaz: they'll be queued up/paused due to the locking required by each, but only for a very brief time period that isn't really noticeable.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/package.scala
```diff
@@ -0,0 +1,426 @@
...
+   *
+   * @param name table name
+   * @param tablespace table space
+   * @param storage storage opts (auto vacuum)
+   */
+  case class TableConfig(name: TableName, tablespace: TableSpace, storage: String)
+
+  object TableConfig {
+    def apply(schema: String, name: String, tablespace: Option[String], vacuum: Boolean = true): TableConfig = {
+      val storage = if (vacuum) { "" } else { " WITH (autovacuum_enabled = false)" }
```
jnh5y, line 161: Hmm.... I wonder if one can turn off autovacuum globally.  If so, should this respect that or be trying to override "bad" system settings?  Either way, it may be worth explaining which tables have autovacuum set and which do not in the docs.

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
jnh5y: Eh, if these are useful to uncomment while running against a live instance, keeping them around would be fine.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/procedures/PartitionMaintenance.scala
elahrvivaz: to keep it simple the one that receives writes is always suffixed with `_writes` and then renamed during the roll, according to a seq. So you'll start with `_writes`, then `_000`, `_001`, etc. we could adopt the final name up front, but then we'd have to update the trigger when we roll tables, as well as track which table is receiving the writes so that we don't move the data out of it yet. the associated indices (e.g. `spatial_geom_000`) start out with the final names so that we don't have to rename them all. i guess it's pretty much equivalent either way.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
jnh5y: *shrugs*  It is fine.  I think all the syntax for this will be weird.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/package.scala
elahrvivaz: all the inputs! specifically, any sql passed into the `execute` methods.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/functions/LogCleaner.scala
jnh5y: If it is lightweight, no worries!

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/tables/PrimaryKeyTable.scala
jnh5y: Is there any locking on createSchema to prevent multiple clients from calling create at the same time?  Same with simultaneous calls to createSchema and deleteSchema?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/package.scala
elahrvivaz: you can disable auto vacuum at a global level, in which case this setting wouldn't affect anything. in that case, you're responsible to managing vacuums on your own. it's not inherently worse, just different. i disabled it on the write ahead tables because they're short-lived, so vacuuming them is needless overhead.

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
jnh5y: Yeah, true, the JDBC DS is locked down.

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/tables/PrimaryKeyTable.scala
elahrvivaz: i don't think there's any explicit locking in java. due to standard sql transactions, i think things would generally be fine. whichever call gets in first will succeed, and the other would end up blocked and then fail and rollback. it would be the same as calling `create table foo` from 2 different sessions, which sql databases should handle fine.

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
elahrvivaz: probably only requires explicit testing when upgrading geotools versions...

jnh5y: That's true.  So we would want to test on major version changes then!  (Assuming that GeoTools version changes only happen on major version changes.)

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/package.scala
elahrvivaz: updated to be clearer

PR #2842, opened 2022-03-07
GEOMESA-3184 Fix root element at  in XmlConverter

elahrvivaz:


elahrvivaz:
will cherry-pick back to 3.3.x

PR #2854, opened 2022-05-10
GEOMESA-3198 Kafka streams integration

elahrvivaz:


docs/user/kafka/streams.rst
```diff
@@ -0,0 +1,119 @@
...
+    with data from GeoMesa 3.4.x or earlier.
+
+For feature types created prior to version 3.5.0, GeoMesa will use a custom partitioner for data
+written to Kafka. When using Kafka Streams operations that require co-partitioning, such as joins, the
+GeoMesa topic will need to be re-partitioned.
+
+Existing feature types can be updated to use the default Kafka partitioner, which will allow for joins without
+re-partitioning. However, note that updates for a given feature may go to the wrong partition, which may result
+in older data being returned when GeoMesa is queried. This will generally resolve itself over time as Kafka log
+retention policies delete the older data (unless Kafka log compaction is enabled for the topic).
```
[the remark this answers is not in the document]
elahrvivaz: ah, how does `compact,delete` work? i didn't see anything about that reading over the compaction docs just now.

geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/utils/GeoMessageSerializer.scala
```diff
@@ -95,6 +94,18 @@ object GeoMessageSerializer {
     new GeoMessageSerializer(sft, serializer, kryoSerializer, avroSerializer, version)
   }
 
+  def partition(numPartitions: Int, toKey: => Array[Byte]): Int = {
+    // use the feature id if available, otherwise (for clear) use random shard
```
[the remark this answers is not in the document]
elahrvivaz: hmm, not sure. `clear` kind of breaks all the paradigms around kafka partitioning. since ordering isn't guaranteed between partitions, even a single consumer might get the clear out of order. we'd almost need to send a clear to _each_ partition, and update the logic to only clear features from that partition. but that seems kind of complicated... going forward, it will always go to the same partition (although it won't necessarily be the first), since `clear` is always an empty key and value and the default partitioner uses a hash algorithm. this method was copied from the old implementation, so i'm inclined to not change the behavior...

docs/user/kafka/streams.rst
```diff
@@ -0,0 +1,119 @@
...
+    val topology = builder.build()
+    // construct the streams app as normal
+
+Joins and Topic Partitioning
+----------------------------
+
+.. warning::
+
+    Kafka streams operations that require co-partitioning will usually require a re-partition step when used
+    with data from GeoMesa 3.4.x or earlier.
```
jnh5y, line 103: *Blinks*  Why?

docs/user/kafka/streams.rst
```diff
@@ -0,0 +1,119 @@
...
+For feature types created prior to version 3.5.0, GeoMesa will use a custom partitioner for data
+written to Kafka. When using Kafka Streams operations that require co-partitioning, such as joins, the
+GeoMesa topic will need to be re-partitioned.
+
+Existing feature types can be updated to use the default Kafka partitioner, which will allow for joins without
+re-partitioning. However, note that updates for a given feature may go to the wrong partition, which may result
+in older data being returned when GeoMesa is queried. This will generally resolve itself over time as Kafka log
+retention policies delete the older data (unless Kafka log compaction is enabled for the topic).
+
+To update an existing feature type, add the user data entry ``geomesa.kafka.partitioning=default``. Through the
```
jnh5y, line 114: I'm unclear what 'default' means here.  Does it mean the partitioning that's used in 3.4.0 and earlier?  

Is it the default for GeoMesa?  Or the default for Kafka?  What are the other options?

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentFeatureSerializer.scala
```diff
@@ -70,38 +70,45 @@ class ConfluentFeatureSerializer(
...
 
-    val feature = featureReader.get.read(id, record)
+    val feature = featureReaders.get.read(id, record)
 
     // set the feature visibility if it exists
-    Option(sft.getUserData.get(GeoMesaAvroVisibilityField.KEY)).map(_.asInstanceOf[String]).foreach { fieldName =>
-      try {
-        SecurityUtils.setFeatureVisibility(feature, record.get(fieldName).toString)
-      } catch {
-        case NonFatal(ex) => throw new IllegalArgumentException(s"Error setting feature visibility using" +
```
jnh5y, line 82: Is there no longer any chance that an exception is thrown?

geomesa-kafka/geomesa-kafka-datastore/src/test/scala/org/locationtech/geomesa/kafka/streams/GeoMesaStreamsBuilderTest.scala
```diff
@@ -0,0 +1,235 @@
...
+      "kafka.consumer.read-back" -> "Inf",
+      "kafka.zk.path"            -> zkPath
+    )
+  }
+
+  "GeoMesaStreamsBuilder" should {
+    "read from geomesa topics" in {
+      val params = getParams("word/count")
+      // write features to the embedded kafka
+      val kryoTopic = WithClose(DataStoreFinder.getDataStore(params.asJava)) { ds =>
```
jnh5y, line 80: Will the streams integration work with Avro topics?

elahrvivaz: yep!

geomesa-kafka/geomesa-kafka-confluent/src/main/scala/org/locationtech/geomesa/kafka/confluent/ConfluentFeatureSerializer.scala
elahrvivaz: I think there could be, but i think the main causes here would be if the vis field was null in the record, and i added a check for that. any exceptions would get bubbled up, instead of being wrapped as they were, but the end result should be similar.

docs/user/kafka/streams.rst
elahrvivaz: default is the default kafka partitioning. there aren't any other options (by convention not having any config will use the old geomesa partitioning for compatibility). this is meant as an opt-in to switch the partitioning if you know what you're doing.

docs/user/kafka/streams.rst
elahrvivaz: we had an old custom partitioner, needed b/c originally we used a timestamp for the kafka message key and we needed to ensure that features with the same FID went to the same topic. We switched to using the FID as the message key a while ago, but we never removed the custom partitioner. with co-partitioning and streams, you have to ensure that the same key in both streams is in the same partition. Since we used a custom partitioner, that won't be the case, so you have to re-partition the stream.

PR #2855, opened 2022-05-13
Fix doc: multiple stats should be deliminated by semi-colon instead of comma

Kontinuation:


elahrvivaz:
thanks for the catch!

PR #2856, opened 2022-05-13
[GEOMESA-3199] String comparison should always comform to lexicographical order

Kontinuation:
The semantics of [GeoTools binary comparison operator](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/CompareFilterImpl.java#L126) is quite strange: when both sides of the operator are string and can be parsed as a number, then it would compare them as if they were numbers. GeoMesa filter optimizer would often replace these filters with a fast version, which always compares strings by lexicographical order.

The default value of binary comparison operators' `matchCase` property is not consistent, the checks for the value of the `matchCase` property in `FastFilterFactory` sometimes prevent from GeoTools filters being replaced with fast filters. Removing these checks makes the behavior of binary comparison operators more consistent.

Indeed there will be some queries exhibiting inconsistent string comparison behavior when the GeoTools comparison operator was not replaced by the query optimizer. However, it is still useful to fix for the most common cases.

### DIFF
--- a/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
+++ b/geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
@@ -112,7 +112,7 @@ class FastFilterFactory private extends org.geotools.filter.FilterFactoryImpl wi
                        exp2: Expression,
                        matchCase: Boolean,
                        matchAction: MatchAction): PropertyIsGreaterThan = {
-    if (matchCase || matchAction != MatchAction.ANY) {
+    if (matchAction != MatchAction.ANY) {
       super.greater(exp1, exp2, matchCase, matchAction)
     } else {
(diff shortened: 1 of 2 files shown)


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/factory/FastFilterFactory.scala
```diff
@@ -224,7 +224,7 @@ class FastFilterFactory private extends org.geotools.filter.FilterFactoryImpl wi
                            exp2: Expression,
                            matchCase: Boolean,
                            matchAction: MatchAction): PropertyIsLessThanOrEqualTo = {
-    if (matchCase || matchAction != MatchAction.ANY) {
+    if (matchAction != MatchAction.ANY) {
```
elahrvivaz, line 227: looking at the geotools filter impls, they seem to ignore `matchCase`, so this seems safe enough.

elahrvivaz (approved):

elahrvivaz:
thanks for the fix!

PR #2857, opened 2022-05-13
GEOMESA-3200 Make filter bounds union and intersection produce non overlapping bounds

Kontinuation:
GeoTools filters were transformed to ranges of values when running queries on indexed attributes. When value ranges overlap with each other, running the query would yield duplicated results.

For example, the CQL query `a < 1 OR a < 2` will be transformed to attribute value ranges `[(-inf, 1), (-inf, 2)]`. These value ranges would be transformed into index key ranges, thus leading to a redundant range scan on the index table.

This patch prevents overlapping attribute value ranges from being populated, thus preventing queries from generating duplicated results.


elahrvivaz (approved):

elahrvivaz:
thanks for the fix!

PR #2875, opened 2022-06-09
GEOMESA-3212 Postgis - convert constant functions to literals for SQL…

elahrvivaz:
… translation

### DIFF
--- a/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/filter/LiteralFunctionVisitor.scala
+++ b/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/filter/LiteralFunctionVisitor.scala
@@ -1,0 +1,38 @@
+/***********************************************************************
+ * Copyright (c) 2013-2022 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/filter/LiteralFunctionVisitor.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.geotools.factory.CommonFactoryFinder
+import org.geotools.filter.FilterAttributeExtractor
+import org.geotools.filter.visitor.DuplicatingFilterVisitor
+import org.opengis.filter.Filter
+import org.opengis.filter.expression.Function
+
+/**
+ * Replaces 'constant' functions with their literal values
+ */
+class LiteralFunctionVisitor extends DuplicatingFilterVisitor {
```
jnh5y, line 20: Hmm...  I wonder if there's already a visitor like this in GeoTools.

Also, I should ask for a unit test... but I'mma :shipit: !

elahrvivaz: there was a similar block of code in the postgis dialect, but it only replaced the arguments to a function and not the function itself.

PR #2882, opened 2022-07-22
GEOMESA-3218 Add scala 2.12 compatibly changes required for 2.13

jfehr67:
These are changes that shouldn't effect building in scala 2.12 but are required for 2.13.


elahrvivaz:
thank you for doing all this!

elahrvivaz:
if you have any more pushes, could you do regular pushes and not force-pushes so that it's easier to compare the change? i will squash them down to a single commit on merge. thanks!

elahrvivaz:
looks like there are still some compile failures, are you planning on fixing them? i can also help out if you'd like.

jfehr67:
Sorry, there was two errors showing up as a result of accidentally including a 2.13 only import that should have been a 2.12 one. The change I made was changing that to the 2.12 import. I'll push separately going forward.

elahrvivaz:
thanks again for getting 2.13 support started!

PR #2904, opened 2022-10-18
GEOMESA-3238 Support for Non-Normalized Heatmaps

mkearns-ccri:
* Adds normalization boolean to DensityProcess
* Updates density process documentation to reflect the change
* Adds scala implementation of the HeatmapSurface class

### DIFF
--- a/geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/GridTransform.scala
+++ b/geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/GridTransform.scala
@@ -1,0 +1,49 @@
+/***********************************************************************
+ * Copyright (c) 2013-2022 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.process.analytic
+
+import org.locationtech.jts.geom.Envelope
+
+class GridTransform(env: Envelope, xSize: Int, ySize: Int, isClamped: Boolean = true) {
+
+  var dx: Double = env.getWidth / (xSize - 1)
+  var dy: Double = env.getHeight / (ySize - 1)
+
+  def x(i: Int): Double =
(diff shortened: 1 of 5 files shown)


mkearns-ccri:
@ronq @jrs53 @elahrvivaz

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/HeatmapSurface.scala
```diff
@@ -0,0 +1,119 @@
...
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.process.analytic
+
+import org.locationtech.jts.geom.Envelope
+
+class HeatmapSurface(kernelRadius: Int,
```
mkearns-ccri, line 13: @elahrvivaz I basically just ripped this from geotools and added a ccri copyright header - not sure if this is okay or not. Unfortunately I had to copy the whole class because java doesn't allow overriding private methods. Thoughts?

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/GridTransform.scala
```diff
@@ -0,0 +1,49 @@
...
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.process.analytic
+
+import org.locationtech.jts.geom.Envelope
+
+class GridTransform(env: Envelope, xSize: Int, ySize: Int, isClamped: Boolean = true) {
```
mkearns-ccri, line 13: same here

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/HeatmapSurface.scala
elahrvivaz: I confirmed with @jnh5y that we're not allowed to copy and re-license geotools as it's LGPL, so we'll need to come up with a different solution here.

elahrvivaz:
Going to close for now, but feel free to re-open if work resumes.

PR #2912, opened 2022-11-14
Doc: Added a subsection for community modules

Kontinuation:
Currently GeoMesa SQL is only known community module, hope there'll be more of them in the future.


elahrvivaz:
thanks!

PR #2918, opened 2022-11-22
3.4.x

BigData-YC:
Using geomesa-hbase ingest is too slow。
And map-reduce process only has one Map task , as well as one Reduce task ..
it is so slow.....
how to set the number of map task and reduce task ???

like this : geomesa-hbase bulk-load -c c-f f --index id --input hdfs://hadoop/file.geojson


BigData-YC:
Or some other way to import it。

elahrvivaz:
map/reduce mappers are based on the input splits - for geojson, you'd have to split up your input file into multiple files so that they can be processed in parallel. If you have any more questions, please open a ticket in JIRA, contact us via the user list or on gitter, or open a stack overflow post. Thanks!

https://github.com/locationtech/geomesa#join-the-community

PR #2922, opened 2022-11-28
GEOMESA-3247 Fix Databricks Spark aggregation/grouping

pashashiz:
If we use spark geomesa library while running in Databricks Runtime (which is the most popular provider of Spark at the moment) via
```scala
spark.withJTS
```
And execute any aggregation or group by operation, there will be the following error:
```
NoSuchMethodException: java.lang.NoSuchMethodError: org.apache.spark.sql.catalyst.plans.logical.Aggregate.<init>
```
The thing is - Geomesa is using Spark private API which does not guarantee any contract stability and turned out that Databricks maintains a fork of the planner and for Aggregate constructor has 4 instead of 3 arguments.

The fix is to allow to the relaxed way of calling that constructor which would work in both cases.

We use that patch for more than a year in production and looks like that is the only issue we noticed so far.


elahrvivaz:
Thanks for the fix! We've tried to engage with databricks, but they won't share their Spark API changes, so things like this crop up.

PR #2923, opened 2022-11-28
GEOMESA-3249 HBase - Fix Spark Kerberos authentication

pashashiz:
Once we started using Geomesa with Kerberos-secured HBase on different runtimes (regular APIs in docker, Flink, Spark on Databricks) we faced a number of issues that prevented Geomesa to work at all. The issues are:

- While running on managed env such as Databricks there is already a current user present on the workers. And simply running `HBaseConnectionPool.configureSecurity` which does `UserGroupInformation.loginUserFromKeytab` does not set the current user as logged-in user and current authorization basically does nothing. There are 2 ways to make a fix:
  - The most simple which affects less code which I applied. Switch from `UserGroupInformation.getCurrentUser` to `UserGroupInformation.getLoginUser` which always would used logged-in user.
  - The best way, yet will require lots of refactoring. The connection to the HBase should be created with an explicit user via `ConnectionFactory.createConnection(conf, user)`. In case there is any nested code that relies on the current user (for example, hadoop hbase data format) is called - that should be wrapped with `user.doAs { ... }`
- `HBaseIndexAdapter` leaks on long-running jobs. There is a known HBase defect and passing the explicit thread pool and closing it manually solves the issue. 
- While writing into HBase that is a good idea to control the flush timeout, otherwise, we cannot guarantee the latency of writing into HBase.
- While working with Spark we need to handle hadoop configuration changes as well as authorization in every `mapPartition` . Note, that explicit reauth is not expensive cause  the user will be statically cached on every node anyway. Also for some reason a piece of code with token auth never worked for us (I am not even sure how it was supposed to be working)
- While ingesting spark DataFrame data into Geomesa if there is an existing Geomesa schema we should prefer it if compatible and not try to force Spark inferred schema which might be slightly different yet compatible and will cause the crash. 

Note, those changes were already working in Geomesa running on Kerberos-secured Azure HD Insights HBase for more than a year on different workloads. We were adding more and more fixes as we faced different situations. I believe the code might be improved a lot yet that will require rewriting a bunch of original code and a new cycle of E2E tests.

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
@@ -21,6 +21,7 @@ import org.apache.hadoop.hbase.regionserver.BloomType
 import org.apache.hadoop.hbase.security.visibility.CellVisibility
 import org.locationtech.geomesa.hbase.HBaseSystemProperties
 import org.locationtech.geomesa.index.api.IndexAdapter.RequiredVisibilityWriter
+import org.locationtech.geomesa.utils.concurrent.ExitingExecutor.NamedThreadFactory
 import org.locationtech.geomesa.utils.io.IsFlushableImplicits
 
 import java.nio.charset.StandardCharsets
@@ -58,6 +59,8 @@ import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 
 import scala.util.Random
 import scala.util.control.NonFatal
+
+import java.util.concurrent.{ExecutorService, Executors, TimeUnit}
 
 class HBaseIndexAdapter(ds: HBaseDataStore) extends IndexAdapter[HBaseDataStore] with StrictLogging {
 
@@ -584,6 +587,12 @@ object HBaseIndexAdapter extends LazyLogging {
     import org.locationtech.geomesa.utils.geotools.RichSimpleFeatureType.RichSimpleFeatureType
 
     private val batchSize = HBaseSystemProperties.WriteBatchSize.toLong
+    private val flushTimeout = HBaseSystemProperties.WriteFlushTimeout.toLong
+
+    val pool: ExecutorService = Executors.newFixedThreadPool(
+      indices.length,
+      new NamedThreadFactory(s"geomesa-hbase-index-writer-%d"))
+
     private val deleteVis = HBaseSystemProperties.DeleteVis.option.map(new CellVisibility(_))
 
     private val mutators = indices.toArray.map { index =>
@@ -593,6 +602,12 @@ object HBaseIndexAdapter extends LazyLogging {
       }
       val params = new BufferedMutatorParams(TableName.valueOf(table))
       batchSize.foreach(params.writeBufferSize)
+      flushTimeout.foreach(params.setWriteBufferPeriodicFlushTimeoutMs)
+
+      // We have to pass a pool explicitly and close it after manually,
+      // cause of HBase issue where pools got leaked and never closed
+      // (in case of long running Spark jobs 24+ hours the workers go out of memory without custom pool)
+      params.pool(pool)
       ds.connection.getBufferedMutator(params)
     }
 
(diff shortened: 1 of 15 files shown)


elahrvivaz:
Thanks for the fixes!

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -59,6 +59,10 @@ import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import scala.util.Random
 import scala.util.control.NonFatal
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder
```
elahrvivaz, line 62: could you use a non-guava class here instead? we have had a lot of issues with incompatible guava versions, so we try to avoid using it wherever possible.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
```diff
@@ -691,7 +707,13 @@ object HBaseIndexAdapter extends LazyLogging {
 
     override def flush(): Unit = FlushWithLogging.raise(mutators)(BufferedMutatorIsFlushable.arrayIsFlushable)
 
-    override def close(): Unit = CloseWithLogging.raise(mutators)
+    override def close(): Unit = {
+      CloseWithLogging.raise(mutators)
+      CloseWithLogging.raise(pool)
```
elahrvivaz, line 712: These should be either changed to `apply` or invoked in the same call, otherwise if the first call fails the second call won't be invoked. I'm not entirely sure the suggestion below will find the right implicits though...
```suggestion
      CloseWithLogging.raise(mutators, pool)
```

pashashiz: Good catch, will fix

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/data/HBaseIndexAdapter.scala
pashashiz: Yep, will fix.

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/package.scala
```diff
@@ -16,6 +16,7 @@ package object hbase {
     val CoprocessorUrl           : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.url")
     val CoprocessorMaxThreads    : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.maximize.threads", "true")
     val WriteBatchSize           : SystemProperty = SystemProperty("geomesa.hbase.write.batch")
+    val WriteFlushTimeout        : SystemProperty = SystemProperty("geomesa.hbase.write.flushTimeout")
```
elahrvivaz, line 19: could you add this to the docs [here](https://raw.githubusercontent.com/locationtech/geomesa/main/docs/user/hbase/configuration.rst) ?

elahrvivaz:
lgtm - @aheyne do you have any reservations about these changes?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/package.scala
pashashiz: Yep

elahrvivaz:
looks like the Scala 2.11 build is dying on the lambda here, which 2.11 doesn't convert into a Runnable like 2.12 does:
```
/home/runner/work/geomesa/geomesa/geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/concurrent/ExitingExecutorTest.scala:26: overloaded method value submit with alternatives:
Error:    (x$1: Runnable)java.util.concurrent.Future[_] <and>
Error:    [T](x$1: Runnable, x$2: T)java.util.concurrent.Future[T] <and>
Error:    [T](x$1: java.util.concurrent.Callable[T])java.util.concurrent.Future[T]
Error:   cannot be applied to (() => String)
Error:          val second = pool.submit(() => Thread.currentThread().getName).get()
```

aheyne:
> lgtm - @aheyne do you have any reservations about these changes?

Looks good to me. I don't think the auth bits will effect anything we're doing. Would need to test to know for sure but I'm comfortable merging this.

PR #2934, opened 2022-12-15
GEOMESA-3245 Upgrade to GeoTools 28, GeoServer 2.22

elahrvivaz:
* Hints are no longer shared when copying a Query, requiring some refactoring to pass along configured hints
* Remove benchmarks module as it's not used or maintained
* Remove 'caching' data store config as unused


elahrvivaz:
rolling this into the massive 4_0_0 pr

PR #2941, opened 2023-01-12
Documentation updates for 4.0

elahrvivaz:


docs/user/upgrade.rst
```diff
@@ -114,7 +123,7 @@ Dependency Version Upgrades
 ---------------------------
 
 The following high-level dependencies have been upgraded. For a full changelist of all the dependency changes,
-see TODO commit hash link on github
+see the following `gist <https://gist.github.com/elahrvivaz/410370ab29e11ac39618c4310701398b/revisions?diff=split>`__.
```
aheyne, line 126: Did you want your personal fork or the geomesa fork here?

elahrvivaz: it's not possible to have organizational gists in github... this was the best way i could come up with to have a viewable diff :/
possibly some other online tool would work?

elahrvivaz: it's also not possible to just show the diff of a single file between 2 revisions on github, which is what i really want.

elahrvivaz: and this diff is now out of date since i updated orc to 1.8.2 :(

PR #2992, opened 2023-08-30
GEOMESA-3292 Work-around for Accumulo VFS classloading

elahrvivaz:


elahrvivaz:
not sure why the dash job started failing without any dependencies changing - i opened a CQ for the one it's complaining about though, so I wouldn't let it hold up the PR.

elahrvivaz:
going to hold off on merging until we can test more thoroughly

aheyne (approved):

elahrvivaz:
I think we can fix this issue just by upgrading indriya along with geotools 31, so going to close this PR.

PR #2999, opened 2023-09-19
GEOMESA-3299 Deprecate query planner threaded hints

elahrvivaz:
* Threaded hints are not working as intended, and usages should be replaced

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureReaderTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureReaderTest.scala
@@ -98,28 +98,6 @@ class AccumuloFeatureReaderTest extends Specification with TestWithFeatureType {
       events.head.asInstanceOf[QueryEvent].hits mustEqual 100
     }
 
-    "be able to count bin results in stats through geoserver" in {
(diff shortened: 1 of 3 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloFeatureReaderTest.scala
```diff
@@ -98,28 +98,6 @@ class AccumuloFeatureReaderTest extends Specification with TestWithFeatureType {
...
-    "be able to count bin results in stats through geoserver" in {
-      val events = ArrayBuffer.empty[AuditedEvent]
-
-      val query = new Query(sftName, filter)
-      val collection = dataStoreWithAudit(events).getFeatureSource(sftName).getFeatures(query)
-
-      // put the hints in the request after getting the feature collection
-      // to mimic the bin output format workflow
-      val hints = Map(QueryHints.BIN_TRACK -> "name", QueryHints.BIN_BATCH_SIZE -> 10)
-      QueryPlanner.setPerThreadQueryHints(hints.asInstanceOf[Map[AnyRef, AnyRef]])
```
elahrvivaz, line 110: these tests weren't replicating geoserver's method calls, which include a call to `getSchema` before setting the hint. when that call is added, these tests fail. since trying to mutate the request query after setting up the scans, etc is dicey, we won't try to support that, and instead will deprecate the functionality.

PR #3009, opened 2023-11-08
GEOMESA-3310 HBase - Expose additional HBase configurations as system…

aheyne:
… properties.

Expose
- hbase.coprocessor.threads
- hbase.ranges.max-per-extended-scan
- hbase.ranges.max-per-coprocessor-scan

### DIFF
--- a/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/package.scala
+++ b/geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/package.scala
@@ -30,6 +30,9 @@ package object hbase {
     val RemoteDensityProperty      : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.density.enable")
     val RemoteStatsProperty        : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.stats.enable")
     val YieldPartialResultsProperty: SystemProperty = SystemProperty("geomesa.hbase.coprocessor.yield.partial.results")
+    val CoprocessorThreadsProperty : SystemProperty = SystemProperty("hbase.coprocessor.threads", "16")
+    val MaxRangesPerExtendedScanProperty   : SystemProperty = SystemProperty("hbase.ranges.max-per-extended-scan", "100")
+    val MaxRangesPerCoprocessorScanProperty: SystemProperty = SystemProperty("hbase.ranges.max-per-coprocessor-scan", Int.MaxValue.toString)
 
     @deprecated("Use coprocessor url")
     val CoprocessorPath: SystemProperty = SystemProperty("geomesa.hbase.coprocessor.path")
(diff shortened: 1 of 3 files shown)


geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/package.scala
```diff
@@ -30,6 +30,9 @@ package object hbase {
     val RemoteDensityProperty      : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.density.enable")
     val RemoteStatsProperty        : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.stats.enable")
     val YieldPartialResultsProperty: SystemProperty = SystemProperty("geomesa.hbase.coprocessor.yield.partial.results")
+    val CoprocessorThreadsProperty : SystemProperty = SystemProperty("hbase.coprocessor.threads", "16")
+    val MaxRangesPerExtendedScanProperty   : SystemProperty = SystemProperty("hbase.ranges.max-per-extended-scan", "100")
+    val MaxRangesPerCoprocessorScanProperty: SystemProperty = SystemProperty("hbase.ranges.max-per-coprocessor-scan", Int.MaxValue.toString)
```
elahrvivaz, line 35: should we make these match the other ones and start with geomesa?
```suggestion
    val CoprocessorThreadsProperty : SystemProperty = SystemProperty("geomesa.hbase.coprocessor.threads", "16")
    val MaxRangesPerExtendedScanProperty   : SystemProperty = SystemProperty("geomesa.hbase.ranges.max-per-extended-scan", "100")
    val MaxRangesPerCoprocessorScanProperty: SystemProperty = SystemProperty("geomesa.hbase.ranges.max-per-coprocessor-scan", Int.MaxValue.toString)
```

elahrvivaz:
could you also add these to https://github.com/locationtech/geomesa/blob/main/docs/user/hbase/configuration.rst ?

geomesa-hbase/geomesa-hbase-datastore/src/main/scala/org/locationtech/geomesa/hbase/package.scala
aheyne: I committed that but I'm actually thinking of reverting. Reason they don't have the `geomesa.` prefix is because the equivalent data store parameters don't as well. e.g.
https://www.geomesa.org/documentation/stable/user/hbase/usage.html#:~:text=hbase.coprocessor.threads

elahrvivaz: yeah, that's been our pattern - e.g. ds param `accumulo.remote.bin.enable` but sys prop `geomesa.accumulo.remote.bin.enable`. i think the idea is that hbase is probably using sys props prefixed with `hbase.` and we don't want to conflict with them. i.e. from a quick look at the hbase docs, `hbase.master.maintenance_mode` is a sys prop they look for.

elahrvivaz:
@aheyne you happy with this?

elahrvivaz:
I'm ready to merge if you're good with it

aheyne:
I'm happy.

elahrvivaz:
ah, i think you didn't squash commits on the merge - i'm going to do a quick rebase/force push to fix it.

PR #3010, opened 2023-11-14
GEOMESA-3311 Add subsampling to GraduatedQueryGuard

aheyne:
- Configurable subsample percentage based on size
- Configurable subsample attribute name
- Mix duration and percentage/attribute guard modes

### DIFF
--- a/docs/user/datastores/query_interceptor.rst
+++ b/docs/user/datastores/query_interceptor.rst
@@ -141,13 +141,13 @@ Durations can be given in a number of days, hours, or minutes. For example:
           "sftName" = [
             { size = 1,  duration = "60 days" }
             { size = 10, duration = "3 days"  }
-            { duration = "1 day" }
+            {            duration = "1 day"   }
           ]
         }
       }
     }
 
-Percentages can be defined in decimal form. e.g. .1 corresponds to 10%. Any query smaller than the first size will
+Sampling percentages can be defined in decimal form. e.g. .1 corresponds to 10%. Any query smaller than the first size will
 return 100% of the records. For example:
 
 .. code-block:: none
@@ -156,15 +156,15 @@ return 100% of the records. For example:
       guard {
         graduated {
           "sftName" = [
-            { size = 1,  percentage = .8 }
-            { size = 10, percentage = .5  }
-            { percentage = .1 }
+            { size = 1,  sampling-percentage = .8 }
+            { size = 10, sampling-percentage = .5 }
+            {            sampling-percentage = .1 }
           ]
         }
       }
     }
 
-It is also possible to specify the attribute to use for the threading key in subsampling:
+It is also possible to specify the sampling attribute to use for the threading key in subsampling:
 
 .. code-block:: none
 
@@ -172,9 +172,9 @@ It is also possible to specify the attribute to use for the threading key in sub
       guard {
         graduated {
           "sftName" = [
-            { size = 1,  percentage = .8, attribute = "name" }
-            { size = 10, percentage = .5, attribute = "name"  }
-            { percentage = .1, attribute = "name" }
+            { size = 1,  sampling-percentage = .8, sampling-attribute = "name" }
+            { size = 10, sampling-percentage = .5, sampling-attribute = "name" }
+            {            sampling-percentage = .1, sampling-attribute = "name" }
           ]
         }
       }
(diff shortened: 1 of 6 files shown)


geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -110,17 +170,29 @@ object GraduatedQueryGuard extends LazyLogging {
...
+      } else None
+
+      val percentageAttr: Option[String] = if (limitsConfig.hasPath("attribute")) {
+        Some(limitsConfig.getString("attribute"))
+      } else None
+
+      new SizeAndLimits(size, duration, percentage, percentageAttr)
     }
-    evaluateLimits(confs.toSeq)
+    evaluateLimits(confs, sft)
```
elahrvivaz, line 196: looks like you need to add back in the .toSeq to fix the 2.13 build

docs/user/datastores/query_interceptor.rst
```diff
@@ -142,5 +147,59 @@ hours, or minutes. For example:
...
+
+.. code-block:: none
+
+    geomesa {
+      guard {
+        graduated {
+          "sftName" = [
+            { size = 1,  percentage = .8, attribute = "name" }
+            { size = 10, percentage = .5, attribute = "name"  }
+            { percentage = .1, attribute = "name" }
```
elahrvivaz, line 177: maybe we could change these name to `sampling-percentage` and `sampling-attribute` or something similar so it's more descriptive? or alternatively have a nested block like `sampling: { percentage: .1, attribute: name}`

docs/user/datastores/query_interceptor.rst
```diff
@@ -142,5 +147,59 @@ hours, or minutes. For example:
...
+            { duration = "1 day", percentage = .1, attribute = "name" }
+          ]
+        }
+      }
+    }
+
+In the above example, Any query with area less than 1 square degree will return all results and allow for any time
+range (given the time range doesn't trigger a full table scan). A query larger than 1 but less than 10 will block any
+query longer than 3 days AND only 50% of the results will be returned. Finally, any query larger than 10 must be less
+than 1 day in length AND only 10% of the results for any "name" will be returned.
```
elahrvivaz, line 202: i find this behavior a little confusing but i think that's just how the graduated guard worked so no need to change it. thanks for the detailed explanation, it helps a lot!

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
```diff
@@ -32,12 +35,25 @@ class GraduatedQueryGuard extends QueryInterceptor with LazyLogging {
...
-    guardLimits = buildLimits(ConfigFactory.load().getConfigList(s"$ConfigPath.${sft.getTypeName}"))
+    guardLimits = buildLimits(ConfigFactory.load().getConfigList(s"$ConfigPath.${sft.getTypeName}"), sft)
     // this should be ensured during the loading of limits, but just double check so we can use `.last` safely below
     require(guardLimits.nonEmpty)
   }
 
-  override def rewrite(query: Query): Unit = {}
+  override def rewrite(query: Query): Unit = {
+    if (!disabled) {
+      val bounds = query.getFilter.accept(ExtractBoundsFilterVisitor.BOUNDS_VISITOR, null).asInstanceOf[Envelope]
```
elahrvivaz, line 45: do you know if this operates on the default geom by default, or maybe we need to pass that in? technically there could be indices on other geometries, but i guess we don't know what index is being used yet at this point in the processing pipeline... maybe just using "any" geom would be best here (not sure if that's what it's currently doing or not)

geomesa-index-api/src/test/scala/org/locationtech/geomesa/index/TestGeoMesaDataStore.scala
```diff
@@ -9,6 +9,7 @@
 package org.locationtech.geomesa.index
 
 import org.geotools.data.Query
+import org.geotools.util.factory.Hints
```
elahrvivaz, line 12: unused import?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/planning/guard/GraduatedQueryGuard.scala
aheyne: The geotools docs say 
> all the literal geometry instances involved if spatial operations - using geom.getEnvelopeInternal().
which I would interpret to mean it considers all geometries included in the query.

elahrvivaz (approved):

PR #3021, opened 2024-01-08
GEOMESA-3268 Partitioned PostGIS - Check for long type name errors

adeet1:


geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
```diff
@@ -112,6 +112,12 @@ class PartitionedPostgisDialect(store: JDBCDataStore) extends PostGISDialect(sto
   }
 
   override def postCreateTable(schemaName: String, sft: SimpleFeatureType, cx: Connection): Unit = {
+    // Throw an error if the sft name is longer than 31 characters
+    if (sft.getTypeName.length() > 31) {
+      val errorMsg = "Failed to create schema: the name of the SimpleFeatureType exceeds 31 characters"
```
elahrvivaz, line 117:
```suggestion
      val errorMsg = "Can't create schema: type name exceeds max supported length of 31 characters"
```
Makes it a little clearer that longer names aren't supported

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -47,7 +47,7 @@ class PartitionedPostgisDataStoreTest extends Specification with BeforeAfterAll
 
   val hours = 1
   val spec =
-    "name:List[String],props:String:json=true,age:Int,dtg:Date,*geom:Point:srid=4326;" +
+    "name:List[String],props:String:json=true,age:Int,dtg_abcdefghijklmnopqrstuvwxyzabcde:Date,*geom_abcdefghijklmnopqrstuvwxyzabcde:Point:srid=4326;" +
```
elahrvivaz, line 50: could we revert this change? we could add another test with long attribute names, but since it passes we could maybe skip it. I think the failure case would be something like `reallylongcommonprefix__________dtg:Date,reallylongcommonprefix__________geom:Point` which is maybe not something we need to worry about.

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -109,14 +109,37 @@ class PartitionedPostgisDataStoreTest extends Specification with BeforeAfterAll
   }
 
   "PartitionedPostgisDataStore" should {
+    "not work" in {
```
elahrvivaz, line 112:
```suggestion
    "fail with a useful error message if type name is too long" in {
```

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -109,14 +109,37 @@ class PartitionedPostgisDataStoreTest extends Specification with BeforeAfterAll
...
+      ds must not(beNull)
+
+      try {
+        ds must beAnInstanceOf[JDBCDataStore]
+
+        // This sft name exceeds 31 characters, so it should fail
+        val sft = SimpleFeatureTypes.renameSft(this.sft, "abcdefghijklmnopqrstuvwxyzabcde_____")
+        ds.getTypeNames.toSeq must not(contain(sft.getTypeName))
+        ds.createSchema(sft) must throwAn[java.io.IOException].like {
+          case e => e.getMessage mustEqual "Error occurred creating table"
```
elahrvivaz, line 123: could we check the e.getCause, or is there some other way we can verify that our custom error message is showing up here?

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -109,14 +109,37 @@ class PartitionedPostgisDataStoreTest extends Specification with BeforeAfterAll
...
+        }
+      } catch {
+        case NonFatal(e) => logger.error("", e); ko
```
elahrvivaz, line 126:
```suggestion
```
can probably delete this and just let the error bubble up - alternatively there is a `fail(e)` method in specs that will pass in the exception.

geomesa-gt/geomesa-gt-partitioning/src/test/scala/org/locationtech/geomesa/gt/partition/postgis/PartitionedPostgisDataStoreTest.scala
```diff
@@ -109,14 +109,37 @@ class PartitionedPostgisDataStoreTest extends Specification with BeforeAfterAll
...
+
     "work" in {
       val ds = DataStoreFinder.getDataStore(params.asJava)
       ds must not(beNull)
 
       try {
         ds must beAnInstanceOf[JDBCDataStore]
 
-        foreach(Seq("test", "test-dash")) { name =>
+        val sftNames: Seq[String] = Seq("abcdefghijklmnopqrstuvwxyzabcde")
```
elahrvivaz, line 140:
```suggestion
        val sftNames: Seq[String] = Seq("test", "test-thirty-chars-with-dashesz")
```
the dash in the `test-dash` actually was a bug at one point, so i'd like to keep it there.

adeet1: I agree - we can keep the dashes. But I think we'd want to test an sft name with 31 chars (rather than 30), since that is the maximum.

PR #3022, opened 2024-01-10
Kafka Streams integration

a0x8o:
Existing feature types can be updated to use the default Kafka partitioner, which will allow for joins without
re-partitioning. However, note that updates for a given feature may go to the wrong partition, which may result
in older data being returned when GeoMesa is queried. This will generally resolve itself over time as Kafka log
retention policies delete the older data (unless Kafka log compaction is enabled for the topic).


elahrvivaz:
@a0x8o did you mean to open this MR? I'm going to assume it was unintentional, given the conflicts and 10k commits involved. thanks!

PR #3024, opened 2024-01-11
GEOMESA-2159 Check Parquet files for data corruption when closing the ParquetFileSystemWriter

adeet1:


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/package.scala
```diff
@@ -23,6 +24,7 @@ package object common {
 
   val RenderOptions: ConfigRenderOptions = ConfigRenderOptions.concise().setFormatted(true)
   val ParseOptions: ConfigParseOptions = ConfigParseOptions.defaults()
+  val fileValidationEnabled: SystemProperty = SystemProperty("geomesa.fs.validate.file", "false")
```
elahrvivaz, line 27:
```suggestion
  val FileValidationEnabled: SystemProperty = SystemProperty("geomesa.fs.validate.file", "false")
```
convention in scala is to upper-case constants (i.e. vals inside objects)

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -81,6 +83,32 @@ object ParquetFileSystemStorage {
       observer.write(f)
     }
     override def flush(): Unit = observer.flush()
-    override def close(): Unit = CloseQuietly(Seq(writer, observer)).foreach(e => throw e)
+    override def close(): Unit = {
+      println(s"Closing file ${file}")
```
elahrvivaz, line 87:
```suggestion
```
should remove the println here

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -81,6 +83,32 @@ object ParquetFileSystemStorage {
...
+      CloseQuietly(Seq(writer, observer)).foreach(e => throw e)
+      validateParquetFile(file)
+    }
+  }
+
+  def validateParquetFile(file: Path): Unit = {
+    if (!fileValidationEnabled.get.toBoolean) {
```
elahrvivaz, line 94: i'd suggest moving this outside the function - that means we only have to convert it to a boolean once, and also makes the function a little more logical i think, i.e. you'd expect the function to validate a file but it actually won't if the sys prop isn't set

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -81,6 +83,32 @@ object ParquetFileSystemStorage {
...
+    val reader = ParquetReader.builder(new GroupReadSupport(), file).build()
+
+    try {
+      // Read Parquet file content
+      var record = reader.read()
+      while (record != null) {
+        // Process the record
+        record = reader.read()
+      }
+      println(s"${file} is a valid Parquet file")
```
elahrvivaz, line 106:
```suggestion
      logger.debug(s"${file} is a valid Parquet file")
```

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -81,6 +83,32 @@ object ParquetFileSystemStorage {
...
+    } catch {
+      case _: Exception => throw new RuntimeException(s"Unable to validate ${file}: File may be corrupted")
```
elahrvivaz, line 108:
```suggestion
      case e: Exception => throw new RuntimeException(s"Unable to validate ${file}: File may be corrupted", e)
```
this will show the underlying error, which is often useful

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/test/scala/org/locationtech/geomesa/parquet/ParquetReadWriteTest.scala
```diff
@@ -88,6 +90,31 @@ class ParquetReadWriteTest extends Specification with AllExpectations {
...
 
+    "fail if a corrupt parquet file is written" >> {
+      val filePath = new Path(f.toUri)
+      WithClose(SimpleFeatureParquetWriter.builder(filePath, sftConf).build()) { writer =>
+        features.foreach(writer.write)
+      }
+
+      // Corrupt the file by writing an invalid byte somewhere
+      val randomAccessFile = new RandomAccessFile(f.toFile, "rw")
+      println(s"File length: ${randomAccessFile.length()}")
```
elahrvivaz, line 101:
```suggestion
      logger.debug(s"File length: ${randomAccessFile.length()}")
```

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
elahrvivaz: yeah, we're not actually testing that the method gets invoked appropriately in the close() method. but i don't think we can test that case unless we can figure out a good way to corrupt the file after the parquet writer is closed but before the file is validated. since the logic we're not testing is fairly simple, i'd prefer having the method be intuitive, rather than testing the boolean if check.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -81,6 +84,29 @@ object ParquetFileSystemStorage {
...
-    override def close(): Unit = CloseQuietly(Seq(writer, observer)).foreach(e => throw e)
+    override def close(): Unit = {
+      CloseQuietly(Seq(writer, observer)).foreach(e => throw e)
+      if (FileValidationEnabled.get.toBoolean) {
+        validateParquetFile(file, logger)
+      }
+    }
+  }
+
+  def validateParquetFile(file: Path, logger: com.typesafe.scalalogging.Logger): Unit = {
```
elahrvivaz, line 95: instead of passing in the logger, you can just make `object ParquetFileSystemStorage extends LazyLogging {` and that will put a logger in scope for the function.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/test/scala/org/locationtech/geomesa/parquet/ParquetReadWriteTest.scala
```diff
@@ -88,6 +91,31 @@ class ParquetReadWriteTest extends Specification with AllExpectations {
...
+      randomAccessFile.close()
+
+      // Set the system property, validate the file, and then unset it
+      FileValidationEnabled.threadLocalValue.set("true")
+      try {
+        validateParquetFile(filePath, logger) must throwA[RuntimeException].like {
+          case e => e.getMessage mustEqual s"Unable to validate ${filePath}: File may be corrupted"
+        }
+      } finally {
+        FileValidationEnabled.threadLocalValue.remove()
```
elahrvivaz, line 115: we can remove the threadLocalValue bits here, since they're no longer used/needed.

elahrvivaz:
could you also add the new configuration value to the docs [here](https://github.com/locationtech/geomesa/blob/main/docs/user/filesystem/configuration.rst?plain=1#L10)? and just add a sentence indicating it's only implemented for parquet. thanks!

docs/user/filesystem/configuration.rst
```diff
@@ -13,6 +13,13 @@ The following properties control the writing of data files.
 
 .. _fsds_size_threshold_prop:
 
+
+geomesa.fs.validate.file
++++++++++++++++++++++++++
```
elahrvivaz, line 18:
```suggestion
++++++++++++++++++++++++
```
need to line up the underlining for sphinx to be happy

PR #3027, opened 2024-01-16
GEOMESA-3323 Remove do/whiles and early initializers

elahrvivaz:
* These constructs are no longer supported in Scala 3
* See https://docs.scala-lang.org/scala3/guides/migration/incompat-dropped-features.html

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/audit/AccumuloEventWriter.scala
+++ b/geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/audit/AccumuloEventWriter.scala
@@ -49,10 +49,10 @@ class AccumuloEventWriter(connector: AccumuloClient, table: String) extends Runn
     var toMutation = queue.poll()
     if (toMutation != null) {
       val writer = getWriter
-      do {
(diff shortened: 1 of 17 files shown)


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/audit/AccumuloEventWriter.scala
```diff
@@ -49,10 +49,10 @@ class AccumuloEventWriter(connector: AccumuloClient, table: String) extends Runn
     var toMutation = queue.poll()
     if (toMutation != null) {
       val writer = getWriter
-      do {
+      while (toMutation != null && running.get) {
```
aheyne, line 52: Is there ever a case where `running.get` can be false so the mutation sin't added?

elahrvivaz: yeah, that's by design. it only runs every 5 seconds, and will drop anything that was queued up when the system shuts down or the datastore is disposed. might be worth reconsidering, but i think the behavior hasn't changed in this PR.

PR #3028, opened 2024-01-17
GEOMESA-3308 Fix DWithin filter to compute distance between points across the antimeridian

adeet1:


geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastDWithin.scala
```diff
@@ -41,14 +41,24 @@ object FastDWithin {
 
     override def evaluate(obj: AnyRef): Boolean = {
       val geom = exp1.evaluate(obj).asInstanceOf[Geometry]
-      if (geom == null || envelope.distance(geom.getEnvelopeInternal) > maxDegrees) { false } else {
```
elahrvivaz, line 44: i'd like to keep this `envelope.distance(geom.getEnvelopeInternal) > maxDegrees` check, as it's very fast compared to the using the geodetic calculator. i think we can translate the envelope bounds in the same way we do for the geodetic calculations below, and if all of them are > maxDegrees we can return false directly.

geomesa-filter/src/main/scala/org/locationtech/geomesa/filter/expression/FastDWithin.scala
```diff
@@ -41,14 +41,24 @@ object FastDWithin {
...
-          calculator.setStartingGeographicPoint(p1.x, p1.y)
+
+          calculator.setStartingGeographicPoint(p1.x - 360, p1.y)
           calculator.setDestinationGeographicPoint(p2.x, p2.y)
-          calculator.getOrthodromicDistance <= meters
+          val dist1 = calculator.getOrthodromicDistance
+
+          calculator.setStartingGeographicPoint(p1.x + 360, p1.y)
+          calculator.setDestinationGeographicPoint(p2.x, p2.y)
+          val dist2 = calculator.getOrthodromicDistance
```
elahrvivaz, line 59: as another optimization here, we could return early if dist1 <= meters and not calculate dist2. but we also need to calculate dist0, which would be the unmodified point. I'm a little surprised none of the existing unit tests failed wit this change :( before changing anything, could you add a test for `(0 0)` to `(1 0)` (or something similar) and verify that it fails? possibly the calculator is accounting for the anti-meridian wrapping?

adeet1: I added a unit test ([db811e3](https://github.com/locationtech/geomesa/pull/3028/commits/db811e3586285e0446f40b1e330449abeb9c4baa)), but it passes, which I'm not sure is the intended behavior.

However, I think the calculator computes the shortest distance between two points on the Earth's surface, so it's likely that it already accounts for the anti-meridian wrapping.

PR #3035, opened 2024-01-22
GEOMESA-3293 Throw ParameterException if trying to ingest from stdin without specifying a converter

adeet1:


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -246,6 +246,10 @@ object IngestCommand extends LazyLogging {
       }
       val opened = ListBuffer.empty[CloseableIterator[InputStream]]
       def open(): InputStream = {
+        if (file.path.last == '-') {
+          throw new ParameterException("Could not infer type from inputs - please specify a converter/sft or ingest from a file")
+        }
```
elahrvivaz, line 251: I think it would be better to move this up to line 242, and change the check to `(inputs == Input.StdInInputs)`, and change the message to `Cannot infer types from stdin - please specify a converter/sft or ingest from a file`. that way, the if check is a bit clearer, and we don't print out the 'will attempt to detect schema' line

elahrvivaz (approved):

PR #3036, opened 2024-01-24
GEOMESA-3326 Ingest from stdin even if an SFT and converter are not specified

adeet1:
* Cache any bytes that the schema inference functions read from the `stdin` input stream - preventing those functions from consuming those bytes and leaving behind nothing for the actual ingest to read
  * Create a new wrapper class for `StdInHandle`
  * Create a new wrapper class for `CopyingInputStream`, which overrides the `close()` method and prevents `stdin` from being prematurely closed
  * Mark the starting position of the `stdin` stream before starting schema inference, and reset the stream back to that point after the inference is complete, so that the ingest can re-read those bytes
* Change unit tests in the `IngestCommandTest` class to use buffered input streams, replicating how the CLI tools process command-line input

Minor changes:
* Flip the lat/lon columns - this fixes a bug where a SimpleFeature wasn't getting ingested due to a latitude < -90
* Add a comment about an unused parameter

### DIFF
--- a/geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/IngestCommandTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/IngestCommandTest.scala
@@ -16,19 +16,15 @@ import org.locationtech.geomesa.accumulo.MiniCluster
 import org.locationtech.geomesa.accumulo.tools.{AccumuloDataStoreCommand, AccumuloRunner}
 import org.locationtech.geomesa.utils.collection.SelfClosingIterator
 import org.locationtech.geomesa.utils.io.WithClose
-import org.locationtech.geomesa.utils.io.fs.LocalDelegate
+import org.locationtech.geomesa.utils.io.fs.LocalDelegate.StdInHandle
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
 
-import java.io.{BufferedInputStream, ByteArrayInputStream, File, IOException, InputStream}
+import java.io.{BufferedInputStream, ByteArrayInputStream, File}
 import java.util.concurrent.atomic.AtomicInteger
-import scala.util.Try
 
 @RunWith(classOf[JUnitRunner])
 class IngestCommandTest extends Specification {
-
-  // This prevents different values of System.in across different tests from interfering with each other
-  sequential
 
   private val sftCounter = new AtomicInteger(0)
 
@@ -86,7 +82,7 @@ class IngestCommandTest extends Specification {
       }
     }
 
-    "fail to ingest from stdin if no sft, converter, and sft name are specified" in {
+    "require sft or sft name to be specified during ingest from stdin with type inference" in {
       val dataFile = WithClose(getClass.getClassLoader.getResourceAsStream("examples/example1.csv")) { in =>
         IOUtils.toByteArray(in)
       }
@@ -96,56 +92,15 @@ class IngestCommandTest extends Specification {
 
       val command = AccumuloRunner.parseCommand(args).asInstanceOf[AccumuloDataStoreCommand]
 
-      val in = System.in
-      in.synchronized {
-        try {
-          System.setIn(input)
-          command.execute() must throwA[ParameterException].like {
-            case e => e.getMessage mustEqual "SimpleFeatureType name not specified. Please ensure the -f or --feature-name flag is set."
-          }
-        } finally {
-          System.setIn(in)
+      StdInHandle.SystemIns.set(input)
+      try {
+        command.execute() must throwA[ParameterException].like { case e =>
+          e.getMessage mustEqual
+              "SimpleFeatureType name not specified. Please ensure the -f or --feature-name flag is set."
         }
+      } finally {
+        StdInHandle.SystemIns.remove()
       }
-    }
-
-    "cache bytes when reading from stdin" in {
-      @throws[IOException]
-      def readNBytes(stream: InputStream, n: Int): Array[Byte] = {
-        val buffer = new Array[Byte](n)
-        var totalBytesRead = 0
-
-        while (totalBytesRead < n) {
-          val bytesRead = stream.read(buffer, totalBytesRead, n - totalBytesRead)
-          if (bytesRead == -1) {
-            throw new IOException("End of stream reached before reading 'n' bytes.")
-          }
-          totalBytesRead += bytesRead
-        }
-
-        buffer
-      }
-
-      val dataFile = WithClose(getClass.getClassLoader.getResourceAsStream("examples/example1.csv")) { in =>
-        IOUtils.toByteArray(in)
-      }
-
-      // Feed all the bytes to stdin
-      val input = new BufferedInputStream(new ByteArrayInputStream(dataFile))
-      System.setIn(input)
-      val inputSize = input.available()
-
-      // Read some bytes
-      val handle = LocalDelegate.CachingStdInHandle.available().getOrElse(throw new RuntimeException("StdInHandle not available"))
-      val (_, stream) = handle.open.next()
-      val numBytesToRead = 3
-      readNBytes(stream, numBytesToRead)
-
-      // Open the handle again, which will create a new stream that has the cached bytes in it
-      handle.open.next()
-      input.available() mustEqual inputSize - numBytesToRead
-      handle.length - Try(System.in.available().toLong).getOrElse(0L) mustEqual numBytesToRead
-      handle.length mustEqual dataFile.length
     }
 
     "ingest from stdin if no sft and converter are specified" in {
@@ -160,14 +115,11 @@ class IngestCommandTest extends Specification {
 
       val command = AccumuloRunner.parseCommand(args).asInstanceOf[AccumuloDataStoreCommand]
 
-      val in = System.in
-      in.synchronized {
-        try {
-          System.setIn(input)
-          command.execute()
-        } finally {
-          System.setIn(in)
-        }
+      StdInHandle.SystemIns.set(input)
+      try {
+        command.execute()
+      } finally {
+        StdInHandle.SystemIns.remove()
       }
 
       command.withDataStore { ds =>
@@ -192,14 +144,11 @@ class IngestCommandTest extends Specification {
 
       val command = AccumuloRunner.parseCommand(args).asInstanceOf[AccumuloDataStoreCommand]
 
-      val in = System.in
-      in.synchronized {
-        try {
-          System.setIn(input)
-          command.execute()
-        } finally {
-          System.setIn(in)
-        }
+      StdInHandle.SystemIns.set(input)
+      try {
+        command.execute()
+      } finally {
+        StdInHandle.SystemIns.remove()
       }
 
       command.withDataStore { ds =>
@@ -224,16 +173,13 @@ class IngestCommandTest extends Specification {
 
       val command = AccumuloRunner.parseCommand(args).asInstanceOf[AccumuloDataStoreCommand]
 
-      val in = System.in
-      in.synchronized {
-        try {
-          System.setIn(input)
-          command.execute() must throwA[ParameterException].like {
-            case e => e.getMessage mustEqual "SimpleFeatureType name and/or specification argument is required"
-          }
-        } finally {
-          System.setIn(in)
+      StdInHandle.SystemIns.set(input)
+      try {
+        command.execute() must throwA[ParameterException].like {
+          case e => e.getMessage mustEqual "SimpleFeatureType name and/or specification argument is required"
         }
+      } finally {
+          StdInHandle.SystemIns.remove()
       }
     }
 
--- a/geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/IngestCommandTest.scala
+++ b/geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/IngestCommandTest.scala
@@ -16,14 +16,19 @@ import org.locationtech.geomesa.accumulo.MiniCluster
 import org.locationtech.geomesa.accumulo.tools.{AccumuloDataStoreCommand, AccumuloRunner}
 import org.locationtech.geomesa.utils.collection.SelfClosingIterator
 import org.locationtech.geomesa.utils.io.WithClose
+import org.locationtech.geomesa.utils.io.fs.LocalDelegate
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
 
-import java.io.{ByteArrayInputStream, File}
+import java.io.{BufferedInputStream, ByteArrayInputStream, File, IOException, InputStream}
 import java.util.concurrent.atomic.AtomicInteger
+import scala.util.Try
 
 @RunWith(classOf[JUnitRunner])
 class IngestCommandTest extends Specification {
+
+  // This prevents different values of System.in across different tests from interfering with each other
+  sequential
 
   private val sftCounter = new AtomicInteger(0)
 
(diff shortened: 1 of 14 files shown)


geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,89 @@ object IngestCommand extends LazyLogging {
...
-        throw new ParameterException("Parameter <files> did not evaluate to anything that could be read")
+      var open: () => InputStream = null
+      var file: FileHandle = null
+
+      // try to infer the schema from stdin
+      if (inputs == Inputs.StdInInputs) {
+        Command.user.info("No converter defined - will attempt to detect schema from stdin")
+
+        // Cache the bytes from the input stream, because we'll need them again for the ingest
+        val inputBytes = System.in.readAllBytes()
```
elahrvivaz, line 251: I'm a little wary about reading in all available bytes, as it might blow out memory. All the processing during ingest is streaming, so we don't generally hold much data in memory. i think ideally we could read in some fixed amount of bytes, run type inference on that, then for the actual ingest create an input stream that first returns the cached bytes, then starts reading from stdin again.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,89 @@ object IngestCommand extends LazyLogging {
...
+        def openStdin(): InputStream = {
+          try {
+            // This prevents the schema and type inference code from consuming all the bytes from stdin and leaving
+            // no bytes for the actual ingest to read
+            System.setIn(new ByteArrayInputStream(inputBytes))
+
+            // Return stdin
+            System.in
+          } catch {
+            case _: Exception => throw new ParameterException("Unable to read from stdin")
```
elahrvivaz, line 263:
```suggestion
            case e: Exception => throw new RuntimeException("Unable to read from stdin", e)
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,89 @@ object IngestCommand extends LazyLogging {
...
-          case Some(fmt) => TypeAwareInference.infer(fmt, open, Option(sft), Option(file.path))
+          case None => SimpleFeatureConverter.infer(open, Option(sft))
+          case Some(fmt) => TypeAwareInference.infer(fmt, open, Option(sft), None)
         }
         opt.getOrElse {
           throw new ParameterException("Could not determine converter from inputs - please specify a converter")
         }
       } finally {
-        CloseWithLogging(opened)
+        CloseWithLogging(open())
```
elahrvivaz, line 299: I think this line needs to be reverted, and the `opened` array needs to be moved up out of the if block so it stays in scope. The way this is currently working is that you're opening up new InputStreams and then closing them, while the input streams opened by type inference do not get closed. Normally the inferencer code that opens the stream would be responsible for closing it, but we need this extra layer of cleanup b/c some files can evaluate to multiple inputs (i.e. a zip file), and the inferencer isn't aware of the outer file that needs to be closed.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,89 @@ object IngestCommand extends LazyLogging {
...
+
+        open = openFile
       }
+
       val (inferredSft, inferredConverter) = try {
         val opt = format match {
-          case None      => SimpleFeatureConverter.infer(open, Option(sft), Option(file.path))
-          case Some(fmt) => TypeAwareInference.infer(fmt, open, Option(sft), Option(file.path))
+          case None => SimpleFeatureConverter.infer(open, Option(sft))
+          case Some(fmt) => TypeAwareInference.infer(fmt, open, Option(sft), None)
```
elahrvivaz, line 293: I think we need to keep the file.path here, probably filtered appropriately for stdin. some converters rely on the file path as well as the input stream.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,89 @@ object IngestCommand extends LazyLogging {
...
-          val fileName = Option(FilenameUtils.getBaseName(file.path))
-          val base = fileName.map(_.trim.replaceAll("[^A-Za-z0-9]+", "_")).filterNot(_.isEmpty).getOrElse("geomesa")
-          var name = base
-          var i = 0
-          while (existing.contains(name)) {
-            name = s"${base}_$i"
-            i += 1
+
+        val typeName = if (inputs == Inputs.StdInInputs) {
+          Option(params.featureName).getOrElse("default")
```
elahrvivaz, line 308: I think it's probably better to make them specify the feature type name here, and not use `default`

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,89 @@ object IngestCommand extends LazyLogging {
...
+        } else {
+          Option(params.featureName).getOrElse {
+            val existing = ds.toSet[DataStore].flatMap(_.getTypeNames)
+            val fileName = Option(FilenameUtils.getBaseName(file.path))
+            val base = fileName.map(_.trim.replaceAll("[^A-Za-z0-9]+", "_")).filterNot(_.isEmpty).getOrElse("geomesa")
```
elahrvivaz, line 313: we should probably get rid of the `getOrElse("geomesa")` here and require that they specify a type name, for the same reasons mentioned above.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,105 @@ object IngestCommand extends LazyLogging {
...
+              SimpleFeatureConverter.infer(open, Option(sft))
+            } else {
+              SimpleFeatureConverter.infer(open, Option(sft), Option(file.path))
+            }
+
+          case Some(fmt) => if (inputs == Inputs.StdInInputs) {
+            TypeAwareInference.infer(fmt, open, Option(sft), None)
+          } else {
+            TypeAwareInference.infer(fmt, open, Option(sft), Option(file.path))
+          }
```
adeet1, line 305: @elahrvivaz I made this change per your request, but the new unit tests will be unaffected because this code indirectly calls the `DelimitedTextConverterFactory.infer(is: InputStream, sft: Option[SimpleFeatureType], path: Option[String])` function, whose `path` argument is unused.

Are you referring to the `infer` functions from other converter classes (such as the `ParquetConverterFactory`) which use this `path` argument? In that case, do you think we need to add more unit tests to ensure this new logic works with other converters?

elahrvivaz: right, I was referring to the other converter factory types. if you'd like to add tests for those, then feel free, but I won't hold up this PR for it. the type inference is currently only implemented for a subset of the converters, so the path may actually be unused in all of them, but we should still pass it through so that it's there if/when we implement additional type inferences that need it.

adeet1: Thank you for clarifying the details. Yes I agree - additional unit tests likely aren't a priority for this ticket, but I've made a new separate ticket for that work in case anyone wants to look into it: https://geomesa.atlassian.net/browse/GEOMESA-3328

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
adeet1: That makes sense, but how would we determine the number of bytes to read in and do type inference on?

In order to infer the SFT, I think we'd need to read in one SimpleFeature from stdin (at minimum), but we wouldn't know the size of that SimpleFeature at compile time.

aheyne: `AbstractConverterFactory.inferSampleSize` provides a configuration for how many lines should be considered in type inference. 

I would suggest something like defining an input iterator near where your `var open` is. You can duplicate the input stream and sample one duplicate for use in the inferencer and save one for ingest later. 
```
val (input, sample) = Iterator.continually(io.StdIn.readLine()).duplicate match { 
  case (i, s) => (i, s.take(AbstractConverterFactory.inferSampleSize).toList)
}
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,47 +238,117 @@ object IngestCommand extends LazyLogging {
...
+
+            // Return this stream
+            opened.flatten.head
+          } catch {
+            case e: Exception => throw new RuntimeException("Unable to read from stdin", e)
+          }
+        }
+
+        open = openStdin
+      }
```
adeet1, line 283: > `AbstractConverterFactory.inferSampleSize` provides a configuration for how many lines should be considered in type inference.
> 
> I would suggest something like defining an input iterator near where your `var open` is. You can duplicate the input stream and sample one duplicate for use in the inferencer and save one for ingest later.
> 
> ```
> val (input, sample) = Iterator.continually(io.StdIn.readLine()).duplicate match { 
>   case (i, s) => (i, s.take(AbstractConverterFactory.inferSampleSize).toList)
> }
> ```

@aheyne I made a change based on your suggestion, but somehow it's unable to parse the SimpleFeature:

```
java.lang.RuntimeException: Unable to read from stdin

	at org.locationtech.geomesa.tools.ingest.IngestCommand$.openStdin$1(IngestCommand.scala:278)
	at org.locationtech.geomesa.tools.ingest.IngestCommand$.$anonfun$getSftAndConverter$5(IngestCommand.scala:282)
	at org.locationtech.geomesa.convert2.SimpleFeatureConverter$.$anonfun$infer$2(SimpleFeatureConverter.scala:136)
	at org.locationtech.geomesa.convert2.SimpleFeatureConverter$.$anonfun$infer$1(SimpleFeatureConverter.scala:136)
	at org.locationtech.geomesa.convert2.SimpleFeatureConverter$.infer(SimpleFeatureConverter.scala:135)
	at org.locationtech.geomesa.tools.ingest.IngestCommand$.$anonfun$getSftAndConverter$1(IngestCommand.scala:307)
	at org.locationtech.geomesa.tools.ingest.IngestCommand$.getSftAndConverter(IngestCommand.scala:233)
	at org.locationtech.geomesa.tools.ingest.IngestCommand.$anonfun$execute$6(IngestCommand.scala:107)
	at org.locationtech.geomesa.tools.ingest.IngestCommand.$anonfun$execute$6$adapted(IngestCommand.scala:105)
	at org.locationtech.geomesa.tools.package$DataStoreCommand.withDataStore(package.scala:68)
	at org.locationtech.geomesa.tools.package$DataStoreCommand.withDataStore$(package.scala:66)
	at org.locationtech.geomesa.accumulo.tools.ingest.AccumuloIngestCommand.withDataStore(AccumuloIngestCommand.scala:20)
	at org.locationtech.geomesa.tools.ingest.IngestCommand.execute(IngestCommand.scala:105)
	at org.locationtech.geomesa.tools.ingest.IngestCommand.execute$(IngestCommand.scala:58)
	at org.locationtech.geomesa.accumulo.tools.ingest.AccumuloIngestCommand.execute(AccumuloIngestCommand.scala:25)
	at org.locationtech.geomesa.accumulo.tools.ingest.IngestCommandTest.$anonfun$new$20(IngestCommandTest.scala:123)
Caused by: java.lang.NumberFormatException: For input string: "26236,Hermione,25,2015-06-07,"Edward, Bill, Harry",40.232,-53.2356"
	at org.locationtech.geomesa.tools.ingest.IngestCommand$.$anonfun$getSftAndConverter$3(IngestCommand.scala:260)
	at org.locationtech.geomesa.tools.ingest.IngestCommand$.openStdin$1(IngestCommand.scala:271)
	... 15 more
```

aheyne: I think it's probably because you're calling .toByte on 260

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: yeah, delegating to inferSampleSize would be ideal since it's already there and documented. you might also be able to create a new wrapper class that implements `() => InputStream` (which is the factory method used by the type inference) that could handle caching the bytes as they get read, then re-setting back to return the cached bytes.

elahrvivaz:
looks like there are some build errors that need fixing:
> /home/runner/work/geomesa/geomesa/geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/IngestCommandTest.scala:120: value readNBytes is not a member of java.io.InputStream
Error:        stream.readNBytes(3)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/ConvertCommand.scala
```diff
@@ -52,13 +53,13 @@ class ConvertCommand extends Command with MethodProfiling with LazyLogging {
...
+    val format = IngestCommand.getDataFormat(params, inputs.paths)
 
     // use .get to re-throw the exception if we fail
-    IngestCommand.getSftAndConverter(params, inputs.toSeq, format, None).get.flatMap { case (sft: SimpleFeatureType, config: com.typesafe.config.Config) =>
-      val files = if (inputs.isEmpty) { StdInHandle.available().iterator } else {
-        inputs.iterator.flatMap(PathUtils.interpretPath)
+    IngestCommand.getSftAndConverter(params, inputs, format, None).get.flatMap { case (sft: SimpleFeatureType, config: com.typesafe.config.Config) =>
+      val files = if (inputs.paths.isEmpty) { StdInHandle.available().iterator } else {
+        inputs.paths.iterator.flatMap(PathUtils.interpretPath)
       }
```
elahrvivaz, line 63: I think this is duplicating logic in `Inputs` now, and can just be `inputs.handles.iterator`
```suggestion
      val files = inputs.handles.iterator
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,52 +238,108 @@ object IngestCommand extends LazyLogging {
...
+
+          case Some(fmt) => if (inputs.paths == Inputs.StdInInputs) {
+            TypeAwareInference.infer(fmt, open, Option(sft), None)
+          } else {
+            TypeAwareInference.infer(fmt, open, Option(sft), Option(file.path))
+          }
         }
+
+        // Make stdin point back to the beginning
+        open()
```
elahrvivaz, line 307: i think this call shouldn't be necessary, as open will be called during the ingest, which would reset the stream at that point?

adeet1: At first, I didn't have this line, and it was causing the ingest to hang, waiting for input forever. It looks like the ingest is reading from the current position in the stream rather than from the beginning of the stream.

Adding this line had fixed the issue for me.

elahrvivaz:
@adeet1 have you tested this through the CLI tools themselves (vs in a unit test)? if not, could you try that out? [using redis](https://www.geomesa.org/documentation/stable/user/redis/install.html#setting-up-the-redis-command-line-tools) might be an easy way, as you can run a redis instance just with `docker run --rm -p 6379:6379 redis:latest`

elahrvivaz:
also, looks like some conflicts from main have crept in... could you merge in main and fix the conflicts? thanks!

adeet1:
@elahrvivaz I've pushed some commits that should now make the CLI tools do the schema inference and run the ingest successfully.

However, the `"cache bytes when reading from stdin"` unit test is passing, with and without these fixes. I think the next step will be to change the unit test such that it fails without the fixes.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -238,52 +238,111 @@ object IngestCommand extends LazyLogging {
...
+      var open: () => InputStream = null
+      var file: FileHandle = null
+      var opened = ListBuffer.empty[CloseableIterator[InputStream]]
+
+      // try to infer the schema from stdin
+      if (inputs.paths == Inputs.StdInInputs) {
+        Command.user.info("No converter defined - will attempt to detect schema from stdin")
+
+        // Mark the beginning of the stdin stream so we can return to it later
+        System.in.mark(8192)
```
elahrvivaz, line 252: seems like we shouldn't need to do this, and it's actually dangerous b/c according to the javadocs if more than 8192 bytes are read then it might not support reset()-ing back to the start.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: i moved the available checks up to the top of this method, instead of down in the LocalIngestWorker - that way it will validate the available input before any caching.

geomesa-accumulo/geomesa-accumulo-tools/src/test/scala/org/locationtech/geomesa/accumulo/tools/ingest/IngestCommandTest.scala
```diff
@@ -16,20 +16,16 @@ import org.locationtech.geomesa.accumulo.MiniCluster
...
+import java.io.{BufferedInputStream, ByteArrayInputStream, File}
 import java.util.concurrent.atomic.AtomicInteger
-import scala.util.Try
 
 @RunWith(classOf[JUnitRunner])
 class IngestCommandTest extends Specification {
 
-  // This prevents different values of System.in across different tests from interfering with each other
-  sequential
-
```
adeet1: Is there a reason why we don't want the unit tests running sequentially?

My experience is that running them in parallel causes the `stdin` ingest tests to fail randomly and unpredictably, whereas running them sequentially will produce stable results.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -231,107 +238,63 @@ object IngestCommand extends LazyLogging {
...
-
-          case Some(fmt) => if (inputs.paths == Inputs.StdInInputs) {
-            TypeAwareInference.infer(fmt, open, Option(sft), None)
-          } else {
-            TypeAwareInference.infer(fmt, open, Option(sft), Option(file.path))
-          }
+        if (is == null) {
+          throw new ParameterException(errorMsg)
+        } else {
+          is
```
adeet1, line 267: I'm thinking that we should change this to something like:
```scala
is.getOrElse({
    throw new ParameterException(errorMsg)
})
```

and change the `open()` function's return type to `Option[InputStream]` to avoid having to return `null` and possibly risking a `NullPointerException`.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
```diff
@@ -231,107 +238,63 @@ object IngestCommand extends LazyLogging {
...
-            val (_, stream) = handle.open.next()
-            opened += CloseableIterator.single(stream)
-            stream
-          } catch {
-            case e: Exception => throw new RuntimeException("Unable to read from stdin", e)
-          }
-        }
-
-        open = openStdin
+    if (converter == null && (inputs.stdin || inputs.paths.nonEmpty)) {
```
adeet1: Is this `if` statement refactoring mainly for the purpose of improving runtime by reducing the amount of branching in the code? Or is it just for readability?

geomesa-utils-parent/geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/io/fs/LocalFileHandleTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    "cache bytes when reading from stdin" in {
+      @throws[IOException]
+      def readNBytes(stream: InputStream, n: Int): Array[Byte] = {
+        val buffer = new Array[Byte](n)
+        var totalBytesRead = 0
+
+        while (totalBytesRead < n) {
+          val bytesRead = stream.read(buffer, totalBytesRead, n - totalBytesRead)
+          if (bytesRead == -1) {
+            throw new IOException("End of stream reached before reading 'n' bytes.")
```
adeet1, line 34: It might be useful to print out the value of `n`:

```scala
throw new IOException(s"End of stream reached before reading ${n} bytes.")
```

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: open has to return an Iterator, not an option, in order for us to pass it to the type inference method. i split up the logic this way so that we could distinguish between exceptions opening the stream (runtime exceptions) vs exceptions where there was no input (parameter exception). another option would have been to throw an exception `if (!streams.hasNext)`, then catch and rethrow the parameter exception, but that seemed more complicated to me. we could make `is` be an Option inside the function, and then use the `getOrElse` as you outlined, but since it's a small localized function the null is easy to reason about, and doesn't leak out, so I'm ok with it. if you feel strongly about it, feel free to change it, though.

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
elahrvivaz: i don't think it's likely to improve runtime, but it avoids a lot of duplicate/very similar code between the 2 branches, and makes it easier to understand (at least IMO)

geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/ingest/IngestCommand.scala
adeet1: Makes sense. I don't feel too strongly about it, so we can keep it as is!

geomesa-utils-parent/geomesa-utils/src/test/scala/org/locationtech/geomesa/utils/io/fs/LocalFileHandleTest.scala
adeet1: https://github.com/locationtech/geomesa/pull/3036/commits/10b6ed596f780781077cad0ff9c47210886abc84

PR #3038, opened 2024-01-29
GEOMESA-3329 FSDS - Receipt time partition scheme

elahrvivaz:
* Also fixes the 'pattern' for date-time schemes, which is only used for display purposes.

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/ReceiptTimeScheme.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/ReceiptTimeScheme.scala
@@ -1,0 +1,294 @@
+/***********************************************************************
+ * Copyright (c) 2013-2024 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.fs.storage.common.partitions
+
+import org.geotools.filter.visitor.DuplicatingFilterVisitor
+import org.geotools.temporal.`object`.{DefaultInstant, DefaultPeriod, DefaultPosition}
+import org.locationtech.geomesa.fs.storage.api.PartitionScheme.SimplifiedFilter
+import org.locationtech.geomesa.fs.storage.api.{NamedOptions, PartitionScheme, PartitionSchemeFactory}
+import org.locationtech.geomesa.fs.storage.common.partitions.ReceiptTimeScheme.BufferingFilterVisitor
+import org.locationtech.geomesa.utils.geotools.converters.FastConverter
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+import org.opengis.filter._
+import org.opengis.filter.expression.{Expression, Literal, PropertyName}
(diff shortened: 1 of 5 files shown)


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/ReceiptTimeScheme.scala
```diff
@@ -0,0 +1,294 @@
...
+  override val pattern: String = delegate.pattern
+
+  override def getPartitionName(feature: SimpleFeature): String = delegate.getPartitionName(feature)
+
+  override def getSimplifiedFilters(filter: Filter, partition: Option[String]): Option[Seq[SimplifiedFilter]] = {
+    delegate.getSimplifiedFilters(buffered(filter), partition).map { filters =>
+      // always use the full filter since our dates are not guaranteed to match the partition bounds
+      filters.map(f => f.copy(filter = filter))
+    }
+  }
```
adeet1, line 48: Is there a particular reason we're using a delegation pattern here, instead of extending the `DateTimeScheme` class? I think we can just inherit `depth`, `pattern`, and `getPartitionName` directly from the `DateTimeScheme` class instead of having to override them.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/ReceiptTimeScheme.scala
```diff
@@ -0,0 +1,294 @@
...
+object ReceiptTimeScheme {
+
+  val Name = "receipt-time"
+
+  object Config {
+    val DateTimeSchemaOpt: String = "datetime-scheme"
+    val BufferOpt        : String = "buffer"
+  }
+
+  class ReceiptTimePartitionSchemeFactory extends PartitionSchemeFactory {
```
adeet1, line 69: Perhaps we should extend the `DateTimePartitionSchemeFactory` class instead, since the receipt-time partitioning scheme is a special case of the date-time scheme?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/ReceiptTimeScheme.scala
elahrvivaz: hmm yeah, no particular reason - i was thinking of this as a separate thing from the date time scheme, rather than an extension, but in practice it would be equivalent. I guess i mainly didn't want to deal with the constructor args required for the date time scheme, i.e. it would need to change to `class ReceiptTimeScheme(buffer: Duration, formatter: DateTimeFormatter, stepUnit: ChronoUnit, step: Int, dtg: String, dtgIndex: Int) extends DateTimeScheme(formatter, stepUnit, step, dtg, dtgIndex)`

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/ReceiptTimeScheme.scala
elahrvivaz: yeah, we could - i think it would mostly just change the call from `PartitionSchemeFactory.load` to `super.load`, but it seems like we would also want to make the change you suggested above in making the scheme extend date time scheme, to be consistent. i think that would have been a valid way to go, but i don't see a compelling reason to change it now.

PR #3040, opened 2024-02-06
Support NiFi expressions for LambdaDataStoreParams

fdfea:


fdfea:
@elahrvivaz We'd also like this to go in the next 4.x release

elahrvivaz (approved):

PR #3041, opened 2024-02-06
GEOMESA-3036 Throw a useful error if attempting to parse non-GeoJSON features

adeet1:
* Throw an error during schema inference if we were able to parse json elements but couldn't convert them to GeoJSON
* Implement unit tests for ingesting/converting non-GeoJSON features

### DIFF
--- a/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
@@ -42,6 +42,7 @@ class JsonConverterFactory extends AbstractConverterFactory[JsonConverter, JsonC
       val reader = new JsonReader(new InputStreamReader(is, StandardCharsets.UTF_8))
       reader.setLenient(true)
 
+      // TODO: We shouldn't use deprecated methods
       val elements = {
         val iter: Iterator[JsonElement] = new Iterator[JsonElement] {
           private val parser = new JsonParser
@@ -54,6 +55,11 @@ class JsonConverterFactory extends AbstractConverterFactory[JsonConverter, JsonC
       val geojson = elements.collect {
         case el if JsonConverter.isFeature(el) => JsonConverter.parseFeature(el)
         case el if JsonConverter.isFeatureCollection(el) => JsonConverter.parseFeatureCollection(el)
+      }
+
+      // If we were able to parse json elements but couldn't convert them to geojson
+      if (geojson.isEmpty) {
+        throw new IllegalArgumentException("The provided JSON is not valid GeoJSON")
       }
 
       geojson.headOption.map { head =>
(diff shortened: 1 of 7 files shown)


geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
```diff
@@ -42,6 +42,7 @@ class JsonConverterFactory extends AbstractConverterFactory[JsonConverter, JsonC
       val reader = new JsonReader(new InputStreamReader(is, StandardCharsets.UTF_8))
       reader.setLenient(true)
 
+      // TODO: We shouldn't use deprecated methods
```
adeet1, line 45: The instantiation of `JsonParser` and the call `parser.parse` are deprecated.

Will replacing these with newer, current methods cause any backwards compatibility issues?

elahrvivaz: hmm not sure... we could probably change it for the 5.0 release

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
```diff
@@ -56,6 +57,11 @@ class JsonConverterFactory extends AbstractConverterFactory[JsonConverter, JsonC
         case el if JsonConverter.isFeatureCollection(el) => JsonConverter.parseFeatureCollection(el)
       }
 
+      // If we were able to parse json elements but couldn't convert them to geojson
+      if (geojson.isEmpty) {
+        throw new IllegalArgumentException("The provided JSON is not valid GeoJSON")
```
elahrvivaz, line 62: I think we probably shouldn't throw an exception here, as this kind of changes the API contract of the `infer` method. Generally in scala you want the return type to encapsulate things like this, as it makes it easier to reason about. I.e., looking at the `infer` method signature, you might not expect it to throw exceptions. Instead, you would change it to a `Try[..]` type, which can return a success or failure (exception), which makes the contract of the method clear, and forces callers to handle the error case.

I think the infer method should probably be changed to return different states, something like:
a) this isn't something i can parse at all (i.e. it's not json)
b) i can parse this, but i can't infer any schema out of it (i.e. it's not geojson)
c) i can parse this and i've inferred this schema/converter

Even then, things are a bit complicated, b/c for example a CSV converter would be able to parse json as a CSV file (since it has commas), but that's not something we would want to highlight. In the CLI tools we sometimes have the file name, but file extensions don't always match the file contents, so that's not a reliable way to determine which converter is most correct. We should probably try to rank them based on file extension and other clues, and then display the results of all the converters, i.e. something like:

```
we couldn't infer anything from your input. The following converters were tried:
JSON - error - only geojson is supported for type inference
CSV - error - could not determine geometry from the provided fields
AVRO - error - not an avro file
... etc
```

elahrvivaz: So, I'm inclined to not merge this, and to pursue that larger refactoring if you want to keep working on it

PR #3043, opened 2024-02-26
GEOMESA-3333 Update to Accumulo 2.1

elahrvivaz:
* Use testcontainers for Accumulo tests
* Move iterators into separate module to allow the distributed runtime to be used in tests
* Upgrade json-path to 2.9.0


geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
```diff
@@ -348,37 +347,26 @@ class AccumuloDataStoreTest extends Specification with TestWithMultipleSfts {
...
-        val modified = SimpleFeatureTypes.createType(sftName, newSchema)
+      def modify(spec: String): Unit = {
+        val modified = SimpleFeatureTypes.createType(sftName, spec)
         modified.getUserData.putAll(ds.getSchema(sftName).getUserData)
         ds.updateSchema(sftName, modified)
-        val retrievedSchema = SimpleFeatureTypes.encodeType(ds.getSchema(sftName))
-        retrievedSchema mustEqual newSchema
       }
+
+      // "prevent changing default geometry" >> {
```
aheyne, line 356: What's the reason for combining these tests?

aheyne (approved):

geomesa-accumulo/geomesa-accumulo-datastore/src/test/scala/org/locationtech/geomesa/accumulo/data/AccumuloDataStoreTest.scala
elahrvivaz: making it so we can remove the `sequential` flag in the tests, which makes them run faster. it was only needed for this small block of tests so i just put them in one block so that they'd run sequentially but all the others could run in parallel

PR #3045, opened 2024-02-29
GEOMESA-3334 Accumulo - Support atomic writes

elahrvivaz:
* Use `org.locationtech.geomesa.index.geotools.AtomicWriteTransaction`


geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/writer/tx/AccumuloAtomicIndexWriter.scala
```diff
@@ -0,0 +1,208 @@
...
+          errors += ConditionalWriteStatus(indices(i), m.name, status)
+        }
+      }
+      i += 1
+    }
+    if (errors.nonEmpty) {
+      // revert any writes we made
+      successes.foreach { case (i, m) =>
+        // note: if these are rejected, some other update has come through and we don't want to overwrite it
+        writers(i).write(m.invert())
```
aheyne, line 139: Does the order in which we revert matter?

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/writer/tx/AccumuloAtomicIndexWriter.scala
```diff
@@ -0,0 +1,208 @@
...
+      val key = new Key(mutation.row)
+      scanner.setRange(new org.apache.accumulo.core.data.Range(key, true, key.followingKey(PartialKey.ROW), false))
+      val found = scanner.iterator().asScala.toList
+      mutation.apply().getUpdates.asScala.forall { update =>
+        if (update.isDeleted) {
+          found.forall { entry =>
+            !compare(entry.getKey.getColumnFamily, update.getColumnFamily) ||
+                !compare(entry.getKey.getColumnQualifier, update.getColumnQualifier) ||
+                !compare(entry.getKey.getColumnVisibility, update.getColumnVisibility) ||
+                entry.getValue.getSize > 0
```
aheyne, line 163: Why do we care that the found entry has a value?

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/api/IndexAdapter.scala
```diff
@@ -150,7 +161,26 @@ object IndexAdapter {
...
+      try {
+        var i = 0
+        // calculate all the mutations up front to ensure that there aren't any validation errors
+        while (i < converters.length) {
+          values(i) = converters(i).convert(writable)
+          previousValues(i) = converters(i).convert(removable)
+          i +=1
+        }
+      } catch {
+        case NonFatal(e) => throw new IllegalArgumentException("Error creating keys for insert:", e)
```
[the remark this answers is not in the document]
elahrvivaz: Yeah, things like ["geom is null"](https://github.com/locationtech/geomesa/blob/08ff48ffa677d2b90827e025d216b21dcd4b24bd/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/index/z3/Z3IndexKeySpace.scala#L69) and ["lat/lon out of bounds"](https://github.com/locationtech/geomesa/blob/08ff48ffa677d2b90827e025d216b21dcd4b24bd/geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/Z3SFC.scala#L40) - this is the same behavior as before.

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/writer/tx/AccumuloAtomicIndexWriter.scala
elahrvivaz: it shouldn't matter since they're all independent tables/rows

geomesa-accumulo/geomesa-accumulo-datastore/src/main/scala/org/locationtech/geomesa/accumulo/data/writer/tx/AccumuloAtomicIndexWriter.scala
elahrvivaz: i think this check is backwards actually... it's checking to see if a row was deleted successfully, so if it has a value then it wasn't deleted (deletes have empty values, although i don't think they should actually come back from a scan anyway...). does changing it to `entry.getValue.getSize == 0` make more sense?

elahrvivaz:
```suggestion
                entry.getValue.getSize == 0
```

elahrvivaz: I think I was basing this on the `putDelete` method in Accumulo, which sets an empty byte array: https://github.com/apache/accumulo/blob/main/core/src/main/java/org/apache/accumulo/core/data/Mutation.java#L423-L425

Actually the key has a deleted check, so I should probably just look at that:

```suggestion
                entry.getKey.isDeleted
```

PR #3054, opened 2024-03-11
GEOMESA-3341 Update to arrow 15.0.2

elahrvivaz:


elahrvivaz:
seems like the 15.0.1 release didn't actually include the intended fixes, so they are working on a 15.0.2 - we'll wait for that.

PR #3064, opened 2024-03-20
GEOMESA-3259 FSDS - Add support for GeoParquet

adeet1:
* Create a bounding box for each geometry, and add it to the GeoParquet metadata (which requires the metadata map to be changed to a mutable data structure)
* Read and write all geometry attributes as binary (a primitive Parquet type) instead of as a pair of x/y doubles (a group Parquet type), using the same converter and attribute writer for all geometry types, while also maintaining backwards compatibility
* Add support for parsing WKB bytes in the Parquet geometry transformer functions
* Use a spatial index instead of a GeoTools filter for bounding box queries


adeet1:
To-do items:

- [ ] Make `FilterConverter.spatial` backwards-compatible
- [ ] Add support for 3D geometries and bounding boxes
- [ ] Add a unit test assert that ensures the file metadata validates against the [GeoParquet metadata schema](https://github.com/opengeospatial/geoparquet/blob/main/format-specs/schema.json)

geomesa-convert/geomesa-convert-parquet/src/main/scala/org/locationtech/geomesa/convert/parquet/ParquetFunctionFactory.scala
```diff
@@ -58,6 +60,8 @@ class ParquetFunctionFactory extends TransformerFunctionFactory {
     }
 
     protected def eval(x: U, y: U): T
+
+    protected def eval(b: Array[Byte]): T = WKBUtils.read(b).asInstanceOf[T]
```
elahrvivaz, line 64: we can probably leave these functions alone, and just note them as deprecated in the user docs - we can get equivalent functionality with a combo of avroPath + our regular geometry parsing functions (from geomesa-convert-common)

geomesa-convert/geomesa-convert-parquet/src/test/scala/org/locationtech/geomesa/convert/parquet/ParquetConverterTest.scala
```diff
@@ -34,6 +34,39 @@ class ParquetConverterTest extends Specification {
...
+      val conf = ConfigFactory.parseString(
+        """
+          | {
+          |   type         = "parquet",
+          |   id-field     = "avroPath($0, '/__fid__')",
+          |   fields = [
+          |     { name = "name",     transform = "avroPath($0, '/name')" },
+          |     { name = "age",      transform = "avroPath($0, '/age')" },
+          |     { name = "dtg",      transform = "avroPath($0, '/dtg')" },
+          |     { name = "position", transform = "parquetPoint($0, '/position')" },
```
elahrvivaz, line 47:
```suggestion
          |     { name = "position", transform = "point(avroPath($0, '/position'))" },
```

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
```diff
@@ -71,7 +71,7 @@ abstract class AbstractFileSystemStorage(
     * @param observer observer to report stats on the data written
     * @return
     */
-  protected def createWriter(file: Path, observer: FileSystemObserver): FileSystemWriter
+  protected def createWriter(file: Path, observer: BoundsObserver): FileSystemWriter
```
elahrvivaz, line 74: i think this needs to stay as FileSystemObserver here, b/c the user can specify custom observers at runtime. we should probably make FileSystemWriter essentially be it's own observer for bounds, and then we can change the bounds observer here to just pull the bounds out of the writer with a callback.

elahrvivaz: the way you have it now does kind of work, but seems a bit confusing and a mixing of concerns. since the writer now *requires* tracking its own bounds, it should just do that itself.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -30,19 +30,20 @@ import org.locationtech.geomesa.utils.text.StringSerialization
...
-  import SimpleFeatureParquetSchema.{CurrentSchemaVersion, SchemaVersionKey}
+  import SimpleFeatureParquetSchema.{CurrentSchemaVersion, GeoParquetSchemaKey, SchemaVersionKey}
 
   import scala.collection.JavaConverters._
 
   /**
     * Parquet file metadata
     */
-  lazy val metadata: java.util.Map[String, String] = Map(
+  lazy val metadata: java.util.Map[String, String] = scala.collection.mutable.HashMap(
```
elahrvivaz, line 42: we probably don't want this to be a mutable map, especially since it's not a private variable. having mutable state makes things harder to reason about, since you don't know if someone has modified it out from under you.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -56,16 +57,64 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
...
+    val geomField = sft.getGeomField
+
+    // If the sft has no geometry field
+    if (geomField == null) {
+      """{"version":"1.0.0"}"""
+    } else {
+      val primaryColumn = alphaNumericSafeString(geomField)
+      val columns = sft.getAttributeDescriptors.asScala.collect { case g: GeometryDescriptor => geoParquetMetadata(g) }.mkString(",")
+
+      s"""{"version":"1.0.0","primary_column":"$primaryColumn","columns":[$columns"""
```
elahrvivaz, line 90: are we missing the end of the array here?
```suggestion
      s"""{"version":"1.0.0","primary_column":"$primaryColumn","columns":[$columns]}"""
```

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -56,16 +57,64 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
...
+   *
+   * @param sft simple feature type
+   * @return
+   */
+  def geoParquetMetadata(sft: SimpleFeatureType): String = {
+    val geomField = sft.getGeomField
+
+    // If the sft has no geometry field
+    if (geomField == null) {
+      """{"version":"1.0.0"}"""
```
elahrvivaz, line 85: this won't be a valid geoparquet file (columns and such are required in the schema)... in this case, we probably just want to omit the geoparquet metadata entirely.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
elahrvivaz: actually, it looks like columns is supposed to be an object, not an array, where the key is the name of the geometry col: https://geoparquet.org/releases/v1.0.0/#:~:text=are%20described%20below.-,File%20metadata,-Field%20Name

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureReadSupport.scala
```diff
@@ -29,12 +30,15 @@ import scala.collection.mutable.ArrayBuffer
 
 class SimpleFeatureReadSupport extends ReadSupport[SimpleFeature] {
 
-  private var schema: SimpleFeatureParquetSchema = _
+  var schema: SimpleFeatureParquetSchema = null
+  var schemaVersion: Integer = null
```
elahrvivaz, line 34: these should stay private if at all possible - we don't want to expose mutable state outside the scope of the class

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureReadSupport.scala
```diff
@@ -179,6 +190,17 @@ object SimpleFeatureReadSupport {
     }
   }
 
+  private def geometryV2(binding: ObjectType, i: Int, callback: Settable): Converter = {
+    val geomTypes = Seq(ObjectType.POINT, ObjectType.LINESTRING, ObjectType.POLYGON, ObjectType.MULTIPOINT,
+      ObjectType.MULTILINESTRING, ObjectType.MULTIPOLYGON, ObjectType.GEOMETRY)
+
+    if (geomTypes.contains(binding)) {
```
elahrvivaz, line 197: don't think we need this check here, this method is only called right after checking the object type.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureWriteSupport.scala
```diff
@@ -96,15 +118,13 @@ object SimpleFeatureWriteSupport {
...
-      case ObjectType.POLYGON         => new PolygonAttributeWriter(name, index)
-      case ObjectType.MULTIPOINT      => new MultiPointAttributeWriter(name, index)
-      case ObjectType.MULTILINESTRING => new MultiLineStringAttributeWriter(name, index)
-      case ObjectType.MULTIPOLYGON    => new MultiPolygonAttributeWriter(name, index)
-      case ObjectType.GEOMETRY        => new GeometryWkbAttributeWriter(name, index)
-      case _ => throw new IllegalArgumentException(s"Can't serialize field '$name' of type $binding")
+    val geomTypes = Seq(ObjectType.POINT, ObjectType.LINESTRING, ObjectType.POLYGON, ObjectType.MULTIPOINT,
+      ObjectType.MULTILINESTRING, ObjectType.MULTIPOLYGON, ObjectType.GEOMETRY)
+
+    if (geomTypes.contains(binding)) {
```
elahrvivaz, line 124: i think this is another case where we won't ever get here unless this check is true, so we can skip it.

pom.xml
```diff
@@ -2780,6 +2780,11 @@
                 <version>${testcontainers.accumulo.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>com.jayway.jsonpath</groupId>
+                <artifactId>json-path</artifactId>
+                <version>2.9.0</version>
+            </dependency>
```
elahrvivaz, line 2787: we already have this dependency declared in the pom, so we shouldn't need to redeclare it here.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
adeet1: I intentionally left the array bracket unclosed, because the `finalizeWrite()` method in SimpleFeatureWriteSupport appends the bounding box to the metadata string and closes any open brackets.

Regarding columns needing to be an object: I'm wondering if we should still have the `def geoParquetMetadata(sft: SimpleFeatureType): String` create the entire metadata string directly? Or maybe if it's better to put everything in a struct (or similar data structure) and then marshal it into a json string?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -56,16 +57,64 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
...
+        case ObjectType.MULTIPOINT          => "MultiPoint"
+        case ObjectType.GEOMETRY_COLLECTION => "GeometryCollection"
+        case ObjectType.GEOMETRY            => null
+      }
+      Seq(types).filter(_ != null)
+    }
+    // note: don't provide crs, as default is EPSG:4326 with longitude first, which is our default/only crs
+
+    def stringify(geomName: String, encoding: String, geometryTypes: Seq[String], orientation: String, edges: String, epoch: Long): String = {
+      s"""{"$geomName":{"encoding":"$encoding","geometry_types":[${geometryTypes.mkString(",")}],"orientation":"$orientation","edges":"$edges","epoch":$epoch"""
```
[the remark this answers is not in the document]
adeet1: Making this change ([c1dda99](https://github.com/locationtech/geomesa/pull/3064/commits/c1dda9968ca474462b3b728e5f550279a76c7f0d)) seems to cause the `"write files with a target size"` unit test in ParquetStorageTest to fail.

Do we need to change the `val targetSize = 2100L` magic number to something else?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
adeet1: Never mind - I added an assert to the `"write parquet files"` unit test in ParquetReadWriteTest that validates the metadata against the json schema, and fixed the json accordingly.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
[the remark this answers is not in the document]
elahrvivaz: what is the failure? if it's just that there's one file that's smaller (due to the remainder) then tweaking the targetSize in the test makes sense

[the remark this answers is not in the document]
adeet1: Fixed in https://github.com/locationtech/geomesa/pull/3064/commits/9770a3a336d7f145260fc4e07adb4770a21e4fb6

geomesa-convert/geomesa-convert-parquet/src/main/scala/org/locationtech/geomesa/convert/parquet/ParquetConverterFactory.scala
```diff
@@ -109,6 +105,40 @@ class ParquetConverterFactory
       case NonFatal(e) => logger.debug(s"Could not infer Parquet converter from input:", e); None
     }
   }
+
+  private def computeTransformFunction(name: String, path: String, descriptor: AttributeDescriptor, schemaVersion: Int): String = {
+    // Make both expression vals lazy so that we only compute the one we need to, once we determine what the schema version is
+    lazy val expressionV2 = ObjectType.selectType(descriptor) match {
```
elahrvivaz, line 111: can we make these functions instead of lazy vals? you can define functions inside your other functions, if you don't want them to be top-level. using functions seems a little clearer, and there is some compute overhead with initializing lazy vals.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/FilterConverter.scala
```diff
@@ -23,14 +23,18 @@ import scala.reflect.ClassTag
...
-  def convert(sft: SimpleFeatureType, filter: Filter): (Option[FilterPredicate], Option[Filter]) = {
-    if (filter == Filter.INCLUDE) { (None, None) } else {
-      FilterHelper.propertyNames(filter).foldLeft((Option.empty[FilterPredicate], Option(filter)))(reduce(sft))
+  def convert(sft: SimpleFeatureType, filter: Filter): (Int => (Option[FilterPredicate], Option[Filter])) = {
+    def thunk(version: Int): (Option[FilterPredicate], Option[Filter]) = {
+      if (filter == Filter.INCLUDE) { (None, None) } else {
+        FilterHelper.propertyNames(filter).foldLeft((Option.empty[FilterPredicate], Option(filter)))(reduce(sft, version))
+      }
     }
+    thunk
```
elahrvivaz, line 32: this is a good place to use memoization - we're going to be calling this repeatedly for a given query, and the only input that changes is (maybe) the version. so we can do something like:
```
if (filter == Filter.INCLUDE) { _ => (None, None) } else {
  val propertyNames = FilterHelper.propertyNames(filter)
  lazy val v1 = propertyNames.foldLeft((Option.empty[FilterPredicate], Option(filter)))(reduce(sft, 1))
  lazy val v2 = propertyNames.foldLeft((Option.empty[FilterPredicate], Option(filter)))(reduce(sft, 2))
  (i) => if (i == 2) { v2 } else { v1 }
}
```

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -93,7 +93,7 @@ object ParquetFileSystemStorage extends LazyLogging {
   }
 
   def validateParquetFile(file: Path): Unit = {
-    val reader = ParquetReader.builder(new GroupReadSupport(), file).build()
+    val reader = ParquetReader.builder(new SimpleFeatureReadSupport, file).build()
```
elahrvivaz, line 96: i think we can stick with GroupReadSupport here - that way we will check that it's a valid parquet file, but we won't deserialize/manifest everything so it saves us some processing

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/ParquetFileSystemStorage.scala
```diff
@@ -22,9 +21,10 @@ import org.locationtech.geomesa.fs.storage.api._
 import org.locationtech.geomesa.fs.storage.common.{AbstractFileSystemStorage, FileValidationEnabled}
 import org.locationtech.geomesa.fs.storage.common.AbstractFileSystemStorage.FileSystemPathReader
 import org.locationtech.geomesa.fs.storage.common.jobs.StorageConfiguration
-import org.locationtech.geomesa.fs.storage.common.observer.FileSystemObserver
+import org.locationtech.geomesa.fs.storage.common.observer.{BoundsObserver, FileSystemObserver}
```
elahrvivaz, line 24: should run optimize imports here

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/SimpleFeatureParquetWriter.scala
```diff
@@ -16,7 +16,9 @@ import org.apache.parquet.hadoop.api.WriteSupport
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
 import org.apache.parquet.hadoop.{ParquetFileWriter, ParquetWriter}
 import org.geotools.api.feature.simple.SimpleFeature
+import org.locationtech.geomesa.fs.storage.common.AbstractFileSystemStorage.MetadataObserver
 import org.locationtech.geomesa.fs.storage.parquet.io.SimpleFeatureWriteSupport
+import org.locationtech.jts.geom.Envelope
```
elahrvivaz, line 21: should be able to revert this file or optimize imports

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -42,9 +43,26 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
...
     StorageConfiguration.SftNameKey -> sft.getTypeName,
     StorageConfiguration.SftSpecKey -> SimpleFeatureTypes.encodeType(sft, includeUserData = true),
-    SchemaVersionKey                -> CurrentSchemaVersion.toString // note: this may not be entirely accurate, but we don't write older versions
+    SchemaVersionKey                -> version.toString,
+    GeoParquetSchemaKey             -> null
   ).asJava
 
+  def addBoundingBoxMetadata(bbox: Envelope): java.util.Map[String, String] = {
+    val bboxString = s"[${bbox.getMinX}, ${bbox.getMaxX}, ${bbox.getMinY}, ${bbox.getMaxY}]"
+    val result = SimpleFeatureParquetSchema.geoParquetMetadata(sft) + s""","bbox":${bboxString}}}}""""
```
elahrvivaz, line 52: we should pass the bbox into `geoParquetMetadata` function and calculate the full json there - that way the logic is all in one place, and the function is more useful b/c it returns a complete json object. we could make bbox argument an `Option[Envelope]` if you think there's a use case for creating the metadata without including the bbox.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -56,16 +74,64 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
...
+   *
+   * @param sft simple feature type
+   * @return
+   */
+  def geoParquetMetadata(sft: SimpleFeatureType): String = {
+    val geomField = sft.getGeomField
+
+    // If the sft has no geometry field, then omit the GeoParquet metadata entirely
+    if (geomField == null) {
+      ""
```
elahrvivaz, line 102: i think in this case, we should probably omit the geoparquet metadata key entirely - so we could make this function return an Option[String], or we could just make this function only work when there's a geom field, and add the check for the geom field before calling this function.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -56,16 +74,64 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
...
+    } else {
+      val primaryColumn = alphaNumericSafeString(geomField)
+      val columns = sft.getAttributeDescriptors.asScala.collect { case g: GeometryDescriptor => geoParquetMetadata(g) }.mkString(",")
+
+      s"""{"version":"1.1.0-dev","primary_column":"$primaryColumn","columns":$columns"""
```
elahrvivaz, line 107: I think version here needs to be `1.0.0` according the the json-schema

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
```diff
@@ -42,9 +43,26 @@ case class SimpleFeatureParquetSchema(sft: SimpleFeatureType, schema: MessageTyp
   lazy val metadata: java.util.Map[String, String] = Map(
     StorageConfiguration.SftNameKey -> sft.getTypeName,
     StorageConfiguration.SftSpecKey -> SimpleFeatureTypes.encodeType(sft, includeUserData = true),
-    SchemaVersionKey                -> CurrentSchemaVersion.toString // note: this may not be entirely accurate, but we don't write older versions
+    SchemaVersionKey                -> version.toString,
+    GeoParquetSchemaKey             -> null
   ).asJava
 
+  def addBoundingBoxMetadata(bbox: Envelope): java.util.Map[String, String] = {
+    val bboxString = s"[${bbox.getMinX}, ${bbox.getMaxX}, ${bbox.getMinY}, ${bbox.getMaxY}]"
```
elahrvivaz, line 51: according to the spec this needs to be `xmin, ymin, xmax, ymax`

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureParquetSchema.scala
elahrvivaz: colums here doesn't seem quite right - it should end up looking like:
```
columns: {
  "geomA": {
    "encoding": "WKB",
    "geometry_types": []
  },
  "geomB": {
    "encoding": "WKB",
    "geometry_types": [ "Point" ],
    "bbox": [0, 0, 1, 1]
  },
}
```
but i think the way you have it now it won't work for multiple geometries.

elahrvivaz: since this json is getting a little complicated, it might be good to model it using case classes and then use something like json4s to convert it to json.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/main/scala/org/locationtech/geomesa/fs/storage/parquet/io/SimpleFeatureWriteSupport.scala
```diff
@@ -24,26 +26,49 @@ import java.util.{Date, UUID}
...
     this.writer = SimpleFeatureWriteSupport.SimpleFeatureWriter(schema.sft)
 
     new WriteContext(schema.schema, schema.metadata)
   }
 
+  // called once at the end after all SimpleFeatures are written
+  override def finalizeWrite(): FinalizedWriteContext = {
+    // Get the bounding box that spans all the SimpleFeatures' geometries
+    val envelope = observer match {
+      case obj: BoundsObserver => Some(obj.getBoundingBox)
```
elahrvivaz, line 55: we shouldn't need a cast here?

elahrvivaz: i think after looking at the geoparquet json that we actually need to track the bounds for each geometry separately, when there are multiple geometries in the schema.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-parquet/src/test/resources/geoparquet-metadata-schema.json
```diff
@@ -0,0 +1,125 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "GeoParquet",
+  "description": "Parquet metadata included in the geo field.",
+  "type": "object",
+  "required": ["version", "primary_column", "columns"],
+  "properties": {
+    "version": {
+      "type": "string",
+      "const": "1.1.0-dev"
```
elahrvivaz, line 10: we should be using https://geoparquet.org/releases/v1.0.0/schema.json here, not the dev one

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
```diff
@@ -235,8 +235,8 @@ abstract class AbstractFileSystemStorage(
       val path = StorageUtils.nextFile(context.root, partition, metadata.leafStorage, extension, fileType)
       PathCache.register(context.fc, path)
       val updateObserver = new UpdateObserver(partition, path, action)
-      val observer = if (observers.isEmpty) { updateObserver } else {
-        new CompositeObserver(observers.map(_.apply(path)).+:(updateObserver))
+      val observer = if (observers.isEmpty) { updateObserver.asInstanceOf[BoundsObserver] } else {
+        new CompositeObserver(observers.map(_.apply(path)).+:(updateObserver)).asInstanceOf[BoundsObserver]
```
elahrvivaz, line 239: i think ideally we could use the bounds that the writer is already calculating, so that we don't have to calculate them twice.

adeet1 (commented):
- [x] When we compact GeoParquet files in a filesystem partition, we need to ensure that the bounding boxes in the metadata of the files get merged correctly (i.e. assert that the union of bounding boxes of the files before compaction is equal to the union of bounding boxes of the newly compacted files).

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/AbstractFileSystemStorage.scala
```diff
@@ -67,11 +67,13 @@ abstract class AbstractFileSystemStorage(
...
     * Create a writer for the given file
     *
+    * @param partition the partition that the file belongs to
+    * @param action whether to append or modify
     * @param file file to write to
     * @param observer observer to report stats on the data written
     * @return
     */
-  protected def createWriter(file: Path, observer: FileSystemObserver): FileSystemWriter
+  protected def createWriter(partition: String, action: StorageFileAction, file: Path, observer: FileSystemObserver): FileSystemWriter
```
elahrvivaz, line 76:
```suggestion
  protected def createWriter(partition: String, action: StorageFileAction, file: Path, observer: Option[FileSystemObserver]): FileSystemWriter
```
since we don't have the updateObserver anymore, we should observer an option. that will let us clean up the code down on line 239 as well, we won't need the NoOpObserver any more.

elahrvivaz: i think that means we can get rid of the BoundsObserver and related changes now too?

elahrvivaz:
With the release of geoparquet 1.1.0, we don't need to encode everything as WKB any more, our regular encoding matches (or almost matches in some cases) the "native" geoparquet encoding. It may still be useful to support WKB encoding, though, as "native" encoding support doesn't seem widely adopted yet.

PR #3088, opened 2024-04-04
GEOMESA-3347 Type inference for xml and (non-geo) json

elahrvivaz:
* Make inference method use a Try
* Change all float inference to doubles
  * floats cause rounding errors when used as doubles in geometries
* Remove creating points from consecutive number fields
* Improve error messages in CLI tools

### DIFF
--- a/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
+++ b/geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
@@ -9,138 +9,200 @@
 package org.locationtech.geomesa.convert.json
 
 import com.google.gson.stream.{JsonReader, JsonToken}
-import com.google.gson.{JsonElement, JsonParser}
+import com.google.gson.{JsonElement, JsonNull, JsonParser}
+import com.jayway.jsonpath.{JsonPath, PathNotFoundException}
 import com.typesafe.config.Config
 import org.geotools.api.feature.simple.SimpleFeatureType
+import org.locationtech.geomesa.convert.EvaluationContext
 import org.locationtech.geomesa.convert.json.GeoJsonParsing.GeoJsonFeature
 import org.locationtech.geomesa.convert.json.JsonConverter._
-import org.locationtech.geomesa.convert.json.JsonConverterFactory.{JsonConfigConvert, JsonFieldConvert}
+import org.locationtech.geomesa.convert.json.JsonConverterFactory.{JsonConfigConvert, JsonFieldConvert, PropNamer}
 import org.locationtech.geomesa.convert2.AbstractConverter.BasicOptions
 import org.locationtech.geomesa.convert2.AbstractConverterFactory.{BasicOptionsConvert, ConverterConfigConvert, FieldConvert, OptionConvert}
-import org.locationtech.geomesa.convert2.TypeInference.{IdentityTransform, InferredType}
+import org.locationtech.geomesa.convert2.TypeInference.{Namer, PathWithValues, TypeWithPath}
 import org.locationtech.geomesa.convert2.transforms.Expression
 import org.locationtech.geomesa.convert2.{AbstractConverterFactory, TypeInference}
-import org.locationtech.geomesa.utils.geotools.ObjectType.ObjectType
-import org.locationtech.geomesa.utils.geotools.{FeatureUtils, ObjectType}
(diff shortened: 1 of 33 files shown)


geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XmlConverterTest.scala
```diff
@@ -638,6 +638,73 @@ class XmlConverterTest extends Specification {
...
+          features must haveLength(1)
+
+          features.head.getAttribute(0) mustEqual "myxml"
+          features.head.getAttribute(1) mustEqual 123
+          features.head.getAttribute(2) mustEqual "red"
+          features.head.getAttribute(3) mustEqual "5'11"
+          features.head.getAttribute(4) mustEqual 127.5
+          features.head.getAttribute(5) mustEqual 21.1
+          features.head.getAttribute(6) mustEqual 45.1
+          features.head.getAttribute(7) mustEqual WKTUtils.read("POINT (45.1 21.1)")
```
adeet1, line 704:
```suggestion
          val expected = Seq("myxml", 123, "red", "5'11", 127.5, 21.1, 45.1, WKTUtils.read("POINT (45.1 21.1)"))
          features.head.getAttributes.asScala must containTheSameElementsAs(expected)
```

geomesa-convert/geomesa-convert-json/src/test/scala/org/locationtech/geomesa/convert/json/JsonConverterTest.scala
```diff
@@ -1542,5 +1544,105 @@ class JsonConverterTest extends Specification {
...
+          |{ "name": "name2", "files":[2,3], "geom": "POINT (42 52)" }
+          |{ "name": "name3", "demographics": { "age": 3 }, "files":[4,5,6], "geom": "POINT (43 53)" }
+        """.stripMargin
+
+      def bytes = new ByteArrayInputStream(json.getBytes(StandardCharsets.UTF_8))
+
+      val factory = new JsonConverterFactory()
+      // test deprecated infer method as well as new one
+      val old = factory.infer(bytes).fold[Try[(SimpleFeatureType, Config)]](Failure(null))(Success(_))
+      foreach(Seq(factory.infer(bytes, None, Map.empty[String, AnyRef]), old)) { inferred =>
```
adeet1, line 1560: I think it would be better if we could split this into two separate unit tests - one for the deprecated infer method and one for the new method - if possible.

geomesa-convert/geomesa-convert-json/src/main/scala/org/locationtech/geomesa/convert/json/JsonConverterFactory.scala
```diff
@@ -9,138 +9,200 @@
...
+            .flatMap(inferGeoJson(_, None, sft))
+      } else if (JsonConverter.isFeatureCollection(elements.head)) {
+        Try(elements.flatMap(JsonConverter.parseFeatureCollection))
+            .flatMap(inferGeoJson(_, Some("$.features[*]"), sft))
+      } else if (elements.head.isJsonObject) {
+        inferJson(elements, featurePath, sft)
+      } else if (elements.head.isJsonArray) {
+        inferJson(elements, featurePath.orElse(Some("$.[*]")), sft)
+      } else {
+        Failure(new RuntimeException("Could not parse the input as a JSON object or array"))
```
adeet1, line 88: We might want to implement a unit test for this failure case, if we don't already have one.

adeet1 (changes requested):

geomesa-convert/geomesa-convert-xml/src/test/scala/org/locationtech/geomesa/convert/xml/XmlConverterTest.scala
elahrvivaz: the tests in this class do `feature.getAttribute("foo")` comparisons, so this is consistent with the rest of the class at least, although as you point out the test classes aren't consistent with each other.

adeet1: Makes sense. I think it would be good to make the test classes consistent with each other, so that we're doing the same kind of asserts everywhere. Plus it makes the code more concise.

elahrvivaz:
@aheyne I'm going to merge this b/c we want to use it for some prototyping this week, but if you have time to review later I can put up another PR with any necessary changes.

PR #3134, opened 2024-07-08
GEOMESA-3375 Kafka admin client consumer props

malinsinbigler:
Ticket: https://geomesa.atlassian.net/browse/GEOMESA-3375

This fix allows for consumer properties set in the geoserver data store to be passed into the kafka admin client.

Use case: SASL connection properties set within geoserver were not making their way to the admin client causing failures communicating to the broker. This was only an issue when not using zookeeper.

### DIFF
--- a/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaMetadata.scala
+++ b/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaMetadata.scala
@@ -112,6 +112,7 @@ class KafkaMetadata[T](val config: KafkaDataStoreConfig, val serializer: Metadat
     val props = new Properties()
     props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, config.brokers)
     config.producers.properties.foreach { case (k, v) => props.put(k, v) }
+    config.consumers.properties.foreach { case (k, v) => props.put(k, v) }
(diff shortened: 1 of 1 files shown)


geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaMetadata.scala
```diff
@@ -112,6 +112,7 @@ class KafkaMetadata[T](val config: KafkaDataStoreConfig, val serializer: Metadat
     val props = new Properties()
     props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, config.brokers)
     config.producers.properties.foreach { case (k, v) => props.put(k, v) }
+    config.consumers.properties.foreach { case (k, v) => props.put(k, v) }
```
elahrvivaz, line 115:
```suggestion
    config.consumers.properties.foreach { case (k, v) => props.put(k, v) }
    config.producers.properties.foreach { case (k, v) => props.put(k, v) }
```

i'd just suggest switching them, so that producer props take precedence (preserving the existing behavior if there are conflicting keys)

elahrvivaz:
Seems fine - I don't know if there would ever be situations where you *wouldn't* want the consumer props there, but we can address that if it ever comes up :)

elahrvivaz:
Could you update the 2 AdminClient usages in KafkaDataStore to have the same logic? that way it's consistent across the board.

elahrvivaz (approved):

PR #3139, opened 2024-07-18
GEOMESA-3378 Fix CLI "find" warning output

amcdonaldccri:
* The warning in RHEL8 is "find: warning you have specified the -maxdepth option after a non-argument -type"


amcdonaldccri:
Ok I updated my email in eclipse to andrew.mcdonald@ga-ccri.com and set my primary email in github to the same
Also e-signed the eclipse document too

PR #3224, opened 2024-10-28
GEOMESA-3409: Arrow: axis order request parameter

epyatkevich:
[GEOMESA-3409](https://geomesa.atlassian.net/browse/GEOMESA-3409): Geomesa currently defaults to always reading and writing Arrow encoding as Lat / Lon axis order.

With the new axis order request parameter allow the user to request the encoding to be explicitly specified per request.

Axis Order parameter supported values:
- `LatLon` translates to: Latitude = Y Axis = North, Longitude = X Axis = East
- `LonLat` translates to: Longitude = X Axis = East, Latitude = Y Axis = North

The request parameter is optional and if not specified axis order defaults to the existing behavior which is `LatLon`.

Notes:
- This has been manually tested and working on the 4.0.4 branch and ported to the main line

### DIFF
--- a/docs/user/datastores/analytic_queries.rst
+++ b/docs/user/datastores/analytic_queries.rst
@@ -273,7 +273,7 @@ following query hints:
 +-------------------------------------+--------------------+------------------------------------+
 | QueryHints.ARROW_FLATTEN_STRUCT     | Boolean (optional) | flattenStruct                      |
 +-------------------------------------+--------------------+------------------------------------+
-| QueryHints.FLIP_AXIS_ORDER          | Boolean (optional) | axisOrder:LatLon / axisOrder:LonLat|
+| QueryHints.FLIP_AXIS_ORDER          | Boolean (optional) | axisOrder                          |
 +-------------------------------------+--------------------+------------------------------------+
 
 Explanation of Hints
@@ -341,7 +341,7 @@ in the RecordBatch. Note that this hint is currently only supported for PostGIS 
 FLIP_AXIS_ORDER
 ^^^^^^^^^^^^^^^
 
-This hint allows the coordinate axis order to be specified as either Lat/Lon or Lon/Lat explicitly to be set for every geometry related field.
+This hint flips the axis order of returned coordinates from latitude/longitude (default) to longitude/latitude.
 
 Example Query
 +++++++++++++
--- a/docs/user/datastores/analytic_queries.rst
+++ b/docs/user/datastores/analytic_queries.rst
@@ -273,6 +273,8 @@ following query hints:
 +-------------------------------------+--------------------+------------------------------------+
 | QueryHints.ARROW_FLATTEN_STRUCT     | Boolean (optional) | flattenStruct                      |
 +-------------------------------------+--------------------+------------------------------------+
+| QueryHints.FLIP_AXIS_ORDER          | Boolean (optional) | axisOrder:LatLon / axisOrder:LonLat|
++-------------------------------------+--------------------+------------------------------------+
 
 Explanation of Hints
 ++++++++++++++++++++
@@ -335,6 +337,11 @@ ARROW_FLATTEN_STRUCT
 
 This hint will remove the outer struct named after the feature type and will instead return the attribute fields directly
 in the RecordBatch. Note that this hint is currently only supported for PostGIS and geoserver native stores.
+
+FLIP_AXIS_ORDER
+^^^^^^^^^^^^^^^
+
+This hint allows the coordinate axis order to be specified as either Lat/Lon or Lon/Lat explicitly to be set for every geometry related field.
 
 Example Query
 +++++++++++++
--- a/docs/user/datastores/analytic_queries.rst
+++ b/docs/user/datastores/analytic_queries.rst
@@ -273,7 +273,7 @@ following query hints:
 +-------------------------------------+--------------------+------------------------------------+
 | QueryHints.ARROW_FLATTEN_STRUCT     | Boolean (optional) | flattenStruct                      |
 +-------------------------------------+--------------------+------------------------------------+
-| QueryHints.FLIP_AXIS_ORDER          | Boolean (optional) | axisOrder                          |
+| QueryHints.FLIP_AXIS_ORDER          | Boolean (optional) | flipAxisOrder                      |
 +-------------------------------------+--------------------+------------------------------------+
 
 Explanation of Hints
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeWriter.scala
@@ -16,7 +16,6 @@ import org.apache.arrow.vector.types.Types.MinorType
 import org.apache.arrow.vector.types.pojo.{ArrowType, FieldType}
 import org.geotools.api.feature.`type`.AttributeDescriptor
 import org.geotools.api.feature.simple.{SimpleFeature, SimpleFeatureType}
-import org.geotools.referencing.CRS.AxisOrder
 import org.locationtech.geomesa.arrow.jts._
 import org.locationtech.geomesa.arrow.vector.SimpleFeatureVector.SimpleFeatureEncoding
 import org.locationtech.geomesa.arrow.vector.SimpleFeatureVector.SimpleFeatureEncoding.Encoding
@@ -270,7 +269,10 @@ object ArrowAttributeWriter {
       case (_, _, FromList(_)) => throw new NotImplementedError("Geometry lists are not supported")
       case _ => throw new IllegalArgumentException(s"Unexpected geometry type $binding")
     }
-    new ArrowGeometryWriter(name, vector.asInstanceOf[GeometryVector[Geometry, FieldVector]], encoding)
+    val geometryVector = vector.asInstanceOf[GeometryVector[Geometry, FieldVector]]
+    geometryVector.setFlipAxisOrder(encoding.flipAxisOrder)
+
+    new ArrowGeometryWriter(name, geometryVector)
   }
 
   trait ArrowDictionaryWriter extends ArrowAttributeWriter {
@@ -465,21 +467,13 @@ object ArrowAttributeWriter {
   /**
     * Writes geometries - delegates to our JTS geometry vectors
     */
-  class ArrowGeometryWriter(val name: String, delegate: GeometryVector[Geometry, FieldVector], encoding: SimpleFeatureEncoding)
+  class ArrowGeometryWriter(val name: String, delegate: GeometryVector[Geometry, FieldVector])
       extends ArrowAttributeWriter {
 
     override def vector: FieldVector = delegate.getVector
 
     // note: delegate handles nulls
-    override def apply(i: Int, value: AnyRef): Unit = {
-      val options = Map("axisOrder" -> (encoding.axisOrder match {
-        case Encoding.Max => AxisOrder.EAST_NORTH /* Lon/Lat */
-        case _ => AxisOrder.NORTH_EAST /* Lat/Lon */
-      })).asInstanceOf[Map[String, AnyRef]].asJava
-
-      delegate.setOptions(options)
-      delegate.set(i, value.asInstanceOf[Geometry])
-    }
+    override def apply(i: Int, value: AnyRef): Unit = delegate.set(i, value.asInstanceOf[Geometry])
 
     override def setValueCount(count: Int): Unit = delegate.setValueCount(count)
   }
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeWriter.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeWriter.scala
@@ -16,6 +16,7 @@ import org.apache.arrow.vector.types.Types.MinorType
 import org.apache.arrow.vector.types.pojo.{ArrowType, FieldType}
 import org.geotools.api.feature.`type`.AttributeDescriptor
 import org.geotools.api.feature.simple.{SimpleFeature, SimpleFeatureType}
+import org.geotools.referencing.CRS.AxisOrder
 import org.locationtech.geomesa.arrow.jts._
 import org.locationtech.geomesa.arrow.vector.SimpleFeatureVector.SimpleFeatureEncoding
 import org.locationtech.geomesa.arrow.vector.SimpleFeatureVector.SimpleFeatureEncoding.Encoding
@@ -192,7 +193,7 @@ object ArrowAttributeWriter {
           case ObjectType.LONG     => new ArrowLongWriter(name, metadata, factory)
           case ObjectType.FLOAT    => new ArrowFloatWriter(name, metadata, factory)
           case ObjectType.DOUBLE   => new ArrowDoubleWriter(name, metadata, factory)
-          case ObjectType.GEOMETRY => geometry(name, bindings(1), encoding.geometry, metadata, factory)
+          case ObjectType.GEOMETRY => geometry(name, bindings(1), encoding, metadata, factory)
           case ObjectType.BOOLEAN  => new ArrowBooleanWriter(name, metadata, factory)
           case ObjectType.LIST     => new ArrowListWriter(name, bindings(1), encoding, metadata, factory)
           case ObjectType.MAP      => new ArrowMapWriter(name, bindings(1), bindings(2), encoding, metadata, factory)
@@ -234,11 +235,11 @@ object ArrowAttributeWriter {
   private def geometry(
       name: String,
       binding: ObjectType,
-      encoding: Encoding,
+      encoding: SimpleFeatureEncoding,
       metadata: Map[String, String],
       factory: VectorFactory): ArrowGeometryWriter = {
     val m = metadata.asJava
-    val vector = (binding, encoding, factory) match {
+    val vector = (binding, encoding.geometry, factory) match {
       case (ObjectType.POINT, Encoding.Min, FromStruct(c))              => new PointFloatVector(name, c, m)
       case (ObjectType.POINT, Encoding.Min, FromAllocator(c))           => new PointFloatVector(name, c, m)
       case (ObjectType.POINT, Encoding.Max, FromStruct(c))              => new PointVector(name, c, m)
@@ -269,7 +270,7 @@ object ArrowAttributeWriter {
       case (_, _, FromList(_)) => throw new NotImplementedError("Geometry lists are not supported")
       case _ => throw new IllegalArgumentException(s"Unexpected geometry type $binding")
     }
-    new ArrowGeometryWriter(name, vector.asInstanceOf[GeometryVector[Geometry, FieldVector]])
+    new ArrowGeometryWriter(name, vector.asInstanceOf[GeometryVector[Geometry, FieldVector]], encoding)
   }
 
   trait ArrowDictionaryWriter extends ArrowAttributeWriter {
@@ -464,13 +465,21 @@ object ArrowAttributeWriter {
   /**
     * Writes geometries - delegates to our JTS geometry vectors
     */
-  class ArrowGeometryWriter(val name: String, delegate: GeometryVector[Geometry, FieldVector])
+  class ArrowGeometryWriter(val name: String, delegate: GeometryVector[Geometry, FieldVector], encoding: SimpleFeatureEncoding)
       extends ArrowAttributeWriter {
 
     override def vector: FieldVector = delegate.getVector
 
     // note: delegate handles nulls
-    override def apply(i: Int, value: AnyRef): Unit = delegate.set(i, value.asInstanceOf[Geometry])
+    override def apply(i: Int, value: AnyRef): Unit = {
+      val options = Map("axisOrder" -> (encoding.axisOrder match {
+        case Encoding.Max => AxisOrder.EAST_NORTH /* Lon/Lat */
+        case _ => AxisOrder.NORTH_EAST /* Lat/Lon */
+      })).asInstanceOf[Map[String, AnyRef]].asJava
+
+      delegate.setOptions(options)
+      delegate.set(i, value.asInstanceOf[Geometry])
+    }
 
     override def setValueCount(count: Int): Unit = delegate.setValueCount(count)
   }
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
@@ -137,7 +137,7 @@ object SimpleFeatureVector {
   object SimpleFeatureEncoding {
 
     val Min: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Min), Encoding.Min, Encoding.Min, flipAxisOrder = false)
-    val Max: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Max), Encoding.Max, Encoding.Max, flipAxisOrder = true)
+    val Max: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Max), Encoding.Max, Encoding.Max, flipAxisOrder = false)
 
     def min(includeFids: Boolean, proxyFids: Boolean = false, flipAxisOrder: Boolean = false): SimpleFeatureEncoding = {
       val fids = if (includeFids) { Some(if (proxyFids) { Encoding.Min } else { Encoding.Max }) } else { None }
--- a/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
+++ b/geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
@@ -14,9 +14,6 @@ import org.apache.arrow.vector.types.FloatingPointPrecision
 import org.apache.arrow.vector.types.pojo.{ArrowType, FieldType}
 import org.apache.arrow.vector.{BigIntVector, FieldVector}
 import org.geotools.api.feature.simple.{SimpleFeature, SimpleFeatureType}
-import org.geotools.referencing.CRS
-import org.geotools.referencing.CRS.AxisOrder
-import org.geotools.util.factory.Hints
 import org.locationtech.geomesa.arrow.ArrowAllocator
 import org.locationtech.geomesa.arrow.features.ArrowSimpleFeature
 import org.locationtech.geomesa.arrow.jts.GeometryFields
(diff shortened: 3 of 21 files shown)


geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeWriter.scala
```diff
@@ -464,13 +465,21 @@ object ArrowAttributeWriter {
...
 
     // note: delegate handles nulls
-    override def apply(i: Int, value: AnyRef): Unit = delegate.set(i, value.asInstanceOf[Geometry])
+    override def apply(i: Int, value: AnyRef): Unit = {
+      val options = Map("axisOrder" -> (encoding.axisOrder match {
+        case Encoding.Max => AxisOrder.EAST_NORTH /* Lon/Lat */
+        case _ => AxisOrder.NORTH_EAST /* Lat/Lon */
+      })).asInstanceOf[Map[String, AnyRef]].asJava
+
+      delegate.setOptions(options)
```
elahrvivaz, line 480: instead of an untyped map, could we just have a `setLongitudeFirst(bool)` method? that way, we get compile time safety, and don't need to worry about magic keys and type casting, etc.
this should also be called just once, probably up on line 272 when the writer is created.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
```diff
@@ -132,16 +135,20 @@ object SimpleFeatureVector {
   val DescriptorKey   = "descriptor"
   val OptionsKey      = "options"
 
-  case class SimpleFeatureEncoding(fids: Option[Encoding], geometry: Encoding, date: Encoding)
+  case class SimpleFeatureEncoding(fids: Option[Encoding], geometry: Encoding, date: Encoding, axisOrder: Encoding)
```
elahrvivaz, line 138: trying to map axisOrder onto min/max doesn't make too much sense here. I think it would be easier just to make it a `longitudeFirst: boolean`. then we also don't need to worry about importing the CRS libraries just to get the AxisOrder enum

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
```diff
@@ -245,7 +252,18 @@ object SimpleFeatureVector {
       val isLong = dateVector.exists(_.isInstanceOf[BigIntVector])
       if (isLong) { Encoding.Max } else { Encoding.Min }
     }
-    val encoding = SimpleFeatureEncoding(fidEncoding, geomPrecision, datePrecision)
+    val axisOrderPrecision = {
+      if (java.lang.Boolean.getBoolean("org.geotools.referencing.forceXY") ||
+        Hints.getSystemDefault(Hints.FORCE_LONGITUDE_FIRST_AXIS_ORDER) == java.lang.Boolean.TRUE) {
```
elahrvivaz, line 257: I think this will always be true, as we set it globally here: https://github.com/locationtech/geomesa/blob/main/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/geotools/MetadataBackedDataStore.scala#L49

the exception would be PostGIS I guess.

How did you decide that this would be the way to check?

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
```diff
@@ -245,7 +252,18 @@ object SimpleFeatureVector {
...
+        Encoding.Max /* Lon/Lat */
+      } else {
+        CRS.getAxisOrder(sft.getCoordinateReferenceSystem) match {
```
elahrvivaz, line 260: we always treat everything as EPSG:4326, so this could probably be hard-coded. I guess the exception could be postgis, again.

elahrvivaz: I'd like to avoid having gt-referencing as a dependency, if possible... just to minimize the surface area of this change.

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/GeometryVector.java
```diff
@@ -32,4 +34,7 @@ public interface GeometryVector<T extends Geometry, V extends FieldVector> exten
   int getNullCount();
 
   void transfer(int fromIndex, int toIndex, GeometryVector<T, V> to);
+
+  Map<String, Object> getOptions();
+  void setOptions(Map<String, Object> map);
```
elahrvivaz, line 39: as mentioned above, instead of making this so general purpose I think it would be better to just make it a boolean

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractPointVector.java
```diff
@@ -53,8 +53,13 @@ public void set(int index, Point geom) {
       vector.setNull(index);
     } else {
       vector.setNotNull(index);
-      writeOrdinal(index * 2, geom.getY());
-      writeOrdinal(index * 2 + 1, geom.getX());
+      if (swapAxisOrder()) {
```
elahrvivaz, line 56: i think the switch away from a map of options to a boolean will fix this, but we don't want to be doing map lookups on each set call.

geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/ArrowScan.scala
```diff
@@ -272,12 +277,12 @@ object ArrowScan extends LazyLogging {
       SimpleFeatureTypes.createType(options(SftKey), options(SpecKey))
 
     def encoding(e: SimpleFeatureEncoding): (String, String) =
-      EncodingKey -> s"${e.fids.getOrElse("")}:${e.geometry}:${e.date}"
+      EncodingKey -> s"${e.fids.getOrElse("")}:${e.geometry}:${e.date}:${e.axisOrder}"
 
     def encoding(options: Map[String, String]): SimpleFeatureEncoding = {
-      val Array(fids, geom, dtg) = options(EncodingKey).split(":")
+      val Array(fids, geom, dtg, axisOrder) = options(EncodingKey).split(":")
```
elahrvivaz, line 283: for iterator code, we want to ensure back-compatibility - so that means handling the case where axis order is not set. the cleanest thing would probably be to put axis order in a separate key.

elahrvivaz:
looks like a good start, thanks for working on it! I think i'd prefer to replace most references to AxisOrder with `flipXy: Boolean` or maybe `longitudeFirst: Boolean`. that should simplify things a bit.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
ejp-raft: Bit of a contradiction then: the current Arrow logic is hard coded to use Lat/Lon (Y/X) order but the `MetadataBackendDataStore` hint is stating that it should always be the reverse (Lon/Lat, X/Y) order. Not seeing the hint being referenced in the `geomesa` repo, can / should it be removed at this point?

> How did you decide that this would be the way to check?

Think I read through this note:
https://docs.geotools.org/stable/userguide/library/referencing/order.html

and given that I didn't see any way to extract the encoding from a `StructVector` or it's derived `SimpleFeatureType` ended up falling back to the environment variables and hints.

Is there a better way to derive the flag here in `SimpleFeatureVector.getFeatureType(vector: StructVector)`?

elahrvivaz: yes, we weren't entirely consistent in the axis order :sweat_smile: . we have a TODO to remove the longitude first hint, but I'm unsure of the repercussions. i think it would probably be easiest to default to lat/lon in arrow (which is the current behavior), and only flip to lon/lat if the hint is explicitly provided. there could still be cases where someone expects a certain axis order based on the CRS, and it won't automatically work, but I'd say that's something we can save for later if it ever comes up.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
epyatkevich: What about instead of using a boolean using the same approach that [BinaryOutputEncoder](https://github.com/locationtech/geomesa/blob/df730a143b22b27cd25f7524aec1f89b6462d612/geomesa-utils-parent/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/bin/BinaryOutputEncoder.scala) takes which is using this [AxisOrder](https://github.com/locationtech/geomesa/blob/df730a143b22b27cd25f7524aec1f89b6462d612/geomesa-utils-parent/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/bin/AxisOrder.scala) enum?

It would be a more consistent approach throughout the Geomesa repo, wdyt?

elahrvivaz: It's a possibility - although I'll note that even though that AxisOrder is defined in the BinaryOutputEncoder, we haven't wired through the ability to actually change it through a query hint or anything else, see e.g. [here](https://github.com/locationtech/geomesa/blob/df730a143b22b27cd25f7524aec1f89b6462d612/geomesa-index-api/src/main/scala/org/locationtech/geomesa/index/iterators/BinAggregatingScan.scala#L41) where it's not passed through.

The main reason I prefer a boolean is that it's simple to encode/decode to string (for passing as user opts), and doesn't require the enum to be on the classpath.

Maybe we could call the hint `flipXy` - that way, it would satisfy the immediate need but would be somewhat tangential to a more complete axis order evaluation, if we ever decided to pursue one. I think the "correct" solution here is to inspect the CRS or other hints, in the same way that geoserver does, and determine axis order based on that. but due to the complexity of that approach (and to keep back-compatibility), just having a toggle seems simplest.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
epyatkevich: Dependency removed

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
epyatkevich: Yup so kept the existing behavior as the default and the encoders just referencing a derived boolean.

Not thrilled that there's no way to derive the axis order from the `StructVector` but from my manual testing this doesn't seems to be an issue at this point.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/ArrowAttributeWriter.scala
epyatkevich: Was originally going for maximum flexibility and to provide a mechanism to snake in any additional options / metadata to the geometry encoders.

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
```diff
@@ -245,7 +245,7 @@ object SimpleFeatureVector {
       val isLong = dateVector.exists(_.isInstanceOf[BigIntVector])
       if (isLong) { Encoding.Max } else { Encoding.Min }
     }
-    val encoding = SimpleFeatureEncoding(fidEncoding, geomPrecision, datePrecision)
+    val encoding = SimpleFeatureEncoding(fidEncoding, geomPrecision, datePrecision, flipAxisOrder = false)
```
epyatkevich, line 248: - Any way to derive the axis order from the `StructVector`?

elahrvivaz: arrow vectors can have metadata, so we could put something in there - but otherwise, probably not.

ejp-raft: That's fine, for the purpose of encoding this is not necessary at this time.

epyatkevich: That's fine, for the purpose of encoding this is not required can keep as is.

epyatkevich:
Added similar logic to all the remaining geometries

docs/user/datastores/analytic_queries.rst
```diff
@@ -336,6 +338,11 @@ ARROW_FLATTEN_STRUCT
 This hint will remove the outer struct named after the feature type and will instead return the attribute fields directly
 in the RecordBatch. Note that this hint is currently only supported for PostGIS and geoserver native stores.
 
+FLIP_AXIS_ORDER
+^^^^^^^^^^^^^^^
+
+This hint allows the coordinate axis order to be specified as either Lat/Lon or Lon/Lat explicitly to be set for every geometry related field.
```
elahrvivaz, line 344:
```suggestion
This hint flips the axis order of returned coordinates from latitude/longitude (default) to longitude/latitude.
```

geomesa-arrow/geomesa-arrow-gt/src/main/scala/org/locationtech/geomesa/arrow/vector/SimpleFeatureVector.scala
```diff
@@ -132,16 +132,16 @@ object SimpleFeatureVector {
...
 
-  case class SimpleFeatureEncoding(fids: Option[Encoding], geometry: Encoding, date: Encoding)
+  case class SimpleFeatureEncoding(fids: Option[Encoding], geometry: Encoding, date: Encoding, flipAxisOrder: Boolean)
 
   object SimpleFeatureEncoding {
 
-    val Min: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Min), Encoding.Min, Encoding.Min)
-    val Max: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Max), Encoding.Max, Encoding.Max)
+    val Min: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Min), Encoding.Min, Encoding.Min, flipAxisOrder = false)
+    val Max: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Max), Encoding.Max, Encoding.Max, flipAxisOrder = true)
```
elahrvivaz, line 140:
```suggestion
    val Max: SimpleFeatureEncoding = SimpleFeatureEncoding(Some(Encoding.Max), Encoding.Max, Encoding.Max, flipAxisOrder = false)
```
not sure this is used anywhere, but i think keeping it false is more consistent.

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/GeometryVector.java
```diff
@@ -32,4 +32,7 @@ public interface GeometryVector<T extends Geometry, V extends FieldVector> exten
   int getNullCount();
 
   void transfer(int fromIndex, int toIndex, GeometryVector<T, V> to);
+
+  Boolean getFlipAxisOrder();
```
elahrvivaz, line 36:
```suggestion
  Boolean isFlipAxisOrder();
```

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/GeometryVector.java
```diff
@@ -32,4 +32,7 @@ public interface GeometryVector<T extends Geometry, V extends FieldVector> exten
...
+  void setFlipAxisOrder(Boolean flip);
```
elahrvivaz, line 37:
```suggestion
  void setFlipAxisOrder(boolean flip);
```

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractGeometryVector.java
```diff
@@ -14,12 +14,14 @@
...
 
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public abstract class AbstractGeometryVector<T extends Geometry, U extends FieldVector, V extends FieldVector>
     implements GeometryVector<T, U> {
 
   private V ordinal;
   protected U vector;
+  protected final AtomicBoolean flipAxisOrder = new AtomicBoolean();
```
elahrvivaz, line 24: same thing here, don't think we need an atomic variable

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractLineStringVector.java
```diff
@@ -81,8 +86,14 @@ public LineString get(int index) {
       final int offsetEnd = vector.getOffsetBuffer().getInt((index + 1) * ListVector.OFFSET_WIDTH);
       final Coordinate[] coordinates = new Coordinate[offsetEnd - offsetStart];
       for (int i = 0; i < coordinates.length; i++) {
-        final double y = readOrdinal((offsetStart + i) * 2);
-        final double x = readOrdinal((offsetStart + i) * 2 + 1);
+        final double y, x;
+        if (getFlipAxisOrder()) {
+          y = readOrdinal(x(offsetStart + i));
```
elahrvivaz, line 91: i think i'd get rid of the `y()` and `x()` methods and just keep the math here like it was, since it's an extra method call (which might be inlined but might not be), and b/c sometimes we're setting y to x() which seems kind of confusing.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
```diff
@@ -67,7 +68,9 @@ class ArrowConversionProcess extends GeoMesaProcess with LazyLogging {
               @DescribeParameter(name = "batchSize", description = "Number of features to include in each record batch", min = 0)
               batchSize: java.lang.Integer,
               @DescribeParameter(name = "flattenStruct", description = "Removes the outer SFT struct yielding top level feature access", min = 0)
-              flattenStruct: java.lang.Boolean
+              flattenStruct: java.lang.Boolean,
+              @DescribeParameter(name = "axisOrder", description = "Geometry axis order: LatLon or LonLat", min = 0, defaultValue = "LatLon")
+              axisOrder: AxisOrder = LatLon
```
elahrvivaz, line 73:
```suggestion
              @DescribeParameter(name = "flipAxisOrder", description = "Flip the axis order of returned coordinates from latitude/longitude (default) to longitude/latitude", min = 0)
              flipAxisOrder: java.lang.Boolean
```

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
```diff
@@ -84,7 +87,11 @@ class ArrowConversionProcess extends GeoMesaProcess with LazyLogging {
       }
     }
 
-    val encoding = SimpleFeatureEncoding.min(includeFids == null || includeFids, proxyFids != null && proxyFids)
+    val flipAxisOrder = axisOrder match {
+      case LonLat => true
+      case _ => false
+    }
+    val encoding = SimpleFeatureEncoding.min(includeFids == null || includeFids, proxyFids != null && proxyFids, flipAxisOrder)
```
elahrvivaz, line 94:
```suggestion
    val flipAxisOrder = axisOrder match {
      case LonLat => true
      case _ => false
    }
    val encoding = SimpleFeatureEncoding.min(includeFids == null || includeFids, proxyFids != null && proxyFids, Option(flipAxisOrder).exists(_.booleanValue))
```

elahrvivaz:
looks good, thanks for bearing with me! I think we can merge after the next round of minor revisions.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
epyatkevich: - So I went with a hybrid approach: making the public API be very explicit, forcing the user to specify an enum `LatLon` or `LonLat` which internal is then translated and treated like a Boolean. The main reasons for doing so were that it communicates the intent clearly and the concern you had about adding additional dependencies for the enum is no longer there since the `utils` package is already being brought in and referenced virtually everywhere.

- I would suggest we leave this as is having the user specify the axis order and making it have a default value, and the internal logic figure out if it needs to flip or not, wdyt?

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractLineStringVector.java
```diff
@@ -64,8 +64,13 @@ public void set(int index, LineString geom) {
       for (int i = 0; i < geom.getNumPoints(); i++) {
         final Coordinate p = geom.getCoordinateN(i);
         tuples.setNotNull(position + i);
-        writeOrdinal((position + i) * 2, p.y);
-        writeOrdinal((position + i) * 2 + 1, p.x);
+        if (getFlipAxisOrder()) {
```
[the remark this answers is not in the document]
epyatkevich: - I would prefer to keep the internal representation of the flip axis order hidden behind the accessor (`isFlipAxisOrder()`) to encapsulate it and make it much easier to modify in one place if need be in the future because the performance trade off of looking up the method and invoking it should be negligent and will most likely be inlined by the JVM assuming that this code path is executed with enough frequency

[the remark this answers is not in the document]
elahrvivaz: sure, in that case, the variable should be private and not protected.

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractLineStringVector.java
elahrvivaz: well, absent some benchmarks that show performance changes one way or another, I would like to avoid extra method calls. this code gets called potentially millions of times in a given request, so I don't think it's premature to optimize it (or at least try to avoid regression). can we accomplish the same thing with better javadoc comments? I think knowing that x/y are interleaved in the single arrow vector makes the offset calculations fairly transparent.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
elahrvivaz: one concern I have here is that these variables get serialized from xml when called through geoserver, e.g. things like [this](https://github.com/locationtech/geomesa/blob/main/docs/user/_static/process/UniqueProcess_wps.xml). I'm not sure how serialization will work for a custom axis order enum. If you want to keep this, could you test that out? But also, I think it's better to favor consistency, i.e. have the hints that trigger an analytic query correspond to the wps parameters that do the same thing. And as you say, we'd need internal logic to calculate the correct axis order, and I don't have a good sense of how to do that correctly, given all the complexities described in the [geoserver docs](https://docs.geoserver.org/main/en/user/services/wfs/axis_order.html#axis-ordering), so I'd rather punt on it.

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractLineStringVector.java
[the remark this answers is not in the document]
elahrvivaz: ah, I see you already changed that :+1:

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractLineStringVector.java
ejp-raft: - Does geomesa have any kind of benchmark baseline methodology (performance suite) and numbers to compare to?

- If you insist I can revert the `y()` and `x()` functions but again I think for a request with a monster sized geometry with millions of points to process this kind of optimization is really derived by the JVM HotSpot at runtime and the run time ceiling will most likely not be affacted by the extra dispatches to `y()` and `x()` but by I/O related tasks.

geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/transform/ArrowConversionProcess.scala
epyatkevich: I'm running this of a Mac and can't run the tests will switch it over to a Boolean

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractLineStringVector.java
elahrvivaz: you're probably right that it won't matter, but I'm cautious about changes - i'd prefer to err on the side of performance for low-level functions like this. we did have some JMH-based benchmarks that we took out b/c we were no longer using them, but if you want to resurrect them it shouldn't be too difficult: https://github.com/locationtech/geomesa/blob/3.5.x/benchmarks/src/main/scala/org/locationtech/geomesa/benchmarks/JsonPathBenchmarks.scala

geomesa-arrow/geomesa-arrow-jts/src/main/java/org/locationtech/geomesa/arrow/jts/impl/AbstractPointVector.java
```diff
@@ -76,8 +87,13 @@ public void transfer(int fromIndex, int toIndex, GeometryVector<Point, FixedSize
       ((FixedSizeListVector) typed.vector).setNull(toIndex);
     } else {
       ((FixedSizeListVector) typed.vector).setNotNull(toIndex);
-      typed.writeOrdinal(toIndex * 2, readOrdinal(fromIndex * 2));
-      typed.writeOrdinal(toIndex * 2 + 1, readOrdinal(fromIndex * 2 + 1));
+      if (isFlipAxisOrder()) {
```
elahrvivaz, line 90: I think we'd want to check if the axis order is _different_ on the destination vector here:
```suggestion
      if (isFlipAxisOrder() != typed.isFlipAccessOrder()) {
```
alternatively, i think the transfer method assumes that the source and destination vectors are the same, so maybe we don't need any logic here about axis order.

elahrvivaz (approved):

PR #3242, opened 2024-12-04
GEOMESA-3424 Update to Arrow 18.1.0

elahrvivaz:


elahrvivaz:
jsr annotations (`javax.annotation.Nullable`) were previously pulled in transitively through arrow-memory. I just removed the annotations and added javadocs, rather than re-add the dependency.

PR #3244, opened 2024-12-10
GEOMESA-3423 - add unlogged user data option

autodidacticon:
@elahrvivaz 

* Added user-data option `pg.wal.log.enabled`, if set to `false` all postgis-partitioned tables will be created as [unlogged](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-UNLOGGED)
* added a test to `PartitionedPostgisDataStoreTest` to verify that tables are unlogged when option is set
* added a test to `PartitionedPostgisDataStoreTest` to verify that tables are logged when option is unset

### DIFF
--- a/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
+++ b/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
@@ -419,7 +419,7 @@ object PartitionedPostgisDialect {
     val MainTableSpace = "pg.partitions.tablespace.main"
 
     // set postgres table wal logging
-    val WalLogEnabled = "pg.wal.log.enabled"
+    val WalLogEnabled = "pg.wal.enabled"
 
     implicit class ConfigConversions(val sft: SimpleFeatureType) extends AnyVal {
       def getIntervalHours: Int = Option(sft.getUserData.get(IntervalHours)).map(int).getOrElse(6)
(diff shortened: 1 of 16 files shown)


elahrvivaz:
thanks for this! could you add the new config to the docs at https://raw.githubusercontent.com/locationtech/geomesa/refs/heads/main/docs/user/postgis/index_config.rst ?

geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
```diff
@@ -418,6 +418,9 @@ object PartitionedPostgisDialect {
     val WriteAheadPartitionsTableSpace = "pg.partitions.tablespace.wa-partitions"
     val MainTableSpace = "pg.partitions.tablespace.main"
 
+    // set postgres table wal logging
+    val WalLogEnabled = "pg.wal.log.enabled"
```
elahrvivaz, line 422: `wal log` is redundant (write ahead log log), could we change this to `pg.wa.log.enabled` or `pg.wal.enabled`?

autodidacticon: changed to `pg.wal.enabled`

8784b23a727736800eb4e7451e95fa162108bf21

elahrvivaz:
a few of these tables are shared between different feature types - to avoid them being created different depending on which feature type is created first, could you revert the changes to `PartitionTablespacesTable`, `PrimaryKeyTable`, `SequenceTable` and `UserDataTable` so that they're always logged? they only contain metadata, which is updated infrequently or never, so logging them shouldn't be a performance issue.

other than that, lgtm, thanks!

autodidacticon:
> a few of these tables are shared between different feature types - to avoid them being created differently depending on which feature type is created first, could you revert the changes to `PartitionTablespacesTable`, `PrimaryKeyTable`, `SequenceTable` and `UserDataTable` so that they're always logged? they only contain metadata, which is updated infrequently or never, so logging them shouldn't be a performance issue.
> 
> other than that, lgtm, thanks!

Done.

elahrvivaz:
ty!

PR #3245, opened 2024-12-12
GEOMESA-3426 FSDS - Support path filtering for converter queries

fdfea:


fdfea:
Reminding myself to update the docs

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/StorageMetadata.scala
```diff
@@ -53,6 +53,13 @@ trait StorageMetadata extends Compactable with Closeable {
     */
   def leafStorage: Boolean
 
+  /**
+    * The path filter factory used to filter paths during queries
+    *
+    * @return path filter factory
+    */
+  def pathFilterFactory: Option[PathFilterFactory]
```
elahrvivaz, line 61: what do you think about moving this to a consructor arg in `ConverterStorage`, and then have it loaded in `ConverterStorageFactory`? then it could be instantiated in `ConverterStorage.createReader` and passed to `ConverterFileSystemReader`. that would avoid having to modify the StorageMetadata api, since it's only implemented for converter storage we could just isolate the changes there. i'm trying to think if this would be generally useful ever for other metadata impls, but i can't think of anything, since in those we control the file paths that are written.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-api/src/main/scala/org/locationtech/geomesa/fs/storage/api/PathFilterFactory.scala
```diff
@@ -0,0 +1,16 @@
...
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.locationtech.geomesa.fs.storage.api
+
+import org.apache.hadoop.fs.PathFilter
+import org.geotools.api.filter.Filter
+
+trait PathFilterFactory {
```
elahrvivaz, line 14: maybe we can rename this to `PathFiltering` (or if you can think of a better name), so we don't have a FactoryFactory :P

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-convert/src/main/scala/org/locationtech/geomesa/fs/storage/converter/ConverterStorageFactory.scala
```diff
@@ -29,9 +34,22 @@ class ConverterStorageFactory extends FileSystemStorageFactory with LazyLogging
...
 
+    val pathFilteringOpts =
+      context.conf.getValByRegex(Pattern.quote(PathFilterOptsPrefix) + ".*").asScala.map {
+        case (k, v) => k.substring(PathFilterOptsPrefix.length) -> v
+      }
+
+    val pathFiltering = Option(context.conf.get(PathFilterName)).flatMap { name =>
+      val factory = PathFilteringFactory.load(NamedOptions(name, pathFilteringOpts.toMap))
+      if (factory.isEmpty) {
+        logger.warn(s"Failed to load ${classOf[PathFiltering].getName} for config '$name'")
```
elahrvivaz, line 46: do you think a warning is ok here, or should we throw an error? i worry people might not notice the warning, and then get worse performance but since things still work correctly they might not realize the mistake

elahrvivaz: lgtm otherwise!

fdfea:
I've tested with real data on my end and it seems to be working expected. Could we also get this in a 5.0.x release?

elahrvivaz:
thanks for remembering the docs!

PR #3271, opened 2025-01-31
GEOMESA-3447 FSDS - Support allow-list for attribute partition scheme

dlgundry:
- Added the ability to provided a comma separated list of allowed paths to AttributeScheme
- Allows AttributeScheme to be configured to only partition specific pre-defined allowed paths in FSDS

### DIFF
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
@@ -36,15 +36,15 @@ case class AttributeScheme(attribute: String, index: Int, binding: Class[_], def
 
   override def getPartitionName(feature: SimpleFeature): String = {
     val value = feature.getAttribute(index)
-    if (value == null)
-      return defaultPartition
-    val encodedValue = AttributeIndexKey.typeEncode(value)
-    if (allowedValues.nonEmpty) {
-      if (allowedValues.contains(value))
-        return encodedValue
+    if (value == null) {
       return defaultPartition
     }
-    encodedValue
+    val encodedValue = AttributeIndexKey.typeEncode(value)
+    if (allowedValues.isEmpty || allowedValues.contains(encodedValue)) {
+      encodedValue
+    } else {
+      defaultPartition
+    }
   }
 
   override def getSimplifiedFilters(filter: Filter, partition: Option[String]): Option[Seq[SimplifiedFilter]] = {
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
@@ -115,6 +115,7 @@ object AttributeScheme {
           s"Invalid type binding '${binding.getName}' of attribute '$attribute'")
         val allowedValues: Seq[String] = config.options.get(Config.AllowedListOpt).map(_.split(',').toSeq).getOrElse(Seq.empty).map(allowed => AttributeIndexKey.encodeForQuery(allowed.trim(),binding))
         val defaultPartition = AttributeIndexKey.encodeForQuery(config.options.getOrElse(Config.DefaultPartitionOpt, allowedValues.headOption.getOrElse("")), binding)
+        require(allowedValues.isEmpty || allowedValues.contains(defaultPartition), "Default partition must be one of the allowed values")
         Some(AttributeScheme(attribute, index, binding, defaultPartition, allowedValues))
       }
     }
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
@@ -113,7 +113,7 @@ object AttributeScheme {
         val binding = sft.getDescriptor(index).getType.getBinding
         require(AttributeIndexKey.encodable(binding),
           s"Invalid type binding '${binding.getName}' of attribute '$attribute'")
-        val allowedValues: Seq[String] = config.options.get(Config.AllowedListOpt).map(_.split(',').toSeq).getOrElse(Seq.empty).map(AttributeIndexKey.encodeForQuery(_,binding))
+        val allowedValues: Seq[String] = config.options.get(Config.AllowedListOpt).map(_.split(',').toSeq).getOrElse(Seq.empty).map(allowed => AttributeIndexKey.encodeForQuery(allowed.trim(),binding))
         val defaultPartition = AttributeIndexKey.encodeForQuery(config.options.getOrElse(Config.DefaultPartitionOpt, allowedValues.headOption.getOrElse("")), binding)
         Some(AttributeScheme(attribute, index, binding, defaultPartition, allowedValues))
       }
--- a/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
+++ b/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
@@ -22,8 +22,9 @@ import org.locationtech.geomesa.index.index.attribute.AttributeIndexKey
   * @param attribute attribute name
   * @param index attribute index in the sft
   * @param binding type binding
+  * @param allowedValues list of allowedValues to partition
   */
-case class AttributeScheme(attribute: String, index: Int, binding: Class[_]) extends PartitionScheme {
+case class AttributeScheme(attribute: String, index: Int, binding: Class[_], defaultPartition: String, allowedValues: Seq[String]) extends PartitionScheme {
 
   import FilterHelper.ff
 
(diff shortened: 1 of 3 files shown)


geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
```diff
@@ -22,8 +22,9 @@ import org.locationtech.geomesa.index.index.attribute.AttributeIndexKey
   * @param attribute attribute name
   * @param index attribute index in the sft
   * @param binding type binding
+  * @param enums list of enums to partition
   */
-case class AttributeScheme(attribute: String, index: Int, binding: Class[_]) extends PartitionScheme {
+case class AttributeScheme(attribute: String, index: Int, binding: Class[_], enums: Seq[String]) extends PartitionScheme {
```
fdfea, line 27: Please update the docs https://github.com/locationtech/geomesa/blob/main/docs/user/filesystem/partition_schemes.rst and add unit tests https://github.com/locationtech/geomesa/blob/main/geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/partitions/PartitionSchemeTest.scala

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
```diff
@@ -89,6 +91,7 @@ object AttributeScheme {
 
   object Config {
     val AttributeOpt: String = "partitioned-attribute"
+    val EnumerationOpt: String = "enums"
```
fdfea, line 94: I think this could have a more descriptive name

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
```diff
@@ -101,7 +104,11 @@ object AttributeScheme {
         val binding = sft.getDescriptor(index).getType.getBinding
         require(AttributeIndexKey.encodable(binding),
           s"Invalid type binding '${binding.getName}' of attribute '$attribute'")
-        Some(AttributeScheme(attribute, index, binding))
+        val enums: Seq[String] = config.options.getOrElse(Config.EnumerationOpt, null) match {
```
fdfea, line 107: pedantic, but slightly cleaner might be `config.options.get(Config.EnumerationOpt).map(s.split(",")).getOrElse(Seq.empty)`

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
elahrvivaz: maybe `allow-list`? and could rename enums to `allowedValues`

elahrvivaz:
seems like we should also update `getPartitionName` (which isn't used in the converter case, but would in the regular case), so that we don't end up with values that can't be read. i think then we'd also need a default value (instead of the empty string like we have now), maybe as another option, and if it's not specified just use the first in the enumerated list.

dlgundry:
> seems like we should also update `getPartitionName` (which isn't used in the converter case, but would in the regular case), so that we don't end up with values that can't be read. i think then we'd also need a default value (instead of the empty string like we have now), maybe as another option, and if it's not specified just use the first in the enumerated list.

I think I have captured your comment with the latest commit

elahrvivaz:
looks good, i think there's just a little bit of inconsistency in when we're dealing with type-encoded values and plan values. i'll try to capture that in another comment in a minute.

elahrvivaz:
So for the attribute index, we run values through a lexicoder, which maps from a normal value to a sorted stringified value. i.e. if you're dealing with ints, 0 maps to something like 80000000 so that negative values sort correctly. there are 2 methods, `AttributeIndexKey.typeEncode` (which assumes a valid typed input value) and `AttributeIndexKey.encodeForQuery` (which will try to convert the input into the correct type. Since we're dealing with strings in the input here, we probably want to run `allowedValues` and `defaultPartition` through `encodeForQuery` in the factory set-up. then i think the only change is we have to run the input value through typeEncode before comparing it in getPartitionName

elahrvivaz:
@dlgundry you'll also need to sign the Eclipse CLA as outlined here: https://github.com/locationtech/geomesa/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
```diff
@@ -101,7 +113,9 @@ object AttributeScheme {
         val binding = sft.getDescriptor(index).getType.getBinding
         require(AttributeIndexKey.encodable(binding),
           s"Invalid type binding '${binding.getName}' of attribute '$attribute'")
-        Some(AttributeScheme(attribute, index, binding))
+        val allowedValues: Seq[String] = config.options.get(Config.AllowedListOpt).map(_.split(',').toSeq).getOrElse(Seq.empty).map(AttributeIndexKey.encodeForQuery(_,binding))
+        val defaultPartition = config.options.getOrElse(Config.DefaultPartitionOpt, AttributeIndexKey.encodeForQuery(if (allowedValues.nonEmpty) allowedValues.head else "", binding))
```
elahrvivaz, line 117: need the encode to be applied to defaultValue too:

```suggestion
        val defaultPartition = AttributeIndexKey.encodeForQuery(config.options.getOrElse(Config.DefaultPartitionOpt, allowedValues.headOption.getOrElse("")), binding)
```

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
```diff
@@ -35,7 +36,15 @@ case class AttributeScheme(attribute: String, index: Int, binding: Class[_]) ext
...
-    if (value == null) { "" } else { AttributeIndexKey.typeEncode(value) }
+    if (value == null)
+      return defaultPartition
+    val encodedValue = AttributeIndexKey.typeEncode(value)
+    if (allowedValues.nonEmpty) {
+      if (allowedValues.contains(value))
+        return encodedValue
+      return defaultPartition
+    }
+    encodedValue
```
elahrvivaz, line 47: comparison needs to be against encodedValue here i believe. Might be a little clearer to not have explicit returns too (generally i think simple up-front checks like the null check followed by a return are fine, but they shouldn't be sprinkled throughout the method). also, we try to use {} around most statements for clarify.

```suggestion
    if (value == null) {
      return defaultPartition
    }
    val encodedValue = AttributeIndexKey.typeEncode(value)
    if (allowedValues.isEmpty || allowedValues.contains(encodedValue)) {
      encodedValue
    } else {
      defaultPartition
    }
```

elahrvivaz:
generally, don't force-push, that makes it easier to track changes - then i'll squash down and clean up the messages on merge.

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/test/scala/org/locationtech/geomesa/fs/storage/common/partitions/PartitionSchemeTest.scala
```diff
@@ -50,6 +50,24 @@ class PartitionSchemeTest extends Specification with AllExpectations {
       ps.getSimplifiedFilters(ECQL.toFilter("bbox(geom,-170,-80,170,80)")) must beNone
     }
 
+    "partition based on allow list containing attribute value" >> {
+      val ps = PartitionSchemeFactory.load(sft, NamedOptions("attribute", Map("partitioned-attribute" -> "name", "allow-list" -> "test, bar")))
```
elahrvivaz, line 54: hmm this test would probably fail for 'bar' - we should add a `.trim` in the parsing, or else update this test to not use spaces after the comma (or both)

dlgundry: It's not currently failing for me, but I can add trim to the parsing

elahrvivaz: yeah, it's not failing b/c we're testing `test` but i think it would fail if we tested filters against `bar`

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
```diff
@@ -101,7 +113,9 @@ object AttributeScheme {
         val binding = sft.getDescriptor(index).getType.getBinding
         require(AttributeIndexKey.encodable(binding),
           s"Invalid type binding '${binding.getName}' of attribute '$attribute'")
-        Some(AttributeScheme(attribute, index, binding))
+        val allowedValues: Seq[String] = config.options.get(Config.AllowedListOpt).map(_.split(',').toSeq).getOrElse(Seq.empty).map(allowed => AttributeIndexKey.encodeForQuery(allowed.trim(),binding))
+        val defaultPartition = AttributeIndexKey.encodeForQuery(config.options.getOrElse(Config.DefaultPartitionOpt, allowedValues.headOption.getOrElse("")), binding)
```
elahrvivaz, line 117: i guess one more thing here would be to verify defaultPartition is in allowedValues:
```suggestion
        val defaultPartition = AttributeIndexKey.encodeForQuery(config.options.getOrElse(Config.DefaultPartitionOpt, allowedValues.headOption.getOrElse("")), binding)
        require(allowedValues.isEmpty || allowedValues.contains(defaultPartition), "Default partition must be one of the allowed values")
```

elahrvivaz:
@fdfea any more comments?

geomesa-fs/geomesa-fs-storage/geomesa-fs-storage-common/src/main/scala/org/locationtech/geomesa/fs/storage/common/partitions/AttributeScheme.scala
dlgundry: the test "partition based on allow list not containing attribute value and default partition is set" fails with this change, should I remove it as it seems we'll not be supporting having a default-partition that is not listed in the allowed list?

fdfea: having a test that confirms an exception is thrown if the default-partition is not in the allow-list seems like a good idea

PR #3278, opened 2025-02-06
GEOMESA-3449,GEOMESA-3358 Lambda - Persist data in batches

elahrvivaz:
* Simplify in-memory cache so refs don't back up and cause memory issues
* Adds new `lambda.persist.batch.size` data store parameter


fdfea:
Just to clarify, the lambda will persist every duration set by `geomesa.lambda.persist.interval`. If during a persistence cycle, the number of expired features exceeds the `lambda.persist.batch.size`, then it will persist `lambda.persist.batch.size` features, starting with those at the lowest offset (and avoiding reading the entire offset).

If we occasionally expect our expired features offsets to get backed up, I assume it would be better to increase batch size, rather than shortening the persist interval?

elahrvivaz:
No, if there are more expired features than the batch size, it will immediately loop again and keep persisting batches of features until it has processed them all, without waiting for `geomesa.lambda.persist.interval`.

fdfea:
Ah I see, thanks, lgtm.

PR #3327, opened 2025-04-25
GEOMESA-4370 - add redis cluster support

autodidacticon:
Addresses: https://geomesa.atlassian.net/browse/GEOMESA-3470


autodidacticon:
Needs a separate test suite that uses a redis cluster.

geomesa-redis/geomesa-redis-datastore/src/main/scala/org/locationtech/geomesa/redis/data/RedisDataStoreFactory.scala
```diff
@@ -85,51 +94,86 @@ object RedisDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogging {
...
-    * Builds configuration from data store parameters
-    *
-    * @param params params
-    * @return
-    */
+   * Builds configuration from data store parameters
+   *
+   * @param params params
+   * @return
+   */
```
elahrvivaz, line 168: i think all the changes from here and below are just formatting changes, could you revert them to minimize the diff?

geomesa-redis/geomesa-redis-datastore/src/main/scala/org/locationtech/geomesa/redis/data/RedisDataStoreFactory.scala
```diff
@@ -6,6 +6,14 @@
...
  ***********************************************************************/
 
+/** *********************************************************************
+ * Copyright (c) 2013-2025 General Atomics Integrated Intelligence, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ * ********************************************************************* */
+
```
elahrvivaz, line 16: looks like a duplicate header line, think we can delete this
```suggestion
```

geomesa-redis/geomesa-redis-datastore/src/main/scala/org/locationtech/geomesa/redis/data/JedisClusterFactory.java
```diff
@@ -0,0 +1,156 @@
...
+import redis.clients.jedis.JedisClientConfig;
+
+import java.time.Duration;
+import java.util.Set;
+
+/**
+ * A factory for creating and managing pooled JedisCluster instances.
+ * This factory handles the lifecycle of JedisCluster objects within a pool.
+ */
+public class JedisClusterFactory extends BasePooledObjectFactory<JedisCluster> {
```
elahrvivaz, line 29: could you make a src/main/java directory, and move this class there?

elahrvivaz (changes requested):
Looks good, thanks for the contribution! Could you also add the new parameter to https://github.com/locationtech/geomesa/blob/main/docs/user/redis/usage.rst ?

elahrvivaz:
looks like it might be feasible to bring up a redis cluster using testcontainers: https://stackoverflow.com/questions/79367322/how-to-bring-up-redis-cluster-using-testcontainers-in-java

autodidacticon:
@elahrvivaz i managed to get around transaction issues with lua scripting. This should be ready for review.

docs/user/redis/usage.rst
```diff
@@ -10,8 +10,10 @@ Parameter                              Type    Description
 ====================================== ======= ====================================================================================
 ``redis.url *``                        String  Redis connection URL. The URL can be used to specify the Redis database and
                                                credentials, if required - for example, 'redis://user:pass@host:port/db'
+                                               If using a Redis cluster, this may be a comma-separated list of redis URLs.
```
elahrvivaz, line 13: i think these get rendered into one line (and then re-split based on browser width), so we'd want a period after the last sentence:

```suggestion
                                               credentials, if required - for example, 'redis://user:pass@host:port/db'.
                                               If using a Redis cluster, this may be a comma-separated list of redis URLs.
```

geomesa-redis/geomesa-redis-datastore/src/main/scala/org/locationtech/geomesa/redis/data/RedisDataStoreFactory.scala
```diff
@@ -90,31 +91,68 @@ object RedisDataStoreFactory extends GeoMesaDataStoreInfo with LazyLogging {
...
+          config.setTestOnBorrow(TestConnectionParam.lookup(params))
+          val timeout = SocketTimeoutParam.lookup(params).toMillis.toInt
+
+          new JedisPool(config, uri, timeout)
+        case true =>
+          val urls = RedisUrlParam.lookup(params).split(",")
+          val uri = URI.create(urls.head)
+          if (urls.isEmpty) {
+            throw new IllegalArgumentException("No Redis URLs provided. Please set 'redis.url'")
+          }
```
elahrvivaz, line 132: i think these 2 lines need to be flipped:

```suggestion
          if (urls.isEmpty) {
            throw new IllegalArgumentException("No Redis URLs provided. Please set 'redis.url'")
          }
          val uri = URI.create(urls.head)
```

geomesa-redis/geomesa-redis-datastore/src/main/scala/org/locationtech/geomesa/redis/data/SingletonJedisClusterPool.scala
```diff
@@ -0,0 +1,60 @@
...
+  override def returnResource(resource: UnifiedJedis): Unit = {
+    // Do nothing - we're always returning the same instance
+  }
+
+  override def returnBrokenResource(resource: UnifiedJedis): Unit = {
+    // Since JedisCluster handles its own connection recovery, we don't need to do anything here
+  }
+
+  override def close(): Unit = {
+    // do nothing - we're always returning the same instance, calling close on the cluster is handled by JedisCluster's internal pool
```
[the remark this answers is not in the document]
elahrvivaz: yeah, blocking the close on the JedisCluster connection returned from the pool makes sense, but i think the pool _itself_ is only closed on datastore.dispose(), and at that point the connection shouldn't be used any more and should probably be closed.

Given the differences in the api, maybe we should take your original approach and create a wrapper interface that can be backed by either a single JedisCluster connection, or a JedisPool of connections (which can be borrowed from for the lifecycle of each method call). But I'm ok with leaving the pool as the abstraction layer, since you've already put a lot of work into this.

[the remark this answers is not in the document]
autodidacticon: > yeah, blocking the close on the JedisCluster connection returned from the pool makes sense, but i think the pool itself is only closed on datastore.dispose(), and at that point the connection shouldn't be used any more and should probably be closed.

This is a crucial insight. In its current, this implementation will leak connections on a per datastore basis.

> But I'm ok with leaving the pool as the abstraction layer, since you've already put a lot of work into this.
I will consider this more. I think `WithClose` may need some attention since these two implementations have differing lifecycles.

Thanks for your insight. I will consider more.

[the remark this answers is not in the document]
autodidacticon:
```
yeah, blocking the close on the JedisCluster connection returned from the pool makes sense, but i think the pool itself is only closed on datastore.dispose(), and at that point the connection shouldn't be used any more and should probably be closed.
```
Ive managed a fix for this [here](https://github.com/locationtech/geomesa/pull/3327/files#diff-7e08b3c1c09d606fc558921eaed220e5089da8a2c01a5def9c04fc8e75f8e816R62-R66).

geomesa-redis/geomesa-redis-datastore/src/main/scala/org/locationtech/geomesa/redis/data/SingletonJedisClusterPool.scala
```diff
@@ -0,0 +1,68 @@
...
+  override def returnBrokenResource(resource: UnifiedJedis): Unit = {
+    // Since JedisCluster handles its own connection recovery, we don't need to do anything here
+  }
+
+  override def close(): Unit = {
+    // `JedisClusterUncloseable` has disabled `close` but we can obtain a reference to the `executor` field
+    // and call `close` on it, which is what UnifiedJedis' close method does
+    // we need to do this when the pool itself is closed
+    val executorField = classOf[UnifiedJedis].getDeclaredField("executor")
+    executorField.setAccessible(true)
```
elahrvivaz, line 63: i'm somewhat worried this will break in newer java versions, where reflection is more locked down, but if you're ok with that (or don't feel like testing) I don't mind for now.

autodidacticon: I've refactored as indicated. The initial approach was overcomplicated.

elahrvivaz:
i will give it one more look-over but i think we can probably merge after that if you're happy with it @autodidacticon ?

autodidacticon:
If you are running these tests on macos w/ docker you will need to enable host networking within Docker Desktop.

elahrvivaz:
thanks for the contribution!

elahrvivaz:
noticed the ticket was wrong so force pushed to c0d7b5f490c634edaa034ae7af0e97f8ed371cd4

autodidacticon:
🤦

PR #3330, opened 2025-04-29
GEOMESA-3432 FSDS - Write files as GeoParquet

elahrvivaz:
* Support both WKB and native GeoParquet encoding for all geometry types
* Add bounding box fields for predicate push-down
* Update GeoParquet metadata to include covering/bbox


adeet1:
@elahrvivaz Is this meant to be an extension of https://github.com/locationtech/geomesa/pull/3064? If so, then we could potentially close out that MR?

elahrvivaz:
> @elahrvivaz Is this meant to be an extension of #3064? If so, then we could potentially close out that MR?

yes, we can close that PR now, thanks.

PR #3351, opened 2025-05-29
GEOMESA-3484-add-cluster-param

autodidacticon:
@elahrvivaz This was missed in 5.3.0. Let me know what additions need to be made.

### DIFF
--- a/geomesa-redis/geomesa-redis-tools/src/main/scala/org/locationtech/geomesa/redis/tools/RedisDataStoreCommand.scala
+++ b/geomesa-redis/geomesa-redis-tools/src/main/scala/org/locationtech/geomesa/redis/tools/RedisDataStoreCommand.scala
@@ -26,7 +26,8 @@ trait RedisDataStoreCommand extends DataStoreCommand[RedisDataStore] {
   override def connection: Map[String, String] = {
     Map(
       RedisDataStoreParams.RedisUrlParam.key     -> params.url,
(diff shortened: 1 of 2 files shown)


geomesa-redis/geomesa-redis-tools/src/main/scala/org/locationtech/geomesa/redis/tools/RedisDataStoreCommand.scala
```diff
@@ -47,5 +48,8 @@ object RedisDataStoreCommand {
   trait RedisDataStoreParams extends CatalogParam {
     @Parameter(names = Array("--url", "-u"), description = "Redis connection URL", required = true)
     var url: String = _
+
+    @Parameter(names = Array("--cluster", "-cl"), description = "Redis cluster enabled?", required = false)
```
elahrvivaz, line 52: I'm not sure it entirely matters, but we have standardized on single `-`'s with single chars, a la `getopts`. could you change the `-cl` to just be a single char flag (or omit it if it's more confusing since `-c` is already taken)? thanks!

autodidacticon:
Changes have been applied.

Can we request a hotfix release?

elahrvivaz:
Thanks for the update!

> Can we request a hotfix release?

Unfortunately, releases still require a few hours of work, but a release shouldn't be necessary since you can just build/publish it locally as needed?

PR #3540, opened 2026-05-06
GEOMESA-3575 FSDS - Iceberg compatibility

elahrvivaz:
* New parquet file schemas that are compatible with iceberg
* Remove support for older parquet file encodings
* Fix some alignment issues between geomesa and iceberg partition schemes
* Standardize attribute names in partition schemes
* New command to register GeoMesa files with iceberg


cwdobbins:
This looks good, and I'll be able to make the recent trino w/ geo support branch I've been working compatible with your parquet format.

PR #3565, opened 2026-06-08
GEOMESA-3587 Kryo - Fix buffer overflow caused by setBuffer locking maxCapacity

EdgarLQS:
When serializing large features (>65KB data), the buffer expansion path calls output.setBuffer(expanded) which locks maxCapacity to the new buffer's physical length. Since Output is a thread-local singleton, the residual maxCapacity prevents subsequent larger features from being serialized on the same thread, causing Buffer overflow.

Fix by using the two-argument form setBuffer(expanded, -1) to preserve the unbounded maxCapacity semantics, consistent with existing usage in package.scala:100.

Adds a regression test that verifies sequential serialization of increasingly large features does not fail.


elahrvivaz:
@EdgarLQS thank you for the fix! sorry to be a hassle, but before I can merge you'll have to sign the Eclipse contributor agreement here, using the same email as your git commits: https://www.eclipse.org/legal/eca/

EdgarLQS:
I have signed the ECA. Could you please re-trigger the validation check?

Issue #3579, opened 2026-07-14
Migrate SfcBridge.scala functionality to geomesa-z3 as an interop package


cwdobbins:
GeoMesa convention supports providing java-interop packages alongside scala counterparts throughout the repository.  Migrating SfcBridge from geomesa-trino -> geomesa-z3 maintains the SFC-based functionality in one location, makes it available across the codebase via the geomesa-z3 core library, and removes scala-sdk requirements from geomesa-trino.

Deferred from [PR #3573](https://github.com/locationtech/geomesa/pull/3573)

PR #3580, opened 2026-07-14
GEOMESA-3579 Add SpatialFillingCurves Java-interop functionality to geomesa

cwdobbins:
## Summary

Migrates geomesa-trino plugin's Scala `SfcBridge` to `org.locationtech.geomesa.curve.interop.SpaceFillingCurves` in **geomesa-z3** — a Java-friendly facade over `Z2SFC`/`XZ2SFC` —  mirroring a pre-existing GeoMesa code convention, centralizing SFC-related code in the geomesa-z3 core library, enabling re-use across modules, and removing SDK dependency from geomesa-trino-plugin, allowing it to be a pure-Java module. 

## Motivation

The SfcBridge scala class was built to support the development of the geomesa-trino-plugin, to present a Java-friendly facade over Z2SFC/XZ2SFC methods, whose signatures Java could not consume.  Hosting the facade in geomesa-z3 makes it available to any JVM consumer and removes the one and only Scala source code from the Trino plugin module.

### geomesa-z3

- New `org.locationtech.geomesa.curve.interop.SpaceFillingCurves`: `z2Index`/`z2Hex`/`z2HexRanges`, `xz2Index`/`xz2Hex`/`xz2HexRanges`. 
- Nested `HexRange(lower, upper)` case class as the Java-nameable analogue of `IndexRange`; range methods return `java.util.List<HexRange>`.
- Hex encoding delegates to the existing `Z2SFC.hexEncode`/`XZ2SFC.hexEncode` which produces left-shifted lexographically-comparable strings
- SpaceFillingCurvesTest`: assures index/hex parity with the underlying SFCs, clamping, range coverage of indexed values, and the `maxRanges` bound.

### geomesa-trino-plugin

- `SfcBridge.scala` removed; call sites use the new facade and the typed `HexRange`
- Module is now pure Java: `src/main/scala` removed, direct `scala-library` dependency dropped (still bundled transitively in the shaded jar), `scala-maven-plugin` configuration removed.

No behavioral changes.

## Testing

`mvn -pl geomesa-z3 -pl geomesa-trino/geomesa-trino-plugin -pl geomesa-trino/geomesa-trino-datastore clean install`


geomesa-trino/geomesa-trino-plugin/pom.xml
```diff
@@ -169,34 +165,6 @@
...
-          </execution>
-          <execution>
-            <id>scala-test-compile</id>
-            <phase>process-test-resources</phase>
-            <goals>
-              <goal>testCompile</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
```
elahrvivaz, line 199: now that there's no scala, in order for javadocs to work, and to entirely skip the scala plugin, you'll need to copy the plugin config from geomesa-arrow-jts: https://github.com/locationtech/geomesa/blob/d342f0780204f8584410b0f47cbbfa202e62b727/geomesa-arrow/geomesa-arrow-jts/pom.xml#L42-L96

elahrvivaz (changes requested):

geomesa-trino/geomesa-trino-plugin/pom.xml
cwdobbins: fixed with commit

elahrvivaz:
closes #3579

Issue #3581, opened 2026-07-14
geomesa-trino-datastore: pass authorizations as a base64-encoded extra credential instead of a pipe-delimited list


cwdobbins:
The datastore currently forwards the caller's authorization set to the `spatial_iceberg` connector as a pipe-delimited Trino extra credential (`auths:first|second`), with the plugin's `ExtraCredentialAuthorizationResolver` splitting on pipes/commas/whitespace. Because the raw
tokens ride the extraCredentials wire format (`name:value,name:value`), tokens containing `, | ; :` or whitespace cannot round-trip, and both ends enforce a matching character restriction (`AuthTokens.isValid`) to fail closed.

Encode the full auth set as a single base64url value instead (i.e. `auths:<base64url(join(tokens))>`):

- lifts token-character restrictions, allowing arbitrary visibility tokens
- collapses the delimiter-validation logic on both ends to a single bas64 encode/decode step
- removes wire-format injection concerns the current validation guards against

Deferred from [PR #3573](https://github.com/locationtech/geomesa/pull/3573)

PR #3585, opened 2026-07-14
GEOMESA-3581 Base64-encode authorization credentials

cwdobbins:
## Summary

Changes the authorization encoding between geomesa-trino-datastore and the `spatial_iceberg` connector from a pipe-delimited list to a single base64url-encoded value: `auths:<base64url(join(tokens, ","))>` 

The Trino JDBC `extraCredentials` property is `name:value` pairs with `;`/`:` as structural delimiters, and the previous pipe-joined value shared its alphabet with that format — requiring delimiter validation on both ends to prevent a spoofed token from being re-split into authorizations that were never issued. The base64url alphabet (`A-Za-z0-9-_`) cannot collide with the delimiters, making the transport injection-proof.

## Changes

**geomesa-trino-datastore**
- `TrinoDataStore.connectionProperties` comma-joins the validated tokens and base64url encodes the result into the `auths` extra credential

**geomesa-trino-plugin**
- `ExtraCredentialAuthorizationResolver` base64url-decodes the credential value  (padded or unpadded accepted) before tokenizing. A value that is not valid base64 resolves to **no authorizations**. The decoded payload is still split liberally on
  comma/pipe/whitespace, so mesh-injected headers can encode whatever list format they already carry.

### DIFF
--- a/geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/AuthTokens.java
+++ b/geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/AuthTokens.java
@@ -15,9 +15,9 @@ import java.util.Collection;
  * meaning somewhere in the auth transport chain:
  *
  * <ul>
- *   <li>{@code |} — joins tokens in the JDBC {@code auths} extra credential;</li>
  *   <li>{@code ,} — joins tokens in the {@code is_visible} SQL literal (and splits them
- *       again inside the UDF), and delimits auth-mapping file lists;</li>
+ *       again inside the UDF), splits the decoded {@code auths} credential payload,
+ *       and delimits auth-mapping file lists;</li>
  *   <li>{@code ;} / {@code :} — delimit the Trino JDBC {@code extraCredentials}
  *       {@code name:value;name:value} wire encoding;</li>
  *   <li>whitespace and non-printable/non-ASCII — rejected by the Trino JDBC driver's
@@ -45,7 +45,7 @@ final class AuthTokens {
         }
         for (int i = 0; i < token.length(); i++) {
             char c = token.charAt(i);
-            if (c <= ' ' || c > '~' || c == ',' || c == '|' || c == ';' || c == ':') {
+            if (c <= ' ' || c > '~' || c == ',' || c == ';' || c == ':') {
                 return false;
             }
         }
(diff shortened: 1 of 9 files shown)


geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/security/ExtraCredentialAuthorizationResolver.java
```diff
@@ -102,12 +105,16 @@ private boolean secretMatches(ConnectorIdentity identity) {
...
+    private static void addTokens(Set<String> out, String encoded) {
+        if (encoded == null) return;
+        String decoded;
+        try {
+            decoded = new String(Base64.getUrlDecoder().decode(encoded.trim()), StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            LOG.warn("Authorization extra credential is not valid base64url; resolving to no auths");
+            return;
+        }
+        for (String token : decoded.split("[\\s,]+")) {
```
elahrvivaz: is there a reason to not just split on comma here? i think this would preclude auths from containing spaces, which i'm not sure if they ever actually would but seems like we could just split on comma

```suggestion
        for (String token : decoded.split(",")) {
```

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/AuthTokens.java
```diff
@@ -45,7 +45,7 @@ static boolean isValid(String token) {
         }
         for (int i = 0; i < token.length(); i++) {
             char c = token.charAt(i);
-            if (c <= ' ' || c > '~' || c == ',' || c == '|' || c == ';' || c == ':') {
+            if (c <= ' ' || c > '~' || c == ',' || c == ';' || c == ':') {
```
elahrvivaz: are these restrictions still necessary? technically auths can be any valid utf-8 string, it just has to be quoted appropriately. we do split on commas in many places already (which would break auths with valid, quoted commas), but we can probably allow the other chars

cwdobbins: per offline discussion settling on just the comma delimiter

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/security/ExtraCredentialAuthorizationResolver.java
cwdobbins: updated

elahrvivaz:
closes #3581

Issue #3582, opened 2026-07-14
geomesa-trino-datastore: emit pure ST_* predicates and consolidate bbox/Z2 pushdown derivation in the geomesa-trino-plugin connector


cwdobbins:
`TrinoFilterToSQL` currently rewrites CQL spatial predicates into composite SQL — bbox-struct overlap prefilters + `CASE WHEN` shortcuts — and the `spatial_iceberg` connector separately derives the same pruning (bbox domains, Z2/XZ2 ranges) from either ST_* calls or the recognized bbox-comparison pattern. As a result, the same logical query can produce different plans depending on whether it arrives through the datastore or as direct Trino SQL, and similar logic is maintained in two places.

Goal: the datastore emits only row-level `ST_*` function calls (plus temporal/attribute predicates), and the connector becomes the single place bbox/Z2/XZ2 pushdown is derived — so datastore queries and hand-written Trino SQL get identical treatment and deterministic plans.

Deferred from [PR #3573](https://github.com/locationtech/geomesa/pull/3573)

PR #3589, opened 2026-07-19
GEOMESA-3582 Trino - Removes BBOX predicates from datastore SQL, adds as feature flag to the connector

cwdobbins:
## GEOMESA-3582: consolidate spatial pushdown in the connector

Splits spatial-filter handling cleanly between the two `geomesa-trino` modules and makes query plans deterministic regardless of how a query arrives.

### Background

Previously the **datastore** rewrote CQL spatial predicates into composite SQL — `__X_bbox__` overlap prefilters plus `CASE WHEN` row-level shortcuts — *and* the `spatial_iceberg` **connector** independently derived the same pruning from either `ST_*` calls or the recognized bbox-comparison pattern. The same logical query could therefore produce different plans depending on whether it came through the datastore or as hand-written Trino SQL, and near-identical logic was maintained in two places.

### 1. Datastore emits pure `ST_*` (commit 1)

`TrinoFilterToSQL` now emits **only** row-level `ST_*` function calls (plus temporal / attribute predicates). Swapped operand forms are normalized column-first via the OGC complement (`within(a,b) ⇔ contains(b,a)`) so the connector's `ST_*(ST_GeomFromBinary(col), literal)` recognizer always matches. The connector is now the **single** place bbox/Z2/XZ2 pushdown is derived, so datastore queries and hand-written Trino SQL get identical treatment and deterministic plans.

### 2. Connector-side bbox short-circuit for rectangle `ST_Intersects` (commit 2, feature-flagged)

Emitting pure `ST_*` removes the datastore's inline `CASE WHEN` row-level shortcut, which had let a row whose bbox was fully inside the query rectangle skip the WKB decode. This commit recovers a **sound** version of that saving entirely inside the connector, with no datastore coupling.  pruning; those domains ride to the worker in the table handle's `unenforcedPredicate`. For the one case the bbox proves *exactly* — a rectangle `ST_Intersects` on a Z2/point geometry column — `SpatialConnectorMetadata` claims the predicate **enforced** (replacing only the `st_intersects` node with `TRUE`, never the `is_visible` row filter or any other conjunct) and wraps the handle so `SpatialPageSourceProvider` becomes the authoritative filter. Reading only the cheap bbox columns, the page source:

- **accepts** rows whose bbox lies fully inside the query rectangle with **no WKB decode**,
- **rejects** rows whose bbox cannot overlap it, and
- decodes the geometry and runs the exact `intersects` test only on the thin envelope-boundary shell.

For **point** geometry columns the stored bbox is degenerate (`xmin == xmax`, `ymin == ymax`), so the classifier reads just the two coordinate columns and the box test collapses to a point-in-range check.

- **Sound / exact**: directional float32 rounding (reject box rounded outward, accept box inward) guarantees any row that a nearest-rounded float32 bbox could misclassify falls through to the exact test, so the result is identical to the engine's exact predicate. Only the `st_intersects` call is claimed — `is_visible` and every other conjunct still execute.
- **Feature-flagged** via `geomesa.spatial.bbox-page-filter` (**on by default**); see commit 3 for the flag's cost/benefit and how to disable it.

### 3. Connector-side bbox reject pre-filter for the general case (commit 3)

Commit 2's short-circuit engages only for the exactly-provable case (rectangle `ST_Intersects` on a Z2/point column). This commit extends the same page source with a **reject-only** mode covering *every other* spatial predicate the connector could not claim enforced — non-rectangular query geometries, non-point columns, `ST_Contains`, the outer envelope of a `DWITHIN` distance query, etc.

In reject-only mode the page source drops any row whose bbox **cannot** overlap the query envelope (via `SourcePage.selectPositions`, **before** the geometry column is materialized) and hands the survivors to the engine's exact `ST_*` residual unchanged. It reuses the same reject-bound machinery as the short-circuit's reject pass. This is a pure pre-filter:

- **Sound / non-destructive**: only rows failing a *necessary* overlap condition are dropped, so no matching row is ever removed; the engine's exact predicate still runs on every survivor. Null bbox values are kept. Validated by cross-checking counts against the stock `iceberg` catalog (no pushdown) across rectangle, triangle, `ST_Within`, `DWITHIN`, and non-spatial filters.
- **Boosts coarsely-pruned, non-point / non-rectangular geometry**: the WKB decode is the dominant per-row cost, and rejecting on the bbox first avoids it for the rows that file/row-group pruning missed.
- **Zero-overhead when inapplicable**: non-spatial queries carry no bbox domains and delegate to Iceberg unchanged.

Both the short-circuit (commit 2) and this reject pre-filter are controlled by the single `geomesa.spatial.bbox-page-filter` catalog property, **on by default**. Its value is inversely proportional to spatial partition quality: on a well-`truncate`-partitioned Z2/XZ2 table, file/row-group pruning already removes most non-matching rows, so the extra bbox-column read can outweigh the saved decodes (most visibly for `DWITHIN`, which must materialize the geometry for the exact distance test regardless); on a coarsely-partitioned table the row-level rejection catches what file pruning misses. Set `geomesa.spatial.bbox-page-filter=false` to fall back to partition + per-file stat pruning plus the engine's exact `ST_*` on every surviving row.

### DIFF
--- a/geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorFactory.java
+++ b/geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorFactory.java
@@ -40,6 +40,19 @@ public class SpatialConnectorFactory implements ConnectorFactory {
     private static final String AUTH_RESOLVER   = SECURITY_PREFIX + "auth-resolver";
     private static final String AUTH_MAPPING    = SECURITY_PREFIX + "auth-mapping-file";
 
+
+    /** Enables connector-side bbox filtering (see {@link BboxFilteringPageSource}).
+     *  - For a rectangle {@code ST_Intersects} on a Z2/point geometry column the connector claims
+     *    the predicate as enforced and the page source filters:
+     *      a) rows whose bbox is inside the rectangle are accepted (skips decoding their geometry WKB),
+     *      b) rows outside are rejected
+     *      c) an exact test runs only on the envelope boundary.
+     *  - ON by default;
+     *  - Disable via {@code geomesa.spatial.bbox-page-filter=false}.
+     */
+    private static final String SPATIAL_PREFIX  = "geomesa.spatial.";
+    private static final String BBOX_PAGE_FILTER = SPATIAL_PREFIX + "bbox-page-filter";
+
     /** Default constructor; instances are created by {@link SpatialIcebergPlugin}. */
     public SpatialConnectorFactory() {}
 
@@ -65,16 +78,17 @@ public class SpatialConnectorFactory implements ConnectorFactory {
     public Connector create(String catalogName, Map<String, String> config,
                             ConnectorContext context) {
         AuthorizationResolver resolver = buildResolver(config);
+        boolean bboxShortCircuit = Boolean.parseBoolean(config.getOrDefault(BBOX_PAGE_FILTER, "true"));
 
-        // Iceberg uses strict config validation; strip our keys so it doesn't
-        // reject them as unused.
+        // Iceberg uses strict config validation; strip our keys so it doesn't reject them as unused.
         Map<String, String> icebergConfig = new LinkedHashMap<>();
-        config.forEach((k, v) -> { if (!k.startsWith(SECURITY_PREFIX)) icebergConfig.put(k, v); });
+        config.forEach((k, v) -> {
+            if (!k.startsWith(SECURITY_PREFIX) && !k.startsWith(SPATIAL_PREFIX)) icebergConfig.put(k, v);
+        });
 
-        ConnectorFactory icebergFactory =
-            new IcebergPlugin().getConnectorFactories().iterator().next();
+        ConnectorFactory icebergFactory = new IcebergPlugin().getConnectorFactories().iterator().next();
         Connector icebergConnector = icebergFactory.create(catalogName, icebergConfig, context);
-        return new SpatialConnector(icebergConnector, catalogName, resolver);
+        return new SpatialConnector(icebergConnector, catalogName, resolver, bboxShortCircuit);
     }
 
     /** Builds the identity→auths resolver from catalog config, or null when no
@@ -89,19 +103,17 @@ public class SpatialConnectorFactory implements ConnectorFactory {
         if ("file".equals(kind)) {
             String path = config.get(AUTH_MAPPING);
             if (path == null) {
-                throw new IllegalArgumentException(
-                    AUTH_RESOLVER + "=file requires " + AUTH_MAPPING);
+                throw new IllegalArgumentException(AUTH_RESOLVER + "=file requires " + AUTH_MAPPING);
             }
             return new FileAuthorizationResolver(Path.of(path));
         }
-        // External implementation named by fully-qualified class with a
-        // (Map<String,String>) constructor.
+        // External implementation named by fully-qualified class with a (Map<String,String>) constructor.
         try {
-            return (AuthorizationResolver) Class.forName(kind)
-                .getConstructor(Map.class).newInstance(config);
+            return (AuthorizationResolver) Class.forName(kind).getConstructor(Map.class).newInstance(config);
         } catch (ReflectiveOperationException e) {
             throw new IllegalArgumentException(
-                "Cannot load AuthorizationResolver '" + kind + "' named by " + AUTH_RESOLVER, e);
+                "Cannot load AuthorizationResolver '" + kind + "' named by " + AUTH_RESOLVER, e
+            );
         }
     }
 }
--- a/geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorMetadata.java
+++ b/geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorMetadata.java
@@ -82,6 +82,20 @@ public class SpatialConnectorMetadata implements ConnectorMetadata {
     /** Ordered sub-field names of a {@code __<X>_bbox__} struct column. */
     private static final List<String> BBOX_FIELDS = List.of("xmin", "ymin", "xmax", "ymax");
 
+    // Trino function names this connector recognizes, lowercased for consistency.
+    // n.b. Trino itself resolves function names case-insensitively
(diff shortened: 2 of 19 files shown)


geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorFactory.java
```diff
@@ -65,16 +76,19 @@ public String getName() {
     public Connector create(String catalogName, Map<String, String> config,
                             ConnectorContext context) {
         AuthorizationResolver resolver = buildResolver(config);
+        boolean bboxPageFilter = Boolean.parseBoolean(config.getOrDefault(BBOX_PAGE_FILTER, "false"));
```
elahrvivaz: seems like in the numbers you showed me, there was a decent speed-up - but you still think it should be off by default?

cwdobbins: I ran it enabled/disabled on rawob_harrisnextgen yesterday for some queries, let me go review those metrics and/or run a few explain queries directly against trino, will report back here

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorFactory.java
```diff
@@ -40,6 +40,17 @@ public class SpatialConnectorFactory implements ConnectorFactory {
...
+
+    /** Enables page-source bbox cheap-reject (see {@link BboxRejectingPageSource}): drops rows
+     *  whose bbox can't overlap the query envelope before their geometry WKB is decoded. Sound and
+     *  safety-net-preserving. OFF by default: its value is inversely proportional to spatial
+     *  partition quality — on a well-Z2/XZ2-partitioned table file/row-group pruning already
+     *  removes most non-matching rows, so the extra bbox-column read can outweigh the saved
+     *  decodes. Enable ({@code geomesa.spatial.bbox-page-filter=true}) for coarsely-partitioned
+     *  tables (few, large files) where row-level rejection catches what file pruning misses. */
+    private static final String SPATIAL_PREFIX  = "geomesa.spatial.";
+    private static final String BBOX_PAGE_FILTER = SPATIAL_PREFIX + "bbox-page-filter";
```
elahrvivaz, line 54: we should document this config in docs/user/trino - looks like we don't yet have a page for configs like this?

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorFactory.java
cwdobbins: Updated the connector-side query optimization, and proved it is faster on the dev datasets with the flag on, setting enabled by default

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorMetadata.java
```diff
@@ -219,13 +246,50 @@ public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(
...
+        // Conditions:
+        //     (a) accept via the envelope requires query == its envelope,
+        //     (b) the shell exact test (JTS intersects) matches Trino's ST_Intersects
+        // exactly for points;
+        //     (c) z2 partitioning ⟺ point data.
+        // Every other predicate keeps the engine's exact residual.
+        if (bboxShortCircuit && matches.size() == 1 && resultHandle instanceof IcebergTableHandle ith) {
+            SpatialMatch m = matches.get(0);
+            GeometryColumn g = geoms.get(m.geomName());
+            if ("st_intersects".equals(m.functionName())
```
elahrvivaz: will the function name always be lower-cased here? maybe do a case insensitive match, or maybe there's a constant somewhere we could use?

elahrvivaz: same comment for the other places in this file where we use function name comparisons

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorMetadata.java
```diff
@@ -425,6 +498,101 @@ private static Method lookupMethod(Class<?> c, String name) {
...
+
+    /** Finds the first {@code st_intersects} call in the expression, descending through ANDs. */
+    private static Call findSpatialCall(ConnectorExpression expr) {
+        if (!(expr instanceof Call call)) {
+            return null;
+        }
+        if ("st_intersects".equals(call.getFunctionName().getName().toLowerCase(Locale.ROOT))) {
+            return call;
+        }
+        if (AND_FUNCTION_NAME.equals(call.getFunctionName())) {
```
elahrvivaz, line 583: as a future enhancement we could consider spatial clauses that are ORs too

elahrvivaz (approved):
just had a question about case-sensitivity of function names, but otherwise looks good! (although I don't know the internals of trino well enough to spot any issues)

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorMetadata.java
cwdobbins: Meant to do that actually- pull the magic strings out and apply case sensitivity for reader sanity.  (it doesn't appear to matter, but definitely had been on my list of things to fix that i... dropped on the floor

geomesa-trino/geomesa-trino-plugin/src/main/java/org/locationtech/geomesa/trino/spatial/iceberg/connector/SpatialConnectorMetadata.java
cwdobbins: agree.   I'll probably do a quick follow-on after this PR gets merged, should be a fairly bite-sized unit of work

elahrvivaz:
closes #3582

PR #3573, opened 2026-07-01
GEOMESA-3591 Added Trino backend support for GeoMesa

cwdobbins:
Added Trino backend support for GeoMesa, via a spatial partition-supporting Trino plugin and a Trino-backed GeoTools DataStore.
  
  This changeset introduces the geomesa-trino module:

  - **geomesa-trino-plugin** — a custom Trino plugin (spatial_iceberg catalog) that adds spatial partition pushdown and row-level visibility filtering on top of Trino's stock Iceberg connector.
  - **geomesa-trino-datastore** — a read-only GeoTools DataStore that lets existing GeoMesa/GeoTools consumers query the same warehouse over JDBC (CQL in, features out), getting the plugin's pruning automatically.

  How it works

  **Plugin.** The connector wraps Trino's iceberg connector, inspects a query's target table's partition specs, and identifies spatial partitions.   SpatialConnectorMetadata.applyFilter() walks the query's constraint expression for spatial predicates (ST_Intersects/ST_Within/ST_Contains and the equivalent bbox-struct comparisons) and injects two kinds of domains: per-file __<geom>_bbox__ REAL-stat domains for Parquet file pruning, and Z2/XZ2 range domains on the truncate-partitioned index column for manifest/partition pruning. Splits, page sources, writes, and transactions all delegate to the stock connector unchanged — there's no per-row scan-time transform. The Z2/XZ2 space-filling curves reuse GeoMesa's own Z2SFC/XZ2SFC (via a small Scala SfcBridge), so partitioning is bit-compatible with the writer. Additionally, visibility filtering is accomplished by providing an is_visible(<delegate-auths>) predicate, which filters rows when a visibility column is present (__vis__ or visibilities, etc, - configurable) to appropriately count/select only those rows the user is entitled to access.

  **DataStore.** A ContentDataStore (query engine) over the Trino JDBC driver. It pushes filter/projection/sort/count/bounds down as SQL and opens a fresh, per-request-authenticated connection per query so each caller's authorizations are forwarded to the connector. It can integrate with _any_ Trino; but require the plugin to be running server-side in order to leverage the spatial pruning and visibility filtering.

  **Build**

  Maven runs on the standard JDK 17. Only geomesa-trino-plugin is compiled by JDK 25 toolchain — forced by Trino 481. The datastore and benchmark are Java-17-clean (the Trino JDBC driver is Java-11 bytecode) and build on the Maven JVM with no toolchain. Everything is in the default reactor.

[...]

PR #3590, opened 2026-07-19
GEOMESA-3584 Derive geometry type + name information from geomesa SFT spec metadata

### LINKED ISSUE #3584: TrinoTypeMapper: bind geometry attributes to their specific subtype via stored SFT metadata
`TrinoTypeMapper` currently binds geometry columns to `Point` | `Geometry`, but consumers of the GeoTools schema would benefit from the precise subtype (`LineString`, `Polygon`, `MultiPolygon`, …).

Goal: when the Iceberg table carries a GeoMesa-encoded SimpleFeatureType , parse it during schema discovery and bind each geometry attribute to
the subtype it declares. Fall back to the current implementation when no SFT metadata is present.

Deferred from [PR #3573](https://github.com/locationtech/geomesa/pull/3573)

cwdobbins:
## GEOMESA-3584: Derive geometry tyes from their stored SFT metadata attibutes

Deferred from #3573.

### Background

`TrinoTypeMapper` bound every geometry column to `Point` or generic `Geometry` — `Point`when the column carried a `__X_z2__` companion, `Geometry` otherwise. Consumers of the discovered GeoTools schema got no more than that, even though the table's geometry is often a precise subtype (`LineString`, `Polygon`, `MultiPolygon`, …).

### Change

When an Iceberg table carries a GeoMesa-encoded SimpleFeatureType, schema discovery now parses it and binds each geometry attribute to the subtype it declares; it falls back to the previous naming heuristic when no SFT metadata is present.

- **`TrinoSchemaDiscovery`** reads the `geomesa.sft.spec` (and `geomesa.sft.name`) table properties from the Iceberg `"<table>$properties"` metadata table.  The spec is parsed with GeoMesa's `SimpleFeatureTypes.createType`, and each geometry attribute's JTS binding drives the descriptor. 
- **`TrinoTypeMapper.toDescriptor`** takes the resolved geometry binding `Class`.
- **failsafe**: an absent property, an unreadable `$properties`, or an unparseable spec all fall back to the Point | Geometry heuristic.
- Declared `geomesa-utils` explicitly in the datastore pom since `SimpleFeatureTypes` is now used directly.

closes #3584

### DIFF
--- a/geomesa-trino/geomesa-trino-datastore/pom.xml
+++ b/geomesa-trino/geomesa-trino-datastore/pom.xml
@@ -64,6 +64,11 @@
       <artifactId>geomesa-security_${scala.binary.version}</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.locationtech.geomesa</groupId>
+      <artifactId>geomesa-utils_${scala.binary.version}</artifactId>
+      <version>${project.version}</version>
+    </dependency>
 
     <dependency>
       <groupId>org.junit.jupiter</groupId>
--- a/geomesa-trino/geomesa-trino-datastore/pom.xml
+++ b/geomesa-trino/geomesa-trino-datastore/pom.xml
@@ -62,14 +62,11 @@
     <dependency>
       <groupId>org.locationtech.geomesa</groupId>
       <artifactId>geomesa-security_${scala.binary.version}</artifactId>
-      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.locationtech.geomesa</groupId>
       <artifactId>geomesa-utils_${scala.binary.version}</artifactId>
-      <version>${project.version}</version>
     </dependency>
-
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
--- a/geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscovery.java
+++ b/geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscovery.java
@@ -22,6 +22,8 @@ import java.io.IOException;
 import java.sql.*;
 import java.util.*;
 
+import static org.locationtech.geomesa.trino.datastore.TrinoDataStore.escapeQuotes;
+
 class TrinoSchemaDiscovery {
 
     private static final Logger LOG = LoggerFactory.getLogger(TrinoSchemaDiscovery.class);
@@ -43,8 +45,12 @@ class TrinoSchemaDiscovery {
         tb.setName(typeName);
         tb.setNamespaceURI(store.getNamespaceURI());
 
-        String sql = String.format("SELECT * FROM \"%s\".\"%s\".\"%s\" LIMIT 0",
-            store.catalog(), store.trinoSchema(), typeName);
+        String sql = String.format(
+            "SELECT * FROM %s.%s.%s LIMIT 0",
+            escapeQuotes(store.catalog()),
+            escapeQuotes(store.trinoSchema()),
+            escapeQuotes(typeName)
+        );
 
         Map<String, String> sftProps = readSftProperties(typeName);
         String sftSpec = sftProps.get(SFT_SPEC_PROPERTY);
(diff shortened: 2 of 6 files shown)


geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscovery.java
```diff
@@ -79,9 +97,66 @@ SimpleFeatureType discover(String typeName) throws IOException {
...
+    }
+
+    /** Reads the {@code geomesa.sft.*} properties (spec + name) from the Iceberg
+     *  {@code <table>$properties} metadata table in one query; empty when the table carries none
+     *  or doesn't expose {@code $properties}. Non-fatal: any failure just disables SFT-driven
+     *  binding/name for this table. */
+    private Map<String, String> readSftProperties(String typeName) {
+        String sql = String.format(
+            "SELECT key, value FROM \"%s\".\"%s\".\"%s$properties\" WHERE key IN ('%s', '%s')",
+            store.catalog(), store.trinoSchema(), typeName, SFT_SPEC_PROPERTY, SFT_NAME_PROPERTY);
```
elahrvivaz: i think we missed escaping quotes here and on line 47

geomesa-trino/geomesa-trino-datastore/src/test/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscoveryTest.java
```diff
@@ -47,4 +51,58 @@ void xz2CompanionIsNotPointColumn() {
...
+        // table: per-attribute options (fs.bounds, default), a *-marked default Point geometry,
+        // and trailing ;-delimited user data. Must parse and yield geom → Point.
+        String spec = "flightId:String:fs.bounds=true,timeUp:Date:default=true:fs.bounds=true,"
+            + "lat:Double,lon:Double,alt:Double,classification:String,course:Double,speed:Double,"
+            + "callsign:String:fs.bounds=true,origin:String,destination:String,mode3a:String,"
+            + "modeS:String,tailNumber:String:fs.bounds=true,aircraftType:String,facilityName:String,"
+            + "facilityHash:String,airGround:String,updateType:String,altChange:String,"
+            + "processTime:Date,rcptTime:Date,iconHeading:Double,*geom:Point:srid=4326,"
+            + "recordHash:String,source:String,trackId:String,trueAirSpeed:Double,windDirection:Double,"
+            + "windSpeed:Double,windIsAccurate:Boolean,ingestLatency:Long;geomesa.index.dtg='timeUp'";
```
elahrvivaz: we should remove these specific references and just use something generic

geomesa-trino/geomesa-trino-datastore/pom.xml
```diff
@@ -55,6 +55,11 @@
       <artifactId>geomesa-security_${scala.binary.version}</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.locationtech.geomesa</groupId>
+      <artifactId>geomesa-utils_${scala.binary.version}</artifactId>
+      <version>${project.version}</version>
```
elahrvivaz:
```suggestion
    </dependency>
    <dependency>
      <groupId>org.locationtech.geomesa</groupId>
      <artifactId>geomesa-utils_${scala.binary.version}</artifactId>
```
these versions should already be defined in the root dependencyManagement

elahrvivaz (changes requested):

geomesa-trino/geomesa-trino-datastore/src/test/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscoveryTest.java
cwdobbins: yeah, 100%

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscovery.java
cwdobbins: fixed

geomesa-trino/geomesa-trino-datastore/src/test/java/org/locationtech/geomesa/trino/datastore/TrinoSchemaDiscoveryTest.java
cwdobbins: fixed

geomesa-trino/geomesa-trino-datastore/pom.xml
cwdobbins: fixed

elahrvivaz:
created https://github.com/locationtech/geomesa/issues/3592 as a related follow-up

Issue #3588, opened 2026-07-17
Create the geomesa-trino-dist module


Matt16060936:
The "*-dist" modules build a tar.gz with tooling, the plugin jar, and the "*-gs-plugin" tar.gz as well.

These archives appear to follow some convention for structure, starting with a top-level directory names after the plugin and version.

PR #3593, opened 2026-07-21
GEOMESA-3587 Trino - Add gs-plugin and dist modules

### LINKED ISSUE #3587: Create the geomesa-trino-gs-plugin module
The "*-gs-plugin" modules build a tar.gz file which contains the datastore for installing into geoserver.

These archives appear to have a flat layout (no subdirectories) and contain various jars needed by that plugin (including dependencies from outside the geomesa repo).

Matt16060936:
closes #3587 #3588

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -75,7 +75,7 @@ jobs:
           - name: utils
             list: geomesa-utils-parent/geomesa-bom,geomesa-utils-parent/geomesa-hadoop-utils,geomesa-utils-parent/geomesa-utils
           - name: trino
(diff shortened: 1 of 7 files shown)


elahrvivaz:
the validate-ci jobs check that all modules are covered in the tests and javadocs - even though these don't have any tests/javadocs... you'll have to add the new module to https://github.com/locationtech/geomesa/blob/ddda0db8314a4e1e765597e4a9719b59e7c6f308/.github/workflows/javadocs.yml#L61 and https://github.com/locationtech/geomesa/blob/ddda0db8314a4e1e765597e4a9719b59e7c6f308/.github/workflows/build-and-test.yml#L78

PR #3594, opened 2026-07-21
added geomesa-trino-dist module

Matt16060936:


elahrvivaz:
we can hold off on this for the moment b/c we'll want to get the gs-plugin into it too

cwdobbins:
> we can hold off on this for the moment b/c we'll want to get the gs-plugin into it too

Agree, there is a dependency chain here

Matt16060936:
Combined with #3593 to facilitate comprehensive testing.

PR #3601, opened 2026-07-24
GEOMESA-3600 Trino - clean up docs

### LINKED ISSUE #3600: Trino - Clean up documentation

elahrvivaz:
closes #3600

### DIFF
--- a/docs/user/trino/configuration.rst
+++ b/docs/user/trino/configuration.rst
@@ -3,28 +3,23 @@
 Trino Connector Configuration
 =============================
 
-The ``spatial_iceberg`` connector is configured through **catalog properties** in its
-``.properties`` file (see :ref:`trino_install`). Any property that is not in a
-``geomesa.*`` namespace is passed through unchanged to the stock Iceberg connector the
-plugin wraps, so all of Iceberg's own catalog properties are available as-is. The
-GeoMesa-specific properties below are consumed by the plugin and stripped before the
-config reaches the Iceberg delegate.
+The ``spatial_iceberg`` connector is configured through catalog properties in its ``.properties`` file (see :ref:`trino_install`).
+Any property that does not start with ``geomesa.*`` is passed through unchanged to the stock
+`Iceberg connector <https://trino.io/docs/current/connector/iceberg.html#>`__. The following GeoMesa-specific properties are
+supported:
 
 .. list-table::
     :header-rows: 1
-    :widths: 32 12 12 44
+    :widths: 32 12 44
 
     * - Catalog property
       - Default
-      - Required
       - Description
     * - ``geomesa.spatial.bbox-page-filter``
       - ``true``
-      - no
       - Enables connector-side, pre-decode bbox filtering in the page source (see below).
     * - ``geomesa.security.*``
       - —
-      - no
       - Trino-layer row-visibility enforcement; see :ref:`trino_security`.
 
 .. _trino_bbox_page_filter:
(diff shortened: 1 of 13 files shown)


geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoDataStoreFactory.java
```diff
@@ -55,9 +55,6 @@ public TrinoDataStoreFactory() {}
     /** Comma-delimited superset of authorizations to use for queries. */
     public static final Param AUTHS = new Param("geomesa.security.auths", String.class,
         "Comma-delimited superset of authorizations to be used for queries", false);
-    /** When true, ignore implicit authorizations from the underlying Trino user. */
-    public static final Param AUTHS_FORCE_EMPTY = new Param("geomesa.security.auths.force-empty",
```
elahrvivaz, line 59: i *think* this isn't used, it was an accumulo-specific thing where we look up default auths based on the connector credential.

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoDataStoreFactory.java
```diff
@@ -87,8 +84,7 @@ public TrinoDataStoreFactory() {}
      * @return parameter descriptors
      */
     @Override public Param[] getParametersInfo() {
-        return new Param[]{HOST, PORT, CATALOG, SCHEMA, NAMESPACE, USER,
-            AUTHS, AUTHS_FORCE_EMPTY, AUTH_PROVIDER, SECRET};
+        return new Param[]{HOST, PORT, SCHEMA, USER, AUTHS, SECRET, NAMESPACE};
```
elahrvivaz, line 87: i removed auth_provider here b/c this list is mainly just to populate the store page on geoserver, and there auth providers would expect to be SPI loaded and configured. but even without auth_provider here, you could still pass in a custom auth provider if you wanted (like for testing), and it would work

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoDataStoreFactory.java
cwdobbins: ~~oh but it is!~~

cwdobbins: ~~Because I looked at the accumulo datastore config an example of what we should implement, force-empty is indeed supported and looked for, but is probably implemented wrong-  it literally overrides the auths to the empty set, which will only return data from tables with no _vis_ column or rows with no visibilities spec'd~~

Nevermind, saw that you yanked this in revisions in this PR.

docs/user/trino/configuration.rst
```diff
@@ -3,72 +3,32 @@
...
-noticeable for ``DWITHIN``, where the geometry must be materialized for the exact distance
-test regardless. On a coarsely-partitioned table the row-level rejection catches what file
-pruning misses and is a clear win. Set ``geomesa.spatial.bbox-page-filter=false`` to turn
-it off; the connector then relies on partition and per-file stat pruning alone and the
-engine's exact ``ST_*`` predicate decodes every surviving row.
+``geomesa.spatial.bbox-page-filter`` controls whether the connector injects a bounding-box filter for spatial predicates. The
+filter can eliminate rows based on fast bounding box comparisons, without having to decode the full binary geometry
+value. But when disabled, the bounding box columns can be skipped, resulting in higher throughput. Thus, the filter is most
+useful when it eliminates many rows, for example when data is coarsely partitioned. Otherwise, regular manifest and
+file-level pruning may be sufficient. See :ref:`trino_design` for more details on query pruning.
```
cwdobbins, line 34: Refreshingly to the point :)

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoDataStoreFactory.java
cwdobbins: good by me

cwdobbins:
LGTM

PR #10798, opened 2026-08-25
Add extended type-mapping support for nested object and json types to geomesa-trino-datastore

cwdobbins:
Add extended type-mapping support for nested object and json types to geomesa-trino-datastore

### DIFF
--- a/geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
+++ b/geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
@@ -8,6 +8,10 @@
 
 package org.locationtech.geomesa.trino.datastore;
 
+import io.trino.jdbc.Row;
+import io.trino.jdbc.RowField;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import org.geotools.api.data.FeatureReader;
 import org.geotools.api.feature.IllegalAttributeException;
 import org.geotools.api.feature.simple.SimpleFeature;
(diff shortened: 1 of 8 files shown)


geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
```diff
@@ -120,6 +137,94 @@ public SimpleFeature next() throws IOException, IllegalAttributeException, NoSuc
...
+     * <p>The Trino driver hands back {@code java.sql.Array} for arrays and {@link Map} for maps, with {@code row}
+     * values arriving as maps of field name to value. Historically these were declared {@code byte[]} and passed through
+     * uncoerced, which happened to work but promised the wrong type to anything reading the schema.
+     *
+     * @param value raw JDBC value
+     * @param desc the attribute it belongs to
+     * @param col column name, for diagnostics
+     * @return a value assignable to the declared binding
+     */
+    private Object coerce(Object value, AttributeDescriptor desc, String col) throws SQLException {
```
elahrvivaz: do you think it's worth memoizing any of this stuff?

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
```diff
@@ -120,6 +137,94 @@ public SimpleFeature next() throws IOException, IllegalAttributeException, NoSuc
...
+            List<Object> out = new ArrayList<>(attrs.length);
+            for (Object o : attrs) {
+                out.add(normalize(o));
+            }
+            return out;
+        }
+        if (value instanceof Array array) {
+            Object raw = array.getArray();
+            List<Object> out = new ArrayList<>();
+            if (raw instanceof Object[] elements) {
```
elahrvivaz, line 234: should probably log a warning here if this isn't the expected type

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
```diff
@@ -120,6 +137,94 @@ public SimpleFeature next() throws IOException, IllegalAttributeException, NoSuc
...
+        if (value instanceof Row row) {
+            Map<Object, Object> out = new LinkedHashMap<>();
+            List<RowField> fields = row.getFields();
+            for (RowField f : fields) {
+                out.put(f.getName().orElseGet(() -> "field" + f.getOrdinal()),
+                        normalize(f.getValue()));
+            }
+            return out;
+        }
+        if (value instanceof Struct struct) {
```
elahrvivaz: it seems like struct should be equivalent to Row, i.e. produce a map? but it seems like it's hard to find the attribute names. will this actually ever be called? claude seems to think they will always come back as trino Row objects

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
cwdobbins: indeed, working that

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
cwdobbins: agree

geomesa-trino/geomesa-trino-datastore/src/main/java/org/locationtech/geomesa/trino/datastore/TrinoFeatureReader.java
cwdobbins: this is valid- that block is effectively a no-op

PR #10801, opened 2026-08-26
FS/Trino - Use variant fields for storing JSON

### LINKED ISSUE #10799: FS/Trino - Use variant types for storing JSON fields
* Requires an update to Iceberg v3 file format
* Is only compatible with Trino 483+

elahrvivaz:
closes #10799


geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergUpdateWriter.scala
```diff
@@ -134,50 +133,40 @@ object IcebergUpdateWriter {
 
     val files: java.util.Set[DeleteFile] = Collections.newSetFromMap(new ConcurrentHashMap[DeleteFile, java.lang.Boolean]())
 
-    override protected def createAction(partition: Partition): DeleteWriter = new DeleteWriter(storage, partition, files)
+    private val closed = new AtomicBoolean(false)
+    private val fileSupplier: Supplier[OutputFile] = () => storage.table.io().newOutputFile(storage.newFilePath("puffin"))
+    // the second parameter is a function to load previous deletes for merging
+    private val writer = new BaseDVFileWriter(fileSupplier, _ => null)
```
cwdobbins: [Claude]: Previous delete vectors are never merged, producing two live DVs per data file

private val writer = new BaseDVFileWriter(fileSupplier, _ => null)

BaseDVFileWriter.close() (iceberg-core 1.11.0, lines 117-126) calls loadPreviousDeletes.apply(path); a null return means the existing DV is neither merged into the new bitmap nor added to rewrittenDeleteFiles. IcebergUpdateWriter.close() then only calls delta.addDeletes(f) — never removeDeletes — so the old DV stays live in the manifest.

Iceberg permits at most one DV per data file. DeleteFileIndex.add (line 528-534) throws Can't index multiple DVs for <path>: .... So: delete a feature, commit, then delete a second feature from the same data file in a separate writer session, commit, then query → scan planning throws. MergingSnapshotProducer.mergeDVs() only dedupes DVs added within one commit, so it doesn't save this.

Fix: pass a real loader (a BaseDeleteLoader over the data file's existing DVs, the way SparkPositionDeltaWrite does) and route writer.result().rewrittenDeleteFiles() into RowDelta.removeDeletes(...)

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergParquetScan.scala
```diff
@@ -149,34 +141,18 @@ class IcebergParquetScan(
...
+            val deleteFilter = if (file.deletes().isEmpty) { None } else {
+              Some(new GenericDeleteFilter(table.io(), file, table.schema(), schema.schema))
             }
-            val projection = if (deleteIndex.isEmpty) { schema.schema } else { deleteProjection }
-
+            val projection = deleteFilter.fold(schema.schema)(_.requiredSchema())
             WithClose(readFile(file, projection)) { iter =>
-              val withDeletes = deleteIndex.fold(iter) { index =>
+              val withDeletes = deleteFilter.fold(iter) { deletes =>
+                val index = deletes.deletedRowPositions()
```
cwdobbins, line 158: [Claude]:  NPE / silently dropped equality deletes

DeleteFilter.deletedRowPositions() (line 253) returns null when posDeletes.isEmpty(). The guard only tests whether there are any deletes, so a task carrying equality deletes NPEs; a task with both kinds silently ignores the equality deletes and returns rows that should be gone.

The previous code had an explicit require(delete.content() == POSITION_DELETES, "Only positional deletes are supported...") that this PR removes — Either restore that check or use deleteFilter.filter(...), which applies both kinds.

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergUpdateWriter.scala
elahrvivaz: thanks, good catch claude!

PR #10806, opened 2026-08-27
Bugfix/geomesa tools example data

cwdobbins:


elahrvivaz:
fixed in https://github.com/locationtech/geomesa/pull/10812

PR #10812, opened 2026-08-31
Partitioned Postgis - Support attributes named 'fid'

### LINKED ISSUE #10811: Partitioned Postgis - Attributes named 'fid' break the schema

elahrvivaz:
closes #10811

### DIFF
--- a/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
+++ b/geomesa-gt/geomesa-gt-partitioning/src/main/scala/org/locationtech/geomesa/gt/partition/postgis/dialect/PartitionedPostgisDialect.scala
@@ -63,9 +63,8 @@ class PartitionedPostgisDialect(store: JDBCDataStore, grants: Seq[RoleName] = Se
   //  postDropTable
 
   // state for checking when we want to use the write_ahead table in place of the main view
-  private val dropping = new ThreadLocal[Boolean]() {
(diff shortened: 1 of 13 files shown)


cwdobbins (commented):
Beep. Boop. Beep:
```
upgrade() will corrupt schemas that have a fid attribute

SftUserData.FidColumn is never persisted. UserDataTable.create writes an explicit allowlist of keys, and pg.fid.col isn't in it, so it's gone on every read-back.

PartitionedPostgisDialect.upgrade (:83) — shipped as the PostgisUpgradeSchemaCommand CLI — calls postCreateTable with an sft from getSchema. On that path, the metadata probe is skipped, and Columns.apply falls back to the "fid" default. Then Commands.foreach(_.create(info)) regenerates everything against the wrong column:

- DeleteTrigger/UpdateTrigger are CREATE OR REPLACE, so their WHERE fid = OLD.fid now matches on the user's attribute instead of the pk fid_1 — wrong rows deleted/updated.
- PrimaryKeyTable.create does a delete-then-insert, re-registering pk_column = 'fid' in gt_pk_metadata — GeoTools then treats the user attribute as the feature id and exposes fid_1 as a regular attribute.
- New WA/main partitions get PRIMARY KEY (fid, dtg), which won't line up with the existing parent's (fid_1, dtg).

Fix: insert(SftUserData.FidColumn, info.cols.fid.raw) in UserDataTable.create. Read-back is automatic via UserDataTable.read. It's backwards compatible — pre-existing schemas couldn't have had a fid attribute, so the "fid" default is correct for them.
```

PR #10817, opened 2026-09-02
Postgis - Add support for per-row visibility filtering

### LINKED ISSUE #10816: Postgis - Support per-row visibility filtering

elahrvivaz:
closes #10816

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     commit-message:
       prefix: "[dependabot] "
     groups:
@@ -16,6 +18,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     commit-message:
       prefix: "[dependabot] "
     allow:
--- a/.github/workflows/assembly.yml
+++ b/.github/workflows/assembly.yml
@@ -62,7 +62,7 @@ jobs:
         run: |
           set -o pipefail
           # retry if the failure was due to transient download errors from maven central
-          if grep -q -e 'Could not transfer artifact' -e 'Failed to read artifact descriptor' -e 'Could not resolve artifact' -e 'Cannot invoke "java.io.File.toURI()" because "x" is null' build.log; then
+          if grep -q -e 'Could not transfer artifact' -e 'Failed to read artifact descriptor' -e 'Could not resolve artifact' build.log; then
             RESUME_FROM="$({ grep --text 'mvn <args> -rf ' build.log || test $? = 1; } | tail -n1 | sed 's/.*-rf/-rf/')"
             mvn $MAVEN_COMPILE_NO_OP_ARGS $MAVEN_CLI_OPTS $RESUME_FROM | tee -a build.log
           else
--- a/.github/workflows/assembly.yml
+++ b/.github/workflows/assembly.yml
@@ -30,6 +30,8 @@ jobs:
         scala-version: [ "2.12", "2.13" ]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
       - uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
         with:
           distribution: temurin
--- a/.github/workflows/assembly.yml
+++ b/.github/workflows/assembly.yml
@@ -60,7 +60,7 @@ jobs:
         run: |
           set -o pipefail
           # retry if the failure was due to transient download errors from maven central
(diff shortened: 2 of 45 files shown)


cwdobbins (commented):
1. Any query with explicit property names throws AIOOBE on a vis-enabled type

toUserFacing copies attributes positionally and reads _vis at index userFacingType.getAttributeCount. But ContentFeatureSource.getReader(Query) (gt-main 35.1, lines 650–658) retypes the reader to exactly the query's property names — so _vis isn't in the physical feature at all and the indices don't line up. Measured:

```
┌─────────────────────────────────────────┬────────────────────────────────────────────────────────────────────┐
│          query property names           │                               result                               │
├─────────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ [geom]                                  │ ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1 │
├─────────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ [name, geom]                            │ Index 2 out of bounds for length 2                                 │
├─────────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ [name, age, dtg, geom] — all user attrs │ Index 4 out of bounds for length 4                                 │
└─────────────────────────────────────────┴────────────────────────────────────────────────────────────────────┘
```
Fails identically through getFeatureReader(query, tx) and getFeatureSource(...).getFeatures(query). That's effectively every GeoServer WMS/WFS request — rendering and propertyName both set property names. A bare new Query(typeName) works, which is why the PR's tests pass. Resolve the _vis index by name off the incoming feature's own type, and retype against the reader's actual type rather than the cached full schema.

cwdobbins:
2. Malformed label is accepted on write, then breaks every read of the type

Nothing validates the label; SecurityUtils.getVisibility(feature) goes straight into _vis. pg_vis doesn't fail closed on an unbalanced paren — it pushes ( onto the output stack and casts it to boolean:
```
ERROR:  invalid input syntax for type boolean: "("
CONTEXT:  PL/pgSQL function pg_vis(character varying,character varying[]) line 119
```
End-to-end through the store: writing a feature with visibility (admin succeeds, and afterwards getFeatureReader(new Query(...)) and getCount both fail with IOException. In raw SQL I confirmed the row can't be repaired or removed through the view either — DELETE, UPDATE and count(*) all error — so recovery needs direct access to the partition tables. This also contradicts the function's own doc comment ("invalid strings will not raise errors but will always evaluate to 'false'"). Validate on write, and treat any non-boolean token on the output stack as false instead of casting.

cwdobbins:
3. Security — pg_vis grants access the reference evaluator denies

accumulo-access rejects mixed &/| without parens, and VisibilityFilterFunction maps that exception to deny. pg_vis accepts it and groups right-leaning:
```
┌────────┬───────┬─────────────────────────┬────────┐
│  expr  │ auths │     accumulo-access     │ pg_vis │
├────────┼───────┼─────────────────────────┼────────┤
│ a|b&c  │ a     │ reject → deny           │ true   │
├────────┼───────┼─────────────────────────┼────────┤
│ a&b|c  │ a,c   │ reject → deny           │ true   │
├────────┼───────┼─────────────────────────┼────────┤
│ a||b   │ a,b   │ reject → deny           │ true   │
├────────┼───────┼─────────────────────────┼────────┤
│ 'user' │ user  │ reject (only " quoting) │ true   │
└────────┴───────┴─────────────────────────┴────────┘
```

a|b&c with only a is the clearest: pg_vis evaluates a OR (b AND c) = true, while both accumulo-access and a left-to-right reading of (a|b)&c give false. The same label is more permissive in Postgres than in every other GeoMesa backend. Cause: the & branch only pops & and the | branch only pops |, so an equal-precedence operator of the other kind is never popped. Either reject mixed operators outright or pop both at equal precedence — and state which grammar is authoritative in the docs.

cwdobbins:
4. Bug — schema cache is never invalidated

The Caffeine cache only expires on write (geomesa.metadata.expiry, default 10 min), and DecoratingDataStore passes removeSchema/updateSchema straight through. After ds.removeSchema("vis_cache") — type gone from getTypeNames — ds.getSchema("vis_cache") still returned the old 4-attribute type. Worse with updateSchema and the partition-update-schema command: once an attribute is appended, _vis is no longer the last physical column and the positional index math in toUserFacing / VisFeatureWriter.write reads and writes visibility from the wrong column. Override both to invalidate, and look _vis up by name.

elahrvivaz:
> 4. Bug — schema cache is never invalidated
> 
> The Caffeine cache only expires on write (geomesa.metadata.expiry, default 10 min), and DecoratingDataStore passes removeSchema/updateSchema straight through. After ds.removeSchema("vis_cache") — type gone from getTypeNames — ds.getSchema("vis_cache") still returned the old 4-attribute type. Worse with updateSchema and the partition-update-schema command: once an attribute is appended, _vis is no longer the last physical column and the positional index math in toUserFacing / VisFeatureWriter.write reads and writes visibility from the wrong column. Override both to invalidate, and look _vis up by name.

the 2nd part of this isn't valid - JDBCDataStore doesn't support updateSchema, and attributes can't be appended - but I'll add a check just in case.

PR #10825, opened 2026-09-08
FS/Trino - Encode structural type column specs as iceberg table metadata.

cwdobbins:
FS/Trino - Encode structural type column specs as iceberg table metadata.


elahrvivaz:
wouldn't this require changes to TrinoSchemaDiscovery ?

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/SimpleFeatureIcebergSchema.scala
```diff
@@ -171,26 +170,21 @@ object SimpleFeatureIcebergSchema extends LazyLogging {
...
-      val doc = SimpleFeatureTypes.encodeDescriptor(sft, d)
+      // no doc is written: an attribute spec does not fit in a column comment once it carries a
+      // structural type definition, so every attribute's spec is a table property - see columnSpecs
       if (objectType.head == ObjectType.GEOMETRY) {
         // TODO supports native geometry encoding
         require(geometries == GeometryEncoding.GeoParquetWkb, "Only WKB encoding is supported for Geometry types")
-        val geomDoc = {
-          // note: geotools AttributeTypeBuilder shares the user data map - reparse instead so we don't change the original
-          val descriptor = SimpleFeatureTypes.createDescriptor(doc)
-          descriptor.getUserData.put(GeometryEncodingKey, geometries.toString)
```
elahrvivaz, line 181: noting that we're losing the geometry encoding serialization - since currently we only support 1 encoding, it's ok for now, but as a follow-on, i'd like to change how we configure geometry encoding so that it's set at the descriptor level and gets serialized there. i'll open a ticket for that

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/parquet/io/StructuralJson.scala
```diff
@@ -23,9 +24,93 @@ import java.util.{Base64, UUID}
...
+  import scala.collection.JavaConverters._
+
   private val encoder = Base64.getEncoder
   private val decoder = Base64.getDecoder
 
+  /**
+   * The field aliases declared by a schema, in the shape of the documents it describes. Opaque
+   * apart from being what [[applyAliases]] takes.
+   */
+  sealed trait Aliases
```
elahrvivaz: are these aliases supposed to be part of this PR? they don't seem to be invoked anywhere except the unit test for them, and seem like they're related to something else

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/SimpleFeatureIcebergSchema.scala
elahrvivaz: https://github.com/locationtech/geomesa/issues/10826

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/parquet/io/StructuralJson.scala
cwdobbins: This was cross-pollinated from another changeset and should be ripped out

elahrvivaz (approved):

PR #10828, opened 2026-09-11
Trino/FS - Add JSON-path query filtering and projections into structural types

elahrvivaz:


geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/SimpleFeatureIcebergSchema.scala
```diff
@@ -88,17 +88,18 @@ class SimpleFeatureIcebergSchema private (val sft: SimpleFeatureType, val schema
...
       case Some(defs) =>
-        val readSftBuilder = new SimpleFeatureTypeBuilder()
+        // use a hash set so we don't add a descriptor twice
+        val readDescriptors = new java.util.LinkedHashSet[AttributeDescriptor]()
 
         Transforms(sft, defs).foreach {
           case t: PropertyTransform =>
             val descriptor = sft.getDescriptor(t.i)
-            readSftBuilder.add(descriptor)
+            readDescriptors.add(descriptor)
```
cwdobbins, line 97: Claudio no likey this last change (aefe1ee5):

The change from readSftBuilder.add(StringUtils.strip(t.name, "\""), t.binding) (a descriptor named $.props.name) to readDescriptors.add(descriptor) (the base props descriptor) makes StructSimpleFeature's descriptor.getLocalName.startsWith("$") branch unreachable — schema.sft only ever comes from SimpleFeatureIcebergSchema.read or the base table sft, and neither now produces a $-prefixed attribute name. 

**StructJsonPathConverter** and **VariantPathAccessor** become dead code.

Performance-wise, with the path-named descriptor, tsft == readSchema.sft held and FileSystemStorage skipped the transformer entirely — the struct accessor did the projection directly off the iceberg row. Now tsft != readSchema.sft, so every returned feature goes: build a gson tree over the projected struct → serialize to a compact JSON string → re-parse with jayway JsonPath in TransformSimpleFeature. Results stay correct, but it's a per-row round trip that the earlier design specifically avoided.

The duplicate-descriptor problem being fixed is real (Array("props", "$.props.name") would add props twice), but the LinkedHashSet dedupe could be applied while keeping the distinctly-named path descriptor.

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergFilterConverter.scala
```diff
@@ -166,9 +204,104 @@ object IcebergFilterConverter extends LazyLogging {
...
+   * @param fieldPath dot-delimited path to the current field being evaluated
+   * @param fieldType the type of the field the path currently points at
+   * @param path the remaining path to navigate
+   * @return pairs of nested field names (matching the stored schema) and the leaf primitive type, or None if the path is not supported
+   */
+  private def navigate(fieldPath: String, fieldType: Type, path: JsonPath): Option[Seq[(String, Type)]] = {
+    if (path.isEmpty) {
+      return Some(Seq(fieldPath -> fieldType).filter(_._2.isPrimitiveType))
+    }
+    path.head match {
```
cwdobbins, line 281: Beep. Boop.  Flibberdigibiting.../

Navigate collapses two very different outcomes into Some(Seq.empty), and jsonPath turns that into a thrown IllegalArgumentException that fails the whole query:

- **Path into a map field**. The test schema has scores as an avro map → iceberg MapType. For "$.props.scores.x" = 42: PathAttribute("scores") navigates into the struct fine, then PathAttribute("x") hits fieldType = MapType.    The **if fieldType.isStructType** guard fails, and it falls to the trailing `case _ => Some(Seq.empty) → matchingFields.isEmpty → throw`.

- **Path terminating at a non-primitive**. navigate with an empty path does `Some(Seq(fieldPath -> fieldType).filter(_._2.isPrimitiveType))`, so "$.props.nested", "$.props.tags", or "$.props.scores" all yield Seq.empty → throw.

Both are queries a user can reasonably write, and both should behave like the None cases (alwaysTrue + client-side residual), not fail. The error message ("does not match any elements of the structural type") is also misleading in these cases — the path does match, it just isn't pushable.

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergFilterConverter.scala
```diff
@@ -166,9 +204,104 @@ object IcebergFilterConverter extends LazyLogging {
...
+        if (children.exists(_.isEmpty)) { None } else { Some(children.flatMap(_.get).toSeq) }
+
+      // predicates we can't evaluate as iceberg expressions
+      case PathDeepScan => None
+      case _: PathIndexRange if fieldType.isListType => None
+      case _: PathIndices if fieldType.isListType => None
+      case PathIndexWildCard if fieldType.isListType => None
+      case _: PathFilter => None
+
+      case _ => Some(Seq.empty) // doesn't match the field type
```
cwdobbins, line 303:
```
// map navigation is a real json path, not a non-match
case PathAttribute(name, _) if fieldType.isMapType =>
  navigate(s"$fieldPath.$name", fieldType.asMapType().valueType(), path.tail)  // or None, if not pushable
case _ => None  // was: Some(Seq.empty) — unsupported traversal, evaluate client-side
```

Suggestion: let Some(Seq.empty) align w/ comment on `caseInsensitiveField(name) == null` branch. The non-primitive-leaf filter (.filter(_._2.isPrimitiveType)) needs the same treatment — a path terminating at a struct should be None rather than "matched nothing."

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/SimpleFeatureIcebergSchema.scala
elahrvivaz: yeah the problematic change was introduced earlier but the point holds - i forget now exactly why the '$' caused a lot of complications, but i'll see if i can fix the unnecessary serialize/parse step.

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergFilterConverter.scala
elahrvivaz: I think the problem with **path terminating at a non-primitive** is that there isn't really any way to operate on a non-primitive in ECQL. you can kind of do string comparisons, since the struct types get returned as json strings, but you lose all json-semantics and just get a plain string compare (e.g. `{"x":1,"y":2} != {"y":2,"x":1}`)

elahrvivaz: fixed the misleading exception, added support for map types as client-side residuals.

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/IcebergFilterConverter.scala
elahrvivaz: i think this is fixed with the previous comment/commit - i believe paths into map types are not pushable since they don't have fixed keys?

geomesa-fs/geomesa-fs-storage/src/main/scala/org/locationtech/geomesa/fs/storage/core/iceberg/SimpleFeatureIcebergSchema.scala
elahrvivaz: i think this is fixed now

